PR #8, opened 2015-08-14
Fix two issues with bundle installation

ejoerns:
*  signature.c: Replaced outcontent parameter with NULL in CMS_verify()
    
    This fixes malloc issues when verifying larger files. The outcontent
    parameter was passed but not used. Passing NULL is an appropriate alternative
    as CMS_verify() replaces this by using BIO_s_null() internally.

 *  bundle.c: set size retrun parameter earlier in check_bundle()
    
    If an error occured in check_bundle() the 'size' return parameter was
    not yet set to an appropriate value, even when this was already
    available. This lead to incorrect outputs of the squashfs size.
    
    Assigning the value as soon as it is available will allow showing
    correct results in case of other failures.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -269,6 +269,10 @@ gboolean check_bundle(const gchar *bundlename, gsize *size, GError **error) {
 	}
 
 	offset -= sigsize;
+
+	if (size)
+		*size = offset;
+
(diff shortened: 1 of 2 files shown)


jluebbe:
I've cherry-picked these changes to fix a typo in the commit message.

PR #9, opened 2015-08-28
Extend tests for signature.c

ejoerns:
Also fixes two issues discovered by tests


jluebbe:
Cherry-pick in reversed order to have the fix first.

PR #15, opened 2016-01-08
install: check if bundle exists instead of looking at file extension

BastianStender:
Check if rauc bundle file exists before installing, no matter the file
extension. Otherwise handle argument as URL.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -225,7 +225,6 @@ out:
 	r_context_end_step("determine_slot_states", res);
 
 	return res;
-
 }
 
 /* Returns the inactive slots for a given slot class */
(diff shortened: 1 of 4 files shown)


src/install.c
```diff
@@ -1182,7 +1183,7 @@ static gpointer install_thread(gpointer data) {
 	g_debug("thread started for %s\n", args->name);
 	install_args_update(args, "started");
 
-	if (g_str_has_suffix(args->name, ".raucb")) {
+	if (access(args->name, F_OK) != -1) {
```
jluebbe: We should use g_access() instead.

jluebbe:
The check for '.raucb' is intentional to decide between bundles and manifests. We should fail early for missing bundles, though.

PR #24, opened 2016-01-21
bundle info enhancements

ejoerns:
Note: This patch stack is based on the command line handling patch stack

* `rauc info` shows more details of a bundle (manifest)
* Only the manifest file is extracted from squashfs, to save disk space
* Option `--no-verify` allows to check bundles without certificate check

Example output of `./rauc info --no-verify test/good-bundle.raucb`:

    ** Message: checking manifest for: test/good-bundle.raucb
    ** Message: system config not found, using default values
    ** Message: Compatible String:	'Test Config'
    ** Message: 2 Images:
    ** Message: (1)	rootfs.img
    ** Message: 	Slotclass: rootfs
    ** Message: 	Checksum:  de2f256064a0af797747c2b97505dc0b9f3df0de4f489eac731c23ae9ca9cc31
    ** Message: (2)	appfs.img
    ** Message: 	Slotclass: appfs
    ** Message: 	Checksum:  c35020473aed1b4642cd726cad727b63fff2824ad68cedd7ffb73c7cbd890479
    ** Message: 0 Files

### DIFF
--- a/include/bundle.h
+++ b/include/bundle.h
@@ -5,6 +5,7 @@
 gboolean create_bundle(const gchar *bundlename, const gchar *contentdir, GError **error);
 gboolean check_bundle(const gchar *bundlename, gsize *size, GError **error);
 gboolean extract_bundle(const gchar *bundlename, const gchar *outputdir, gboolean verify, GError **error);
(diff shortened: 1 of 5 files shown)


jluebbe:
Reorder the bundle file extract support before the patch which uses it. Also squash the fixup.

PR #30, opened 2016-02-08
main: handle return code of r_boot_set_state() in status mark-good/bad

ejoerns:
Otherwise a failure of r_boot_set_state() will not be propagated.

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -287,8 +287,8 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good) {
 		return grub_set_state(slot, good);
 	}
 
-	g_print("Warning: Your bootloader '%s' is not supported yet\n", r_context()->config->system_bootloader);
-	return TRUE;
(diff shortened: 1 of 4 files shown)


src/bootchooser.c
```diff
@@ -287,8 +287,8 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good) {
 		return grub_set_state(slot, good);
 	}
 
-	g_print("Warning: Your bootloader '%s' is not supported yet\n", r_context()->config->system_bootloader);
-	return TRUE;
+	g_warning("bootloader type '%s' not supported yet", r_context()->config->system_bootloader);
```
jluebbe: Check support for the bootloader during config parsing and use an g_error here.

ejoerns:
Updated pullrequest as requested. The bootloder setting is now checked during config file parsing.
There are also new tests added that check for error reporting in case of missing or invalid bootloader setting.

PR #31, opened 2016-02-08
Check for slot existence

ejoerns:
This adds the SlotState ST_NOTFOUND that will be set when the
destination slot path did not exist when calling determine_slot_states().

When trying to install into a slot with state ST_NOTFUND, the default
handler will report an error.

### DIFF
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -35,6 +35,7 @@ typedef enum {
 	ST_ACTIVE = 1,
 	ST_INACTIVE = 2,
 	ST_BOOTED = 4 | ST_ACTIVE,
+	ST_NOTFOUND = 8,
 } SlotState;
 
 typedef struct _RaucSlot {
--- a/src/install.c
+++ b/src/install.c
@@ -198,6 +198,12 @@ gboolean determine_slot_states(GError **error) {
 	for (GList *l = slotlist; l != NULL; l = l->next) {
 		RaucSlot *s = (RaucSlot*) g_hash_table_lookup(r_context()->config->slots, l->data);
 
+		if (!g_file_test(s->device, G_FILE_TEST_EXISTS)) {
+			g_message("Device for slot %s not found (%s)", s->name, s->device);
+			s->state = ST_NOTFOUND;
+			continue;
+		}
+
 		if (s->parent) {
 			if (s->parent->state & ST_ACTIVE) {
 				s->state |= ST_ACTIVE;
--- a/src/main.c
+++ b/src/main.c
@@ -518,9 +518,12 @@ static gboolean status_start(int argc, char **argv)
 			state = "booted";
 			booted = slot;
 			break;
+		case ST_NOTFOUND:
+			state = "not found";
+			break;
 		case ST_UNKNOWN:
 		default:
-			g_error("invalid slot status");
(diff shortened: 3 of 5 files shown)


include/config_file.h
```diff
@@ -28,6 +28,7 @@ typedef enum {
 	ST_ACTIVE = 1,
 	ST_INACTIVE = 2,
 	ST_BOOTED = 4 | ST_ACTIVE,
+	ST_NOTFOUND = 8 | ST_INACTIVE,
```
jluebbe: ST_NOTFOUND should not imply ST_INACTIVE

src/install.c
```diff
@@ -191,6 +191,11 @@ gboolean determine_slot_states(void) {
 			if (s->state == ST_UNKNOWN)
 				s->state |= ST_INACTIVE;
 		}
+
+		if (!g_file_test(s->device, G_FILE_TEST_EXISTS)) {
```
jluebbe: Perform this check first, before checking if this slot has a parent.

src/install.c
```diff
@@ -250,7 +255,7 @@ GHashTable* determine_target_install_group(RaucManifest *manifest) {
 	for (GList *l = slotmembers; l != NULL; l = l->next) {
 		RaucSlot *s = (RaucSlot*) g_hash_table_lookup(r_context()->config->slots, l->data);
 
-		if (s->state == ST_INACTIVE) {
+		if (s->state & ST_INACTIVE) {
```
jluebbe: We can keep this as == because we don't overload ST_INACTIVE anymore.

src/install.c
```diff
@@ -613,6 +618,17 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* cw
 	GHashTableIter iter;
 	gpointer class, member;
 
+	g_hash_table_iter_init(&iter, target_group);
```
jluebbe: Drop this. We won't have ST_NOTFOUND slots in the target_group.

src/main.c
```diff
@@ -354,7 +357,7 @@ static gboolean status_start(int argc, char **argv)
 	}
 
 	if (argc < 3) {
-		r_exit_status = 1;
+		r_exit_status = 0;
```
jluebbe: Keep this as error code 1.

test/rauc.t
```diff
@@ -69,12 +69,24 @@ test_expect_success "rauc bundle" "
 "
 
 
+TMPDIR=$(mktemp -d)
```
jluebbe: Check if this change is still needed.

src/install.c
```diff
@@ -198,6 +198,12 @@ gboolean determine_slot_states(GError **error) {
 	for (GList *l = slotlist; l != NULL; l = l->next) {
 		RaucSlot *s = (RaucSlot*) g_hash_table_lookup(r_context()->config->slots, l->data);
 
+		if (!g_file_test(s->device, G_FILE_TEST_EXISTS)) {
+			g_message("Device for slot %s not found (%s)", s->name, s->device);
```
jluebbe: This should be a warning.

ejoerns:
This one is pretty outdated and it is unsure if we still need this.

Anyway, if required, it might be worth simply re-implementing this on top of current code base.

Thus we close this for now.

PR #50, opened 2016-06-10
Introduce slot description and new dbus properties

ejoerns:
Also transforms former signal 'ProgressUpdated' to property to allow reading its whenever required.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -154,6 +154,10 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 			slot->name = g_intern_string(value);
 			g_free(value);
 
+			slot->description = g_intern_string(g_key_file_get_string(key_file, groups[i], "description", NULL));
+			if (!slot->description)
+				slot->description = g_strdup("");
+
 			slot->sclass = g_intern_string(groupsplit[1]);
 
 			value = resolve_path(filename,
--- a/src/service.c
+++ b/src/service.c
@@ -8,6 +8,7 @@
 #include <context.h>
 #include <install.h>
 #include <service.h>
+#include <mount.h>
 #include "rauc-installer-generated.h"
 
 GMainLoop *service_loop = NULL;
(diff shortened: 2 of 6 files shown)


src/config_file.c
```diff
@@ -154,6 +154,10 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 			slot->name = g_intern_string(value);
 			g_free(value);
 
+			slot->description = g_intern_string(g_key_file_get_string(key_file, groups[i], "description", NULL));
```
jluebbe: This should not be an interned string. It will need freeing in free_slot().

src/service.c
```diff
@@ -161,6 +225,8 @@ static void r_on_name_acquired(GDBusConnection *connection,
 	if (r_context()->config->autoinstall_path)
 		auto_install(r_context()->config->autoinstall_path);
 
+	set_slot_status();
+
```
jluebbe: This should be done after loading the config, so that the slot information is always available.

src/service.c
```diff
@@ -110,6 +111,69 @@ static gboolean auto_install(const gchar *source) {
...
+	slot_status_tuples = g_new(GVariant*, slot_number);
+
+	res = determine_slot_states(NULL);
+	if (!res)
+		return FALSE;
+
+	g_hash_table_iter_init(&iter, r_context()->config->slots);
+	while (g_hash_table_iter_next(&iter, &key, &value)) {
+		gchar *name = key;
```
jluebbe: Create a per slot function from the body of this loop.

src/service.c
```diff
@@ -110,6 +111,69 @@ static gboolean auto_install(const gchar *source) {
...
+
+			}
+			r_umount_slot(slot, NULL);
+			g_clear_pointer(&slotstatuspath, g_free);
+		}
+
+		if (slot->state == ST_BOOTED)
+			booted = slot;
+
+		slot_status = g_new(GVariant*, 3);
```
jluebbe: Make the dbus-property update a separate function.

ejoerns:
Replaced by #57

Issue #27, opened 2016-01-22
Auto-detect nfs boot


ejoerns:
when booting your target from nfs, currently it is not possible to use rauc, as it cannot detect to be booted from any configured slot.
The only solution is to add an nfs dummy entry to the system.conf.

As detecting an nfs is relatively easy and a common use case for testing, rauc should provide out-of-the box support for handling nfs as fallback if no other active slot could be detected.

ejoerns:
Solved by #51

PR #51, opened 2016-06-10
install: ignore missing active slot for nfs boots

ejoerns:
When booting from nfs, no valid active slot will be detected if nfs is
not mentioned in system.conf.
Ignoring this when seeing `/dev/nfs` in the command line will ease
working with rauc from nfs for testing and debugging purpose.

Signed-off-by: Enrico Joerns <mail@ejoerns.de>


ejoerns:
Belongs to #27

PR #52, opened 2016-06-13
install: disconnect all signal handlers that use args before destroying it

ejoerns:
This will prevent the client handlers from accessing args when might
have already been cleared.

This fixes a segfault that occured while rauc terminated with an error
code.
And, this should also fix the following error:

| Unexpected error from C library during 'pthread_mutex_lock': Invalid
| argument.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -36,15 +36,6 @@ static gboolean install_notify(gpointer data) {
 
 static gboolean install_cleanup(gpointer data)
 {
-	RaucInstallArgs *args = data;
-
-	g_mutex_lock(&args->status_mutex);
-	g_message("installing %s done: %d", args->name, args->status_result);
-	r_exit_status = args->status_result;
-	g_mutex_unlock(&args->status_mutex);
-
-	install_args_free(args);
-
 	g_main_loop_quit(r_loop);
 
 	return G_SOURCE_REMOVE;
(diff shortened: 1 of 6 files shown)


src/main.c
```diff
@@ -142,7 +133,15 @@ static gboolean install_start(int argc, char **argv)
 	g_main_loop_run(r_loop);
 out:
 	g_clear_pointer(&r_loop, g_main_loop_unref);
+
+	g_signal_handlers_disconnect_by_data(installer, args);
 	g_clear_pointer(&installer, g_object_unref);
+
+	g_message("installing %s done: %d", args->name, args->status_result);
+	r_exit_status = args->status_result;
+
```
jluebbe: move out here

src/main.c
```diff
@@ -142,7 +133,15 @@ static gboolean install_start(int argc, char **argv)
 	g_main_loop_run(r_loop);
 out:
```
jluebbe: rename this to out_loop

src/main.c
```diff
@@ -142,7 +133,15 @@ static gboolean install_start(int argc, char **argv)
...
 	g_clear_pointer(&r_loop, g_main_loop_unref);
+
+	g_signal_handlers_disconnect_by_data(installer, args);
 	g_clear_pointer(&installer, g_object_unref);
+
+	g_message("installing %s done: %d", args->name, args->status_result);
```
jluebbe: move this directly after the g_main_loop_run and set r_exit_status early in the function

PR #57, opened 2016-06-30
Introduce slot description, make slot status/state information available anytime and add new dbus properties

BastianStender:
This is an improved and reworked version of #50.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -1,6 +1,8 @@
 #include <config_file.h>
 #include <network.h>
 #include <signature.h>
+#include <mount.h>
+#include <install.h>
 #include <gio/gio.h>
 
 #include "context.h"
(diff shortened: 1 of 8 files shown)


src/rauc-installer.xml
```diff
@@ -6,12 +6,7 @@
 
     <property name="Operation" type="s" access="read"/>
     <property name="LastError" type="s" access="read"/>
-
-    <signal name="ProgressUpdated">
-      <arg name="percentage" type="i"/>
-      <arg name="description" type="s"/>
-      <arg name="nesting_depth" type="i"/>
-    </signal>
+    <property name="ProgressUpdated" type="(isi)" access="read"/>
```
jluebbe: "ProgressUpdated" doesn't fit as a name for a property. Just use "Progress".

src/context.c
```diff
@@ -72,6 +74,50 @@ static gboolean launch_and_wait_variables_handler(gchar *handler_name, GHashTabl
 	return res;
 }
 
+/*
+ * Mounts slot (if necessary), tries to read status file and saves information
+ * in GVariant structure. If the given slot is booted, the booted argument is
```
jluebbe: We don't use GVariants here.

BastianStender:
config_file: add slot description was cherry-picked.

Replaced by #59.

PR #59, opened 2016-07-13
Prepare context explicitly and make slot status/state information available

BastianStender:
This is a reworked version of #57.

### DIFF
--- a/src/rauc-installer.xml
+++ b/src/rauc-installer.xml
@@ -7,6 +7,8 @@
     <property name="Operation" type="s" access="read"/>
     <property name="LastError" type="s" access="read"/>
     <property name="Progress" type="(isi)" access="read"/>
+    <property name="SlotStatus" type="a(sss)" access="read"/>
+    <property name="BootedSlot" type="s" access="read"/>
 
     <signal name="Completed">
       <arg name="result" type="i"/>
--- a/src/service.c
+++ b/src/service.c
@@ -8,6 +8,7 @@
 #include <context.h>
 #include <install.h>
 #include <service.h>
+#include <mount.h>
 #include "rauc-installer-generated.h"
 
 GMainLoop *service_loop = NULL;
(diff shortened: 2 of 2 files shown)


src/context.c
```diff
@@ -72,6 +74,48 @@ static gboolean launch_and_wait_variables_handler(gchar *handler_name, GHashTabl
...
+
+/*
+ * Sets slot status information.
+ */
+static gboolean set_slots_status(void) {
+	GHashTableIter iter;
+	gpointer key, value;
+	gboolean res = FALSE;
+
+	res = determine_slot_states(NULL);
```
ejoerns: Use and propagate error information provided by `determine_slot_states()` here

src/context.c
```diff
@@ -126,6 +170,10 @@ void r_context_prepare(void) {
 	}
 
 	context->pending = FALSE;
+
+	if (!set_slots_status()) {
```
ejoerns: should be called before setting pending = FALSE

src/context.c
```diff
@@ -126,6 +170,10 @@ void r_context_prepare(void) {
...
+		g_debug("could not determine slot states");
```
ejoerns: logging level should be "warning" and display corresponding GError message

src/context.c
```diff
@@ -72,6 +74,48 @@ static gboolean launch_and_wait_variables_handler(gchar *handler_name, GHashTabl
 	return res;
 }
 
+/*
+ * Mounts slot (if necessary), tries to read status file and saves information
+ * in RaucSlot struct.
+ */
+static void set_slot_status(RaucSlot *slot) {
```
ejoerns: naming it `set_` might be a bit confusing here, what about `obtain_` or something similar?

src/context.c
```diff
@@ -72,6 +74,48 @@ static gboolean launch_and_wait_variables_handler(gchar *handler_name, GHashTabl
...
+			r_umount_slot(slot, NULL);
+
+		g_clear_pointer(&slot_status_path, g_free);
+	}
+}
+
+/*
+ * Sets slot status information.
+ */
+static gboolean set_slots_status(void) {
```
ejoerns: see note about `set_slot_status`

ejoerns:
Instead of changing the context setup completely, wouldn't it make sense to remove the progress calls from determine_slot_states() if it is not called by the install process exclusively, anymore?

Something like this maybe?:
```diff
diff --git a/src/install.c b/src/install.c
index 4bf28a3..0272f8e 100644
--- a/src/install.c
+++ b/src/install.c
@@ -135,7 +135,6 @@ gboolean determine_slot_states(GError **error) {
 
        g_assert_nonnull(r_context()->config);
 
-       r_context_begin_step("determine_slot_states", "Determining slot states", 0);
 
        if (r_context()->config->slots == NULL) {
                g_set_error_literal(
@@ -233,7 +232,6 @@ gboolean determine_slot_states(GError **error) {
 
 out:
        g_clear_pointer(&slotlist, g_list_free);
-       r_context_end_step("determine_slot_states", res);
 
        return res;
 }
@@ -967,7 +965,9 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error) {
        g_assert_nonnull(bundlefile);
 
        r_context_begin_step("do_install_bundle", "Installing", 5);
+       r_context_begin_step("determine_slot_states", "Determining slot states", 0);
        res = determine_slot_states(&ierror);
+       r_context_end_step("determine_slot_states", res);
        if (!res) {
                g_propagate_error(error, ierror);
                goto out;
```

ejoerns:
I guess this series is quite obsolete and we should close it. Any doubts?

ejoerns:
Obsolete now. Functionality introduced here is a subset of stuff contained in #166.

PR #60, opened 2016-08-05
Error handling fixes

ejoerns:
This adds some fixes and enhancements for error handling conerning key / cert file parsing and handling.
It also corrects a warning message in bootchooser code.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -186,6 +186,7 @@ gboolean create_bundle(const gchar *bundlename, const gchar *contentdir, GError 
 				error,
 				ierror,
 				"failed signing bundle: ");
+		res = FALSE;
 		goto out;
 	}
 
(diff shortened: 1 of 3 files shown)


src/bundle.c
```diff
@@ -165,7 +165,7 @@ gboolean create_bundle(const gchar *bundlename, const gchar *contentdir, GError
 	GBytes *sig = NULL;
 	GFile *bundlefile = NULL;
 	GFileOutputStream *bundlestream = NULL;
-	gboolean res = FALSE;
+	gboolean res = TRUE;
```
jluebbe: We initialize res to FALSE in all other funcitons. So keep this as-is.

ejoerns:
``` diff
diff --git a/src/bundle.c b/src/bundle.c
index 2f3f37d..72e25a7 100644
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -165,7 +165,7 @@ gboolean create_bundle(const gchar *bundlename, const gchar *contentdir, GError
        GBytes *sig = NULL;
        GFile *bundlefile = NULL;
        GFileOutputStream *bundlestream = NULL;
-       gboolean res = TRUE;
+       gboolean res = FALSE;
        guint64 offset;
 
        g_assert_nonnull(r_context()->certpath);

```

PR #61, opened 2016-08-05
Adaptions for upcoming bootchooser implementation

ejoerns:


src/install.c
```diff
@@ -75,9 +75,9 @@ static const gchar* get_cmdline_bootname(void) {
 	/* For barebox, we check if the bootstate code set the active slot name
 	 * in the command line */
 	if (g_strcmp0(r_context()->config->system_bootloader, "barebox") == 0) {
-		regex = g_regex_new("bootstate\\.active=(\\S+)", 0, 0, NULL);
+		regex = g_regex_new("(bootstate|bootchooser)\\.active=(\\S+)", 0, 0, NULL);
```
ukleinek: If you write this as
    regex = g_regex_new("(?:bootstate|bootchooser)\\.active=(\\S+)", 0, 0, NULL);
only a single match group is created. This is probably a bit more effective. You must not touch line 80 then.
I just tested this, works as expected (with the new bootchooser implementation).

ejoerns: :+1:  updated my PR

PR #26, opened 2016-01-21
Allow to choose output style of `rauc status`

ejoerns:
Note: Preview PR based on other pull requests

Allows to select between standard and 'shell' output for `rauc status` to ease parsing of informations for shell scripts.

PR #68, opened 2016-08-21
rauc status: allow to choose output style

ejoerns:
This introduces a `output-format` optional parameter for `rauc status` which allows to select the desired printout style.
This will be helpful to either get human readable or parsable output.

Currently supported formats are:
* human readable
* shell-parsable
* json
* json-pretty (with line breaks and indention)

Note that having 'json-glib' on the system is required for building with json support. The configure switch `--disable-json` allows building rauc without json support.

This obsoletes PR #26

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -505,17 +505,53 @@ static gchar* slotstate_to_str(SlotState slotstate)
 	return state;
 }
 
+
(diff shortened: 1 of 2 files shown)


jluebbe:
The should be rebased after PR #70 is merged.

PR #78, opened 2016-10-05
manifest: use checksum for hook file

jremmet:
Verifying the integrity of the hook file in the manifest is
necessary for security reasons.

Handle hook files like other images or files.
Add test for manifest with hooks and checksum

Signed-off-by: Jan Remmet j.remmet@phytec.de

### DIFF
--- a/include/manifest.h
+++ b/include/manifest.h
@@ -49,6 +49,7 @@ typedef struct {
 	gchar *handler_args;
 
 	gchar *hook_name;
+	RaucChecksum hook_checksum;
 	InstallHooks hooks;
 
 	GList *images;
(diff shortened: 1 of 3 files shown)


jluebbe:
Not having a checksum for the hook script is not a security problem, because the whole bundle is checked before use.

Nevertheless, we should calculate and add the checksum to the manifest. This will prepare for using the hooks script in network mode.

ejoerns:
I've taken the liberty of proposing a patch for this PR to allow it getting merged. Any concerns about that?

ejoerns:
:warning: Note that this feature change cannot be merged directly to prevent breaking Bundle compatibility.

This patch results in bundles unconditionally generated with having new keys 'sha256' and 'size' in the Manifest. Now, older versions of RAUC will reject these bundles due to unknown Manifest entries.

In order to allow this change to go into RAUC, a further step would be required that adds optional parsing of the new keys. Then, in a later release, one would add the full support, finally.
This would potentially allow upgrading from an old version to the interim version and then to the final version including this change without breaking compatibility.

jremmet:
That's fine by now.

PR #100, opened 2017-02-20
doc: using: dbus case sensitivity

jremmet:
use correct upper and lower case for dbus interface and method/property


ejoerns:
Congratulations, you've created our :100:th pull request :+1: :tada:

PR #101, opened 2017-02-20
bootchooser: put magic values in defines

jremmet:

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -7,6 +7,10 @@
 #include "install.h"
 
 #define BAREBOX_STATE_NAME "barebox-state"
+#define BAREBOX_STATE_DEFAULT_ATTEMPS	3
+#define BAREBOX_STATE_ATTEMPS_PRIMARY	3
+#define BAREBOX_STATE_DEFAULT_PRIORITY	10
+#define BAREBOX_STATE_PRIORITY_PRIMARY	20
 #define UBOOT_FWSETENV_NAME "fw_setenv"
 
 #if 0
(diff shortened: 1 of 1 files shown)


jremmet:
Maybe storing this values in system.conf is an option? We could also fall-back to the now hard coded.

src/bootchooser.c
```diff
@@ -152,16 +155,16 @@ static gboolean barebox_set_primary(RaucSlot *slot) {
...
-			prio = prio2;
+			prio = BAREBOX_STATE_DEFAULT_PRIORITY;
 		}
 		g_ptr_array_add(pairs, g_strdup_printf("%s.%s.priority=%i",
 				BOOTSTATE_PREFIX, s->bootname, prio));
 	}
 
 	g_ptr_array_add(pairs, g_strdup_printf("%s.%s.remaining_attempts=%i",
-			BOOTSTATE_PREFIX, slot->bootname, 3));
+			BOOTSTATE_PREFIX, slot->bootname, BAREBOX_STATE_NEW_ATTEMPS));
```
ejoerns: This should be `BAREBOX_STATE_DEFAULT_ATTEMPS`, too. There is no difference in meaning here.

src/bootchooser.c
```diff
@@ -152,16 +155,16 @@ static gboolean barebox_set_primary(RaucSlot *slot) {
 			continue;
 
 		if (s == slot) {
-			prio = prio1;
+			prio = BAREBOX_STATE_NEW_PRIORITY;
```
ejoerns: I would prefer something like `BAREBOX_STATE_PRIORITY_PRIMARY` as it better reflects the purpose of this value.

src/bootchooser.c
jremmet: I have a case were the customer want's different behaviour after an update. After Update there should be less attempts than later if the update works fine.
rauc mark-good will use BAREBOX_STATE_DEFAULT_ATTEMPS

src/bootchooser.c
jremmet: changed this

PR #112, opened 2017-04-27
config_file: enhance parsing boolean "readonly" key

OnkelUlla:
Return an error instead of interpreting a typo in the readonly key's value as
"false".

Signed-off-by: Ulrich Ölmann <u.oelmann@pengutronix.de>


OnkelUlla:
Force-pushed a new version of the branch rebased on top of master that adds the requested test.
This test fails on master and does not fail anymore after the proposed enhancement in the config file parser.

OnkelUlla:
Again force-pushed a new version of the branch which
  - incorporates the changes you asked for,
  - uses the function g_error_matches() instead of checking a handmade expression and
  - extends an already existing test case to additionally check for correctly parsing the value "false".

PR #113, opened 2017-05-08
Add handler section for manifest.raucm reference docs

artizirk:
A full day was wasted on trying to figure out how to fully customize the update process. 

This pull request will add a small section about the handler to the docs.


docs/reference.rst
```diff
@@ -165,6 +165,13 @@ A valid manifest file must have the file extension ``.raucm``.
   List of hooks enabled for this bundle.
 
 
+**[handler] section**
+
+``filename``
+  Handler script path name, relative to the bundle content. Used to fully
+  replace default update process.
```
ejoerns: The `[handler]` section also takes an `args` argument which should be described here, too.

What about adding at least a single line:

    ``args``
    Arguments to pass to the handler script, such as ``args=--verbose``

ejoerns:
Btw., in case of troubles and questions feel free to use GitHub issues or post to the mailing list (rauc@pengutronix.de). If you are lucky, you will not have to waste a whole day ;)

docs/reference.rst
artizirk: Seems like I missed args options when going through the source code. Documentation about it has been added in the next commit.

artizirk:
I went with `[handler]` mostly because I couldn't figure out why rauc wasn't copying files defined in the manifest `[file.<slot>/filename]` sections and I wanted to get at least something working before end of the day :). I will create a issue to discuss my special case further.

ejoerns:
As a final step, please add a Signed-off-by line to you commits (git commit -s will do that automagically) and squash it to a single one, then we can merge it. I've added some contribution guidelines in #118 that will also describe the requirement and background of the Signed-off-by.

artizirk:
Commits are now squashed and Signed-off-by has been added.

ejoerns:
:+1:

PR #119, opened 2017-05-11
add documentation to the release tarball

jluebbe:


Makefile.am
```diff
@@ -193,6 +193,10 @@ doc:
 	@echo
 	@echo "Build finished. The HTML pages are in $(SPHINXBUILDDIR)/html."
 
+EXTRA_DIST += docs
+dist-hook:
+	rm -rf "$(distdir)/docs/build" "$(distdir)/docs/.gitignore"
```
ejoerns: This will conflict with `dist-hook` target defined in line 9. Either squash them or use an extra target: 

``` diff
diff --git a/Makefile.am b/Makefile.am
index cc1f3d1..146ab26 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -6,7 +6,7 @@ EXTRA_DIST = $(top_srcdir)/.version
 BUILT_SOURCES = $(top_srcdir)/.version
 $(top_srcdir)/.version:
        echo $(VERSION) > $@-t && mv $@-t $@
-dist-hook:
+dist-hook: clean-doc
        echo $(VERSION) > $(distdir)/.tarball-version
 
 CODE_COVERAGE_IGNORE_PATTERN = "*-generated.c"
@@ -194,7 +194,7 @@ doc:
        @echo "Build finished. The HTML pages are in $(SPHINXBUILDDIR)/html."
 
 EXTRA_DIST += docs
-dist-hook:
+clean-doc:
        rm -rf "$(distdir)/docs/build" "$(distdir)/docs/.gitignore"
 
 AM_DISTCHECK_CONFIGURE_FLAGS = "--without-systemdunitdir"
```

Makefile.am
```diff
@@ -193,6 +193,10 @@ doc:
 	@echo
 	@echo "Build finished. The HTML pages are in $(SPHINXBUILDDIR)/html."
 
+EXTRA_DIST += docs
```
ejoerns: I'm not sure whether we want to pull in every file that lies under `docs/`. Despite we cleaned the generated html files there may be other files located in this directory in your working copy that will pollute the distribution.

ejoerns:
Updated PR and also added basic test for building doc.

The dist-hook for cleaning is not required anymore, as we only include the conf.py and the rst files.

PR #120, opened 2017-05-11
Use C's string concat instead of runtime copy

leiflm:
Just a minor simplification.

Shouldn't those defines be typed constants instead anyway?


leiflm:
Thanks for the super quick review. I know it's trivial.

I forcibly pushed an amended (signed + author change) commit to this branch.

PR #124, opened 2017-06-06
Allow passing booted slot via cmdline and context

ejoerns:
* Adds `--override-boot-slot` cmdline argument to allow faking a detected boot slot

* Moves bootslot detection to context

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -622,7 +622,7 @@ static gchar* r_status_formatter_readable(void)
 	GString *text = g_string_new(NULL);
 
 	g_string_append_printf(text, "Compatible:  %s\n", r_context()->config->system_compatible);
-	g_string_append_printf(text, "booted from: %s\n", get_bootname());
+	g_string_append_printf(text, "booted from: %s\n", r_context()->bootslot);
 
 	g_string_append(text, "slot states:\n");
 	g_hash_table_iter_init(&iter, r_context()->config->slots);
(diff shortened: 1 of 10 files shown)


src/main.c
```diff
@@ -895,14 +895,15 @@ static void cmdline_handler(int argc, char **argv)
...
-	      *handlerextra = NULL;
+	      *handlerextra = NULL, *bootslot = NULL;
 	char *cmdarg = NULL;
 	GOptionContext *context = NULL;
 	GOptionEntry entries[] = {
 		{"conf", 'c', 0, G_OPTION_ARG_FILENAME, &confpath, "config file", "FILENAME"},
 		{"cert", '\0', 0, G_OPTION_ARG_FILENAME, &certpath, "cert file", "PEMFILE"},
 		{"key", '\0', 0, G_OPTION_ARG_FILENAME, &keypath, "key file", "PEMFILE"},
 		{"mount", '\0', 0, G_OPTION_ARG_FILENAME, &mount, "mount prefix", "PATH"},
+		{"boot-slot", '\0', 0, G_OPTION_ARG_STRING, &bootslot, "overwrite booted slot", "SLOTNAME"},
```
jluebbe: Please rename the option to --override-boot-slot and use "override auto-detection of booted slot" as help message.

ejoerns:
``` diff
index 8dab205..774bdda 100644
--- a/src/main.c
+++ b/src/main.c
@@ -903,7 +903,7 @@ static void cmdline_handler(int argc, char **argv)
                {"cert", '\0', 0, G_OPTION_ARG_FILENAME, &certpath, "cert file", "PEMFILE"},
                {"key", '\0', 0, G_OPTION_ARG_FILENAME, &keypath, "key file", "PEMFILE"},
                {"mount", '\0', 0, G_OPTION_ARG_FILENAME, &mount, "mount prefix", "PATH"},
-               {"boot-slot", '\0', 0, G_OPTION_ARG_STRING, &bootslot, "overwrite booted slot", "SLOTNAME"},
+               {"override-boot-slot", '\0', 0, G_OPTION_ARG_STRING, &bootslot, "override auto-detection of booted slot", "SLOTNAME"},
                {"handler-args", '\0', 0, G_OPTION_ARG_STRING, &handlerextra, "extra handler arguments", "ARGS"},
                {"debug", 'd', 0, G_OPTION_ARG_NONE, &debug, "enable debug output", NULL},
                {"version", '\0', 0, G_OPTION_ARG_NONE, &version, "display version", NULL},
```

PR #127, opened 2017-06-08
u-boot multislot configurations

hundeboll:
On a system using three slots (i.e. A, B, and R/recovery), the boot
order would be "A B R". When updating B from A, the bootorder is changed
to "B A", when it should be changed to "B A R".

Fix this by changing how the boot order environment string is
generated, so that:

  1. non-active slots are appended
  2. active slots are prepended
  3. the updated slot is prepended as the final step


ejoerns:
@hundeboll do you have any updates on this topic? Would you like to adapt the PR to get it mainline?

mnhu:
@ejoerns I will get back to this within the next few weeks. Quite busy at the moment, so feel free to pick it up :)

hundeboll:
@ejoerns I have updated the PR to use `fw_printenv` in both `src/` and `test/`. Please have a look...

src/bootchooser.c
```diff
@@ -284,6 +286,57 @@ static gboolean grub_set_primary(RaucSlot *slot) {
...
+
+	res = g_subprocess_get_if_exited(sub);
+	if (!res) {
+		g_warning(UBOOT_FWGETENV_NAME " did not exit normally");
+		goto out;
+	}
+
+	ret = g_subprocess_get_exit_status(sub);
+	if (ret != 0) {
+		g_warning(UBOOT_FWGETENV_NAME " failed with exit code: %i", ret);
```
ejoerns: We need to have a `res = FALSE` here, otherwise `uboot_env_get()` would return TRUE, as previously set by `res = g_subprocess_get_if_exited(sub);`

src/bootchooser.c
```diff
@@ -284,6 +286,57 @@ static gboolean grub_set_primary(RaucSlot *slot) {
 	return res;
 }
 
+static gboolean uboot_env_get(const gchar *key, GString **value) {
```
ejoerns: Might be a bit inconsistent here to have `GString **` in get and `gchar *` in set function. As we actually pass a GString to `uboot_env_set()`, anyway, it might be the simplest solution to change its arg to `GString *`.

mnhu: Are we sure we want to change the `value` argument to `GString *`? Doesn't that require instantiation and freeing of temporary `GString` objects for every place where `uboot_env_set()` is called with a string literal as value? (e.g. `res = uboot_env_set(key, "3");` in line 421.)

To gain consistency between `uboot_env_{set,get}()` I can change the value argument of `uboot_env_get()` to `char *` and add a length argument, e.g `static gboolean uboot_env_get(const char *key, char *value, size_t len)` and return FALSE is the value buffer is too small to hold the env-string?

src/bootchooser.c
mnhu: Pushed an update...

src/bootchooser.c
hundeboll: Ping ?

ejoerns: Sorry for the delay. I wanted to test this a bit on hardware but do not have a board with u-boot present atm. Maybe I have to drop the hardware test and trust code reading and testing. But I would like to have @jluebbe's eyes on that before merging, as this touches critical infrastructure.

Well, calling `uboot_env_set` with fixed strings is more ore less a simplification that I am not yet really happy with, yet. Maybe one does not want 3 but only 2 or only 1 attempt for newly activated slots. Thus in the future this could become configurable.
Concerning your remarks for the return type of uboot_env_get I would not like to instancaite and pass buffers from outside. We could do the same as with the `GString`, returning a newly allocated `gchar*`. This is what all other functions do, too (sometimes as return value, sometimes as point-pointer argument).
But in general, I would not see this discussion as a blocker. There is no public API designed, thus we are free to refactor the code and methods if that should be required.

mnhu:
I've updated the commit title. Please re-check.

mnhu:
Anything I can do to get this merged ?

mnhu:
Ping

src/bootchooser.c
```diff
@@ -284,6 +286,58 @@ static gboolean grub_set_primary(RaucSlot *slot) {
...
+	GError *error = NULL;
+	GBytes *stdout_buf = NULL;
+	const char *data;
+	gsize offset;
+	gsize size;
+	gboolean res = FALSE;
+	gint ret;
+
+	g_assert_nonnull(key);
+
```
jluebbe: Check that value is not null as well.

src/bootchooser.c
```diff
@@ -284,6 +286,58 @@ static gboolean grub_set_primary(RaucSlot *slot) {
...
+	}
+
+	ret = g_subprocess_get_exit_status(sub);
+	if (ret != 0) {
+		g_warning(UBOOT_FWGETENV_NAME " failed with exit code: %i", ret);
+		res = FALSE;
+		goto out;
+	}
+
+	offset = strlen(key) + 1;
```
jluebbe: Please add a comment why this fixed indexing (instead of string parsing) works.

src/bootchooser.c
```diff
@@ -333,26 +387,33 @@ static gboolean uboot_set_state(RaucSlot *slot, gboolean good) {
...
 
-	g_string_append(order, slot->bootname);
+	g_string_append(order_new, slot->bootname);
 
-	/* Iterate over class members */
-	slots = g_hash_table_get_values(r_context()->config->slots);
-	for (GList *l = slots; l != NULL; l = l->next) {
-		RaucSlot *s = l->data;
-		if (s == slot)
+	res = uboot_env_get("BOOT_ORDER", &order_current);
```
jluebbe: Currently, we don't require that BOOT_ORDER already exists in the env. Just use the list of bootable slots as a default for order_current.

ejoerns:
Just saw you updated the commits, we will have a look at it. It's better to leave a note after pushing updates so we are notified as GitHub ist very silent about force-pushes (which of course has its advantages and disadvantages.. ;) )

PR #130, opened 2017-06-23
D-Bus Testing

ejoerns:
This adds tests for installing rauc bundle via D-Bus.

The series includes the patch from #129 in order to make it succeed.


ejoerns:
Rebase it and added SoB.

PR #132, opened 2017-06-27
docs: using: adjust description of mark-good/bad

OnkelUlla:
It is not the boot that gets marked, it is the slot.

Signed-off-by: Ulrich Ölmann <u.oelmann@pengutronix.de>

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -80,29 +80,30 @@ default it will print a human readable representation of your system.
 Alternatively you can obtain a shell-parsable description, or a JSON
 representation of the system status.
 
-Marking Boot as Successful / Failed
-------------------------------------
+React to a Successfully Booted System/Failed Boot
+-------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/using.rst
```diff
@@ -80,29 +80,30 @@ default it will print a human readable representation of your system.
...
-------------------------------------
+React to a Successfully Booted System/Failed Boot
+-------------------------------------------------
 
 Normally, the full system update chain is not completed before being sure the
 newly installed system runs without any errors.
 As the definition and detection of a `successful` operation is really
-system-depended, RAUC provides a command for marking a boot as either
-successful or failed.
+system-dependend, RAUC provides commands to preserve a slot as being the
```
jluebbe: dependend is spelled dependent

OnkelUlla:
Good catch!
I just corrected the spelling and force-pushed a new version.

PR #134, opened 2017-07-04
activate-installed & mark-active

OnkelUlla:
Today RAUC bundles are copied to the local filesystem, then installed via `rauc install <PATH/TO/BUNDLE.raucb>` and finally the system is rebooted into the new software.

Additionally support for the following use case is needed: after installation the freshly installed slot is not activated automatically, but only after an explict decision made by a user. Before this activation reboots shall only start the old slot. This comes from practical demands of a service technician who wants to push new software to some dozens of systems at the same time, but take them into operation individually and under supervision.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,6 +41,7 @@ librauc_la_SOURCES = \
 	src/context.c \
 	src/install.c \
 	src/manifest.c \
+	src/mark.c \
 	src/mount.c \
 	src/service.c \
 	src/signature.c \
@@ -53,6 +54,7 @@ librauc_la_SOURCES = \
 	include/context.h \
 	include/install.h \
 	include/manifest.h \
+	include/mark.h \
 	include/mount.h \
 	include/service.h \
 	include/signature.h \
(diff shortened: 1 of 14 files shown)


OnkelUlla:
Modified the topmost commit to adapt the tests to this new situation as well by starting the RAUC D-Bus service on the system or the session bus according to whether they are running under user mode linux or conventionally.

OnkelUlla:
Explicitely use a newly created session bus dedicated only to the tests when being run conventionally.

src/context.c
```diff
@@ -183,6 +310,11 @@ void r_context_prepare(void) {
 	}
 
 	context->pending = FALSE;
+
+	if (!determine_slot_states(&error)) {
+		g_warning("Could not determine slot states: %s", error->message);
+		g_clear_error(&error);
+	}
```
OnkelUlla: You are completely right: a warning is not sufficient here.
Deeper analysis with today's experience suggests going a different way. I will push a refactored series soon.

PR #135, opened 2017-07-05
src/update_handler: allow uncompressd tarballs

hundeboll:
Since update files are compressed in the bundle with squashfs, there is no need to
require them to be compressed prior to being bundles.

This commit removes the dot from the tar entries in the updatepairs
array, so that *.tar files are handled just like e.g. .tar.gz files.


jluebbe:
In some cases, using .tar.xz results in higher compression ratios, because it uses independent blocks of a fixed size to allow random access. Nevertheless, this is a useful change. Thanks!

PR #139, opened 2017-08-07
Fixes for error handling and printout

ejoerns:


src/install.c
```diff
@@ -1135,7 +1135,16 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error) {
 	res = TRUE;
 
 umount:
-	umount_bundle(mountpoint, NULL);
+	{
+		/* Do not cover previous error but report if no other occurred, yet */
+		GError *iierror = NULL;
+		if (!umount_bundle(mountpoint, &iierror)) {
+			if (!ierror)
+				ierror = iierror;
```
jluebbe: ierror is not used below this point. Did you want to propagate it?
This patch is also missing a SoB line.

ejoerns: Yes, this does not seem to be ready to merge, yet. Nevertheless, maybe there are better options for solving this than this handling (saving mount state in variable, skipping unmount if mount failed, etc.). Thus I will push the series again with this commit omitted for now.

ejoerns:
Updated

PR #143, opened 2017-08-08
Fix and enhance command line handling

ejoerns:
Some misleading argument number parsings were fixed and commands now can print a summary giving a brief description of what they will do.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -875,6 +875,7 @@ typedef struct {
 	const RaucCommandType type;
 	const gchar* name;
 	const gchar* usage;
+	const gchar* summary;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -221,7 +221,7 @@ static gboolean bundle_start(int argc, char **argv)
 		goto out;
 	}
 
-	if (argc != 4) {
+	if (argc < 4) {
```
jluebbe: We should still reject additional arguments (below as well).

ejoerns:
Updated

src/main.c
ejoerns: You're right.

PR #150, opened 2017-08-22
src/update_handler: add tar_to_vfat handler

hundeboll:
Some boards boots directly from a FAT partition, in which case it is
convenient to update the boot partition from a tar ball (i.e. Yocto/poky
doesn't include a FAT image type).

Add support for tar* -> vfat filesystems by formatting the partition
with `mkfs.vfat` and calling the generic `untar_image` function
afterwards.


ejoerns:
Good work! Having a vfat slot type will increase the potential application areas of RAUC. Also great that you added the tests by yourself, this would have been the first item on my list to mention.
As you introduce a new slot type, at least quite a few more things need to be done for the current RAUC version:

``` diff
diff --git a/docs/integration.rst b/docs/integration.rst
index 64b926c..792d2a0 100644
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -118,6 +118,8 @@ A list of slot storage types currently supported by RAUC:
 +----------+-------------------------------------------------------------------+-------------+
 | ubifs    | An UBI volume containing an UBIFS in NAND.                        |     x       |
 +----------+-------------------------------------------------------------------+-------------+
+| vfat     | A block device holding a vfat filesystem.                         |     x       |
++----------+-------------------------------------------------------------------+-------------+
 
 Grouping Slots
 ^^^^^^^^^^^^^^
diff --git a/docs/reference.rst b/docs/reference.rst
index 73abf6c..4ec1f12 100644
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -137,7 +137,7 @@ set of slots.
 
 ``type``
   The type describing the slot. Currently supported values are ``raw``,
-  ``nand``, ``ubivol``, ``ubifs``, ``ext4``.
+  ``nand``, ``ubivol``, ``ubifs``, ``ext4``, ``vfat``.
 
 ``bootname``
diff --git a/src/config_file.c b/src/config_file.c
index 35bdef8..64c0db2 100644
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -40,6 +40,7 @@ RaucSlotType supported_slot_types[] = {
        {"ubifs", TRUE},
        {"ubivol", FALSE},
        {"nand", FALSE},
+       {"vfat", TRUE},
        {}
 };
```
 
Maybe it's also worth thinking about an img->vfat handler? But I would see this as an 'extra'.

PR #153, opened 2017-08-24
Add basic bundle resigning support

ejoerns:
This adds the command `rauc resign` which allows to replace the bundle signature without modifying the bundle content.

Based on PR #152

PR #154, opened 2017-08-24
Support for extracting and dumping certificate and trust chain info

ejoerns:
This series performs some refactoring of bundle handling functions. It introduces a `RaucBundle` structure to support passing configuration and meta information to bundle handling functions.
These changes form the base for the also included extraction of certificate and trust chain information obtained during bunde verification.
`rauc info` will now print trust chain information if signature verification succeeded.
A new option for `rauc info` ( ``--dump-cert``) lets you dump useful certificate information about a bundle.

Series is based on PR #152 and #153

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -445,13 +445,25 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 	}
 
 	if (verify) {
+		CMS_ContentInfo *cms = NULL;
+		X509_STORE *store = NULL;
+
 		g_message("Verifying bundle... ");
 		/* the squashfs image size is in offset */
-		res = cms_verify_file(bundlename, sig, offset, NULL, NULL, &ierror);
+		res = cms_verify_file(bundlename, sig, offset, &cms, &store, &ierror);
 		if (!res) {
 			g_propagate_error(error, ierror);
 			goto out;
 		}
+
+		res = cms_get_cert_chain(cms, store, &ibundle->verified_chain, &ierror);
+		if (!res) {
+			g_propagate_error(error, ierror);
+			goto out;
+		}
+
+		X509_STORE_free(store);
+		CMS_ContentInfo_free(cms);
 	}
 
 	if (bundle)
@@ -542,4 +554,6 @@ out:
 void free_bundle(RaucBundle *bundle) {
 
 	g_free(bundle->path);
+	if (bundle->verified_chain)
+		sk_X509_pop_free(bundle->verified_chain, X509_free);
 }
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -324,15 +324,15 @@ out:
 gboolean resign_bundle(const gchar *inpath, const gchar *outpath, GError **error) {
 	GError *ierror = NULL;
 	gboolean res = FALSE;
-	gsize size;
+	RaucBundle *bundle;
 
-	res = check_bundle(inpath, &size, TRUE, &ierror);
+	res = check_bundle(inpath, &bundle, TRUE, &ierror);
 	if (!res) {
 		g_propagate_error(error, ierror);
 		goto out;
 	}
 
-	res = truncate_bundle(inpath, outpath, size, &ierror);
+	res = truncate_bundle(inpath, outpath, bundle->size, &ierror);
 	if (!res) {
 		g_propagate_error(error, ierror);
 		goto out;
@@ -349,7 +349,7 @@ out:
 	return res;
 }
 
-gboolean check_bundle(const gchar *bundlename, gsize *size, gboolean verify, GError **error) {
+gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean verify, GError **error) {
 	GError *ierror = NULL;
 	GBytes *sig = NULL;
 	GFile *bundlefile = NULL;
@@ -357,6 +357,11 @@ gboolean check_bundle(const gchar *bundlename, gsize *size, gboolean verify, GEr
 	guint64 sigsize;
 	goffset offset;
 	gboolean res = FALSE;
+	RaucBundle *ibundle = g_new0(RaucBundle, 1);
+
+	g_return_val_if_fail (bundle == NULL || *bundle == NULL, FALSE);
+
+	ibundle->path = g_strdup(bundlename);
 
 	r_context_begin_step("check_bundle", "Checking bundle", verify);
 
@@ -422,8 +427,7 @@ gboolean check_bundle(const gchar *bundlename, gsize *size, gboolean verify, GEr
 
 	offset -= sigsize;
 
-	if (size)
-		*size = offset;
+	ibundle->size = offset;
 
 	res = g_seekable_seek(G_SEEKABLE(bundlestream),
 			      offset, G_SEEK_SET, NULL, &ierror);
@@ -455,8 +459,13 @@ gboolean check_bundle(const gchar *bundlename, gsize *size, gboolean verify, GEr
 		}
 	}
 
+	if (bundle)
+		*bundle = ibundle;
+
 	res = TRUE;
 out:
+	if (!bundle)
+		free_bundle(ibundle);
 	g_clear_object(&bundlestream);
 	g_clear_object(&bundlefile);
 	g_clear_pointer(&sig, g_bytes_unref);
@@ -519,16 +528,16 @@ out:
 
 gboolean mount_bundle(const gchar *bundlename, const gchar *mountpoint, gboolean verify, GError **error) {
 	GError *ierror = NULL;
-	gsize size;
+	RaucBundle *bundle = NULL;
 	gboolean res = FALSE;
 
-	res = check_bundle(bundlename, &size, verify, &ierror);
+	res = check_bundle(bundlename, &bundle, verify, &ierror);
 	if (!res) {
 		g_propagate_error(error, ierror);
 		goto out;
 	}
 
-	res = r_mount_loop(bundlename, mountpoint, size, &ierror);
+	res = r_mount_loop(bundlename, mountpoint, bundle->size, &ierror);
 	if (!res) {
 		g_propagate_error(error, ierror);
 		goto out;
@@ -552,4 +561,9 @@ gboolean umount_bundle(const gchar *bundlename, GError **error) {
 	res = TRUE;
 out:
 	return res;
+}
+
+void free_bundle(RaucBundle *bundle) {
+
+	g_free(bundle->path);
 }
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -321,18 +321,13 @@ out:
 	return res;
 }
 
-gboolean resign_bundle(const gchar *inpath, const gchar *outpath, GError **error) {
+gboolean resign_bundle(RaucBundle *bundle, const gchar *outpath, GError **error) {
 	GError *ierror = NULL;
 	gboolean res = FALSE;
-	RaucBundle *bundle;
 
-	res = check_bundle(inpath, &bundle, TRUE, &ierror);
-	if (!res) {
-		g_propagate_error(error, ierror);
-		goto out;
-	}
+	g_return_val_if_fail(bundle != NULL, FALSE);
 
-	res = truncate_bundle(inpath, outpath, bundle->size, &ierror);
+	res = truncate_bundle(bundle->path, outpath, bundle->size, &ierror);
 	if (!res) {
 		g_propagate_error(error, ierror);
 		goto out;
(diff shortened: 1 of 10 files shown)


src/bundle.c
```diff
@@ -347,15 +444,24 @@ gboolean check_bundle(const gchar *bundlename, gsize *size, gboolean verify, GEr
...
 			g_propagate_error(error, ierror);
 			goto out;
 		}
+
+		ibundle->signer = cms_get_signer_info(cms, store,
+				&ibundle->verified_chain);
 	}
 
+	if (bundle)
+		*bundle = ibundle;
```
mnhu: Shouldn't `ibundle` (and `ibundle->signer`) be free'd in case `bundle` evaluates to false?

src/service.c
```diff
@@ -97,6 +97,7 @@ static gboolean r_on_handle_info(RInstaller *interface,
 	gchar* bundledir = NULL;
 	gchar* manifestpath = NULL;
 	RaucManifest *manifest = NULL;
+	RaucBundle *bundle = NULL;
```
mnhu: Shouldn't `bundle` be freed after calling `check_bundle()`?

src/main.c
```diff
@@ -600,6 +658,12 @@ static gboolean info_start(int argc, char **argv)
 
 	g_print("%s\n", formatter(manifest));
 
+	if (info_dumpchain)
+		g_print("%s\n", cert_chain_to_text(bundle->verified_chain));
```
mnhu: Shouldn't the pointer returned from `cert_chain_to_text()` bee freed?

src/main.c
```diff
@@ -600,6 +658,12 @@ static gboolean info_start(int argc, char **argv)
...
+
+	if (info_dumpcert)
+		g_print("%s\n", cert_to_text(bundle->signer));
```
mnhu: Shouldn't the pointer returned from `cert_to_text()` bee freed?

src/signature.c
```diff
@@ -170,12 +171,70 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		X509_verify_cert(cert_ctx);
+
+		*verified_chain = X509_STORE_CTX_get1_chain(cert_ctx);
+
+		g_debug("Got %d chain elements", sk_X509_num(*verified_chain));
+	}
+
+out:
+
+	return signers ? sk_X509_value(signers, 0) : NULL;
```
mnhu: `sk_X509_value()` returns a member of `signers` but doesn't free it?

ejoerns:
Yeah, good points. Some relate to the prior PR's which were part of this patch stack, but I added a fix commit on top of the stack for those.
At least the bundle cleanup I had on my list, but somehow I must have dropped this from my mental stack...

src/signature.c
```diff
@@ -230,12 +231,20 @@ gboolean cms_verify(GBytes *content, GBytes *sig, GError **error) {
...
+
+
 	res = TRUE;
 out:
 	ERR_print_errors_fp(stdout);
 	BIO_free_all(incontent);
 	BIO_free_all(insig);
-	X509_STORE_free(store);
+	if (!store)
+		X509_STORE_free(istore);
```
jluebbe: Do we need to free icms as well if we don't pass it along?

src/bundle.c
jluebbe: This is handled below in the updated version.

src/bundle.c
```diff
@@ -553,3 +560,11 @@ gboolean umount_bundle(const gchar *bundlename, GError **error) {
 out:
 	return res;
 }
+
+void free_bundle(RaucBundle *bundle) {
+
+	g_free(bundle->path);
+	g_clear_pointer(&bundle->signer, X509_free);
```
jluebbe: There are no signer or verified_chain fields at this point in the patch series.

src/bundle.c
```diff
@@ -349,14 +349,17 @@ gboolean resign_bundle(const gchar *inpath, const gchar *outpath, GError **error
...
 	GError *ierror = NULL;
 	GBytes *sig = NULL;
 	GFile *bundlefile = NULL;
 	GFileInputStream *bundlestream = NULL;
 	guint64 sigsize;
 	goffset offset;
 	gboolean res = FALSE;
+	RaucBundle *ibundle = g_new0(RaucBundle, 1);
+
+	ibundle->path = g_strdup(bundlename);
```
jluebbe: We should check that bundle points to a NULL pointer or is NULL itself. Otherwise we'd leak a RaucBundle later when passing the new RaucBundle.

test/bundle.c
```diff
@@ -92,18 +92,23 @@ static void bundle_test_create_extract(BundleFixture *fixture,
...
 
 	outputdir = g_build_filename(fixture->tmpdir, "output", NULL);
 	g_assert_nonnull(outputdir);
 
-	g_assert_true(extract_bundle(fixture->bundlename, outputdir, TRUE, NULL));
+	g_assert_true(check_bundle(fixture->bundlename, &bundle, TRUE, NULL));
+	g_assert_nonnull(bundle);
+
+	g_assert_true(extract_bundle(bundle, outputdir, NULL));
 	g_assert_true(verify_manifest(outputdir, NULL, FALSE, NULL));
```
jluebbe: Free the RaucBundle (in the other tests as well).

src/bundle.c
```diff
@@ -443,13 +443,19 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
...
 		/* the squashfs image size is in offset */
-		res = cms_verify_file(bundlename, sig, offset, NULL, NULL, &ierror);
+		res = cms_verify_file(bundlename, sig, offset, &cms, &store, &ierror);
 		if (!res) {
 			g_propagate_error(error, ierror);
 			goto out;
 		}
+
+		ibundle->signer = cms_get_signer_info(cms, store,
+				&ibundle->verified_chain);
```
jluebbe: Do we need to free cms and store here?

src/signature.c
```diff
@@ -171,6 +171,69 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		g_string_append_printf(text, "   i:%s\n", buf);
+	}
+
+	return g_string_free(text, FALSE);
+}
+
+X509 *cms_get_signer_info(CMS_ContentInfo *cms, X509_STORE *store, STACK_OF(X509) **verified_chain) {
+	STACK_OF(X509) *signers = NULL;
+	X509 *signer = NULL;
+
```
jluebbe: Add asserts for the input pointers.

src/signature.c
```diff
@@ -171,6 +171,69 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+	if (verified_chain) {
+		X509_STORE_CTX *cert_ctx;
+
+		cert_ctx = X509_STORE_CTX_new();
+		if (cert_ctx == NULL)
+			goto out;
+		if (!X509_STORE_CTX_init(cert_ctx, store, sk_X509_value(signers, 0), NULL))
+			goto out;
+		X509_verify_cert(cert_ctx);
```
jluebbe: Check that this returns 1, otherwise X509_STORE_CTX_get1_chain() may return a bad chain.

src/signature.c
```diff
@@ -171,6 +171,69 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+	X509 *signer = NULL;
+
+	signers = CMS_get0_signers(cms);
+
+	g_debug("Got %d signers:", sk_X509_num(signers));
+
+	if (verified_chain) {
+		X509_STORE_CTX *cert_ctx;
+
+		cert_ctx = X509_STORE_CTX_new();
```
jluebbe: We need to free cert_ctx as well.

src/signature.c
```diff
@@ -171,6 +171,69 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		if (cert_ctx == NULL)
+			goto out;
+		if (!X509_STORE_CTX_init(cert_ctx, store, sk_X509_value(signers, 0), NULL))
```
jluebbe: If I remember correctly, the signature may contain additional certificates for completing the chain. Do we need to pass them here via the chain argument?

src/signature.c
ejoerns: Added.

src/signature.c
ejoerns: Fixed.

src/signature.c
ejoerns: It is possible to include the certs in the signature, but we don't do that, yet. When required, we should implement it both for creation and verification in one series.

src/signature.c
ejoerns: Checked now. Also added GError handling to `cms_get_signer_info()` function.

src/signature.c
ejoerns: We did not free former `cms` variable, yet. Added a separate fix commit for this and adapted further patches hande `icms` properly.

src/bundle.c
ejoerns: Added `g_return_val_if_fail()` check for this.

src/bundle.c
ejoerns: Fixed by reordering patches.

test/bundle.c
ejoerns: Added missing `free_bundle()` calls

src/bundle.c
ejoerns: Seems as if we should, yes.

ejoerns:
Pushed an updated version of the patch series including the proposed fixes. Initial 2 commits are new, other changes are squashed into the existing commits.

ejoerns:
A Travis test fail because of a double free made me aware that returning both the entire certificate chain and the signer certificate itself is redundant. `cms_get_signer_info` was renamed to `cms_get_cert_chain` and now only returns gboolean instead of the redundant signer cert.

Being more strict in input verification also revealed some uninitialized pointers in tests that are fixed now.

src/service.c
jluebbe: This is not yet fixed.

src/signature.c
```diff
@@ -157,6 +157,118 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+
+gchar* print_cert_chain(STACK_OF(X509) *verified_chain) {
+	GString *text = g_string_new(NULL);
+	char buf[BUFSIZ];
+
+	g_string_append(text, "Certificate Chain:\n");
+	for (int i = 0; i < sk_X509_num(verified_chain); i++) {
+		X509_NAME_oneline(X509_get_subject_name(sk_X509_value(verified_chain, i)),
+				buf, sizeof buf);
+		g_string_append_printf(text, "%2d s:%s\n", i, buf);
```
jluebbe: use "Subject" instead of "s"

src/signature.c
```diff
@@ -157,6 +157,118 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		X509_NAME_oneline(X509_get_issuer_name(sk_X509_value(verified_chain, i)),
+				buf, sizeof buf);
+		g_string_append_printf(text, "   i:%s\n", buf);
```
jluebbe: Use "Issuer" instead of "i"

src/signature.c
```diff
@@ -157,6 +157,118 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+
+	/* The first element in the chain must be the signer certificate */
+	g_assert(sk_X509_value(signers, 0) == sk_X509_value(*verified_chain, 0));
+
+	g_debug("Got %d chain elements", sk_X509_num(*verified_chain));
+
+	res = TRUE;
+out:
+	if (cert_ctx)
+		X509_STORE_CTX_free(cert_ctx);
```
jluebbe: Remove the X509_STORE_CTX_free calls above, otherwise we will try to free it twice.

ejoerns: Oh, this seems to be a remnant from the prior version where the context was still local to the now removed `if (verified_chain)...`. Will fix it.

src/signature.c
ejoerns: Ok, got that change in my follow-up PR #155 but I agree we can do it here already

ejoerns:
Well, these are my latest changes now:

``` diff
diff --git a/src/service.c b/src/service.c
index b44c029..0e6ce38 100644
--- a/src/service.c
+++ b/src/service.c
@@ -161,6 +161,8 @@ out:
                                                      "rauc info error");
        }
 
+       g_clear_pointer(&bundle, free_bundle);
+
        return TRUE;
 }
 
diff --git a/src/signature.c b/src/signature.c
index 2daa47c..0594aba 100644
--- a/src/signature.c
+++ b/src/signature.c
@@ -182,10 +182,10 @@ gchar* print_cert_chain(STACK_OF(X509) *verified_chain) {
        for (int i = 0; i < sk_X509_num(verified_chain); i++) {
                X509_NAME_oneline(X509_get_subject_name(sk_X509_value(verified_chain, i)),
                                buf, sizeof buf);
-               g_string_append_printf(text, "%2d s:%s\n", i, buf);
+               g_string_append_printf(text, "%2d Subject: %s\n", i, buf);
                X509_NAME_oneline(X509_get_issuer_name(sk_X509_value(verified_chain, i)),
                                buf, sizeof buf);
-               g_string_append_printf(text, "   i:%s\n", buf);
+               g_string_append_printf(text, "   Issuer: %s\n", buf);
        }
 
        return g_string_free(text, FALSE);
@@ -238,7 +238,6 @@ gboolean cms_get_cert_chain(CMS_ContentInfo *cms, X509_STORE *store, STACK_OF(X5
                                R_SIGNATURE_ERROR,
                                R_SIGNATURE_ERROR_X509_CTX_INIT,
                                "Failed to init new X509 CTX store");
-               X509_STORE_CTX_free(cert_ctx);
                goto out;
        }
 
@@ -248,7 +247,6 @@ gboolean cms_get_cert_chain(CMS_ContentInfo *cms, X509_STORE *store, STACK_OF(X5
                                R_SIGNATURE_ERROR,
                                R_SIGNATURE_ERROR_VERIFY_CERT,
                                "Failed to verify X509 cert");
-               X509_STORE_CTX_free(cert_ctx);
                goto out;
        }
 

```

PR #155, opened 2017-08-30
Support for extracting and handling SPKI hashes

ejoerns:
This series adds functions for extracting public key hashes of certificates involved in signing and the trust chain.

This potentially allows to perform decisions about key chains to use on the target depending on which certificate was used to sign the bundle. A common use case for this is to switch from a development-key-signed bundle to one signed with a release key without changing the bundle content.

Based on PR #154

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -6,9 +6,15 @@
 
 
 #include "config_file.h"
+#include "bundle.h"
 
 typedef void (*progress_callback) (gint percentage, const gchar *message,
 				   gint nesting_depth);
+
+typedef struct {
+	/* The bundle currently mounted by RAUC */
+	RaucBundle *mounted_bundle;
+} RContextInstallationInfo;
 
 typedef struct {
 	/* a busy context must not be reconfigured */
@@ -36,6 +42,9 @@ typedef struct {
 	gchar *handlerextra;
 	/* ignore compatible check */
 	gboolean ignore_compatible;
+
+	/* for storing installation runtime informations */
+	RContextInstallationInfo *install_info;
 } RaucContext;
 
 typedef struct {
--- a/src/signature.c
+++ b/src/signature.c
@@ -5,6 +5,7 @@
 #include <openssl/err.h>
 #include <openssl/evp.h>
 #include <openssl/pem.h>
+#include <openssl/crypto.h>
 
 #include "context.h"
 #include "signature.h"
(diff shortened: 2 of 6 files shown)


src/signature.c
```diff
@@ -157,6 +158,63 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		BIO *bio;
+		unsigned char *ibuf = NULL;
+		unsigned char *buf2;
+		unsigned int len = 0;
+		unsigned int n = 0;
+		unsigned char md[SHA256_DIGEST_LENGTH];
+
+		bio = BIO_new(BIO_s_mem());
+
+		len = i2d_X509_PUBKEY(X509_get_X509_PUBKEY(cert), NULL);
```
jluebbe: Can this fail or return 0?

src/signature.c
```diff
@@ -157,6 +158,63 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		buf2 = ibuf = g_malloc(len);
+		if (ibuf) i2d_X509_PUBKEY(X509_get_X509_PUBKEY(cert), &buf2);
```
jluebbe: g_malloc only returns NULL if len is 0. We shouldn't even try in that case.

src/signature.c
```diff
@@ -157,6 +158,63 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		unsigned char *ibuf = NULL;
+		unsigned char *buf2;
+		unsigned int len = 0;
+		unsigned int n = 0;
+		unsigned char md[SHA256_DIGEST_LENGTH];
+
+		bio = BIO_new(BIO_s_mem());
+
+		len = i2d_X509_PUBKEY(X509_get_X509_PUBKEY(cert), NULL);
+		buf2 = ibuf = g_malloc(len);
```
jluebbe: Please use a more descriptive name instead of buf2.

src/signature.c
```diff
@@ -157,6 +158,63 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		if (ibuf) i2d_X509_PUBKEY(X509_get_X509_PUBKEY(cert), &buf2);
+
+		g_assert(buf2 - ibuf == len);
+
+		if (!EVP_Digest(ibuf, len, md, &n, EVP_sha256(), NULL)) {
+			g_warning("Error in EVP_Digest\n");
+			goto out;
+		}
+
```
jluebbe: g_assert that the n matches what we expect.

src/signature.c
```diff
@@ -157,6 +158,63 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+		for (int j = 0; j < (int)n; j++) {
```
jluebbe: Maybe use g_string_sized_new(), g_string_append_printf() and g_string_truncate instead of a BIO and the ?: ?

src/signature.c
```diff
@@ -157,6 +158,63 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, G
...
+
+		BIO_set_close(bio, BIO_NOCLOSE); /* So BIO_free() leaves BUF_MEM alone */
+		BIO_free(bio);
+out:
+		g_free(ibuf);
+		return data;
+}
+
+gchar** get_pubkey_hashes(STACK_OF(X509) *verified_chain) {
+	GPtrArray *hashes = g_ptr_array_new_full(1, g_free);
```
jluebbe: We probably have more than one hash in most cases. 3 certs + 1 NULL pointer seems plausible.

include/context.h
```diff
@@ -6,11 +6,17 @@
...
 #include "config_file.h"
+#include "bundle.h"
 
 typedef void (*progress_callback) (gint percentage, const gchar *message,
 				   gint nesting_depth);
 
 typedef struct {
+	/* The bundle currently mounted by RAUC */
+	RaucBundle *mounted_bundle;
+} RContextRuntimeInfo;
```
jluebbe: Maybe call it RContextInstallationInfo and zero it at clean it at the start of each installation?

src/update_handler.c
```diff
@@ -382,6 +384,20 @@ static gboolean run_slot_hook(const gchar *hook_name, const gchar *hook_cmd, Rau
...
+
+		hashes = get_pubkey_hashes(r_context()->runtime->mounted_bundle->verified_chain);
+
+		while (*hashes != NULL) {
+			g_string_append_printf(cert_pkey_hashes, "%s", (gchar*) *hashes);
+			g_free(*hashes);
+			g_string_append_printf(cert_pkey_hashes, "%c", *++hashes ? ' ' : '\0');
+		}
+
+		g_subprocess_launcher_setenv(launcher, "RAUC_BUNDLE_PKEY_HASHES", g_string_free(cert_pkey_hashes, FALSE), FALSE);
```
jluebbe: Please use RAUC_BUNDLE_SPKI_HASHES instead.

src/update_handler.c
```diff
@@ -382,6 +384,20 @@ static gboolean run_slot_hook(const gchar *hook_name, const gchar *hook_cmd, Rau
...
+	if (r_context()->runtime->mounted_bundle) {
+		gchar **hashes = NULL;
+
+		hashes = get_pubkey_hashes(r_context()->runtime->mounted_bundle->verified_chain);
+
+		while (*hashes != NULL) {
+			g_string_append_printf(cert_pkey_hashes, "%s", (gchar*) *hashes);
+			g_free(*hashes);
+			g_string_append_printf(cert_pkey_hashes, "%c", *++hashes ? ' ' : '\0');
+		}
```
jluebbe: Use g_strjoinv() instead. By switching get_pubkey_hashes to return the g_ptr_array, you could simplify memory management here.

src/update_handler.c
ejoerns: Ok.

src/signature.c
ejoerns: Ok, will replace it by `der_buf` and `tmp_buf` and add a note why we need them

src/signature.c
ejoerns: Yes, `len < 0` indicates an error

src/signature.c
ejoerns: Ok, will test for `len <= 0` before and remove the ibuf check

src/signature.c
ejoerns: Ok.

src/signature.c
ejoerns: `g_string_sized_new()` sounds good. Will save some LOC.

src/signature.c
ejoerns: Ok, I will init it to 4.

src/update_handler.c
ejoerns: `g_strjoinv()` popped from my mental stack, yes we should use it! I'm not sure about the second point, yet.

include/context.h
ejoerns: `RContextInstallationInfo` sounds ok. Not sure about the second point, yet.

ejoerns:
Rebased and updated PR.

PR #158, opened 2017-09-02
Miscellaneous

OnkelUlla:
Again some small patches in the context of RAUC's destroy functions.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -552,6 +552,7 @@ out:
 }
 
 void free_bundle(RaucBundle *bundle) {
+	g_return_if_fail(bundle);
 
 	g_free(bundle->path);
 	if (bundle->verified_chain)
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -556,4 +556,5 @@ void free_bundle(RaucBundle *bundle) {
 	g_free(bundle->path);
 	if (bundle->verified_chain)
 		sk_X509_pop_free(bundle->verified_chain, X509_free);
+	g_free(bundle);
 }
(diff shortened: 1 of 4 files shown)


src/bundle.c
```diff
@@ -552,9 +552,10 @@ gboolean umount_bundle(RaucBundle *bundle, GError **error) {
 }
 
 void free_bundle(RaucBundle *bundle) {
-
-	g_free(bundle->path);
-	if (bundle->verified_chain)
-		sk_X509_pop_free(bundle->verified_chain, X509_free);
-	g_free(bundle);
+	if (G_LIKELY(bundle)) {
```
jluebbe: Can we use the inverse? if (!bundle) return? That would avoid indenting the whole function.

ejoerns: The glib functions use `g_return_if_fail()` in these cases (to check for passing NULL pointers). The downside of this is that we would have to use `g_clear_pointer()` for calling the free function on potential NULL pointers, but this is what we have to do for glib free functions (Except `g_free()`) anyway.

OnkelUlla: I changed it to `g_return_if_fail()` as requested.
Digging a little bit more through GLib, reveals that this scheme is not used consitently, though. See [gchecksum](https://github.com/GNOME/glib/blob/master/glib/gchecksum.c#L1552) for example.

OnkelUlla:
Did the change to `g_return_if_fail()` as requested and rebased to current master.

ejoerns:
Travis is now fixed by #167. Please rebase and push again.

OnkelUlla:
Rebased and pushed again. :-)

OnkelUlla:
Rebased onto Enrico's coveralls fix.

PR #164, opened 2017-09-07
Fix and encapsulate bundle mount handling

ejoerns:
This moves mointpoint handling to bundle mount functions and allows fixing unmounting bundles for busybox mount by using again the bundle mountpoint as path argument instead of the source bundle path.


src/bundle.c
```diff
@@ -556,6 +581,7 @@ gboolean umount_bundle(RaucBundle *bundle, GError **error) {
 void free_bundle(RaucBundle *bundle) {
 
 	g_free(bundle->path);
+	g_clear_pointer(&bundle->mount_point, g_free);
```
OnkelUlla: No need to use `g_clear_pointer()` here, instead directly use `g_free()` which handles NULL pointers.

ejoerns: You are right. I've pushed a fix.

ejoerns:
Rebased on #167 to let Travis do its job again.

PR #166, opened 2017-09-12
Central persistent status file

OnkelUlla:
In the current version of RAUC the slot status can only be stored in a per-slot-individual file on the slot itself (if the slot contains a writable filesystem), with these changes, they can now be stored jointly in a central place by adding the key

    [system]
    statusfile=/some_persistent_writable_filesystem/etc/rauc/system.raucs

to /etc/rauc/system.conf. This allows to storing status information also for read-only slots that for example contain a SquashFS, kernel, bootloader or DTB, but need a persistent writable filesystem surviving updates as a prerequisite.

Please note that this branch is based on pull requests #158 ("Miscellaneous") and #164 ("Fix and encapsulate bundle mount handling").


OnkelUlla:
Rebased and pushed again. :-)

OnkelUlla:
Rebased onto Enrico's coveralls fix.

docs/reference.rst
```diff
@@ -246,20 +259,56 @@ A valid manifest file must have the file extension ``.raucm``.
...
+
+There are some slot specific metadata that are of interest for RAUC, e.g. a hash
+value of the slot's content (SHA-256 per default) that is matched against its
+counterpart of an image inside a bundle to decide if an update of the slot has
+to be performed or can be skipped. These slot metadata can be persisted in one
+of two ways: either in a slot status file that is created inside each slot which
+contains a writable filesystem or in a single, central and comprehensive status
+file that lives on a persistent filesystem surviving updates. The former is
+RAUC's default whereas the latter mechanism is chosen by making use of the
+optional key :ref:`statusfile <statusfile>` in the system.conf file. Both are
```
mnhu:
```
s/system.conf/`system.conf`
```

ejoerns:
For now, I've just rebased the original branch onto current master which omits the patches that were already merged in other PRs and allows us to focus on the relevant ones.

docs/reference.rst
OnkelUlla: Fixed and re-pushed.

src/config_file.c
```diff
@@ -524,15 +689,59 @@ gboolean save_slot_status(RaucSlot *dest_slot, RaucImage *mfimage, GError **erro
...
+	GError *ierror = NULL;
+	gboolean res;
+	gchar *group;
+
+	g_return_val_if_fail(dest_slot, FALSE);
+	g_return_val_if_fail(dest_slot->status, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+	g_return_val_if_fail(r_context()->config->statusfile_path, FALSE);
+
+	g_key_file_load_from_file(key_file, r_context()->config->statusfile_path, G_KEY_FILE_NONE, &ierror);
```
jluebbe: I'd expect the global slot status file to be regenerated from the in-memory-data. That would avoid having to read it every time. We should drop the dest_slot argument here as well.

OnkelUlla: That's okay with me and as you already added commit a39dcb10dc9f ("src/config_file: save status data for all slots when using a global file"), I consider your change request as done.

ejoerns:
Rebased onto current master, ready for merge if tests pass (from my point of view).

PR #169, opened 2017-09-14
Rework Target Slot Selection and Image Matching Algorithm

ejoerns:
This series prepares, introduces and tests a simplified and improved target slot selection handling as the current implementation had a couple of drawbacks.

It is also preparing structures that allow future extension regarding image selection.

Based on #158


ejoerns:
Rebased onto master. Patch stack is reduced by the patches that were merged with #158.

PR #178, opened 2017-09-27
Add Intermediate Certificate Support and Extended Signature Test & Error Reporting

ejoerns:
Apart from a few enhancements in error message forwarding form Openssl, this series adds support for handling intermediate certificates. To be detailed it enables to generate bundle signatures that may include intermediate certificates as well as closes gaps in allowing to verify such bundles.

To be able to test this, we have to switch the example openssl test setup generation script back to create a full PKI hierarchy where we have a single PKI and the dev and rel certs derived of it.
In the context of these changes and the required adaptions in the signature test, some enhancements and extensions in the general signature test infrastructure were applied, too.

To finally close the recent topics on signature enhancements, there is also some extension in documentation that describes the latest features and how they can be used.


test/signature.c
```diff
@@ -89,42 +89,72 @@ static void signature_sign_file(void)
...
+	g_assert_nonnull(cms);
+	g_assert_nonnull(store);
 
-	// Test non-existing file
-	g_assert_false(cms_verify_file("path/to/nonexisting/file", sig, 0, NULL, NULL, &error));
-	g_assert_error(error, G_FILE_ERROR, G_FILE_ERROR_NOENT);
+	g_clear_pointer(&store, X509_STORE_free);
+	g_clear_pointer(&cms, CMS_ContentInfo_free);
+
+	// Test valid manifest with invalid offset
```
jluebbe: The limit argument to cms_verify_file is not an offset, but a size limit (which causes a truncation here).

test/signature.c
```diff
@@ -174,6 +174,88 @@ static void signature_loopback(void)
...
+	CMS_ContentInfo *cms = NULL;
+	X509_STORE *store = NULL;
+	STACK_OF(X509) *verified_chain = NULL;
+
+	GBytes *content = read_file("test/openssl-ca/manifest", NULL);
+	GBytes *sig = read_file("test/openssl-ca/manifest-r1.sig", NULL);
+	g_assert_nonnull(content);
+	g_assert_nonnull(sig);
+
+	/* We sign with release-1 CA key and cert */
```
jluebbe: We don't actually sign in this test case, so why specific key and cert?

test/signature.c
```diff
@@ -196,6 +278,8 @@ int main(int argc, char *argv[])
 	g_test_add_func("/signature/verify", signature_verify);
 	g_test_add_func("/signature/verify_file", signature_verify_file);
 	g_test_add_func("/signature/loopback", signature_loopback);
+	g_test_add_func("/signature/get_cert_chain", signature_get_cert_chain);
+	g_test_add_func("/signature/selfsigned", signature_selfsigned);
```
jluebbe: The commit message states that we also check the intermediate CA case, which seems to be missing.

test/bundle.c
```diff
@@ -163,23 +163,20 @@ static void bundle_test_resign(BundleFixture *fixture,
 	resignbundle = g_build_filename(fixture->tmpdir, "resigned-bundle.raucb", NULL);
 	g_assert_nonnull(resignbundle);
 
-	/* Switch to release key pair */
-	r_context_conf()->certpath = g_strdup("test/openssl-ca/rel/release-1.cert.pem");
-	r_context_conf()->keypath = g_strdup("test/openssl-ca/rel/private/release-1.pem");
+	/* Input bundle must *not* verify against 'rel' keyring */
+	r_context()->config->keyring_path = g_strdup("test/openssl-ca/rel-ca.pem");
```
jluebbe: Please explain why we need to use r_context instead of r_context_conf in this case.

docs/advanced.rst
```diff
@@ -99,6 +99,71 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
 Note that you can specify the ``--intermediate`` argument multiple times to
 include multiple intermediate certificates to your bundle signature.
 
+Resigning Bundles
+~~~~~~~~~~~~~~~~~
+
+RAUC allows to exchange the signature of a bundle.
```
jluebbe: s/exchange/replace/

docs/advanced.rst
```diff
@@ -99,6 +99,71 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
...
+A typical use case for this is if a bundle that was generated by an autobuilder
+and signed with a development certificate was tested sucessfully on your target
+and should become a release bundle now.
+Fo this it should get resigned with a specific release key without modifying
+the content of the bundle itself.
```
jluebbe: Use:
> A typical use case for this is if a bundle that was generated by an autobuilder
> and signed with a development certificate was tested successfully on your target
> and should now become a release bundle.
> For this it needs to be resigned with the release key without modifying
> the content of the bundle itself.

docs/advanced.rst
```diff
@@ -99,6 +99,71 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
...
+  rauc resign --cert=<certfile> --key=<keyfile> --keyring=<keyring> <input-bundle> <output-bundle>
+
+It verifies the bundle against the given keyring, strips the old signature and
+attaches a new one based on the key and cert files provided.
+
+Switching the Keyring -- SPKI hashes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When switching form a development to a release signature, typically it is
+also required to equip the rootfs with a different keyring file.
```
jluebbe: > When switching from a development to a release signature, it is typically
> required to also equip the rootfs with a different keyring file.

docs/advanced.rst
```diff
@@ -99,6 +99,71 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
...
+
+While the development system should accept both development and release
+certificates, the release system should accept release certificates only.
```
jluebbe: > certificates, the release system should only accept release certificates.

docs/advanced.rst
```diff
@@ -99,6 +99,71 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
...
+     Issuer: /O=Test Org/CN=Test Org Provisioning CA Root
+     SPKI sha256: 00:34:F8:FE:5A:DC:3B:0D:FE:64:24:07:27:5D:14:4D:E2:39:8C:68:CC:9A:86:DD:67:03:D7:15:11:16:B4:4E
+
+A post-install hook instead can access the SPKI hashes via the environment
+variable ``RAUC_BUNDLE_SPKI_HASHES`` that will be set by RAUC when invoking the
+hook script.
+This variable will contain space-separated list of the hashes in the same order
+they are listed in ``rauc info``.
+This list can be used to define a condition in the hook for either installing
+one or the other keyring file on the target.
```
jluebbe: Maybe add some small shell example here?

test/signature.c
ejoerns: Will have to remove this item from the commit message, as the commit re-introducing the 3-layer PKI comes afterwards. Intermediate cert tests for 2-layer PKI are useless.

docs/advanced.rst
ejoerns: Added an example post-install section for a hook script handling copying either development or release keyring to the final keyring path

ejoerns:
I've applied the changes and extensions you requested and updated the PR series.

test/signature.c
```diff
@@ -261,6 +261,71 @@ static void signature_selfsigned(void)
...
+	g_assert_nonnull(sig);
+	g_assert_no_error(error);
+
+	/* Without explicit intermediate certificate, this must fail */
+	g_assert_false(cms_verify(content, sig, &cms, &store, &error));
+	g_assert_error(error, R_SIGNATURE_ERROR, R_SIGNATURE_ERROR_INVALID);
+	g_assert_null(cms);
+	g_assert_null(store);
+
+	g_clear_error(&error);
```
jluebbe: Free sig before reusing the pointer.

test/signature.c
```diff
@@ -261,6 +261,71 @@ static void signature_selfsigned(void)
...
+	g_assert_nonnull(store);
+
+	/* Verify obtaining cert chain works */
+	g_assert_true(cms_get_cert_chain(cms, store, &verified_chain, &error));
+	g_assert_no_error(error);
+	g_assert_nonnull(verified_chain);
+
+	g_clear_pointer(&store, X509_STORE_free);
+	g_clear_pointer(&cms, CMS_ContentInfo_free);
+	g_clear_error(&error);
```
jluebbe: Free sig (and maybe verified_chain?) here as well

PR #179, opened 2017-09-28
Add Support for Image Variants

ejoerns:
Based on #169 this adds support for specifying variants of an image in RAUC.

By assigning an extra 'variant' to a board, it is then possible to create bundles that have different images for the same slot on different hardware variants.

The basic syntax is (``system.conf``):

``` ini
[system]
variant=my-variant-name
```

And in your bundles manifest:

``` cfg
[image.kernel.my-variant-name]
filename=my-variant.img

[image.kernel.other-variant]
filename=other-variant.img
```


src/manifest.c
```diff
@@ -85,11 +85,17 @@ static gboolean parse_image(GKeyFile *key_file, const gchar *group, RaucImage **
 	g_return_val_if_fail(image == NULL || *image == NULL, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
-	groupsplit = g_strsplit(group, ".", 2);
-	g_assert_cmpint(g_strv_length(groupsplit), ==, 2);
+	groupsplit = g_strsplit(group, ".", -1);
```
jluebbe: We need to allow at most 3 groups, because DT compatibles may contain '.'.

docs/advanced.rst
```diff
@@ -176,6 +176,57 @@ are listed below:
 +-----------+--------------------------+---------------------------+
 
 
+Handling Board Variants With a Single Bundle
+--------------------------------------------
+
+If you have variants of hardware that require to install different variants of
```
jluebbe: > If you have hardware variants that require installing different images
> (e.g. for the kernel or for an FPGA bitstream), but have other slots 
> that are common (such as the rootfs) between all hardware variants,
> RAUC allows you to put multiple different variants of these images in the
> same bundle.
> RAUC calls this feature 'image variants'.

docs/advanced.rst
```diff
@@ -176,6 +176,57 @@ are listed below:
...
+variant your specific board is. You can do this in your ``system.conf`` by
+setting the ``variant`` key:
+
+.. code-block:: cfg
+
+  [system]
+  ...
+  variant=variant-1
+
+The ``variant`` key accepts a special value ``dtb`` that allows (on device-tree
```
jluebbe: Update this as well when changing the corresponding code.

docs/advanced.rst
```diff
@@ -176,6 +176,57 @@ are listed below:
...
+  filename=variant1.img
+  ...
+
+  [image.kernel.variant-2]
+  filename=variant1.img
+  ...
+
+It is allowed to have both a specific variant as well as a default image in the
+same bundle.
+Then, on a system that matches a specific variant an image is available for
```
jluebbe: > If a specific variant of the image is available, it will be used on that system.
> ￼On all other systems, the default image will be used instead.

Also document that if an image exists in the manifest, it should be available for all variants. Either using a default or by listing all variants.

ejoerns:
Pushed a new approach for specifying variants without magic key `variant` but dedicated key `variant-name`, `variant-dtb`, `variant-file` instead.

jluebbe:
I folded in some doc and coding style changes.

ejoerns:
Looks good for me :+1:

PR #181, opened 2017-10-04
Skippable

jremmet:
Add flag to system.conf to avoid skipping of slots. This force installation even if hashes are unchanged.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -167,6 +167,11 @@ hierarchical separator.
   Marks the slot as existing but not updatable. May be used for sanity checking
   or informative purpose. A ``readonly`` slot cannot be a target slot.
 
+``ignore-checksum``
+  If set to ``true`` this will bypass the default hash comparison for this slot
+  and force RAUC to unconditionally update it. The default value is ``false``,
+  which means that updating this slot will be skipped if new image's hash
+  matches hash of installed one.
 
 Manifest
 --------
@@ -478,7 +483,7 @@ Performing an update using the default RAUC mechanism will work as follows:
    A. Determine update handler (based on image and slot type)
    #. Try to mount slot and read slot status information
 
-      a. Skip update if new image hash matches hash of insalled one
+      a. Skip update if new image hash matches hash of installed one
 
    #. Perform slot update (image copy / mkfs+tar extract / ...)
    #. Try to write slot status information
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -74,6 +74,8 @@ typedef struct _RaucSlot {
 	gchar *bootname;
 	/** flag indicating if the slot is updatable */
 	gboolean readonly;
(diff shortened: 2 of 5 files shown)


docs/reference.rst
```diff
@@ -166,6 +166,11 @@ set of slots.
   Marks the slot as existing but not updatable. May be used for sanity checking
   or informative purpose. A ``readonly`` slot cannot be a target slot.
 
+``skippable``
+  Marks the slot as skippabel if slot status file shows same image hash. Its
+  default value is  ``true`` which means that this slot may be skipped if new
+  image hash matches hash of installed one. If the value of this parameter is
+  ``false`` the slot is updated installed independently from the hash.
```
ejoerns: the slot is _always_ updated ~~installed~~ independently

docs/reference.rst
```diff
@@ -166,6 +166,11 @@ set of slots.
   Marks the slot as existing but not updatable. May be used for sanity checking
   or informative purpose. A ``readonly`` slot cannot be a target slot.
 
+``skippable``
+  Marks the slot as skippabel if slot status file shows same image hash. Its
```
ejoerns: s/skippabel/skippable

jremmet:
typos fixed

jluebbe:
What is the use case for this? Do you have cases where you don't want to rely on the ``slot.raucs`` checksum, because the actual contents of the slot may have been modified? Or do you want to always write these slots, even if the contents are actually identical?

In the future, I'd like to have the slot hash calculated on the target (image or file based) to prepare for delta updates. In that case, modifications would be detected and the intended behavior of a "skippable" slot wouldn't match the first use case.

Maybe call this 'write-always' or 'ignore-checksum', depending on the intended use case?

jremmet:
The slot content may be altered by the application. Maybe there are even new Files written which are not under control of rauc. So the wish was to always write it.

ejoerns:
Ok, thus the use case is not unconditionally writing but work-around the lack of runtime hash calculation.

I would vote for using `ignore-checksum=true` which seems to fit best as description for the behavior you intend. If you agree please update you PR accordingly so that we can merge it.

jremmet:
ok will rework PR

docs/reference.rst
```diff
@@ -167,6 +167,12 @@ hierarchical separator.
   Marks the slot as existing but not updatable. May be used for sanity checking
   or informative purpose. A ``readonly`` slot cannot be a target slot.
 
+``ignore-checksum``
```
ejoerns: Maybe:

> If set to ``true`` this will bypass the default hash comparison for this slot and force RAUC to unconditionally update it. The default value is ``false`` which means that updating this slot will be skipped if new image's hash matches hash of installed one.

jremmet: Yes much better :)

PR #186, opened 2017-11-10
src/install: add support for already mounted devices

hundeboll:
In cases where a slot device is mounted at boot by the system being
updated, rauc should umount the device before doing its update, and then
remount when done.

This becomes usefull on boards like RPi and Beaglebone Black, where the
boot partition must be mounted at boot to read/write the u-boot
environment, but also updated by rauc (i.e. to update the bootloader).

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -858,7 +858,14 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 			goto out;
 		}
 
-		if (dest_slot->mount_point || dest_slot->ext_mount_point) {
+		if (dest_slot->ext_mount_point && dest_slot->system_mounted) {
+			if (!r_umount(dest_slot->ext_mount_point, &ierror)) {
+				res = FALSE;
+				g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_MOUNTED,
+						"Could not umount destination device: '%s'", dest_slot->device);
+				goto out;
+			}
+		} else if (dest_slot->mount_point || dest_slot->ext_mount_point) {
 			res = FALSE;
 			g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_MOUNTED,
 					"Destination device '%s' already mounted", dest_slot->device);
@@ -975,6 +982,18 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 		}
 
 image_out:
+
+		if (dest_slot->ext_mount_point && dest_slot->system_mounted) {
+			res = r_mount_full(dest_slot->device, dest_slot->ext_mount_point,
+					dest_slot->type, 0, &ierror);
+			if (!res) {
+				g_propagate_prefixed_error(
+						error,
+						ierror,
+						"Remounting failed: ");
+				goto out;
+			}
+		}
 
 		install_args_update(args, g_strdup_printf("Updating slot %s done", dest_slot->name));
 	}
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -190,6 +190,11 @@ hierarchical separator.
   active slots, so that the inactive one can be selected as the update target.
   The parent slot is referenced using the form ``<slot-class>.<idx>``.
 
+``mounted``
+  Setting this entry to ``true`` tells RAUC to umount the slot before
+  updating it, and to remount it when done. Used for slots that are
+  mounted at boot by the system.
+
 ``readonly``
   Marks the slot as existing but not updatable. May be used for sanity checking
   or informative purpose. A ``readonly`` slot cannot be a target slot.
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -88,6 +88,8 @@ typedef struct _RaucSlot {
 	gchar *type;
 	/** the name this slot is known to the bootloader */
 	gchar *bootname;
+	/** flag to indicate that this slot is always mounted */
+	gboolean system_mounted;
 	/** flag indicating if the slot is updatable */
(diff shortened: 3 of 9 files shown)


hundeboll:
I am not sure how to add the tests needed for this. Can you give me some pointers?

src/install.c
```diff
@@ -630,6 +630,13 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 		mfimage = l->data;
 		dest_slot = g_hash_table_lookup(target_group, mfimage->slotclass);
 
+		if (dest_slot->external_mount_point && !r_umount(dest_slot->device, &ierror)) {
```
ejoerns: It might not work calling unmount with the device path rather than the actual mount point. At least busybox mount could not handle this properly when I tested it for the last time.

src/install.c
```diff
@@ -88,7 +88,7 @@ gboolean determine_slot_states(GError **error) {
 		gchar *devicepath = resolve_loop_device(g_unix_mount_get_device_path(m));
 		RaucSlot *s = find_config_slot_by_device(r_context()->config,
 				devicepath);
-		if (s) {
+		if (s && !s->external_mount_point) {
```
ejoerns: Wouldn't it be safer to look at the actual mount point?

jluebbe:
How would your system.conf look for this use-case, one boot and two root slots?
Would you use a different bundle to update the bootloader instead of the rootfs? If so, I think something like http://rauc.readthedocs.io/en/latest/advanced.html#updating-the-bootloader could be better.

mnhu:
> Could you point out if there is any advantage in explicitly specifying the mount point rather than only specifying a boolean flag indicating that an unmount and remount should be performed?

I changed the patch to use a simple boolean. Also updated it to use the mount-point found with `g_unix_mounts_get()` and store in a separate slot member variable.

mnhu:
> How would your system.conf look for this use-case, one boot and two root slots?
> Would you use a different bundle to update the bootloader instead of the rootfs? If so, I think something like http://rauc.readthedocs.io/en/latest/advanced.html#updating-the-bootloader could be better.

Hmm.. I think we'd prefer to keep bootloader updates explicit and separate from rootfs updates. We do test it regularly in our CI, but only do it rarely in production.
Another point is the reboots needed in both AB and AR setups. With bootloader updates embedded in the rootfs, it requires two reboots to know if worked: one from "current" to "new" after updating the rootfs, and one from "new" to "new" after updating the bootloader.

mnhu:
What are your thoughts on this? Should we move forward with adding tests etc, or is there no chance of getting it merged?

Thanks,

ejoerns:
There is a chance, yes.
We came across different disadvantages of the current way we handle mount point detection and how we differentiate external mounts from 'internal' mounts. As I fear the changes that will be made there will interfere with your patch, I would like to keep it blocked until we are sure about the future handling of mounts.
A proposal for changing the mount handling is in my casync patch stack: https://github.com/rauc/rauc/pull/217/commits/97e74f9b6b40b4a819702a75e53412d89c65f31e

mnhu:
That's fine. We will take a look at the casync feature soon...

ejoerns:
With #232 we've merged the mount patch from the casync stack I mentioned thus it might be worth rebasing and adapting this patch onto current master.

ejoerns:
Do you still plan to put effort into this?

mnhu:
@ejoerns Pushed an updated version including tests of the feature. Thanks for the patience.

mnhu:
@ejoerns Care to take a look at this again some time ?

mnhu:
@ejoerns One last bump :)

ejoerns:
Shame on me, I'm still scrolling through a backlog of things to to in constantly changing priority...

From my point of view this can go into mainline. I will test it and you'll finally get new feedback on this!

@jluebbe Please also check and note if you have any reservations

ejoerns:
Went through the patches and they basically look really fine for me.

I felt free to add a further patch to assure bootslot is actually remounted, please have a look at it and say if you agree with it.

The only thing I am not fully sure is if the key `mounted` in the config file is unmistakable enough to know what it does an not expect something else. The variable is named `system_mounted`, maybe the name `system-mounted` would be an option too and make it more clear?

I  would like to hear if there are other opinions on this...

src/install.c
```diff
@@ -858,7 +858,14 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 			goto out;
 		}
 
-		if (dest_slot->mount_point || dest_slot->ext_mount_point) {
+		if (dest_slot->ext_mount_point && dest_slot->system_mounted) {
+			if (!r_umount(dest_slot->ext_mount_point, &ierror)) {
```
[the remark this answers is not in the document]
hundeboll: Yes, we saw that when switching to systemd, I think patched the service unit to remove the `MountFlags=slave` line.

hundeboll:
I no longer work with this setup, so I currently don't have a setup to test it. If you leave it hanging around, I'll get back to it eventually.

Otherwise, we should just close it. Then it can be reopened if someone takes the time to look at it.

Villemoes:
I've sent an alternative PR (#326) that avoids the mount propagation issue, and which also works for our use case.

jluebbe:
> I've sent an alternative PR (#326) that avoids the mount propagation issue, and which also works for our use case.

I *finally* found some time to go through the old, dusty PRs. #326 is a cleaner solution and nicely avoid the problems with mount namespaces and even something else keeping the mount busy. I'll close this one and update #326. Nevertheless, thanks for the groundwork @hundeboll!

PR #191, opened 2017-12-01
Fixes and error handling rework for RAUC boot selection layer

ejoerns:
This series targets different issues in Barebox bootchooser (boot selection layer) implementation.

The two main ones are:

* This is one of the components that still lacks appropriate GError handling, giving us only insufficient information about failures
* barebox-state backend did not properly handle disabled slots for setting slots as primary. This also additionally requires reading state info instead of only writing it

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -21,8 +21,8 @@ static gboolean bootchooser_order_primay(RaucSlot *slot, GString **value) {
 	GString *order = g_string_sized_new(10);
 	GList *slots;
 
-	g_assert_nonnull(slot);
-	g_assert_nonnull(value);
+	g_return_val_if_fail(slot, FALSE);
+	g_return_val_if_fail(value, FALSE);
 
 	g_string_append(order, slot->bootname);
 
@@ -55,6 +55,9 @@ static gboolean barebox_state_get_int(const gchar* name, int *value) {
 	guint64 result = 0;
 	GPtrArray *args = g_ptr_array_new_full(10, g_free);
 	
+	g_return_val_if_fail(bootname, FALSE);
+	g_return_val_if_fail(bb_state, FALSE);
+
 	g_ptr_array_add(args, g_strdup(BAREBOX_STATE_NAME));
 	g_ptr_array_add(args, g_strdup("-g"));
 	g_ptr_array_add(args, g_strdup(name));
@@ -107,6 +110,8 @@ static gboolean barebox_state_set(GPtrArray *pairs) {
 	gboolean res = FALSE;
 	GPtrArray *args = g_ptr_array_new_full(2*pairs->len+2, g_free);
 
+	g_return_val_if_fail(pairs, FALSE);
+
 	g_assert_cmpuint(pairs->len, >, 0);
 	
 	g_ptr_array_add(args, g_strdup(BAREBOX_STATE_NAME));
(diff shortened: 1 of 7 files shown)


src/bootchooser.c
```diff
@@ -44,27 +44,36 @@ static gboolean bootchooser_order_primay(RaucSlot *slot, GString **value) {
...
 }
 
+typedef struct {
+	guint32 prio;
+	guint32 attempts;
+} BareboxSlotState;
+
+#define BOOTSTATE_PREFIX "bootstate"
+
 #if 0
```
jluebbe: According to the commit message, the function should be reenabled.

src/bootchooser.c
```diff
@@ -22,12 +22,11 @@ GQuark r_bootchooser_error_quark (void)
 #define UBOOT_FWSETENV_NAME "fw_setenv"
 #define UBOOT_FWGETENV_NAME "fw_printenv"
 
-static gboolean bootchooser_order_primay(RaucSlot *slot, GString **value) {
+static GString *bootchooser_order_primay(RaucSlot *slot) {
 	GString *order = g_string_sized_new(10);
 	GList *slots;
 
 	g_return_val_if_fail(slot, FALSE);
```
jluebbe: We can't return FALSE here anymore.

src/bootchooser.c
```diff
@@ -508,9 +501,13 @@ static gboolean uboot_set_primary(RaucSlot *slot) {
 	/* Add updated slot as first entry in new boot order */
 	g_string_append(order_new, slot->bootname);
 
-	res = uboot_env_get("BOOT_ORDER", &order_current, NULL);
-	if (!res && !bootchooser_order_primay(slot, &order_current))
-		goto out;
+	res = uboot_env_get("BOOT_ORDER", &order_current, &ierror);
+	if (!res) {
+		g_message("Unable to obtain BOOT_ORDER");
```
jluebbe: Maybe indicate that this is not cirtical and we fall back to a default order?

test/install-content/custom_handler.sh
```diff
@@ -24,6 +24,14 @@ done
 exit_if_empty $RAUC_MOUNT_PREFIX
 mkdir -p $RAUC_MOUNT_PREFIX/image
 
+# This is only for testing
```
jluebbe: Should this change be squashed into the one changing the code?

src/bootchooser.c
ejoerns: This would require squashing commits if we intend to keep the commit compilable.

jluebbe: Hmm, OK. Then explain that in the commit message.

src/bootchooser.c
ejoerns: I assume changing it to 
``` c
g_message("Unable to obtain BOOT_ORDER, using defaults");
```

would be sufficient, as the log level indicates being non-critical?

test/install-content/custom_handler.sh
ejoerns: Yes, could, was more ore less a post-fix.

ejoerns:
Updated commit messages, squashed and fixed as requested.

``` diff
diff --git a/src/bootchooser.c b/src/bootchooser.c
index bf8888b..504dfa6 100644
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -26,7 +26,7 @@ static GString *bootchooser_order_primay(RaucSlot *slot) {
        GString *order = g_string_sized_new(10);
        GList *slots;
 
-       g_return_val_if_fail(slot, FALSE);
+       g_return_val_if_fail(slot, NULL);
 
        g_string_append(order, slot->bootname);
 
@@ -504,7 +504,7 @@ static gboolean uboot_set_primary(RaucSlot *slot, GError **error) {
 
        res = uboot_env_get("BOOT_ORDER", &order_current, &ierror);
        if (!res) {
-               g_message("Unable to obtain BOOT_ORDER");
+               g_message("Unable to obtain BOOT_ORDER, using defaults");
                g_clear_error(&ierror);
 
                order_current = bootchooser_order_primay(slot);
```

PR #196, opened 2017-12-06
.travis.yml: use gcc-7

ejoerns:
We already have GCC 7 which might find more issues by improved compile-time checks. Thus enable it for testing.


ejoerns:
I've adapted the cov-configure call to use gcc-7.

PR #202, opened 2017-12-14
save_manifest_file for variants

jremmet:
make save_manifest_file variant aware

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -417,6 +417,12 @@ gboolean save_manifest_file(const gchar *filename, RaucManifest *mf, GError **er
 
 		group = g_strconcat(RAUC_IMAGE_PREFIX ".", image->slotclass, NULL);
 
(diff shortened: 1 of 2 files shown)


src/manifest.c
```diff
@@ -417,6 +417,9 @@ gboolean save_manifest_file(const gchar *filename, RaucManifest *mf, GError **er
 
 		group = g_strconcat(RAUC_IMAGE_PREFIX ".", image->slotclass, NULL);
 
+		if (image->variant)
+			 group = g_strconcat(group, ".", image->variant, NULL);
```
ejoerns: `g_strconcat()` allocates a new string but we have to free the old one before.

Something like 

    if (image->variant) {
       gchar *tmp = group;
       group = g_strconcat(group, ...)
       g_free(tmp);
    }

should do it.

PR #203, opened 2017-12-15
Improvements and fixes for image variants

ejoerns:
* Fixes progress step calculation
* Adds variant information to `rauc info` printouts
* Adds some useful logging output about variant usage


ejoerns:
@jremmet please check if that makes image variant handling working and better usable for you

jremmet:
Yes it works now as expected :)

PR #208, opened 2018-01-05
Support resolving device symlinks using `realpath()` for more flexible matching

ejoerns:
Especially in cases where standard device paths (such as /dev/sdb) are
not persistent over reboots, one might need to use symlink pathes
generated by standard or custom udev rules. A good example for this are
the /dev/disk-by-* symlinked device paths.

As RAUC uses the device path to string-compare against the `root=`
kernel command line argument it will not match against a system config
slot entry with a symlink device path such as

    [slot.rootfs.0]
    device=/dev/disk/by-path/pci-0000:00:17.0-ata-1-part1
    ...

This patch resolves this by calling `realpath()` to always get the
canonicalized absolute pathname of the `device` parameter when comparing
against boot slot device name as provided by root= kernel command line.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -10,30 +10,13 @@
 
 RaucContext *context = NULL;
 
-/* resolve PARTUUID=xxx to real device name */
-static gchar* resolve_partuuid_arg(gchar *arg) {
-	gchar buf[PATH_MAX + 1];
-	gchar *res = NULL;
-
-	gchar *partuuidpath = g_build_filename(
-			"/dev/disk/by-partuuid/",
-			&arg[9],
-			NULL);
-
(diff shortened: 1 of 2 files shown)


jluebbe:
When using an initramfs, having the symlink as the root= argument is also possible. Should we use realpath() on both sides before the comparison?

ejoerns:
We already resolve `root=PARTUUID=xxx` entries and use `realpath()` for getting the device of the `/dev/disk/by-partuuid` entry. This is done in `context.c` function `resolved_partuuid_arg()`.
I would agree extending this to always resolve paths if possible and directly store the resolved path in `context->bootslot`.

ejoerns:
I've extended the series with the modification I proposed.

PR #215, opened 2018-01-16
Rework manifest verification handling

ejoerns:
Step by step this removes the code from `verify_manifest()` by either moving or simply dropping it.

While doing this, some drawbacks and issues of the current handling are resolved while preparing code for cool new features.


ejoerns:
Note: Pushed an update for this that keeps `verify_manifest` function as we may need to use it for some sort of `rauc verify` functionality later on.

PR #217, opened 2018-01-19
Introduce casync support

ejoerns:
This series of patches adds support for using the [casync](https://github.com/systemd/casync) tool for image and archive installation.

By using casync, way we gain a couple of cool new capabilities:

 * Direct streaming of Updates from remote servers (without intermediate download of images)
 * Binary Delta-like Updating to reduce network load (e.g. for constrained links)
 * Simpilified and storage friendly artifact pooling on deployment server side
 * CDN friendly chunk fetching

Creating a casync-supporting bundle is a two-step process:

1. Create your conventional RAUC bundle
2. Convert it

       rauc convert old-bundle-raucb casync-bundle.raucb

This will create a `casync-bundle.raucb` bundle file (only containing index files instead of real images)  and a separate casync chunk store `casync-bundle-castr/`.

Place these two on a web server or start one (such as `python3 -m http.server`) in the directory you extracted them to.

Then you can install it remotely on your target device by running

    rauc install http://192.168.0.1:8000/casync-bundle.raucb

This will install the images described by the index files placed in the bundle with using appropriate slots as seeds on target side and only fetching chunks remotely that cannot be obtained from the target itself.

Note that you have to have casync compiled an in your path, both on target as well as on host side!

PR is based on PR's #214 and #215

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -568,6 +568,21 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 		ibundle->path = g_strdup(bundlename);
 	}
 
+	/* Determine store path for casync, defaults to bundle */
+	if (r_context()->config->store_path) {
+		ibundle->storepath = r_context()->config->store_path;
+	} else {
+		gchar *strprfx;
+
+		if (ibundle->origpath)
+			strprfx = g_strndup(ibundle->origpath, strlen(ibundle->origpath) - 6);
+		else
+			strprfx = g_strndup(ibundle->path, strlen(ibundle->path) - 6);
+		ibundle->storepath = g_strconcat(strprfx, ".castr", NULL);
+
+		g_free(strprfx);
+	}
+
 	if (verify && !r_context()->config->keyring_path) {
 		g_set_error(error, R_BUNDLE_ERROR, R_BUNDLE_ERROR_KEYRING, "No keyring file provided");
 		res = FALSE;
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -7,6 +7,11 @@
 #include "context.h"
 #include "mount.h"
 #include "signature.h"
+#include "utils.h"
+#include "network.h"
+
+/* Maximum downloadable bundle size (8MB) */
+#define BUNDLE_DL_MAX_SIZE 8*1024*1024
 
 GQuark
 r_bundle_error_quark (void)
@@ -346,6 +351,13 @@ out:
 	return res;
 }
 
+static gboolean is_remote_scheme(const gchar *scheme) {
+	return (g_strcmp0(scheme, "http") == 0) ||
+		(g_strcmp0(scheme, "https") == 0) ||
+		(g_strcmp0(scheme, "sftp") == 0) ||
+		(g_strcmp0(scheme, "ftp") == 0);
+}
+
 gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean verify, GError **error) {
 	GError *ierror = NULL;
 	GBytes *sig = NULL;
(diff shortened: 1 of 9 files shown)


ejoerns:
Rebased on master

src/update_handler.c
```diff
@@ -343,6 +515,14 @@ static gboolean untar_image(RaucImage *image, gchar *dest, GError **error)
 	return res;
 }
 
+static gboolean unpack_image(RaucImage *image, gchar *dest, GError **error)
```
jluebbe: Shouldn't this be called unpack_archive?

src/update_handler.c
```diff
@@ -634,7 +814,71 @@ static gboolean tar_to_ubifs_handler(RaucImage *image, RaucSlot *dest_slot, cons
...
+		} else {
+			g_propagate_error(error, ierror);
+		}
+	}
+
+out:
+	return res;
+}
+
+static gboolean catar_to_ext4_handler(RaucImage *image, RaucSlot *dest_slot, const gchar *hook_name, GError **error) {
```
jluebbe: Couldn't this also be handled by arch(ive)_to_ext4_handler?

src/update_handler.c
```diff
@@ -916,12 +1140,25 @@ typedef struct {
...
 
 RaucUpdatePair updatepairs[] = {
+	{"*.ext4.caibx", "ext4", img_to_fs_handler},
+	{"*.ext4.caibx", "raw", img_to_raw_handler},
+	{"*.vfat.caibx", "raw", img_to_raw_handler},
+	//{"*.ubifs", "ubivol", img_to_ubivol_handler}, /* unsupported */
+	//{"*.ubifs", "ubifs", img_to_ubifs_handler}, /* unsupported */
+	//{"*.img", "nand", img_to_nand_handler}, /* unsupported */
+	//{"*.img", "ubivol", img_to_ubivol_handler}, /* unsupported */
+	//{"*.squashfs", "ubivol", img_to_ubivol_handler}, /* unsupported */
```
jluebbe: Why are these unsupported?

src/bundle.c
```diff
@@ -346,6 +456,174 @@ gboolean resign_bundle(RaucBundle *bundle, const gchar *outpath, GError **error)
...
+
+	/* Iterate over each image and convert */
+	for (GList *l = manifest->images; l != NULL; l = l->next) {
+		RaucImage *image = l->data;
+		gchar *imgpath = NULL, *idxfile = NULL, *idxpath = NULL;
+
+		imgpath = g_build_filename(contentdir, image->filename, NULL);
+
+		if (image_is_archive(image)) {
+
```
jluebbe: Remove empty line.

src/bundle.c
```diff
@@ -103,6 +109,110 @@ static gboolean unsquashfs(const gchar *bundlename, const gchar *contentdir, con
...
 
+static gboolean casync_make_arch(const gchar *idxpath, const gchar *contentpath, const gchar *store, GError **error) {
+	GSubprocess *sproc = NULL;
+	GError *ierror = NULL;
+	gboolean res = FALSE;
+	GPtrArray *args = g_ptr_array_new_full(15, g_free);
+	GPtrArray *iargs = g_ptr_array_new_full(15, g_free);
+	const gchar *tmpdir = NULL;
+
+	tmpdir = g_dir_make_tmp("arch-XXXXXX", &ierror);
```
jluebbe: Does this need to be configurable? Root filesystems could be some GBs.

ejoerns: Might be useful, I fear we have other things we should make configurable, too. Thus I would like to postpone this.

ejoerns:
Applied fixups for your comments. What do you think.. Is it sufficient to give it a try?

src/bundle.c
jluebbe: OK.

jluebbe:
Looks good to me now, can you squash the fixups?

PR #220, opened 2018-01-24
src/bootchooser: support re-sorted EFI boot entries in efi_get_primary()

ejoerns:
Some EFI implementations might choose to have a default EFI boot entry
always placed on first position in order to execute it first and then
continue selecting the next boot target to boot the actual system.

In this case, the simple approach on looking at the first element in
EFI BootOrder for finding the primary slot will not work.
Also having BootNext set must not necessarily mean that this will match
one of the bootnames handeled by RAUC.

A more generic approach for this would be to scan the entire BootOrder
list until finding the first element we know as a bootname for any of
our slots configured and consider this to be the primary one.

The patch changes the efi_get_primary() method to gain this behavior.

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -995,33 +995,26 @@ static RaucSlot *efi_get_primary(GError **error) {
 		goto out;
 	}
 
-	/* We only need to look further if bootnext is not set */
+	/* We prepend the content of BootNext if set */
 	if (bootnext) {
-		g_debug("Skip scanning for primary as Bootnext is set to %s", bootnext->name);
-		goto get_slot;
+		g_debug("Detected BootNext set to %s", bootnext->name);
+		bootorder_entries = g_list_prepend(bootorder_entries, bootnext);
 	}
 
-	if (g_list_first(bootorder_entries) == NULL) {
-		g_set_error_literal(
-				error,
-				R_BOOTCHOOSER_ERROR,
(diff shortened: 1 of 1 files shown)


src/bootchooser.c
```diff
@@ -1011,18 +1012,22 @@ static RaucSlot *efi_get_primary(GError **error) {
...
+		g_hash_table_iter_init(&iter, r_context()->config->slots);
+		while (g_hash_table_iter_next(&iter, NULL, (gpointer *)&slot)) {
+			if (g_strcmp0(bootnext->name, slot->bootname) == 0) {
+				primary = slot;
+				break;
+			}
 		}
-	}
+
+		bootnext = g_list_nth_data(bootorder_entries, idx++);
```
jluebbe: We should be able to just use g_list_next() for iteration, which seems more ideomatic.

ejoerns: I obtain `bootnext` from `efi_bootorder_get()` and thus use g_list_nth_data() to directly get the an alternative for `bootnext` if `BootNext` EFI var was not set.
As I cannot run g_list_next() on the data object, I chose to iterate here this way.

Otherwise I have to change to code to use a GList for iterating and add some more extra handling to use the `efi_bootentry` `bootnext` I get from `efi_bootorder_get()`.

ejoerns:
According to our discussion I've pushed an updated version that now prepends bootnext to the list and simply iterates over all elements.

PR #233, opened 2018-02-21
Add option  "--detailed" to rauc status

OnkelUlla:
There hasn't been a way to get hold of a slot's status information (which bundle has been used to install it, when has it been installed last, has it been activated, ...). Now this data can be extracted via ``rauc status --detailed`` supporting all present output-formats to have it machine-parsable if required.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1074,6 +1074,86 @@ static gchar* r_status_formatter_json(gboolean pretty)
 #endif
 }
 
+static RaucSlotStatus* r_variant_get_slot_state(GVariant *vardict)
+{
+	RaucSlotStatus *slot_state = g_new0(RaucSlotStatus, 1);
+	GVariantDict dict;
+
+	g_variant_dict_init(&dict, vardict);
+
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -1074,6 +1178,86 @@ static gchar* r_status_formatter_json(gboolean pretty)
...
 }
 
+static RaucSlotStatus* r_variant_get_slot_state(GVariant *vardict)
+{
+	RaucSlotStatus *slot_state = g_new0(RaucSlotStatus, 1);
+	GVariantDict dict;
+
+	g_variant_dict_init(&dict, vardict);
+
+	g_variant_dict_lookup (&dict, "bundle.compatible", "s", &slot_state->bundle_compatible);
```
jluebbe: Remove the space before (, and in the following lines as well.

OnkelUlla: Oh, I missed that originally. Fixed it now!

OnkelUlla:
Yes, I successfully tested both: configured with "--enable-service" and "--disable-service".

PR #234, opened 2018-02-23
uncrustify: run on include, src and test

jluebbe:
Signed-off-by: Jan Luebbe <jlu@pengutronix.de>


test/update_handler.c
```diff
@@ -15,15 +15,15 @@ typedef struct {
...
-	TEST_UPDATE_HANDLER_EXPECT_FAIL   = 1 << 0,
-	TEST_UPDATE_HANDLER_NO_IMAGE_FILE = 1 << 1,
-	TEST_UPDATE_HANDLER_NO_TARGET_DEV = 1 << 2,
-	TEST_UPDATE_HANDLER_HOOKS         = 1 << 3,
-	TEST_UPDATE_HANDLER_PRE_HOOK      = 1 << 4,
-	TEST_UPDATE_HANDLER_POST_HOOK     = 1 << 5,
-	TEST_UPDATE_HANDLER_INSTALL_HOOK  = 1 << 6,
-	TEST_UPDATE_HANDLER_NO_HOOK_FILE  = 1 << 7,
-	TEST_UPDATE_HANDLER_HOOK_FAIL     = 1 << 8,
+	        TEST_UPDATE_HANDLER_EXPECT_FAIL   = 1 << 0,
```
ejoerns: This is broken

test/bootchooser.c
```diff
@@ -166,7 +166,7 @@ bootstate.system1.priority=20\n\
 }
 
 static void bootchooser_grub(BootchooserFixture *fixture,
-		gconstpointer user_data)
+                             gconstpointer user_data)
```
ejoerns: Mh, doesn't this change *all* occurrences of multi-line function parameter definitions? But I also couldn't find any pattern for this in the code we had. Just notice that it is different from function call parameter alignment, but that might be ok.

.uncrustify.cfg
```diff
@@ -10,6 +10,7 @@ indent_func_def_param           = true
 indent_func_proto_param         = true
 indent_func_param_double        = true
 indent_switch_case              = 8
+nl_fdef_brace                   = add
```
ejoerns: with `nl_fdef_brace = add` I get

    22 files changed, 438 insertions(+), 219 deletions(-)

while with `nl_fdef_brace = remove` I get

    21 files changed, 165 insertions(+), 327 deletions(-)

which seems less invasive. Thus we explicitly change coding style?

ejoerns:
Just some final steps required for this. I've added a fixup to call the build script automatically if executable was not found, yet.
Then we should squash the last commits and add a short description why we need to build uncrustify locally.

PR #238, opened 2018-02-28
Dbus doc methods

jremmet:
Add methods Info, Mark and GetSlotStatus

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -441,6 +441,12 @@ Methods
 ~~~~~~~
 :ref:`Install <gdbus-method-de-pengutronix-rauc-Installer.Install>` (IN  s source);
 
+:ref:`Info <gdbus-method-de-pengutronix-rauc-Installer.Info>` (IN  s bundle, s compatible, s version);
+
+:ref:`Mark <gdbus-method-de-pengutronix-rauc-Installer.Mark>` (IN  s state, IN  s slot_identifier, s slot_name, s message);
+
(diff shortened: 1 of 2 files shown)


docs/reference.rst
```diff
@@ -474,6 +480,69 @@ Triggers the installation of a bundle.
...
+
+The Info() Method
+^^^^^^^^^^^^^^^^^
+
+.. code::
+
+  de.pengutronix.rauc.Installer.Info()
+  Info (IN  s bundle, s compatible, s version);
+
+Triggers the installation of a bundle.
```
jluebbe: This looks like a leftover from copying the Install() method.

jremmet:
Thanks for the review, copy pasted mistake fixed.

ejoerns:
Good work!

Just one question about the `Info()` method that we intentionally left undocumented because we were unsure if it is the right way to go or if we should have made a string-variant-dict-based interface for this, too.
Thus the question if you actually use the Info method and, if, how that fits your needs?

jremmet:
Our customer asks about a D-BUS Interface to get the "rauc status" information. So I looked at  GetSlotStatus. There was no need for the Info method now. But a  string-variant-dict-based interface would allow more flexibility in future. So +1

ejoerns:
I assume you chose the wrong button for answering?

ejoerns:
I've started #248 as a discussion base for further D-Bus enhancements and will merge this one for now.

PR #239, opened 2018-03-01
Dbus getslotstatus entries

jremmet:
I want to get the output from "rauc status" also vie dbus.
GetSlotStatus sends only the "detailed" status information in terms of "rauc status --detailed"

Is this the right place for this information or should it go to an own method?

### DIFF
--- a/src/service.c
+++ b/src/service.c
@@ -4,6 +4,8 @@
 #include <stdio.h>
 
 #include "bundle.h"
+#include "bootchooser.h"
+#include "config_file.h"
 #include "context.h"
 #include "install.h"
 #include "mark.h"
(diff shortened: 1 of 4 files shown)


jremmet:
First I want to be sure that this is the right method here. There are status not linked to a slot like:
"Compatible", "Variant", "Booted from" and "Activated"
I'll send a patch also answering these in string array in a own out parameter.

src/service.c
```diff
@@ -214,13 +216,39 @@ static gboolean r_on_handle_mark(RInstaller *interface,
...
+	if (slot->state)
+		g_variant_dict_insert(&dict, "state", "s", slotstate_to_str(slot->state));
+	if (slot->description)
+		g_variant_dict_insert(&dict, "description", "s", slot->description);
+	if (slot->parent)
+		g_variant_dict_insert(&dict, "parent", "s", slot->parent->name);
+	if (slot->mount_point)
+		g_variant_dict_insert(&dict, "mountpoint", "s", slot->mount_point);
+	if (slot->bootname)
+		g_variant_dict_insert(&dict, "boot status", "s", good ? "good" : "bad");
```
jluebbe: I'd prefer "boot-status" instead (we don't have spaces in field names so far).

PR #242, opened 2018-03-08
Add support for updating eMMC boot partitions 

BastianStender:
This enables atomic updates of eMMC boot partitions for bootloaders:

- read active boot partition
- empty the inactive boot partition
- set inactive boot partition rw
- write bootloader to inactive partition
- set inactive boot partition ro
- set inactive boot partition active
- check that the previous inactive boot partition is active now

Also the docs were restructured and now reflect the first bootloader update possibility via RAUC.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -481,6 +481,8 @@ RAUC will report an error if for example a booloader image is only present for
 variant A when you try to install on variant B.
 This should prevent from bricking your device by unintentional partial updates.
 
+.. _sec-manual-write:
+
 Manually Writing Images to Slots
 --------------------------------
 
@@ -517,14 +519,39 @@ non-bootable system caused by an error or power-loss during the update.
 
 Whether atomic bootloader updates can be implemented depends on your
 SoC/firmware and storage medium.
-For example eMMCs have two dedicated boot partitions (see the JEDEC standard
-JESD84-B51_ for details), one of which can be enabled atomically via
-configuration registers in the eMMC.
+
+Update eMMC Boot Partitions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+RAUC supports updating eMMC boot partitions (see the JEDEC standard JESD84-B51_
+for details), one of which gets enabled atomically via configuration registers
+in the eMMC (*ext_csd registers*).
 
 .. _JESD84-B51: http://www.jedec.org/standards-documents/results/jesd84-b51
 
-As a further example, the NXP i.MX6 supports up to four bootloader copies when
-booting from NAND flash.
+The required slot type is ``boot-emmc``.
+The device to be specified is expected to be the root device.
+The boot partitions are derived automatically.
+A ``system.conf`` could look like this:
+
+.. code-block:: cfg
+
+  [slot.bootloader.0]
+  device=/dev/mmcblk1
+  type=boot-emmc
+
+.. important::
+
+  A kernel bug may prevent consistent toggling of the eMMC ext_csd boot
+  partition register.
+  Be sure your kernel contains this patch:
+  https://www.spinics.net/lists/linux-mmc/msg48271.html
+
+Bootloader Update Ideas
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The NXP i.MX6 supports up to four bootloader copies when booting from NAND
+flash.
 The ROM code will try each copy in turn until it finds one which is readable
 without uncorrectable ECC errors and has a correct header.
 By using the trait of NAND flash that interrupted writes cause ECC errors and
@@ -532,14 +559,15 @@ writing the first page (containing the header) last, the bootloader images can
 be replaced one after the other, while ensuring that the system will boot even in
 case of a crash or power failure.
 
-Currently, independent of whether you are able to update your bootloader with
-fallback, atomically or with some risk of an unbootable system, our suggestion
-is to handle updates for it outside of RAUC.
-The main reason is to avoid booting an old system with a new bootloader, as this
-combination is usually not tested during development, increasing the risk of
-problems appearing only in the field.
+The slot type could be called "boot-imx6-nand" analogous to eMMC.
 
-One possible approach to this is:
+Considerations When Updating the Bootloader
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Booting an old system with a new bootloader is usually not tested during
+development, increasing the risk of problems appearing only in the field.
+If you want to address this issue do not add the bootloader to your bundle, but
+rather use an approach like this:
 
 * Store a copy of the bootloader in the rootfs.
 * Use RAUC only to update the rootfs. The combinations to test
--- a/include/emmc.h
+++ b/include/emmc.h
@@ -1,0 +1,79 @@
+#include <glib.h>
+#include <linux/types.h>
+
+#define EMMC_BOOT_PARTITIONS			2
+#define INACTIVE_BOOT_PARTITION(part_active)	((part_active + 1) % EMMC_BOOT_PARTITIONS)
+
+#define EXT_CSD_PART_CONFIG			179
+#define EXT_CSD_CMD_SET_NORMAL			(1 << 0)
+#define EXT_CSD_BOOT_CFG_ACC			(0x07)
+
+/* From kernel linux/mmc/mmc.h */
+#define MMC_SWITCH				6               /* ac	[31:0] See below	R1b */
+#define MMC_SEND_EXT_CSD			8               /* adtc				R1  */
+#define MMC_SWITCH_MODE_WRITE_BYTE		0x03            /* Set target to value */
+
+/* From kernel linux/mmc/core.h */
+#define MMC_RSP_PRESENT				(1 << 0)
+#define MMC_RSP_CRC				(1 << 2)        /* expect valid crc */
+#define MMC_RSP_BUSY				(1 << 3)        /* card may send busy */
+#define MMC_RSP_OPCODE				(1 << 4)        /* response contains opcode */
+
+#define MMC_CMD_AC				(0 << 5)
+#define MMC_CMD_ADTC				(1 << 5)
+
+#define MMC_RSP_SPI_S1				(1 << 7)        /* one status byte */
+#define MMC_RSP_SPI_BUSY			(1 << 10)       /* card may send busy */
+
+#define MMC_RSP_SPI_R1				(MMC_RSP_SPI_S1)
+#define MMC_RSP_SPI_R1B				(MMC_RSP_SPI_S1|MMC_RSP_SPI_BUSY)
+
+#define MMC_RSP_R1				(MMC_RSP_PRESENT|MMC_RSP_CRC|MMC_RSP_OPCODE)
+#define MMC_RSP_R1B				(MMC_RSP_PRESENT|MMC_RSP_CRC|MMC_RSP_OPCODE|MMC_RSP_BUSY)
+
+
+/**
+ * Reads the active eMMC boot partition index of given eMMC device into the
+ * given variable.
+ *
+ * @param device eMMC /dev path (/dev/mmcblkX)
+ * @param bootpart_active will contain the active boot partition
+ * (0 for mmcblkXboot0, 1 for mmcblkXboot1, 6 for mmcblkX user partition and
+ * -1 for no active boot partition)
+ * for user partition)
+ * @param error return location for a GError, or NULL
+ *
+ * @return True if succeeded, False if failed
+ */
+gboolean r_emmc_read_bootpart(const gchar *device, gint *bootpart_active, GError **error);
+
+/**
+ * Set the given boot partition (by index) active.
+ *
+ * @param device eMMC /dev path (/dev/mmcblkX)
+ * @param bootpart_active boot partition to set active (0 or 1)
+ * @param error return location for a GError, or NULL
+ *
+ * @return True if succeeded, False if failed
+ */
+gboolean r_emmc_write_bootpart(const gchar *device, gint bootpart_active, GError **error);
+
+/**
+ * Set eMMC boot partition to read-only.
+ *
+ * @param device eMMC boot partition /dev path (/dev/mmcblkXbootY)
+ * @param error return location for a GError, or NULL
+ *
+ * @return True if succeeded, False if failed
+ */
+gboolean r_emmc_force_part_ro(const gchar *device, GError **error);
+
+/**
+ * Set eMMC boot partition to read-write.
+ *
+ * @param device eMMC boot partition /dev path (/dev/mmcblkXbootY)
+ * @param error return location for a GError, or NULL
+ *
+ * @return True if succeeded, False if failed
+ */
+gboolean r_emmc_force_part_rw(const gchar *device, GError **error);
--- a/src/emmc.c
+++ b/src/emmc.c
@@ -1,0 +1,178 @@
+#include <config.h>
+
+#include <errno.h>
+#include <fcntl.h>
+#include <glib/gstdio.h>
+#include <linux/major.h>
+#include <linux/mmc/ioctl.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include "emmc.h"
+#include "update_handler.h"
+
+
+static int r_emmc_read_extcsd(int fd, guint8 extcsd[512]) {
+	struct mmc_ioc_cmd cmd = {};
+
+	cmd.write_flag = 0;
+	cmd.opcode = MMC_SEND_EXT_CSD;
+	cmd.arg = 0;
+	cmd.flags = MMC_RSP_SPI_R1 | MMC_RSP_R1 | MMC_CMD_ADTC;
+	cmd.blksz = 512;
+	cmd.blocks = 1;
+	mmc_ioc_cmd_set_data(cmd, extcsd);
+
+	return ioctl(fd, MMC_IOC_CMD, &cmd);
+}
+
+gboolean r_emmc_read_bootpart(const gchar *device, gint *bootpart_active, GError **error) {
+	gint ret;
+	gboolean res = FALSE;
+	guint8 extcsd[512];
+	int fd;
+	/* count from 1 */
+	gint active_partition = -1;
+
+	fd = g_open(device, O_RDONLY);
+	if (fd == -1) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"opening eMMC device failed: %s", g_strerror(errno));
+		goto out;
+	}
+
+	ret = r_emmc_read_extcsd(fd, extcsd);
+	if (ret) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Could not read from extcsd register %d in %s\n",
+				EXT_CSD_PART_CONFIG, device);
+		goto out;
+	}
+
+	/* retrieve active partition from BOOT_PART register */
+	active_partition = (extcsd[EXT_CSD_PART_CONFIG] & 0x38) >> 3;
+
+	/* return the partitions counted from 0 */
+	*bootpart_active = active_partition - 1;
+
+	res = TRUE;
+
+out:
+	if (fd >= 0)
+		g_close(fd, NULL);
+
+	return res;
+}
+
+static gint r_emmc_write_extcsd(int fd, guint8 index, guint8 value) {
+	struct mmc_ioc_cmd cmd;
+
+	memset(&cmd, 0, sizeof(cmd));
+
+	cmd.write_flag = 1;
+	cmd.opcode = MMC_SWITCH;
+	cmd.arg = (MMC_SWITCH_MODE_WRITE_BYTE << 24) | (index << 16) |
+	          (value << 8) | EXT_CSD_CMD_SET_NORMAL;
+	cmd.flags = MMC_RSP_SPI_R1B | MMC_RSP_R1B | MMC_CMD_AC;
+
+	return ioctl(fd, MMC_IOC_CMD, &cmd);
+}
+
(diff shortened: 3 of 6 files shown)


include/emmc.h
```diff
@@ -0,0 +1,111 @@
...
+
+#define EMMC_BOOT_PARTITIONS			2
+#define INACTIVE_BOOT_PARTITION(part_active)	((part_active + 1) % EMMC_BOOT_PARTITIONS)
+
+#define EXT_CSD_PART_CONFIG			179
+#define EXT_CSD_CMD_SET_NORMAL			(1 << 0)
+#define EXT_CSD_BOOT_CFG_ACC			(0x07)
+
+/* From kernel linux/major.h */
+#define MMC_BLOCK_MAJOR				179
```
jluebbe: This is unused.

src/emmc.c
```diff
@@ -0,0 +1,189 @@
...
+#include <linux/mmc/ioctl.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include "emmc.h"
+#include "update_handler.h"
+
+
+int r_emmc_read_extcsd(int fd, __u8 *extcsd) {
```
jluebbe: This can be static. Remove it from the header es well.

src/emmc.c
```diff
@@ -0,0 +1,189 @@
...
+	res = TRUE;
+
+out:
+	if (fd)
+		close(fd);
+
+	return res;
+}
+
+gboolean r_emmc_force_part_write(const gchar *device, gchar value, GError **error) {
```
jluebbe: This seems to be internal as well, so make it static.

src/update_handler.c
```diff
@@ -1034,6 +1083,160 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+	gboolean res = FALSE;
+	int out_fd;
+	int ret = 0;
+	int part_active;
+	int part_active_after;
+	GOutputStream *outstream = NULL;
+	GError *ierror = NULL;
+	RaucSlot *part_slot = NULL;
+
+	g_debug("img_to_boot_emmc_handler");
```
jluebbe: Remove this.

src/update_handler.c
```diff
@@ -56,6 +57,54 @@ static GOutputStream* open_slot_device(RaucSlot *slot, int *fd, GError **error)
 	return outstream;
 }
 
+static gboolean empty_slot(RaucSlot *slot, GError **error)
```
jluebbe: Maybe rename this to "clear_slot"?

src/update_handler.c
```diff
@@ -56,6 +57,54 @@ static GOutputStream* open_slot_device(RaucSlot *slot, int *fd, GError **error)
...
+{
+	GError *ierror = NULL;
+	GInputStream *instream;
+	GOutputStream *outstream = NULL;
+	GFile *zerosrc = g_file_new_for_path("/dev/zero");
```
jluebbe: Using /dev/zero from C seems strange. We could just write a zeroed page from memory until the device is full.

BastianStender:
Pushed:
- squashed @jluebbe's doc fixup
- removed `MMC_BLOCK_MAJOR`, included `linux/major.h` instead (needed to make `MMC_IOC_CMD` macro work)
- made `r_emmc_read_extcsd()`, `r_emmc_write_extcsd()`, `r_emmc_force_part_write()` static, removed from header
- renamed `empty_slot()` to `clear_slot()`
- `clear_slot()`: write zero buffer instead of GInputStream from `/dev/zero`
- unnecessary debug output removed

BastianStender:
Pushed:
- `clear_slot()`: fix indent (uncrustified)

docs/advanced.rst
```diff
@@ -517,39 +519,66 @@ non-bootable system caused by an error or power-loss during the update.
...
-* Use RAUC only to update the rootfs. The combinations to test
+* Use RAUC to update the rootfs. The combinations to test
   can be reduced by limiting which old versions are supported by an update.
 * Reboot into the new system.
 * On boot, before starting the application, check that the current slot
   is 'sane'. Then check if the installed bootloader is older than the
   version shipped in the (new) rootfs. In that case:
 
-  * Disable the old rootfs slot and update the bootloader.
+  * Disable the old rootfs slot and update the bootloader via
```
ejoerns: Mh... not sure if this isn't a bit broken. Doesn't write-slot require you to define the slot in the system.conf despite we intentionally do not want to handle this slot with RAUC?..

src/emmc.c
```diff
@@ -0,0 +1,190 @@
...
+	if (fd)
+		close(fd);
+
+	return res;
+}
+
+static int r_emmc_write_extcsd(int fd, __u8 index, __u8 value) {
+	struct mmc_ioc_cmd cmd;
+
+	memset(&cmd, 0, sizeof(cmd));
```
jluebbe: You can use struct mmc_ioc_cmd cmd = {}; instead of the memset.

src/emmc.c
```diff
@@ -0,0 +1,190 @@
...
+static gboolean r_emmc_force_part_write(const gchar *device, gchar value, GError **error) {
+	gboolean ret = FALSE;
+	gchar **device_split = NULL;
+	const gchar *sysfs_path = NULL;
+	FILE *f = NULL;
+
+	g_assert(value == '0' || value == '1');
+
+	/* extract mmcblkXbootY from /dev/mmcblkXbootY */
+	device_split = g_strsplit(device, "/", 3);
```
jluebbe: You could use g_path_get_basename(), which never fails.

src/emmc.c
```diff
@@ -0,0 +1,190 @@
...
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Could not generate sysfs path from device: split failed");
+		goto out;
+	}
+
+	sysfs_path = g_strdup_printf("/sys/block/%s/force_ro", device_split[2]);
+
+	if (!g_file_test(sysfs_path, G_FILE_TEST_EXISTS)) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Could not generate sysfs path from device: %s does not exist",
```
jluebbe: The problem is not generating the path here, but the configured block devices doesn't have the force_ro attribute.

src/emmc.c
```diff
@@ -0,0 +1,190 @@
...
+	cmd.blocks = 1;
+	mmc_ioc_cmd_set_data(cmd, extcsd);
+
+	return ioctl(fd, MMC_IOC_CMD, &cmd);
+}
+
+gboolean r_emmc_read_bootpart(const gchar *device, gint *bootpart_active, GError **error) {
+	int ret;
+	gboolean res = FALSE;
+	__u8 extcsd[512];
```
ejoerns: maybe prefer using guint8 here as we normally use glib types.

src/update_handler.c
```diff
@@ -56,6 +59,51 @@ static GOutputStream* open_slot_device(RaucSlot *slot, int *fd, GError **error)
...
+	while (write_count != -1) {
+		write_count = g_output_stream_write(outstream, zerobuf, CLEAR_BLOCK_SIZE, NULL,
+				&ierror);
+		/*
+		 * G_IO_ERROR_NO_SPACE is expected here, because the block
+		 * device is cleared completely
+		 */
+		if (write_count == -1 &&
+		    !g_error_matches(ierror, G_IO_ERROR, G_IO_ERROR_NO_SPACE)) {
+
```
jluebbe: Remove empty line.

src/emmc.c
```diff
@@ -0,0 +1,190 @@
...
+	int fd;
+	/* count from 1 */
+	int active_partition = -1;
+
+	fd = open(device, O_RDONLY);
+
```
ejoerns: please remove newline here and in similar cases (in some you already did that).

src/update_handler.c
```diff
@@ -56,6 +59,51 @@ static GOutputStream* open_slot_device(RaucSlot *slot, int *fd, GError **error)
...
+	}
+
+	res = TRUE;
+
+out:
+	g_clear_object(&outstream);
+	return res;
+}
+
+
```
jluebbe: Remove empty line.

src/update_handler.c
```diff
@@ -1034,6 +1082,158 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+				NULL,
+				dest_slot,
+				&ierror);
+		if (!res) {
+			g_propagate_error(error, ierror);
+			goto out;
+		}
+	}
+
+	/* read active boot partition vom ext_csd */
```
jluebbe: "from" instead of "vom"

src/update_handler.c
```diff
@@ -1034,6 +1082,158 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+
+	/* create a temporary RaucSlot with the actual partition as device */
+	part_slot = g_new0(RaucSlot, 1);
+
+	/* determine active boot partition and set the other to be written */
+	part_slot->device = g_strdup_printf(
+			"%sboot%d",
+			dest_slot->device,
+			INACTIVE_BOOT_PARTITION(part_active));
+
```
jluebbe: Remove empty line.

src/emmc.c
```diff
@@ -0,0 +1,190 @@
...
+	if (fd == -1) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"opening output device failed: %s", strerror(errno));
```
ejoerns: use `g_strerror()` here

docs/advanced.rst
BastianStender: True. Any suggestions? Either define the slot in the system.conf without providing updates in bundles, or write the bootloader without `rauc slot-write`.

src/emmc.c
BastianStender: Mh, okay. `strerror()` is used in other places too, e.g. [here](https://github.com/rauc/rauc/blob/master/src/update_handler.c#L41) and [here](https://github.com/rauc/rauc/blob/master/src/update_handler.c#L68). More patch opportunities :)

BastianStender:
Pushed changes:
- use glib types where appropriate
- remove empty lines where appropriate
- `r_emmc_read_extcsd()`: expect `guint8 extcsd[512]` as argument to allow size checking by compiler
- `r_emmc_read_bootpart()`/`r_emmc_read_bootpart()`: mention eMMC device in error messages
- `r_emmc_read_bootpart()`/`r_emmc_read_bootpart()`: close file only if `fd > 0`
- docs/advanced: do not recommend `rauc write-slot` if RAUC should not handle a slot in "old system with a new bootloader" scenario

BastianStender:
Pushed changes:
- `r_emmc_read_extcsd`: simplify `mmc_ioc_cmd` struct init with `{}`
- `clear_slot()`: use shorter struct init `{}` instead of `{ 0 }`
- `r_emmc_read_bootpart()`/`r_emmc_read_bootpart()`: close file only if `fd >= 0`
- src/emmc.c: include `config.h`

BastianStender:
Pushed changes:
- `r_emmc_force_part_write()`: simplify basename extraction via `g_path_get_basename`

BastianStender:
Pushed changes:
- src/emmc.c: drop const keyword as it is not really needed here and complicates freeing; add missing `g_free()`

BastianStender:
Ah, thanks for that :)

BastianStender:
Pushed changes:
- squashed @jluebbe's const fixup

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+				dest_slot,
+				&ierror);
+		if (!res) {
+			g_propagate_error(error, ierror);
+			goto out;
+		}
+	}
+
+	/* read active boot partition from ext_csd */
+	ret = r_emmc_read_bootpart(dest_slot->device, &part_active, &ierror);
```
ejoerns: return type of `r_emmc_read_bootpart()` is gboolean, thus do not assign it to a `gint` variable. Use `res` here instead, too, which would also save a line of code (`res = FALSE`).

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+		res = FALSE;
+		goto out;
+	}
+	g_message("Found active eMMC boot partition %sboot%d",
+			dest_slot->device, part_active);
+
+	/* create a temporary RaucSlot with the actual partition as device */
+	part_slot = g_new0(RaucSlot, 1);
+
+	/* determine active boot partition and set the other to be written */
```
ejoerns: misleading as we determined the active one already. Maybe the prior comment already describes both and you could just remove the empty line between them.

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+	ret = r_emmc_read_bootpart(dest_slot->device, &part_active, &ierror);
+	if (!ret) {
+		g_propagate_error(error, ierror);
+		res = FALSE;
+		goto out;
+	}
+	g_message("Found active eMMC boot partition %sboot%d",
+			dest_slot->device, part_active);
+
+	/* create a temporary RaucSlot with the actual partition as device */
```
ejoerns: maybe s/actual partition/actual (currently inactive) boot partition/ ?

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+	g_message("Disabling read-only mode of slot device partition %s",
+			part_slot->device);
+	res = r_emmc_force_part_rw(part_slot->device, &ierror);
+	if (!res) {
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	/* clear block device */
+	g_message("Clearing slot device %s", part_slot->device);
```
ejoerns: The messages are a bit inconsistent when speaking about `part_slot->device` as either *slot device* or *slot device partition*.

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	/* sanity check: read active boot partition from ext_csd
+	 *
+	 * Read explicitly from root device (this forces another kernel
+	 * partition switch and should trigger the ext_csd bug more reliably.
+	 */
+	ret = r_emmc_read_bootpart(dest_slot->device, &part_active_after, &ierror);
```
ejoerns: See note about `gboolean` above.

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+	res = r_emmc_write_bootpart(
+			part_slot->device,
+			INACTIVE_BOOT_PARTITION(part_active),
+			&ierror);
+	if (!res) {
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	/* sanity check: read active boot partition from ext_csd
```
ejoerns: s/read/readd/

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+	 *
+	 * Read explicitly from root device (this forces another kernel
+	 * partition switch and should trigger the ext_csd bug more reliably.
```
ejoerns: close bracket here

src/emmc.c
```diff
@@ -0,0 +1,178 @@
...
+	return res;
+}
+
+static gboolean r_emmc_force_part_write(const gchar *device, gchar value, GError **error) {
+	gboolean ret = FALSE;
+	gchar *device_basename = g_path_get_basename(device);
+	gchar *sysfs_path = NULL;
+	FILE *f = NULL;
+
+	g_assert(value == '0' || value == '1');
```
ejoerns: Please use `g_return_{val_,}if_fail()` for checking input args of functions. Also consider checking other inputs, such as GError, too.

include/emmc.h
```diff
@@ -0,0 +1,76 @@
...
+#define MMC_RSP_R1				(MMC_RSP_PRESENT|MMC_RSP_CRC|MMC_RSP_OPCODE)
+#define MMC_RSP_R1B				(MMC_RSP_PRESENT|MMC_RSP_CRC|MMC_RSP_OPCODE|MMC_RSP_BUSY)
+
+
+/**
+ * Reads the active eMMC boot partition index of given eMMC device into the
+ * given variable.
+ *
+ * @param device eMMC /dev path (/dev/mmcblkX)
+ * @param bootpart_active will contain the active boot partition (0 or 1)
```
ejoerns: If I get it right, bootpart_active may also contain value `6` when the user area was enabled for boot

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+	}
+
+	/* read active boot partition from ext_csd */
+	ret = r_emmc_read_bootpart(dest_slot->device, &part_active, &ierror);
+	if (!ret) {
+		g_propagate_error(error, ierror);
+		res = FALSE;
+		goto out;
+	}
+	g_message("Found active eMMC boot partition %sboot%d",
```
ejoerns: This printout will be wrong if `part_active` is 6.

src/update_handler.c
```diff
@@ -1034,6 +1080,157 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+			dest_slot->device, part_active);
+
+	/* create a temporary RaucSlot with the actual partition as device */
+	part_slot = g_new0(RaucSlot, 1);
+
+	/* determine active boot partition and set the other to be written */
+	part_slot->device = g_strdup_printf(
+			"%sboot%d",
+			dest_slot->device,
+			INACTIVE_BOOT_PARTITION(part_active));
```
ejoerns: This will silently switch booting from user partittion to boot1. Might be intended but should at least get an info message, what do you think?

src/update_handler.c
BastianStender: What does this mean?

src/emmc.c
BastianStender: I see the point. Is this something you are trying to enforce everywhere now? There are different variants of g_assert used for argument checking in the very same file, e.g. [here](https://github.com/rauc/rauc/blob/master/src/update_handler.c#L557).

src/update_handler.c
ejoerns: Skip that :D seems I confused while scrolling through the code and assumed this to be a re-add from 'adding again' but its indeed a simple 'read' ;)

src/emmc.c
ejoerns: Yes, we started with `g_assert` in some points but we would like to stick to glib convention and use g_return.. functions here. We tried to enforce this in the last patches added (while not being sure if we succeeded). The intention is to fix this with all newly touched code and later on also in existing code. Anyway, there a still a bunch of functions where we don't do any checks at all. Good patch opportunities.. ;)

src/update_handler.c
BastianStender: Yes, that's refactor fallout, will fix.

src/update_handler.c
BastianStender: Ok.

src/update_handler.c
BastianStender: Ok.

src/update_handler.c
BastianStender: Ok, will clarify.

src/update_handler.c
BastianStender: True.

src/update_handler.c
BastianStender: Ok.

include/emmc.h
BastianStender: Yes, the docstring is wrong.

src/update_handler.c
BastianStender: Yes, that's misleading. Will fix it.

src/update_handler.c
BastianStender: Yes, it's intended, will add comment + appropriate message

BastianStender:
> One genereal concern is that the img_to_boot_emmc_handler looks a bit too verbose compared to the others. Maybe we should consider moving some of the messages to debug level?

Yep, will do.

> One last thing to mention in general is that the usage of error class and number is a bit too untargeted imho. You use the tuple R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED for every single error which basically is not what the error classes and numbers are designed for. Might it be worth adding an eMMC error class here? For defining further R_UPDATE errors, the errors might be a bit too specific
> 
> I could image something like
> R_EMMC_EXTCSD_{READ,WRITE}_FAILED, R_EMMC_ERROR, ...
> 
> @jluebbe what do you think about the error case?

We discussed this internally. If error handling is really used anywhere various functions need to be adapter. So we will leave it like this.

BastianStender:
Pushed changes as fixup:
- `r_emmc_read_bootpart()`: document values for "user partition" and "disabled"
- `r_emmc_read_bootpart()`: replace `g_assert()` with `g_return_val_if_fail()`
- `img_to_boot_emmc_handler()`: use correct return types and simplify
- `img_to_boot_emmc_handler()`: fix messages about user/disabled boot partition
- `img_to_boot_emmc_handler()`: make comments about device/partition more comprehensive
- `img_to_boot_emmc_handler()`: make it less verbose (rw/ro mode, toggle message are now debug)

Also: rebased on current master

BastianStender:
Pushed changes:
- `img_to_boot_emmc_handler()`: fix comment copy/paste error
- squashed

PR #245, opened 2018-03-09
Add of required parameters to create casync bundle.

potens1:
Signed-off-by: Nicolas Di Pietro <potens1@gmail.com>


potens1:
I don't know if those parameters are also required for the `install` operation, if those are required I can do another commit for that

ejoerns:
No, the install option naturally only need the keyring (for verifying the udpate). This is normally performed on the target system which has a `system.conf` that sets the proper keyring path.

PR #256, opened 2018-04-03
use g_auto helpers for automatic cleanup 

pH5:
Bump the required GLib version to 2.45.8 and use g_auto helpers for automatic cleanup.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -687,7 +687,6 @@ static gchar* info_formatter_json_base(RaucManifest *manifest, gboolean pretty)
 #if ENABLE_JSON
 	g_autoptr(JsonGenerator) gen = NULL;
 	g_autoptr(JsonNode) root = NULL;
-	gchar *str;
 	g_autoptr(JsonBuilder) builder = json_builder_new();
 
 	json_builder_begin_object(builder);
@@ -753,9 +752,7 @@ static gchar* info_formatter_json_base(RaucManifest *manifest, gboolean pretty)
 	root = json_builder_get_root(builder);
 	json_generator_set_root(gen, root);
 	json_generator_set_pretty(gen, pretty);
-	str = json_generator_to_data(gen, NULL);
-
-	return str;
+	return json_generator_to_data(gen, NULL);
 #else
 	g_error("json support is disabled");
 	return NULL;
@@ -1045,7 +1042,6 @@ static gchar* r_status_formatter_json(gboolean pretty)
 	g_autoptr(JsonGenerator) gen = NULL;
 	g_autoptr(JsonNode) root = NULL;
 	GHashTableIter iter;
-	gchar *str;
 	g_autoptr(JsonBuilder) builder = json_builder_new();
 	GError *ierror = NULL;
 	RaucSlot *slot, *primary = NULL;
@@ -1170,9 +1166,7 @@ static gchar* r_status_formatter_json(gboolean pretty)
 	root = json_builder_get_root(builder);
 	json_generator_set_root(gen, root);
 	json_generator_set_pretty(gen, pretty);
-	str = json_generator_to_data(gen, NULL);
-
-	return str;
+	return json_generator_to_data(gen, NULL);
 #else
 	g_error("json support is disabled");
 	return NULL;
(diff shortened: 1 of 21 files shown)


src/main.c
```diff
@@ -768,8 +756,6 @@ static gchar* info_formatter_json_base(RaucManifest *manifest, gboolean pretty)
 	str = json_generator_to_data(gen, NULL);
 
 	json_node_free(root);
```
mnhu: Without looking into the json-api: Should the root pointer still be freed when declared as autoptr?

src/main.c
```diff
@@ -1193,10 +1174,6 @@ static gchar* r_status_formatter_json(gboolean pretty)
 	json_generator_set_pretty(gen, pretty);
 	str = json_generator_to_data(gen, NULL);
 
-	json_node_free(root);
```
mnhu: Here the json root node pointer is freed as opposed to my previous comment.

src/main.c
```diff
@@ -1654,7 +1628,7 @@ static void cmdline_handler(int argc, char **argv)
 	g_free(text);
 
 done:
-	g_clear_pointer(&context, g_option_context_free);;
+	return;
```
mnhu: Why the return statement here at the very end of a void function?

src/main.c
pH5: Absolutely not, thanks for catching this!
json_node_free is the autoptr cleanup function registered for the JsonNode type, so I have to remove this as well.

src/main.c
pH5: So I can keep the "done:" label and not have to change all the "goto done;"s in this function in this patch.

ejoerns:
Just added two further fixes that I found during second review

src/manifest.c
```diff
@@ -136,12 +136,9 @@ static gboolean parse_image(GKeyFile *key_file, const gchar *group, RaucImage **
 	g_key_file_remove_group(key_file, group, NULL);
 
 	res = TRUE;
-	*image = iimage;
+	*image = g_steal_pointer(&iimage);
 
 out:
```
michaelolbrich: the label is no longer necessary. Just return above. And remove `res`.

src/manifest.c
```diff
@@ -299,20 +296,15 @@ static gboolean parse_manifest(GKeyFile *key_file, RaucManifest **manifest, GErr
 	g_strfreev(groups);
 
 	res = TRUE;
+	*manifest = g_steal_pointer(&raucm);
 free:
```
michaelolbrich: same here.

src/manifest.c
```diff
@@ -338,14 +330,13 @@ gboolean load_manifest_mem(GBytes *mem, RaucManifest **manifest, GError **error)
 	}
 
 out:
```
michaelolbrich: and here.

src/manifest.c
```diff
@@ -473,17 +460,13 @@ gboolean save_manifest_file(const gchar *filename, RaucManifest *mf, GError **er
 
 		if (file->filename)
 			g_key_file_set_string(key_file, group, "filename", file->filename);
-
-		g_free(group);
 	}
 
 	res = g_key_file_save_to_file(key_file, filename, NULL);
```
michaelolbrich: just return here.

michaelolbrich:
There are several more res / goto out combinations that are no longer needed.

ejoerns:
@michaelolbrich Yes, I've seen them but decided to remove them step by step in separate patches to not block this more than required and to not let the patches grow too much.

But maybe we can add fixes for some very obvious ones with low loc impact.

ejoerns:
@michaelolbrich, @pH5 I've pushed a fixup now. Please let me know if your are ok with this and if we can merge.

michaelolbrich:
The fixup is ok.
I still feel that the now unnecessary gotos should be removed as well, but nothing looks broken.

ejoerns:
@michaelolbrich I've already prepared some patches for that, but not entirely finished. If you promise to have a look on these too when I finish and push them, we can still do it in this PR :)

michaelolbrich:
@ejoerns I think we can merge this PR. Just ping me when the followup PR is ready

ejoerns:
Ok, sounds like a plan! Then approve it, to give me a good feeling when pushing the big button ;)

ejoerns:
Squashed and rebased onto master

PR #260, opened 2018-04-06
service: report system status via GetSlotStatus

jremmet:
"raus status" reports also status not linked to a single slot:
"compatible", "variant", "booted", "boot_primary"
Add them to GetSlotStatus. This adds a extra parameter to the method.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -451,7 +451,7 @@ Methods
 
 :ref:`Mark <gdbus-method-de-pengutronix-rauc-Installer.Mark>` (IN  s state, IN  s slot_identifier, s slot_name, s message);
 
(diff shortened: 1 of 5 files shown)


BastianStender:
We are currently investigating whether and how to restructure the D-BUS API. So it might take a while until we decide what to do with this.

Nonetheless `GetSlotStatus` is not the correct method to add global system status information like these. I would expect a separate method/separate properties for this.

jremmet:
I will rework it to use separate properties for this "static" status properties.

jremmet:
I will rework it to use separate properties for this  status properties.

jremmet:
Reworked using static properties #269

PR #261, opened 2018-04-09
src/update_handler: support img to ext4

mhei:
This adds support for writing .img files to ext4 partitions.

The .img extension is traditionally used for various image types,
so for ext4 filesystem images, too.

Let's add this specific update pair to force usage of
img_to_fs_handler, and thus to allow running hooks with
target filesystem mounted.


ejoerns:
So far we have tried to use the default image extensions for dealing with image type matching. It's not my strongest opinion but I would expect that to be a task that one can give back to the image creation mechanism that should create `.ext4` extensions instead of `.img` ones.

Do you have an actual use case for that?

@jluebbe What do you think about that?

mhei:
At I2SE, we use a self-made image creation "tool" we named [STIB](https://github.com/I2SE/stib) - basically  some handy Makefiles to assemble a Debian based root filesystem. Currently, it produces `.img` files, but yes, this could also be changed to create `.ext4` files. However, I still think, that it would be more intuive for rauc users, if this pair would be added, so this is the reason for this PR.

jluebbe:
ext4 is not special (compared to vfat or ubifs, for example), so if we support this generic extension for ext4, it should also be supported for the others. We'd also need documentation for this new behavior, suggesting to use explicit extensions when possible, as this allows checking during installation that the slot type is correct.

ejoerns:
@mhei do you still require this? Do you plan to spent further effort on an extended patch including support for all slot types with added documentation?
Otherwise I would like to close this PR in favor of the current 'explicit' solution.

mhei:
@ejoerns I've update the PR, please have a look whether I understood your comments correctly. As already stated, it's not a strong requirement on my side. However, I did not find the current 'explicit' behavior documented? This was the reason I used `.img` and then wondered why the hooks run without mounted filesystem.

jluebbe:
@mhei I've changed tothe img_to_ubifs_handler for `*.img` on `ubifs` slots. Otherwise it seems good now.

mhei:
Thank you all for fixing my PR, seems it was little bit too late already...

PR #262, opened 2018-04-09
doc: document unsquashfs requirement

mhei:

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -23,7 +23,7 @@ In principle, these are:
 * The bootloader
 * The Linux kernel
 * The init system
-* System utilities (mount, mkfs, ...)
+* System utilities (mount, mkfs, unsquashfs, ...)
 * The update tool, RAUC itself
 
 .. note::
@@ -171,6 +171,8 @@ Note that build systems may handle parts of these dependencies automatically,
 but also in this case you will have to select some of them manually as RAUC
 cannot fully know how you intend to use your system.
 
+:General: unsquashfs (from `Squashfs
+          <http://squashfs.sourceforge.net/>`_)
 :NAND Flash: nandwrite (from `mtd-utils
              <git://git.infradead.org/mtd-utils.git>`_)
 :UBIFS: mkfs.ubifs (from `mtd-utils


docs/integration.rst
```diff
@@ -171,6 +171,8 @@ Note that build systems may handle parts of these dependencies automatically,
 but also in this case you will have to select some of them manually as RAUC
 cannot fully know how you intend to use your system.
 
+:General: unsquashfs (from `Squashfs
+          <http://squashfs.sourceforge.net/>`_)
```
ejoerns: I fear this mixes things that should not be mixed. The other tools are required to extract or install the update on the target. Instead, `unsquashfs` is neither used nor required for this.

The only situations where unsquashfs is used are calls on `rauc info` (for extracting the manifest without unpacking the bundle)  and `rauc convert` (for converting it into a casync bundle). During normal extraction, the bundle only becomes mounted to extract the actual images.

mhei:
I understand your points. My background is that I tried to create a Debian package ([see this branch](https://github.com/mhei/rauc/tree/debianize), binary packages can be found at [packagecloud.io](https://packagecloud.io/i2se/debian-bsp-addons)) and then tested all `rauc ` commands. `rauc info` failed, so I simply added unsquashfs to the dependencies and thought it would be worth to mention it here, too.

But yes, it feels "tinkered" at the moment...

ejoerns:
To lead this to a proper solution, I would propose adding something like

``` patch
diff --git a/docs/integration.rst b/docs/integration.rst
index a2f56341..b2f7b46a 100644
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -188,6 +188,8 @@ cannot fully know how you intend to use your system.
 :vfat: mkfs.vfat (from `dosfstools
                   <https://github.com/dosfstools/dosfstools>`_)
 
+Note that for running `rauc info` on the target (as well as on the host), you
+also need to have ``unsquashfs`` tool installed.
 
 Interfacing with the Bootloader
 -------------------------------
```

What do you think?

mhei:
LGTM 👍

ejoerns:
I've opened #279 to resolve this and mentioned you as 'Suggested-by' if that is ok for you. Will close this one for now

PR #270, opened 2018-05-03
Try to unmount mounted target slots instead of straight failure

leiflm:
**Summary**: This tries to work around a case where a targeted destination partition might be mounted when rauc (service) was started for some reason (`rauc status mark-good`).

**Old behavior**: rauc aborts.

**New behavior**: rauc tries to `umount` the slot's device before giving up.

**Note**: rauc *should* watch mtab during runtime so that an unmount could be handled via a hook script. It breaks the mount iso-/(insu?)lation.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -860,9 +860,18 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 
 		if (dest_slot->mount_point || dest_slot->ext_mount_point) {
 			res = FALSE;
-			g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_MOUNTED,
-					"Destination device '%s' already mounted", dest_slot->device);
-			goto out;
+			if (dest_slot->state == ST_INACTIVE) {
+				res = r_umount(dest_slot->device, error);
+			}
+			if (res) {
+				g_clear_pointer(&dest_slot->mount_point, g_free);
+				g_clear_pointer(&dest_slot->ext_mount_point, g_free);
+			} else {
+				g_propagate_error(error, ierror);
+				g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_MOUNTED,
+						"Destination device '%s' already mounted", dest_slot->device);
+				goto out;
+			}
 		}
 
 		/* Verify image checksum (for non-casync images) */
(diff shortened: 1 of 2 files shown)


src/install.c
```diff
@@ -860,9 +860,18 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
...
-					"Destination device '%s' already mounted", dest_slot->device);
-			goto out;
+			if (dest_slot->state == ST_INACTIVE) {
+				res = r_umount(dest_slot->device, error);
+			}
+			if (res) {
+				g_clear_pointer(&dest_slot->mount_point, g_free);
+				g_clear_pointer(&dest_slot->ext_mount_point, g_free);
+			} else {
+				g_propagate_error(error, ierror);
```
leiflm: I'm not really sure about this call/its necessity.

ejoerns:
This basically solves a similar problem as #186 intends to do.

I prefer the handling of #186 because it is explicit.

The general idea of partition handling is, that the (inactive) partitions handled by RAUC are under control of RAUC only. Thus, by default, the target system should not access an inactive partition from the active system.
If it does instead, it would be interesting why! I assume it to be really unsafe to simply call an unmount on a partition when we do not know why it was mounted despite we expect it to be not!

A proper use case is what @mnhu describes in #186, a non-redundant boot partition of an RPI that *needs* to be mounted by the active system, too. But, in this case we know that this is a valid case and can explicitly specify this in the config file. This is baiscally what #186 does.

What is your user case for the patch above? When did you hit the problem that an inactive slot was mounted when RAUC wanted to update it?

leiflm:
> What is your user case for the patch above? When did you hit the problem that an inactive slot was mounted when RAUC wanted to update it?

We're in the transition of a whole redundant system to a shared data-partition system. As it involves automatic repartitioning, we're mount-binding the data part of one specific rootfs-slot in the meanwhile .

Aside the issues [mentioned by @jluebbe](https://github.com/rauc/rauc/pull/186/files#r193800250) about systemd-`Mountflags`, I'm fine with this solution, as it is slot-specific configurable. Feel free to close this PR.

ejoerns:
Your case really sounds as if one would like to have a purpose-built hook for this at it is expected to happen only once during the devices life cylce and is not a standard behavior we want to handle.

As you suggested, I will close this one in favor of #186 despite already assuming that also #186 might not yet end in a mainline solution due to the unresolved systemd handling issue.

PR #271, opened 2018-05-13
emmc: add workaround for older kernels

yann-morin-1998:
Kernels up to (and including 3.3) forgot to include types.h in their
mmc/ioctl.h, and thus __u32 (and others) are not defined, causing
compilation errors:

    http://autobuild.buildroot.org/results/621/621587906bd2bb27c43b6fcbb051d75f20e32e7c/build-end.log

Fix that by explicitly including types.h before mmc/ioctl.h.

Signed-off-by: "Yann E. MORIN" <yann.morin.1998@free.fr>

### DIFF
--- a/src/emmc.c
+++ b/src/emmc.c
@@ -4,6 +4,7 @@
 #include <fcntl.h>
 #include <glib/gstdio.h>
 #include <linux/major.h>
+#include <linux/types.h> /* kernel < 3.4 forgot that in mmc/ioctl.h */
 #include <linux/mmc/ioctl.h>
 #include <string.h>
 #include <sys/ioctl.h>


yann-morin-1998:
Hmm... The check for the DCO is incomplete/incorrect...

As far as I understand, the SoB-line is expected to have an RFC2822-compliant email
address. RFC2822 states that a 'dot' is not an atom, so it but be in a double-quoted string.

My human-readable name contains a dot, so my name is double-quoted.

Emantor:
Hello, I looked through the docs for git-config and DCO, but could only find references to RFC2822 for some date formats in git. Can you link a source for RFC2822-compliant email addresses for either git or DCO?

Thanks

BastianStender:
The patch looks good to me.

The actual issue regarding the "Signed-off-by" is that your name (`git config --get user.name`) does not match the name in the SoB (without vs. with quotes), see [dco's source](https://github.com/probot/dco/blob/master/lib/dco.js#L32). Looking at other peoples SoBs I don't think you should quote your name. Just use `git commit --signoff` and you'll be fine.

Bastian

ejoerns:
@yann-morin-1998 Would you like to resend with the Signed-off-by fixed? Then we could simply merge it.

ejoerns:
Ping. Should not be that complicated ;)

yann-morin-1998:
@ejoerns Pong! Not complex, but only 24h per day... ;-)

ejoerns:
@yann-morin-1998 yes, truste me, I know about this limitation.. ;) Thanks for updating!

PR #274, opened 2018-05-18
contrib: cgi: add simple stand-alone CGI interface for RAUC

BastianStender:
The CGI binary is meant as a basic D-Bus example and starting point for
controlling the update client RAUC via very basic web mechanisms.


contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+ */
+static gint64 fread_chunked(char *ptr, size_t size, FILE *stream, GError **error)
+{
+	gint64 items_read = 0;
+	gint64 bytes_read = 0;
+	gint percentage = 0;
+	gint item_size = 8;
+
+	while (bytes_read < (gint64) size) {
+		items_read = fread(ptr + bytes_read, item_size, 1024, stream);
```
ejoerns: This will not work except for files which have a size that is multiple of item_size (8) bytes!

Either set `item_size` to 1 or use an appropriate glib method

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+			/* ?progress shows progress information */
+			if (!progress_status(installer, &error))
+				goto error;
+
+		} else if (g_strcmp0("status", query_string) == 0) {
+			/* ?status shows slot information */
+			if (!slot_status(installer, &error))
+				goto error;
+		} else {
+			g_set_error(&error, CGI_ERROR, CGI_ERROR_BAD_REQUEST, "Resource unavailable");
```
ejoerns: I would like a `"Resource '%s' unavailable", query_string` here

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+/*
+ * Retrieve ProgressUpdated property via rauc D-Bus service.
+ */
+static gboolean get_progress_updated(RInstaller *installer, gint *percentage, gchar **description, GError **error)
+{
+	gboolean res = FALSE;
+	GVariant* progress_updated;
+	GVariant* g_percentage;
+	GVariant* g_description;
+
```
ejoerns: Would be worth adding

```
        g_return_val_if_fail(percentage != NULL, FALSE);
        g_return_val_if_fail(description == NULL || *description == NULL, FALSE);
```

This also reveals that the `get_progress_updated()` method breaks the convention that GError functions returning false should have GError set.

Glib manual say:

> Avoid creating functions which have a boolean return value and a GError parameter, but where the boolean does something other than signal whether the GError is set. Among other problems, it requires C callers to allocate a temporary error. Instead, provide a "gboolean *" out parameter. There are functions in GLib itself such as g_key_file_has_key() that are deprecated because of this.

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+out:
+	return res;
+}
+
+/*
+ * Last installation was successfull if LastError D-Bus property is an empty
+ * string.
+ */
+static gboolean get_last_installation_success(RInstaller *installer, GError **error)
```
ejoerns: See my note about gboolean return with GError ** argument above.

contrib/cgi/README.rst
```diff
@@ -0,0 +1,201 @@
...
+like a web server would do. For better readability the JSON output is formatted
+*here only*:
+
+Status
+~~~~~~
+
+The slot status structure is directly based on the GVariant structure of
+`GetSlotStatus <http://rauc.reayydthedocs.io/en/latest/reference.html#the-getslotstatus-method>`_.
+
+::
```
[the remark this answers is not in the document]
BastianStender: There is a blank line. Did you accidentally review an old version of this PR? I fixed this line nearly 3 weeks ago..

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+}
+
+/*
+ * Handle CGI env variables and trigger requested operations.
+ */
+static gint cgi_handler(int argc, char **argv)
+{
+	gint ret = 1;
+	RInstaller *installer;
+	const gchar *method = g_getenv("REQUEST_METHOD");
```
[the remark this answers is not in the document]
BastianStender: Assuming the value of the `REQUEST_METHOD` env variable changes, yes. But I do not see why this should happen? CGI is pretty dumb, a web server calling `rauc-cgi` will not change it and we do not manipulate it neither.

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+	return ret;
+}
+
+/*
+ * Handle CGI env variables and trigger requested operations.
+ */
+static gint cgi_handler(int argc, char **argv)
+{
+	gint ret = 1;
+	RInstaller *installer;
```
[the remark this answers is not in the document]
BastianStender: Will do.

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+
+	return ret;
+}
+
+/*
+ * Handle CGI env variables and trigger requested operations.
+ */
+static gint cgi_handler(int argc, char **argv)
+{
+	gint ret = 1;
```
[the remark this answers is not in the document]
BastianStender: Yes, I decided to stick to gint here in order to return this value by `int main(int argc, char **argv)` directly. Should I change it nonetheless?

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+
+		/* no upload in progress */
+		if (!write_upload_status(-1, &error))
+			goto error;
+
+		print_headers("200 OK", "text/plain");
+		g_print("Upload and install trigger executed successfully.\n");
+
+	} else {
+		g_set_error(&error, CGI_ERROR, CGI_ERROR_METHOD_NOT_ALLOWED, "Unsupported method");
```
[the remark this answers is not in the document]
BastianStender: Okay, will add the unsupported method to the error message.

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+
+	res = TRUE;
+
+out:
+	json_node_free(root);
+	g_object_unref(gen);
+	g_object_unref(builder);
+	g_free(header_name_upper);
+	g_free(full_header_name);
+	g_free(header_value);
```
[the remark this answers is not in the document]
BastianStender: Will do.

contrib/cgi/README.rst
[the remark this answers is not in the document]
ejoerns: I initially intended to start review 3 weeks ago. Seems that GitHub automagic was not smart enough to drop the outdated comment from the review stack ;)
No fear, I reviewed your latest patches.

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,683 @@
...
+static gboolean stdin_to_file(GError **error)
+{
+	gint ret = FALSE;
+	gint64 read_len = 0;
+	gint64 bytes_read = 0;
+	gchar *file_buf = NULL;
+	const gchar *content_length = g_getenv("CONTENT_LENGTH");
+
+	if (!content_length) {
+		g_set_error(error, CGI_ERROR, CGI_ERROR_LENGTH_REQUIRED, "Please provide a Content-Length header.");
```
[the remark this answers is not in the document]
BastianStender: Okay.

contrib/cgi/src/cgi.c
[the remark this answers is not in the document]
ejoerns: Pinned one line too high, maybe. The code is

``` c
const gchar *method = g_getenv("REQUEST_METHOD");
const gchar *query_string = g_getenv("QUERY_STRING");
```

How I read the doc is that you cannot trust memory pointed at by `method` anymore after having read `QUERY_STRING`. Note that there is no string memory allocation involved.

contrib/cgi/src/cgi.c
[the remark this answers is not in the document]
ejoerns: I'm not religious in this case. You may also keep it as it is.

contrib/cgi/src/cgi.c
BastianStender: Oops, thanks for catching this.

contrib/cgi/src/cgi.c
BastianStender: Good point. I am not sure whether `503 Service Unavailable` is actually the correct HTTP code for this..

contrib/cgi/src/cgi.c
BastianStender: > Avoid creating functions which have a boolean return value and a GError parameter, but where the boolean does something other than signal whether the GError is set.

If I am not mistaken that is actually what the function does. The function has a boolean return value and a GError parameter and the booleans does signal whether the GError is set. So what's wrong with that?

contrib/cgi/src/cgi.c
BastianStender: Here your note from does actually apply, will fix it.

contrib/cgi/src/cgi.c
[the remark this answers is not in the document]
BastianStender: Ok, will `g_strdup()` to be sure.

contrib/cgi/src/cgi.c
ejoerns: Mh, I must have looked at another function while writing this...

Ah... it was the user of `get_progress_updated()` (I stumbled accross this when testing with `g_return_val_if_fail()`).

It is actually `get_will_reboot()` I was talking about.

BastianStender: Adding..
```
g_return_val_if_fail(percentage != NULL, FALSE);
g_return_val_if_fail(description == NULL || *description == NULL, FALSE);
```
..here would break the mentioned rule from GLib manual pages..

> Avoid creating functions which have a boolean return value and a GError parameter, but where the boolean does something other than signal whether the GError is set.

ejoerns: Yesss, but this is mainly for catching invalid API usage and not detecting and handling runtime errors. It does not signal a GError is set but at least an error condition. Let me say, if that produces an error, the missing content of GError is your smallest problem :laughing: 

As glib does it this way too I would expect it to be meant to do it this way.

BastianStender: Alright.

BastianStender:
Pushed requested changes:
- avoid boolean non-status return value with GError parameter
- catch invalid API usage with g_return_val_if_fail()
- fix `fread_chunked()` in case bundle size is not multiple of `item_size`
- `g_strdup()` const env variables as they are unstable
- add more meaningful error messages and use `g_set_error_literal()` where applicable
- free `RInstaller` after usage

Furthermore some improvements were pushed also:
- handle stale bundle/status files in `try_cleanup_status_on_install_failure()`
- do not set upload progress to `-1` when upload finished
- write negative progress in case of error
- README: fix small syntax errors

BastianStender:
Gentle ping

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,723 @@
...
+				print_headers("503 Service Unavailable", "text/plain");
+				break;
+			default:
+				print_headers("500 Internal Server Error", "text/plain");
+		}
+		g_printerr("%s\n", error->message);
+		g_error_free(error);
+	}
+
+out:
```
ejoerns: Not used and can be removed.

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,723 @@
...
+	json_builder_add_int_value(builder, upload_progress);
+
+	/*
+	 * value of a custom HTTP header that may be provided by the client
+	 * performing the upload used to identify who is currently uploading a
+	 * firmware
+	 */
+	header_ptr = &HTTP_HEADERS[0];
+	while (*header_ptr) {
+		header_name_upper = g_ascii_strup(*header_ptr, -1);
```
ejoerns: Scope of `header_name_upper`, `full_header_name` and `header_value` variables can and should be limited to while loop.

contrib/cgi/src/cgi.c
BastianStender: Yes, I left it there for structural reasons. Will remove it.

contrib/cgi/src/cgi.c
BastianStender: Right, will do.

BastianStender:
Pushed changes:
- removed unused `out` label
- moved `header_name_upper`, `full_header_name` and `header_value` in while loop scope

BastianStender:
Pushed changes:
- travis: build contrib/cgi/

ejoerns:
Pushed changes:

* uncrustify: also run on contrib/cgi

BastianStender:
@jluebbe What do you think about this?

contrib/cgi/src/cgi.c
```diff
@@ -0,0 +1,720 @@
...
+{
+	gint64 items_read = 0;
+	gint64 bytes_read = 0;
+	gint percentage = 0;
+
+	while (bytes_read < (gint64) size) {
+		items_read = fread(ptr + bytes_read, 1, 1024, stream);
+		bytes_read += items_read;
+		percentage = (bytes_read * 100) / size;
+		if (!write_upload_status(percentage, error)) {
```
jluebbe: We should call this only if the percentage has actually changed to reduce the overhead.

jluebbe: We should call this only if the percentage has actually changed to reduce the overhead.

PR #278, opened 2018-06-08
src/update_handler: add fsync for eMMC buffer before force_ro

jbrennan-impinj:
This fixes the eMMC so that the bootloader is completely written to
the boot partition before going to read only mode.  Otherwise,
the boot loader can get truncated.

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -1166,6 +1166,9 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot, 
 		goto out;
 	}
 
+	/* disable closing outstream file descriptor */
+	g_unix_output_stream_set_close_fd((GUnixOutputStream *)outstream, FALSE);
+
 	/* copy */
 	g_message("Copying image to slot device partition %s",
 			part_slot->device);
@@ -1174,6 +1177,20 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot, 
 		g_propagate_error(error, ierror);
 		goto out;
 	}
+
+	/* flush to block device before making eMMC boot partition read only */
+	if (fsync(out_fd) == -1) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED, "Output device synchronization failed: %s", strerror(errno));
+		res = FALSE;
+		goto out;
+	}
+
+	if (close(out_fd) == -1) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED, "Close output device failed: %s", strerror(errno));
+		res = FALSE;
+		goto out;
+	}
+	out_fd = 0;
 
 	/* reenable read-only on determined eMMC boot partition */
 	g_debug("Reenabling read-only mode of slot device partition %s",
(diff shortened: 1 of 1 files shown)


EugenWiens:
how does the error become noticeable?

src/update_handler.c
```diff
@@ -1175,6 +1178,10 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
 		goto out;
 	}
 
+	/* flush to block device before making eMMC boot partition read only */
+	fsync(out_fd);
```
ejoerns: `fsync()` may fail, thus please also check the return value here and report an error if we fail

jbrennan-impinj:
Yes, this started becoming a problem after I switched to 4.16.x kernel when writing to the eMMC.  About 30% of the time the eMMC would not finish flushing data before setting force_ro flag.   I was able to reproduce the problem by calling 'rauc write-slot bootloader.0 [bootloader file]' multiple times. 

I had also thought of placing this logic inside of copy_raw_image(), but wasn't sure if I could assume in the future that all callers would pass in a GUnixOutputStream.  If I can make that assumption , then I could extract the file descriptor from g_unix_output_stream_get_fd() and call fsync().

ejoerns:
I would say this is a valid assumption as we target files and devices on unix systems. But then we should make this explicit and use `GUnixOutputStream` in the function signature of `copy_raw_image`.

src/update_handler.c
```diff
@@ -1246,6 +1249,9 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
 	if (!res && part_slot)
 		r_emmc_force_part_ro(part_slot->device, NULL);
 
+	if ((out_fd > 0) && (close(out_fd) == -1))
+		g_error("Close output device failed: %s", strerror(errno));
```
ejoerns: g_error() is configured to abort() and core-dump. Thus it should not be used for handling possible runtime issues.

I would propose doing an uncontidional `close(out_fd)` in the error handling path of `fsync()` above and a normal `close(out_fd)` with normal error handling here.
This will prevent you from overriding a previous error caused by `fsync()` with a follow-up error caused by a failed `close()` while still having proper error reporting for both cases.

ejoerns:
@jbrennan-impinj would you like to rework it and provide a generic solution?

ejoerns:
We've merged the autoptr patch, thus please rebase on current master. The only change that affects you is that the free statements at the end of the function were removed.

One issue your code still has is that you will set an error but return `TRUE` if `close(out_fd)` fails.
Simply also set `res=FALSE` after the `g_set_error()` call.
Maybe it is also simpler for the structure to add a `close(out_fd)` to the `copy_raw_image()` error path, too and then move do

    if (close(out_fd) == -1) {
            g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED, "Close output device failed: %s", strerror(errno));
            res = FALSE;
            goto out;
    }

directly after the `fsync()` call.

But I would prefer the `GUnixOutputStream` solution anyway, which will then of course have a slightly different error handling structure. Thus, feel free to move it to `copy_raw_image()`.

ejoerns:
@jbrennan-impinj we think about doing a stable release v0.4.1 soon. This patch would be one of those I'd like to see in. Do you have time to care for it?

jbrennan-impinj:
I made the requested changes for the error path and tested against the emmc.  Now  passes every time.

By the way, I attempted to do the fsync inside of _copy_raw_image_() and was getting strange errors with the fsync.   Almost like the g_unix_output_stream_get_fd() doesn't return the proper file descriptor.   If you or anyone would like to go in this direction, the branch can be found at https://github.com/jbrennan-impinj/rauc/tree/jbrennan/emmc-fsync-copy-raw-image

ejoerns:
@jbrennan-impinj Isn't it that you simply forgot what you had in the above version of the fix already?

The `g_unix_output_stream_set_close_fd()` isn't set and thus the `g_output_stream_splice()` just closes the stream and the `out_fd` too early

ejoerns:
This is deprecated now by #296 which is based on the pre-work of @jbrennan-impinj and fixes the issue directly inside the `copy_raw_image()` method.

@jbrennan-impinj Would you like to have a look at it and possibly test it for your use-case?

PR #279, opened 2018-06-18
docs: integration: document need for 'unsquashfs' tool for 'rauc info'

ejoerns:
Suggested-by: Michael Heimpold michael.heimpold@i2se.com

Fixes: #262

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -161,6 +161,12 @@ following Busybox feature:
 
 -  ``CONFIG_FEATURE_TAR_AUTODETECT=y``
 
+Depending on the actual storage type and/or filesystem used, further target
+tool might be required.
+The documentation chapter
+`Required Target Tools <http://rauc.readthedocs.io/en/latest/integration.html#required-target-tools>`_
+gives a more detailed list on these.
+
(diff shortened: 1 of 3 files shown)


BastianStender:
Looks good, but shouldn't this get added to the `README.rst` (prerequisites section) as well?

ejoerns:
Mh, it is neither a build dependency nor a real target dependency.

We don't list required target tools there either, thus I would say it is ok to leave this out for the README.rst. Maybe one should add a generic note that depending on the underlying storage and file system types used further target tools can be required?

BastianStender:
I see. What about a link in the `README.rst` pointing to https://rauc.readthedocs.io/en/latest/integration.html#required-target-tools ?

BastianStender:
By the way, `git grep -C10 g_subprocess` reveals more external tools. At least "fakeroot" is not documented anywhere or am I missing something?

ejoerns:
@BastianStender is that more like what you would have expected? ;)

BastianStender:
Yes, very nice :+1:

PR #283, opened 2018-07-05
src/install: fix invalid return value in case of failed mark_active()

ejoerns:
935346f replaced the function `r_boot_set_primary()` with the function
`mark_active()`. While `r_boot_set_primary()` returned a gboolean that was
assigned to the return variable 'res', mark_active() does not.
As `res` will have been set to `TRUE` by the code flow before, this
absolutely requires to set 'res' manually to `FALSE` in case of an error
of `mark_active()`.
As this was missing, the installation was reported successfully even if
the actual slot activation failed.

This patch now fixes the return value of
`launch_and_wait_default_handler()` in case of a failed `mark_active()`
call by setting `res` to `FALSE` manually.


UVV-gh:
Thanks for the fix

PR #285, opened 2018-07-05
src/config_file: do not attempt to mount an already mounted slot

ejoerns:
So far, when loading the slot status file, we unconditionally attempt to
mount the slot to the default RAUC mount path, read the content and
remount it.
This basically works for most cases, as also if the slot is already
mounted (because it is the root slot, for example), a further mount will
implicitly only create a bind mount in the system.

However, there are sever cases where this will fail. A common example
is when the rootfs slot we currently run from is mounted readonly.
Then an attempt to (bind)mount it (rw) in RAUC will result in an error
like:

> mount: /dev/sda4 is already mounted or /run/rauc/rootfs.1 busy
>        /dev/sda4 is already mounted on /

As a result of this a
```
  rauc status --detailed
```
call will not give any data from the currently running rootfs.

The safest way to avoid issues like this is to use the already existing
information about detected external mounts and simply skip mounting if
the respective slot is already reported as being mounted.


src/config_file.c
```diff
@@ -588,26 +588,31 @@ static void load_slot_status_locally(RaucSlot *dest_slot)
...
+		}
 	}
 
 	slotstatuspath = g_build_filename(dest_slot->mount_point, "slot.raucs", NULL);
 
 	res = read_slot_status(slotstatuspath, dest_slot->status, &ierror);
 	if (!res) {
 		g_message("Failed to load status file %s: %s", slotstatuspath, ierror->message);
-		r_umount_slot(dest_slot, NULL);
+		if (!dest_slot->ext_mount_point)
```
jluebbe: Can you simplify the error path here? Maybe reuse the r_umount_slot() below?

ejoerns:
Maybe this way?

PR #290, opened 2018-07-13
elboulangero/casync extract fixes

elboulangero:


src/update_handler.c
```diff
@@ -275,6 +275,7 @@ static gboolean casync_extract_image(RaucImage *image, gchar *dest, GError **err
 
 	res = TRUE;
 out:
+	g_free(seed);
```
ejoerns: Good catch! But should be solved in a different way as we've just introduced usage of glib automatic variable cleanup feature. Thus please use instead:
``` diff
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -219,7 +219,7 @@ static gboolean casync_extract_image(RaucImage *image, gchar *dest, GError **err
        GError *ierror = NULL;
        gboolean res = FALSE;
        RaucSlot *seedslot = NULL;
-       gchar *seed = NULL;
+       g_autofree gchar *seed = NULL;
        gchar *store = NULL;
        gboolean seed_mounted = FALSE;
```

src/update_handler.c
```diff
@@ -253,10 +253,11 @@ static gboolean casync_extract_image(RaucImage *image, gchar *dest, GError **err
 		seed = g_strdup(seedslot->device);
 	}
 
+extract:
+	/* Set store */
```
ejoerns: Good catch, but please be a bit more verbose in your patch discription on why it fixes things. The actual thing that was wrong here is that `casync_extract()` would have been called without having initiallized the `store` variable.

elboulangero:
Pull request updated, thanks for the feedback

ejoerns:
Thanks for the update! Looks fine for me now. :+1:

PR #291, opened 2018-07-22
Introduce max-download-size config setting (fixes #287)

### LINKED ISSUE #287: Introduce config setting for max bundle download size
At the moment, a hard-coded BUNDLE_DL_MAX_SIZE define is used for a safety check. I think it would be better to introduce this as config parameter with the current value as default, to be more flexible at runtime.

mhei:
At the moment, a hard-coded BUNDLE_DL_MAX_SIZE define is
used for a safety check when a bundle is downloaded.
To be more flexible at runtime, introduce a new config
parameter 'max-download-size'.
The current define is used as default value when the
config setting is not specified.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -87,6 +87,11 @@ Example configuration:
   key not exists, the bootchooser framework searches per default for ``/state``
   or ``/aliases/state``.
 
+``max-bundle-download-size``
+  Defines the maximum downloadable bundle size in bytes, and thus must be
+  a simple integer value (without unit) greater than zero.
+  It overwrites the compiled-in default value of 8 MiB.
+
 **[keyring] section**
 
 The ``keyring`` section refers to the trusted keyring used for signature
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -31,6 +31,7 @@ gboolean default_config(RaucConfig **config)
 {
 	RaucConfig *c = g_new0(RaucConfig, 1);
 
+	c->max_bundle_download_size = DEFAULT_MAX_BUNDLE_DOWNLOAD_SIZE;
 	c->mount_prefix = g_strdup("/mnt/rauc/");
 
 	*config = c;
@@ -128,6 +129,27 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 
 	if (g_strcmp0(c->system_bootloader, "barebox") == 0) {
 		c->system_bb_statename = g_key_file_get_string(key_file, "system", "barebox-statename", NULL);
+	}
+
+	c->max_bundle_download_size = g_key_file_get_uint64(key_file, "system", "max-bundle-download-size", &ierror);
+	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
+		g_debug("No value for key \"max-bundle-download-size\" in [system] defined "
+				"- using default value of %d bytes.", DEFAULT_MAX_BUNDLE_DOWNLOAD_SIZE);
+		c->max_bundle_download_size = DEFAULT_MAX_BUNDLE_DOWNLOAD_SIZE;
+		g_clear_error(&ierror);
+	} else if (ierror) {
+		g_propagate_error(error, ierror);
+		res = FALSE;
+		goto free;
+	}
+	if (c->max_bundle_download_size == 0) {
+		g_set_error(
+				error,
+				R_CONFIG_ERROR,
+				R_CONFIG_ERROR_MAX_BUNDLE_DOWNLOAD_SIZE,
+				"Invalid value (%" G_GUINT64_FORMAT ") for key \"max-bundle-download-size\" in system config", c->max_bundle_download_size);
+		res = FALSE;
+		goto free;
 	}
 
 	c->mount_prefix = g_key_file_get_string(key_file, "system", "mountprefix", NULL);
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -5,10 +5,14 @@
 #include <checksum.h>
 #include "manifest.h"
 
+/* Default maximum downloadable bundle size (8 MiB) */
+#define DEFAULT_MAX_BUNDLE_DOWNLOAD_SIZE 8*1024*1024
+
 typedef enum {
 	R_CONFIG_ERROR_INVALID_FORMAT,
 	R_CONFIG_ERROR_BOOTLOADER,
-	R_CONFIG_ERROR_PARENT
+	R_CONFIG_ERROR_PARENT,
+	R_CONFIG_ERROR_MAX_BUNDLE_DOWNLOAD_SIZE
 } RConfigError;
 
 #define R_CONFIG_ERROR r_config_error_quark()
@@ -28,6 +32,8 @@ typedef struct {
 	gchar *system_variant;
 	gchar *system_bootloader;
 	gchar *system_bb_statename;
+	/* maximum filesize to download in bytes */
+	guint64 max_bundle_download_size;
 	/* path prefix where rauc may create mount directories */
 	gchar *mount_prefix;
 	gchar *store_path;
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -10,9 +10,6 @@
 #include "utils.h"
 #include "network.h"
 
-/* Maximum downloadable bundle size (8MB) */
-#define BUNDLE_DL_MAX_SIZE 8*1024*1024
-
 GQuark
 r_bundle_error_quark(void)
 {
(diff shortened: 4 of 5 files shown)


src/config_file.c
```diff
@@ -130,6 +134,12 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 		c->system_bb_statename = g_key_file_get_string(key_file, "system", "barebox-statename", NULL);
 	}
 
+	c->max_download_size = g_key_file_get_uint64(key_file, "system", "max-download-size", NULL);
+	if (!c->max_download_size) {
```
OnkelUlla: `g_key_file_get_uint64()` distinguishes among others between `G_KEY_FILE_ERROR_KEY_NOT_FOUND` and `G_KEY_FILE_ERROR_INVALID_VALUE`, so please utilize the function's error argument instead of its return value and apply a finer grained differentiation with different debug messages, but the same default value (see [here](https://github.com/rauc/rauc/blob/master/src/config_file.c#L456) and [here](https://github.com/rauc/rauc/blob/master/src/install.c#L967) for example).

mhei:
@OnkelUlla: Thanks for your review. I'll update this PR soon.

ejoerns:
@mhei Just a general interest from my side: Do you use the download feature for casync support? Or for the (fairly old) 'network mode' of RAUC?
We actually had several plans to remove the latter one while we intend to extend support for the first one. In case you use casync and you already hit the 8MiB boundary, we should also think about this default value ;).

mhei:
First, sorry for my delayed answers, I'm currently traveling (on holidays)...

@ejoerns: We (I2SE) are just starting with rauc and using it for our charging station firmware. Our distributed (traditional) bundle is around 300 MiB at the moment, and it has not yet been decided whether the download of the bundle to the device is managed by an external application or whether rauc should handle it directly using the network mode. We also started investigating casync support. It is still an ongoing discussion with our customers, which method is used or preferred for the final field rollout.

mhei:
@OnkelUlla: I updated the PR, would be great if you could have a second look.

src/config_file.c
```diff
@@ -130,6 +131,24 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 		c->system_bb_statename = g_key_file_get_string(key_file, "system", "barebox-statename", NULL);
 	}
 
+	c->max_bundle_download_size = g_key_file_get_uint64(key_file, "system", "max-bundle-download-size", &ierror);
+	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
+		g_message("No value for key \"max-bundle-download-size\" in [system] defined "
```
ejoerns: This should be `g_debug()` to not pollute the log for those who don't use it.

src/config_file.c
```diff
@@ -130,6 +131,24 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
+				"- using default value of %d bytes.", DEFAULT_MAX_DOWNLOAD_BUNDLE_SIZE);
+		c->max_bundle_download_size = DEFAULT_MAX_DOWNLOAD_BUNDLE_SIZE;
+	}
+	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE)) {
+		g_message("Value of key \"max-bundle-download-size\" in [system] "
```
ejoerns: We are always very strict in config file parsing. Thus if there is an error in config file, RAUC should also report an error to avoid unintentionally writing configuration that has no effect.

src/config_file.c
```diff
@@ -130,6 +131,24 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
+				"is no valid unsigned integer - using default value of %d bytes.", DEFAULT_MAX_DOWNLOAD_BUNDLE_SIZE);
+		c->max_bundle_download_size = DEFAULT_MAX_DOWNLOAD_BUNDLE_SIZE;
+	}
+	g_clear_error(&ierror);
+	if (c->max_bundle_download_size == 0) {
```
ejoerns: Good sanity check, but should also result in an error.

src/config_file.c
```diff
@@ -130,6 +131,24 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 		c->system_bb_statename = g_key_file_get_string(key_file, "system", "barebox-statename", NULL);
 	}
 
+	c->max_bundle_download_size = g_key_file_get_uint64(key_file, "system", "max-bundle-download-size", &ierror);
```
ejoerns: using `max-bundle-download-size` could add irritation about the actual unit used here MiB? KiB? Bytes?

I would suggest naming it `max-bundle-download-size-bytes` as I fear introducing units will add too much overhead.

src/config_file.c
mhei: Ok

src/config_file.c
mhei: Agreed.

src/config_file.c
mhei: Ok.

src/config_file.c
mhei: Puh... I understand your point, but I think appending -bytes to an already longish name is not nice either. As you said, I also guess that this config option is rarely used, so people should RTFM to be sure about the unit. Obmitting the -bytes suffix also allows to add an unit parsing later if it ever will become important...
So, I'll update my PR without this change first, but if it is a no-go for you to get this in, I'll change it...

BTW: What are the reasons for the considerations to mark this obsolete? I think this is a really cool feature, because there is no need to temporary store the whole file locally. And while I see the pro's of casync, I'm not sure that everybody wants to use it. One problem I see, is that distribution of a single file to customers web server for deployment is simpler than merging files of a chuck store.

ejoerns: Ok, I understand your concerns regarding the longish name, but than at least your have to WTFM :) Cause atm it does not indicate which unit you have to use or if a value of "8MiB" is a valid or invalid one.
Adding something like "Specify maximum bundle download size in number of bytes, e.g. ``max-bundle-download-size =8*1024*1024``" should be sufficient.

Regarding the network mode: The real 'network mode ' was invented in the early days of RAUC but did not have too many users until now (meaning we know of a single one!). Since the initial patches this was actually basically tested but did not get many of the changes and features the normal bundle mode has.

When I talk about network mode, this also has nothing to do with bundles. It means having a single manifest downloaded from the web server and downloading the individual images described in it separately. It does not cover downloading a normal bundle over network. Is that actually how you use it? Using it for downloading normal bundles is not intended to work at the moment.

OnkelUlla: I agree: leaving the door open to add a unit parser later is a good thing.
@mhei: As you are presumably going to prepare an updated PR with respect to the documentation would you mind renaming the macro `DEFAULT_MAX_DOWNLOAD_BUNDLE_SIZE` to `DEFAULT_MAX_BUNDLE_DOWNLOAD_SIZE` to have it consistent with the variable's name? (This is no must, but just a gentle request).

mhei:
Thanks for all your feedback, I've updated the PR again.
It seems that I missunderstood the 'network mode': when I talked about it, I means that I simply use `rauc install <url>` (with url being the reference to a usual rauc bundle - not casync one) instead of a downloaded file which is already locally present on the target. In this case, rauc downloads the file into TMP directory and then processes the bundle same as "manually" downloaded one - so my argument that no tempory storage is needed does not apply - we just do not need to care about the download ourself.

src/config_file.c
```diff
@@ -130,6 +131,28 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
+	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
+		g_debug("No value for key \"max-bundle-download-size\" in [system] defined "
+				"- using default value of %d bytes.", DEFAULT_MAX_BUNDLE_DOWNLOAD_SIZE);
+		c->max_bundle_download_size = DEFAULT_MAX_BUNDLE_DOWNLOAD_SIZE;
+		g_clear_error(&ierror);
+	}
+	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE)) {
```
ejoerns: When I look at this again, I would prefer making this consistent with the other tests by using an else if and testing for all errors here, cause other we might potentially miss errors here. The handling is already generic, anyway.
``` diff
-	}
-	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE)) {
+	} else if (ierror) {
```

docs/reference.rst
```diff
@@ -87,6 +87,12 @@ Example configuration:
   key not exists, the bootchooser framework searches per default for ``/state``
   or ``/aliases/state``.
 
+``max-bundle-download-size``
+  Defines the maximum downloadable bundle size in bytes, and thus must be
+  a simple integer value (without unit) greater than zero.
+  It overwrites the compiled-in default value of 8 MiB.
+
```
ejoerns: please remove extra line

ejoerns:
Ok, so no added dependency on what I called the `network mode` ;) Thanks for explanation!
The ability to download full bundles via network by RAUC directly was not what it was intended for and should be handled carefully because RAUC normally runs as root.
We added the size limit originally to prevent unintentional downloading non-casync bundles with that mechanism, but having the ability to do this if one would like to is fine for me, too.
Despite I would not strongly encourage people to do so ;)

mhei:
Fixed the two points :)

PR #293, opened 2018-07-30
Implement uboot_get_state() and uboot_get_primary()

timothytylee:
This patch allows "rauc status" to report which slot is activated, and slot status when used with U-Boot.

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -549,6 +549,52 @@ out:
 	return res;
 }
 
+/* We assume bootstate to be good if slot is listed in 'BOOT_ORDER' and its
+ * remaining attempts counter is > 0 */
+static gboolean uboot_get_state(RaucSlot* slot, gboolean *good, GError **error)
+{
+	g_autoptr(GString) order = NULL;
+	g_autoptr(GString) attempts = NULL;
+	g_auto(GStrv) bootnames = NULL;
+	g_autofree gchar *key = NULL;
+	GError *ierror = NULL;
+	gboolean found = FALSE;
+	gboolean res = FALSE;
+
+	g_return_val_if_fail(slot, FALSE);
+	g_return_val_if_fail(good, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	res = uboot_env_get("BOOT_ORDER", &order, &ierror);
+	if (!res) {
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	/* Scan boot order list for given slot */
+	bootnames = g_strsplit(order->str, " ", -1);
+	for (gchar **bootname = bootnames; *bootname; bootname++) {
+		if (g_strcmp0(*bootname, slot->bootname) == 0) {
+			found = TRUE;
+			break;
+		}
+	}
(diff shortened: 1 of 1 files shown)


src/bootchooser.c
```diff
@@ -549,6 +549,52 @@ static gboolean uboot_env_set(const gchar *key, const gchar *value, GError **err
...
+
+	/* Scan boot order list for given slot */
+	bootnames = g_strsplit(order->str, " ", -1);
+	for (gchar **bootname = bootnames; *bootname; bootname++) {
+		if (g_strcmp0(*bootname, slot->bootname) == 0) {
+			found = TRUE;
+			break;
+		}
+	}
+	if (!found)
```
ejoerns: This actually means the boot state is bad. Thus you need to set `*good` here properly,

src/bootchooser.c
```diff
@@ -571,6 +617,45 @@ static gboolean uboot_set_state(RaucSlot *slot, gboolean good, GError **error)
...
+	}
+
+	/* Iterate over current boot order */
+	bootnames = g_strsplit(order->str, " ", -1);
+	for (gchar **bootname = bootnames; *bootname; bootname++) {
+		/* find matching slot entry */
+		g_hash_table_iter_init(&iter, r_context()->config->slots);
+		while (g_hash_table_iter_next(&iter, NULL, (gpointer*) &slot)) {
+			if (g_strcmp0(*bootname, slot->bootname) == 0) {
+				primary = slot;
```
ejoerns: It won't be sufficient here to look if the key is in `BOOT_ORDER`. Consider the case we have `BOOT_ORDER=A B` and `BOOT_A_DEFAULT=0`, `BOOT_B_DEFAULT=3`. In this case, the boot selection won't boot A but B instead and thus we cannot return A as the 'primary' slot from this function, too.

ejoerns:
In https://github.com/ejoerns/rauc/commits/topic/u-boot-testing-next I've rebase your work on my PR #295 and added a minimal fixup (https://github.com/ejoerns/rauc/commit/dd6bd5d6877be2523d800e31f17ef8a61426fd51) for the first issue I described and a test (https://github.com/ejoerns/rauc/commit/ab4ae30ab3ffd09f6105cb1fde80b9c86dcf47c9) for the second.
Maybe you would like to use this for continuing work on it.

ejoerns:
@timothytylee do you plan to work on a fix for this? If not, I would probably pick this up and add the changes required to make this fully work.

timothytylee:
Hi Jorns,

Thank you for reviewing the code.  Sorry, I was busy with other matters, and could not spare the time.  I'll let you continue with the fix then.

Regards,
Timothy

ejoerns:
@timothytylee Thanks for your reply! I guess I've managed to fix the last remaining item. I now have a branch rebased onto master (including already added uboot tests) with your commit extended by my fixups and another commit adding tests. Would you like me to open a new PR or should I force-push into your existing one?

timothytylee:
I guess a new PR from be more appropriate then.  I'll close this PR.

ejoerns:
Ok. I've opened #311 for this now. Please test and review if you find the time.

PR #296, opened 2018-08-08
Add fsync to copy_raw_image()

ejoerns:
This is a more generic approach of #278 as it fixes `fsync()` handling in `copy_raw_image()` directly which itself is used by different handlers.


ejoerns:
Great, thanks for testing! Then I will merge it now.

PR #298, opened 2018-08-09
Some Collected Minor Fixes and Cleanups

ejoerns:

PR #269, opened 2018-05-03
service: add system info properties

jremmet:
Make system information availabile via dbus. This adds the properties
Compatible, Variant, and Bootslot

Signed-off-by: Jan Remmet <j.remmet@phytec.de>

PR #299, opened 2018-08-09
RFC: Obtain rauc status fully over D-Bus

ejoerns:
> By default, i.e. when compiled with --enable-service, we have a
> separation of RAUC into the service controlling the installation and the
> command line interface used to control it.
> The communication between these two works via D-Bus as it would be for
> any other client or frontend.
> 
> While for `rauc install` there is a clear seperation according to the
> handling described above, for `rauc status` this was always not that
> strict but more or less a 'hybrid' solution, mainly due to histroical
> reasons. The (detailed) status information the client retrived via D-Bus
> while all other information was examined by calls from the client, not
> the server.
> This not only breaks the design pattern but also adds some drawbacks
> such as that the client also had to read the RAUC configuration file in
> ordert to obtain the slot configuration and it did unnecessarily make its on
> interaction with the bootloader, etc. which is a potential conflict with
> what the service does at the same time.
> 
> After some pre-work this patch now finally fully implements the client
> mode for rauc status and obtains all information via D-Bus only.
> Only for the non-service case, i.e. without D-Bus the old behavior will
> be retained.

This is a preview of the support, still missing some points to implement but showing a general view of where I would like to go.

Currently based on #269 and #298


ejoerns:
Rebased onto current master

ejoerns:
Rebased onto current master and added missing implementation details such as valid parent hierarchy and obtaining of primary slot.

Depends on #269

src/service.c
```diff
@@ -298,6 +291,12 @@ static GVariant* create_slotstatus_array(void)
 		g_clear_error(&ierror);
 	}
 
+	res = determine_boot_states(&ierror);
```
jluebbe: Do we need to call the bootloader interface every time someone calls the dbus method? Currently we even call the bootloader tool multiple times (once for each slot with a bootname).

ejoerns:
Rebased onto current master to fix conflict.

``` diff
index 6f5c3281..1c140e97 100644
--- a/src/main.c
+++ b/src/main.c
@@ -938,7 +938,7 @@ static gchar* r_status_formatter_readable(RaucStatusPrint *status)
        g_string_append_printf(text, "Compatible:  %s\n", status->compatible);
        g_string_append_printf(text, "Variant:     %s\n", status->variant);
        g_string_append_printf(text, "Booted from: %s\n", status->bootslot);
-       g_string_append_printf(text, "Activated:   %s\n", status->primary ? status->primary->name : NULL);
+       g_string_append_printf(text, "Activated:   %s (%s)\n", status->primary ? status->primary->name : NULL, status->primary ? status->primary->bootname : NULL);
 
        g_string_append(text, "slot states:\n");
        g_hash_table_iter_init(&iter, status->slots);

```

PR #302, opened 2018-08-10
Fix missing error handling

OnkelUlla:


OnkelUlla:
Pushed a slightly updated version where `ierror` is not set to `NULL` after the return of `g_propagate_prefixed_error()` and a superfluous `g_clear_error()` is removed.

PR #305, opened 2018-08-13
src/install: actually handle readonly slots

### LINKED ISSUE #161: system config slot key `readonly` ignored
Both the documentation and the parser handle a `readonly` key that can be set in a slot configuration in a system.conf as follows:

``` cfg
[slot.rootfs.0]
...
readonly=true
```
But, there is no code that further processes or respects this setting. Thus, currently it is simply unhandled.

The `readonly` key must either be used or should be removed from documentation and from parser.

ejoerns:
So far we had the slot config booloan key `readonly` that allows to mark
a slot as being detected but not being writable by RAUC.

But, this option was only of informational nature and did not really
prevent from accidently installing content to a slot marked as readonly.

This patch now slightly extends the target slot selection algorithm to
actually care about the readonly flag by not adding the slot to the
hash map of target slots.

This way it will not be possible to install data to 'readonly' slots,
but they are still considered e.g. for parent slot detection, etc.
An extra test to assure this was added.

Fixes #161

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -396,14 +396,21 @@ GList* get_install_images(const RaucManifest *manifest, GHashTable *target_group
 
 	slotclasses = get_all_manifest_slot_classes(manifest);
 
+	/* Find exactly 1 image for each class listed in manifest */
 	for (gchar **cls = slotclasses; *cls != NULL; cls++) {
 		RaucImage *matching_img = NULL;
 
 		for (GList *l = manifest->images; l != NULL; l = l->next) {
 			RaucImage *lookup_image = l->data;
+			RaucSlot *target_slot = NULL;
+
+			/* Not interested in slots of other classes */
+			if (!g_strcmp0(lookup_image->slotclass, *cls) == 0)
+				continue;
 
 			/* Check if target_group contains an appropriate slot for this image */
-			if (!g_hash_table_contains(target_group, lookup_image->slotclass)) {
+			target_slot = g_hash_table_lookup(target_group, lookup_image->slotclass);
+			if (!target_slot) {
 				g_set_error(error,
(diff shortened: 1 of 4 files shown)


ejoerns:
@BastianStender Good question. Possibly not, or at least not without anything that 'forces' this.

One might use this for debugging / hacking and thus circumvent the default 'readonly' safety flag, but one might also be protected from unintentionally deleting content.

BastianStender:
If I'm not mistaken `rauc write-slot` will not respect the readonly flag currently. In my opinion this should be changed.

src/install.c
```diff
@@ -375,6 +375,10 @@ GHashTable* determine_target_install_group(void)
 
 
 		if (slot_list_contains(selected_root_slots, parent)) {
+			if (iterslot->readonly) {
+				g_debug("\tNot adding readonly slot");
```
jluebbe: Should we have a more noticeable warning in this case?

ejoerns: That would mean we do not allow readonly slots.

The continue here is simply to not add slots marked as 'readonly' to the mapping of classes to images. This means we will not generate a 'bootloader -> botloader.0' mapping if 'bootloader.0' is marked readonly.

Where this will fail is when someone comes with an bundle containing an image for the 'bootloader' class, then it will detect there is no mapping for this and exit with an error.

ejoerns:
Pushed an update:

* Move the readonly check to `get_install_images()` and return an explicit error
* Add test for this
* Add check to `write-slot` function
* Add test for this

jluebbe:
@ejoerns this now conflicts in test/rauc.t.

ejoerns:
Again rebased onto master and resolved conflicts in rauc.t

PR #306, opened 2018-08-13
src/install: Not finding an appropriate target slot must be fatal

ejoerns:
Otherwise, a manifest containing an image for e.g. both a rootfs and an
appfs will install successfully also if the target system does not have
an appfs configured.

This is against the rule of RAUC of being unambiguous in the
installation description.


BastianStender:
`get_install_images()` returns NULL if `target_group` contains no appropriate slot for each image or if the image for a class in the manifest does not match the variant - at least if I understand `get_install_images()` correctly.

ejoerns:
Pushed an update to fix the commit message.

PR #300, opened 2018-08-09
add support for PKCS#11 tokens

jluebbe:
This allows using HSMs or other tokens to store the signing keys.

PR #308, opened 2018-08-14
Add Support for Verifying Signature During Bundle Creation

ejoerns:
To detect signing errors and expired certificates early, this series adds trust chain verificiation during creation of a bundle if the  `--keyring` argument is provided or a keyring configured.

Additionally a check is implemented to warn the signer about using certificates in the chain that will expire within the period of a month.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -31,7 +31,7 @@ sub-command:
 
 .. code-block:: sh
 
-  rauc bundle --cert=<certfile> --key=<keyfile> <input-dir> <output-file>
+  rauc bundle --cert=<certfile> --key=<keyfile> --keyring=<keyringfile> <input-dir> <output-file>
 
 Where ``<input-dir>`` must be a directory containing all images and scripts the
 bundle should include, as well as a manifest file ``manifest.raucm`` that
@@ -45,6 +45,13 @@ Instead of the ``certfile`` and ``keyfile`` arguments, PKCS#11 URLs such as
 ``'pkcs11:token=rauc;object=autobuilder-1'`` can be used to avoid storing
 sensitive key material as files (see :ref:`PKCS#11 Support <pkcs11-support>`
 for details).
+
+While the ``--cert`` and ``--key`` argument are mandatory for signing and must
+provide the certificate and private key that should be used for creating the
+signature, the ``--keyring`` argument is optional and (if given) will be used
+for verifying the trust chain validity of the signature after creation.
+Note that this is very useful to prevent from signing with obsolete
+certificates, etc.
 
 Obtaining Bundle Information
 ----------------------------
--- a/src/signature.c
+++ b/src/signature.c
@@ -11,6 +11,14 @@
 #include "context.h"
 #include "signature.h"
 
+/* Define for OpenSSL 1.0.x backwards compatiblity.
+ * We use newer get0 names to be clear about memory ownership and to not use
+ * API deprecated in OpenSSL 1.1.x */
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L)
+#define X509_get0_notAfter X509_get_notAfter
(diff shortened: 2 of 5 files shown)


src/signature.c
```diff
@@ -180,6 +180,28 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, g
 				"Read zero bytes");
 		goto out;
 	}
+
+	/* keyring was given, perform verification to obtain trust chain */
+	if (r_context()->config->keyring_path) {
+		CMS_ContentInfo *vcms = NULL;
+		X509_STORE *store = NULL;
+		STACK_OF(X509) *verified_chain = NULL;
```
jluebbe: Should `store` and `verified_chain` be freed?

docs/using.rst
```diff
@@ -41,6 +41,13 @@ which image to install to which slot, which scripts to execute etc.
 bundles must always have a ``.raucb`` file name suffix in order to ensure that
 RAUC treats them as bundles.
 
+While the ``--cert`` and ``--key`` argument are mandatory for signing and must
+provide the certificate and private key that should be used for creating the
+signature, the ``--keyring`` argument is optional and (if given) will be used
+for verifying the trust chain validity of the created signature after creation.
```
jluebbe: The 'created' before signature is redundant.

.travis.yml
```diff
@@ -6,6 +6,7 @@ before_install:
   - sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu/ xenial main restricted universe multiverse" -y
   - sudo apt-get update -qq
   - sudo apt-get install -qq libglib2.0-dev libjson-glib-dev
+  - sudo apt-get install -qq faketime
```
jluebbe: If PR #300 is merged first, this should move to the normal packages list.

src/signature.c
ejoerns: Ack. Used g_autoptr for these now.

docs/using.rst
ejoerns: removed

.travis.yml
ejoerns: That's the case, thus moved faketime to packages list.

src/signature.c
```diff
@@ -353,6 +353,52 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, g
 				"Read zero bytes");
 		goto out;
 	}
+
+	/* keyring was given, perform verification to obtain trust chain */
+	if (r_context()->config->keyring_path) {
+		g_autoptr(CMS_ContentInfo) vcms = NULL;
+		g_autoptr(X509_STORE) store = NULL;
+		STACK_OF(X509) *verified_chain = NULL;
```
jluebbe: Why is it not necessary to free verified_chain? A comment would help.

src/signature.c
```diff
@@ -353,6 +353,52 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, g
...
+		}
+
+		if (!cms_get_cert_chain(vcms, store, &verified_chain, &ierror)) {
+			g_propagate_error(error, ierror);
+			res = NULL;
+			goto out;
+		}
+
+		for (int i = 0; i < sk_X509_num(verified_chain); i++) {
+			ASN1_TIME *expiry_time;
```
jluebbe: Do we need to free expiry_time?

ejoerns:
Pushed changes:
* Rebased onto master
* Resolved conflicts in rauc.t
* use X509_get0_not{Before,After} functions for clearer memory ownership, added backward compatibility to OpenSSL 1.0.x
* added missing free for `verified_chain`
* exemplarily documented memory ownership of `cms_get_cert_chain()`

PR #309, opened 2018-08-14
Add Option to use Signing Time For Bundle Verification Instead of Current

ejoerns:
In some security models it might be desired to use a different validity
model for signature verification.

While the default one used by OpenSSL and RAUC is to check if the
signature is valid at the time of verification, an alternative approach
is to check if the certificates were valid at the time of signing.

The newly introduced config option `use-signing-time` allows to select the
latter behavior if required.


src/config_file.c
```diff
@@ -249,6 +249,17 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 	c->keyring_path = resolve_path(filename,
 			g_key_file_get_string(key_file, "keyring", "path", NULL));
 
+	c->use_cert_signing_time = g_key_file_get_boolean(key_file, "keyring", "use-signing-time", &ierror);
+	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND) ||
+			g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_GROUP_NOT_FOUND)) {
+		c->use_cert_signing_time = FALSE;
```
jluebbe: Why do you have  `use-signing-time` as the config option, but `use_cert_signing_time` as the variable name? The time is not store in the certificate's signature but in the bundle signature. Maybe `use-bundle-signing-time` for both?

src/signature.c
```diff
@@ -396,6 +398,31 @@ gboolean cms_get_cert_chain(CMS_ContentInfo *cms, X509_STORE *store, STACK_OF(X5
 	return res;
 }
 
+static gboolean asn1_time_to_tm(const ASN1_TIME *intime, struct tm *tm)
```
jluebbe: Please add a short comment why this is needed.

src/signature.c
```diff
@@ -450,6 +477,38 @@ gboolean cms_verify(GBytes *content, GBytes *sig, CMS_ContentInfo **cms, X509_ST
 		goto out;
 	}
 
+	/* Optionally use certificate signing timestamp for verification */
+	if (r_context()->config->use_cert_signing_time) {
+		STACK_OF(CMS_SignerInfo) *sinfos;
+		CMS_SignerInfo *si;
+		X509_ATTRIBUTE *xa;
+		ASN1_TYPE *so;
+		X509_VERIFY_PARAM *param = X509_VERIFY_PARAM_new();
```
jluebbe: These are missing the corresponding `free`s.

src/signature.c
```diff
@@ -450,6 +477,38 @@ gboolean cms_verify(GBytes *content, GBytes *sig, CMS_ContentInfo **cms, X509_ST
...
+		xa = CMS_signed_get_attr(si, CMS_signed_get_attr_by_NID(si, NID_pkcs9_signingTime, -1));
+		so = X509_ATTRIBUTE_get0_type(xa, 0);
+
+		/* convert to time_t to make it usable for seting verify parameter */
+		if (!asn1_time_to_tm(so->value.utctime, &tm)) {
+			g_set_error(
+					error,
+					R_SIGNATURE_ERROR,
+					R_SIGNATURE_ERROR_UNKNOWN,
+					"Failed converting signing time");
```
jluebbe: "Failed to convert bundle signing time"

.travis.yml
```diff
@@ -6,6 +6,7 @@ before_install:
   - sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu/ xenial main restricted universe multiverse" -y
   - sudo apt-get update -qq
   - sudo apt-get install -qq libglib2.0-dev libjson-glib-dev
+  - sudo apt-get install -qq faketime
```
jluebbe: If PR #300 is merged first, this should move to the normal packages list.

src/signature.c
ejoerns: Not sure but I assume only param needs to be freed as others are only pointers.

src/config_file.c
ejoerns: Sounds good.

ejoerns:
Pushed changes:
* Renamed arg to `use-bundle-signing-time`
* added note to `asn1_time_to_tm()`
* Added missing free of X509_VERIFY_PARAM
* Fixed error message string
* Rebased onto master
* Adapted for new strict config file parsing
* Used faketime from default xenial packages

ejoerns:
Pushed changes:
* Fixed option name in commit message

PR #317, opened 2018-08-25
build: make eMMC boot partition support optional

tpetazzoni:
The eMMC boot partition support, added in commit
ea5cc7ff606c65536da218bd1ef6d0ca279c9b17 ("src/update_handler: add
support for updating eMMC boot partitions"), requires
<linux/mmc/ioctl.h>, only available starting from kernel headers 3.0.

Even though it is pretty likely that people are going to use Linux >=
3.0 on their embedded systems these days, RAUC also needs to be built
natively on the build machine to produce update artifacts, and the
build machine is sometimes using an ancient Linux system, especially
in an enterprise contexts.

In order to make sure that RAUC builds fine in this context, this
commit makes the eMMC boot partition support optional, by verifying
the availability of <linux/mmc/ioctl.h>.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@bootlin.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -39,7 +39,6 @@ librauc_la_SOURCES = \
 	src/checksum.c \
 	src/config_file.c \
 	src/context.c \
-	src/emmc.c \
 	src/install.c \
 	src/manifest.c \
 	src/mark.c \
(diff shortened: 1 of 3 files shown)


BastianStender:
Wow, what build system runs on such an ancient system state?

tpetazzoni:
We have a Buildroot user running on RHEL6.5 (in a corporate environment, as you can imagine), and he has set up an autobuilder that tests all Buildroot packages on this old system. See rauc failing to build (when building for the host machine): http://autobuild.buildroot.net/?reason=host-rauc-0.4 (look only at build failures starting from August 3rd).

jluebbe:
Thanks for your contribution!

PR #318, opened 2018-08-27
update: add squashfs to raw handling.

eroullit:
My image is formatted as `squashfs` and at first I could not perform an update via `rauc install`.
With the help of this changeset, setting `RAUC_IMAGE_FSTYPE=squashfs` and setting the slots type to `raw`, it was then possible to update the slot.

Do you have any experience using `casync` to update a `squashfs`-based slot ?


jluebbe:
Regarding squashfs with casync, I don't have any hard numbers, but the casync algorithm should be able to find reused data. You may need to play a bit with squashfs and casync block sizes to get a good result.

PR #324, opened 2018-09-07
src/bootchooser: fix segfault in efi_get_primary()

elboulangero:
bootnext might null

Signed-off-by: Arnaud Rebillout <arnaud.rebillout@collabora.com>


elboulangero:
> Maybe you also like to note that the message was misleading / wrong.

Sure, I just did that, thanks for the feedback!

ejoerns:
excellent :)

PR #327, opened 2018-09-13
src/install.c: Add network mode deprecation warning

ejoerns:
This is the initial step of deprecating the unused and
slightly outdated RAUC network (non-bundle) mode.

If someone actively uses this mode (Note: Affects only 'real' network mode, without bundles, not the remote functionality of RAUC)

PR #328, opened 2018-09-13
Remove need for .raucb extension

### LINKED ISSUE #264: Question: is fixed bundle filename suffix really required?
> Note that RAUC bundles must always have a `.raucb` file name suffix in order
> to ensure that RAUC treats them as bundles.

When looking at the code, it seems that this is only used to switch between
local filesystem access or network download mode, but there is no other
technical requirement?

So my question is, whether it would be possible to remove this requirement as
I would prefer to distribute the update image files with .image extension.
This way, we could "hide" the internally used update implementation without
renaming the distributed files forth and back. (This is what I would do if you
tell me that this is a fixed design decision)

ejoerns:
In #264 and other discussions the question came up if the current way of forcing a `.raucb` extension for RAUC bundles is really required.

Background is that some may simply want to hide how they update, do not like the extension or have other requirements in their infrastructure that needs them to use a special extensions.

However, the main reasons that RAUC forces a `.raucb` extensions is to be explicit what we want to handle as a bundle, but the more to have a distinction between the historical RAUC 'network mode' that uses manifests instead of bundles and the conventional bundle mode that is effectively the one-and-only today.

When removing the need for a `.raucb` extensions, this does not mean we encourage everyone to use whatever extension he likes, but, we allow it. Note that it is probably seen very inconvenient to give a .tar.gz file the extension `.archive` or naming `.deb` files as `.mydist`. But you may have a reason for it.
And additionally, the majority of tools handling names with a common file name extension do not actually force this extension. Thus we adapt to this behavior, we will not be something 'special'.

As we want to get rid of the old network mode, anyway (as you may have noted in #327 that I originally prepared as part of this series but split out for allowing separate discussion) I have honored the fact that having a `.raucb` bundle is the default and do the switch for network mode now only if there is a `.raucm` extension detected.

The reason for doing this distinction still file extension-based is that this has less impact on structure and code in case we remove the network mode and that it would need larger restructuring of current code when doing a content-based check on a bundle/manifest that we potentially need to download first.

Beside this obvious change, this also fixes support for `file://` URIs and adds a little more checks and debug output to know what went wrong when picking the a faulty bundle/manifest location.

Note: Fixes #264

### DIFF
--- a/include/bundle.h
+++ b/include/bundle.h
@@ -8,7 +8,8 @@ GQuark r_bundle_error_quark(void);
 
 typedef enum {
 	R_BUNDLE_ERROR_SIGNATURE,
-	R_BUNDLE_ERROR_KEYRING
+	R_BUNDLE_ERROR_KEYRING,
+	R_BUNDLE_ERROR_IDENTIFIER
 } RBundleError;
 
 typedef struct {
--- a/include/network.h
+++ b/include/network.h
@@ -16,4 +16,4 @@ static inline void network_init(void)
 gboolean download_file(const gchar *target, const gchar *url, gsize limit, GError **error);
 gboolean download_file_checksum(const gchar *target, const gchar *url,
 		const RaucChecksum *checksum);
-gboolean download_mem(GBytes **data, const gchar *url, gsize limit);
+gboolean download_mem(GBytes **data, const gchar *url, gsize limit, GError **error);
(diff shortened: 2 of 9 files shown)


mhei:
Just a quick review, not tested - but LGTM. Thank you very much for taking care of this!

ejoerns:
Note that the update I've pushed some days ago also harmonizes the path/URI handling between `rauc install` and `rauc info` and adds some basic tests.

PR #332, opened 2018-09-25
Improve error detection in sharness tests (rauc.t)

ejoerns:


ejoerns:
This would have detected #331

PR #346, opened 2018-10-11
 Add disable_counting logic

patrislav1:
When the boot process is interrupted for any reason (power issues, brown-out resets, users unplugging the gadget while it boots, etc.) more than three times in a row, then bootchooser will happily switch to the fall-back target, even though there's nothing wrong with the actual target at all.

This patch limits the scope of remaining_attempts countdown to the phase between flash update and verification after reboot (if the variable "disable_counting" is present in the bootstate node).

Additionally, it limits the number of flash writes to the barebox-state partition. (there used to be two flash writes for each and every boot cycle, but with this patch no more flash writes take place after the verification of the last update).

[Barebox mailing list thread with background
](http://lists.infradead.org/pipermail/barebox/2018-October/034961.html)

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -436,6 +436,30 @@ boot target is required::
           };
   };
 
+If an extra variable "disable_counting" is provided, then RAUC can communicate to
+the bootloader, if a firmware update has been marked as good and doesn't need to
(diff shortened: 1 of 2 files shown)


jluebbe:
This seems to be an unusual case and could be implemented via pre-/post-handlers, without additional complexity in RAUC itself.

Issue #325, opened 2018-09-10
install hook and other bundled files are not taken into account in checksum


Villemoes:
When implementing a custom install hook, changing the logic in that script should result in a "checksum mismatch", even if a bundle with the same image file has already been installed. More generally, if any custom hooks are in play, any file in the bundle can affect the result, so should somehow be taken into account before deciding that the update should/may be skipped.

I realize it's not that easy to implement. Thinking out loud: Perhaps the manifest would have to have a per-slot "extra_sha256", and a corresponding "extra_files" indicating which files in the bundle should go into computing that (computed as "cat $extra_files | sha256sum" or something). Normally, extra_sha256 would be just sha256 of the empty string, and that's what would be used for comparison when an existing slot status file does not have an extra_sha256.

Perhaps a simpler thing to implement is supporting force-install-same on a per-manifest basis.

Villemoes:
Actually, there's no need for an extra_sha256, only an extra_files (or whatever name is appropriate) that would also be included in the existing sha256. And looking closer, it actually seems rather straight-forward to implement (just have update_checksum and verify_checksum take a list of files instead of just one), but I'll wait a bit more for people to comment before making a POC patch.

ejoerns:
Well, first of all I agree with you that when using hooks the relation between the sha and the actual content that will be installed is not that clear. As you pointed out

1) The hook script might change logic
2) The image(s) that will be installed may change
3) Other scripts executed by the hook script may change

Taking care of the first one seems straight forward as we only use a single (base) hook script/binary. Thus each slot that has a hook defined will take the hook script into account for calculating its sha.

But, beside 'slot hooks' there are also 'install hooks' that are global and can affect the entire installation. I am not sure how to handle these, yet. At the moment we only have the 'install-check' hook that would only have impact on *if* something will be installed or not. But thinking of per-install post-install hooks I am not fully sure if we would have to take them into account or can argue that they only change what will happen before/after installation and will be executed anyway...

For images there is no drawback when using only a single one, the one we reflect in the `filename` key.
But as you pointed out, multiple files/images may be used by a single slot hook where you can do whatever you like. Maybe you have a bootloader + bitstream that you want to write into the same slot, or have an additional script being called by the hook script that performs data migration for you slot.

Adding these as `extra_files` to the `[image]` section in the manifest they belong to and allow tracking them sounds reasonable for me.

The remaining point to discuss is if one would use a separate sha for this or include this in the existing sha.
[...]

PR #348, opened 2018-10-12
Preparations for multi-file-checksums

Villemoes:
This is preparation for allowing multiple files to go into computing the sha256, cf. #325 . From this, it would mostly be a matter of figuring out how one would pass a list of file names and perhaps provide convenience functions for passing a single file. It also tries to fix something I've noticed a couple of times: That memory use of the RAUC process temporarily goes through the roof during an update.

For now only compile-tested, as I don't have any boards handy, and it is anyway mostly a POC.

### DIFF
--- a/include/checksum.h
+++ b/include/checksum.h
@@ -27,7 +27,7 @@ typedef struct {
  * @param error return location for a GError, or NULL
  * @return TRUE on success, FALSE if an error occurred
  */
-gboolean update_checksum(RaucChecksum *checksum, const gchar *filename, GError **error);
+gboolean compute_checksum(RaucChecksum *checksum, const gchar *filename, GError **error);
 
 /**
  * Verifies provided file checksum.
--- a/include/utils.h
+++ b/include/utils.h
@@ -2,6 +2,18 @@
 
 #include <glib.h>
 
+/* Use
+ *
+ *   g_auto(filedesc) fd = -1
+ *
+ * to declare a file descriptor that will be automatically closed when
+ * fd goes out of scope. The desctructor is guaranteed to preserve
+ * errno.
+ */
+typedef int filedesc;
+void close_preserve_errno(filedesc fd);
+G_DEFINE_AUTO_CLEANUP_FREE_FUNC(filedesc, close_preserve_errno, -1)
+
 #define R_LOG_DOMAIN_SUBPROCESS "rauc-subprocess"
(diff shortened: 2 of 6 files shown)


ejoerns:
@Villemoes Thanks for this useful pre-work! For finally reviewing it and allowing us to merge, you should care of fixing both the test and the style errors that Travis currently reveals.

ejoerns:
We've turned our testing infrastructure inside out. Thus please rebase onto current master and let us see what works then.

Villemoes:
> We've turned our testing infrastructure inside out. Thus please rebase onto current master and let us see what works then.

It rebased cleanly, and now it seems to pass. There's no functional change here, which is why I didn't add new tests (the existing tests serve to check that my refactoring was indeed a noop). Obviously new tests are needed if and when we do multiple files. But the avoid-huge-mmap thing is actually somewhat of an issue, so I hope this can get in without waiting for us to hash out the details of how to determine which files should be part of the checksum.

jluebbe:
While I'm not sue about the multi-file-checksum feature, these changes seem useful as they are. Just some questions:

I'd have expected mmap()-based hashing to be faster. Did you notice any performance changes? Which problems did you experience with the current approach on your device?

Actually, the checksum check of individual files during installation is not required for security, as the full bundle is already verified before mounting the squashfs. It's mainly a sanity check that the bundle was created correctly. So we could make it optional (for example via the system.conf). What's your opinion on that?

Villemoes:
> 
> I'd have expected mmap()-based hashing to be faster. Did you notice any performance changes? Which problems did you experience with the current approach on your device?

No, if anything mmap() is usually slower than accessing a file through a reasonable read-loop - but in simple benchmarks the difference is usually in the noise. Avoiding mmap() is not for performance, not of the RAUC part anyway, but as I wrote, allocating perhaps 500+MB of memory can be quite a hit on an embedded system, if it's even possible. So it's more for future-proofing, and avoiding _other_ processes' working set from being evicted.

> Actually, the checksum check of individual files during installation is not required for security, as the full bundle is already verified before mounting the squashfs. It's mainly a sanity check that the bundle was created correctly. So we could make it optional (for example via the system.conf). What's your opinion on that?

Hm. I do like sanity checks, and since we already have all the infrastructure on both the bundle-creating and bundle-consuming sides, I don't see why one would want to disable that check. Who knows, it _might_ catch some subtle bug in the squashfs decompression.

Issue #323, opened 2018-09-04
Travis xenial build environment seems to be broken.


ejoerns:
While claiming to be a beta, current Travis xenial environment that we moved too seems to behave more like a pre-alpha.

We face a lot of build issues that are not related to RAUC but look like a screwed up build environment.

We have to consider going back to trusty that only had one distinct randomly appearing issue but with much less frequency...

ejoerns:
To leave some historical notes here for future visitors: We moved to build custom docker imgas we spawn inside the Travis test and perform all test in this.
One further advantage of this method will be that it potentially allows us now to test again different custom defined build environments.

PR #350, opened 2018-10-19
Migrate to Using Docker for Testing

ejoerns:
This uses the custom build ejoerns/rauc-testing docker image for
running tests.

It aims to eliminate the current drawbacks (random errors probably caused by broken images) and pains we see when testing with Travis on current xenial (beta) environment.

Currently this uses my personal docker test environment build, thus we might want to move this to the 'rauc' GitHub user as well.

Fixes: #323

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,12 @@ language: c
 sudo: required
 services:
   - docker
+env:
+  - ci_env=`bash <(curl -s https://codecov.io/env)`
 
 before_script:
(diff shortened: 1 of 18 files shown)


ejoerns:
I've rebased and partly squashed the commits onto current master to now build with images generated by the rauc/testing docker repo https://hub.docker.com/r/rauc/testing/

ejoerns:
Some commit message fixes. Also added 2 patches for handling initialization errors properly.

Issue #360, opened 2018-11-09
Udpating appfs does not change the boot order


stobajas:
Hi !
I'm using rauc version v0.1.1 present in the meta-rauc for Yocto Pyro version.

I have a similar configuration to this one:
  [system]
  compatible=FooCorp Super BarBazzer
  bootloader=uboot

  [slot.rootfs.0]
  device=/dev/sda0
  type=ext4
  bootname=system0

  [slot.rootfs.1]
  device=/dev/sda1
  type=ext4
  bootname=system1

  [slot.appfs.0]
  device=/dev/sda2
  type=ext4
  parent=rootfs.0

  [slot.appfs.1]
  device=/dev/sda3
  type=ext4
  parent=rootfs.1

I have multiple slots and my use case is:
- update at least one bundle with one slot. The main reason is to reduce a amount of data downloaded if we don't need to.

For that, before each install, i'm mirroring the filesystems of the all the slots and then overwrite with the bundles.

The problem i'm facing is that when i am updating a bundle with only the appfs slot, Rauc does not modify the BOOT_ORDER u-boot parameter. I thought the "parent" parameter linked to rootfs's bootname in the appfs slot would be modify it.

Do you have faced already this issue and patched it in the newer versions ?

Thanks for your comments.

stobajas:
After a little investigation, in the v0.1.1 version, all happens in the function launch_and_wait_network_handler in src/install.c 
```
	/* Mark all parent destination slots bootable */
	g_message("Marking slots as bootable...");
	g_hash_table_iter_init(&iter, target_group);
	while (g_hash_table_iter_next(&iter, NULL, (gpointer *)&slot)) {
		if (slot->parent || !slot->bootname)
			continue;

		res = r_boot_set_primary(slot);

		if (!res) {
			g_warning("Failed marking slot %s bootable", slot->name);
			goto out;
		}
	}
```

Should i change with this patch ?

```
        g_message("Marking slots as bootable...");
        g_hash_table_iter_init(&iter, target_group);
        while (g_hash_table_iter_next(&iter, NULL, (gpointer *)&dest_slot)) {
-               if (dest_slot->parent || !dest_slot->bootname)
+               if (!dest_slot->parent || !dest_slot->bootname)
                        continue;
 
                res = r_boot_set_primary(dest_slot);
```
Or maybe it is better to set dest_slot->bootname of the parent slot at the initialization ?

Thanks for your comments

stobajas:
Hi, 
[...]

PR #362, opened 2018-11-12
src/install: fix how bootname slots are marked active/inactive 

ejoerns:
Fixes: #360

Preface
-------

When performing an installation, we need to deactivate booting the
relevant set of slots before actually updating them, and after a
successful installation, we must (by default) mark the set of slots as
active (primary) again.

Activating and deactivating here means that we tell the boot selection
layer (and thus the underlying bootloader used) to switch order,
priority, attempts counter, etc. in order to assure the selected slot is
not booted anymore (deactivate) or booted as the primary on (activated).

As a bootloader only boots exactly *one single* boot target.
A target group, as computed and used by RAUC, must contain exactly *one*
slot that serves as the interface to the bootloader, i.e. that has a
'bootname' set.

We can neither atomically install if we cannot deactivate the slot
group, nor we can install to slot groups where multiple slots are
activatable by the bootloader.

Issue Story
-----------

The current implementation of marking a slot 'bootable' / 'non-bootable'
simply iterates over a loop of all 'install_images' and calls the the
mark function for each slot having a bootname and no parent set.

Despite this works in most cases, this is wrong on several levels:

* 'install_images' contains exactly one entry for each slot for that an
  image was defined in the bundle Manifest.

  This means, if we have a 'rootfs' and an 'appfs' slot and the bundle
  only contains an 'appfs' image, the 'install_images' will only contain
  an entry for the 'appfs' slot.

  As a result of this, the mark_active() function will not be called for
  the 'rootfs' slot, which is the actual activatable slot of the install
  group formed of 'rootfs.x and appfs.x' and thus the slot will neither
  be deactivated nor reactivated (marked primary).

  This is a severe issue as it reports a successful update despite it
  was neither atomic nor lead to booting the newly installed system!

* The loop iteration potentially also allows that

  a) when we have no bootname set, silently no slot will be activated
     (similar to what we have above).

  b) when we have multiple slots with a bootname set (misconfiguration),
  then the mark_active() call will silently be executed for all of them,
  which is as wrong as not executing it.

Happy End
---------

We solve this all at once now by introducing the function
get_bootname_slot():

It returns *exactly one* slot from an target install group that has a
'bootname' set.
As we use the full 'target install group' instead of the reduced 'image
install group', we assure iterating over all relevant slots.

For cases where none or mutliple have set a bootname, the fuction
returns with a GError set.

Then we simply call `mark_active()` / `r_boot_set_state()` on the
(single) slot returned.


src/install.c
```diff
@@ -343,19 +343,30 @@ static gboolean slot_list_contains(GList *slotlist, const RaucSlot *testslot)
...
 		if (!iterslot->bootname)
-			continue;
+				continue;
 
+		/* If we've detected a bootname slot while already having found
+		 * a bootname slot, this is an error and we must abort.
+		 * This can be caused either by system misconfiguration or by
+		 * cases where we have bad bundles with images for multiple
+		 * concurrent bootname slots. */
 		if (bootnameslot) {
```
OnkelUlla: What if the bundle contains images for more than one slot of the targetgroup (e.g. _appfs_ and _rootfs_)? Shouldn't this be
```C
if (bootnameslot && bootnameslot != iterslot) {
```
instead?

src/install.c
```diff
@@ -343,19 +343,30 @@ static gboolean slot_list_contains(GList *slotlist, const RaucSlot *testslot)
...
 
-	/* Sanity check: only one bootname allowed */
-	g_hash_table_iter_init(&iter, targetgroup);
-	while (g_hash_table_iter_next(&iter, NULL, (gpointer*) &iterslot)) {
+	/* Iterate over all slots we have an image to install for */
+	for (GList *l = install_images; l != NULL; l = l->next) {
+		RaucSlot *iterslot = g_hash_table_lookup(targetgroup, ((RaucImage*) l->data)->slotclass);
+
+		/* If our slot has a parent with bootname set, use this for
+		 * testing (setting both parent and bootname is not allowed) */
```
OnkelUlla: I do not understand the supplementary explanation in parentheses - could you rephrase it a little bit?

ejoerns:
There is nothing wrong, but as the topic (or the solution) is still contentious, I just rebased the current state to throw away all unnecessary things to not overlook the essential.
Your comments are valid as far as I can see, but I decided to not yet invest time into a patch that I may have to rework in major parts. But I haven't forgotten them and they did not get lost obvioulsy.

To indicate this is not a ready PR I've changed the subject to 'RFC' and kept in some dirty squash comments. But maybe I should have left an explicit note.

I'd like to wait for @jluebbe's thoughts about this and the discussion in #360 before continuing here.

ejoerns:
I've updated the PR to remove the parent slot handling and reduce the impact as much as possible to have a chance getting it merged into 1.0.

@jluebbe what do you think?

ejoerns:
@OnkelUlla Thanks for review, I've fixed the commit message as requested.

OnkelUlla (dismissed):
Good work, you even found an additional typo! ;-)

PR #363, opened 2018-11-13
install: mark active child slot with its parent

stobajas:
If a slot has no bootname but its parent does, then replace the
dest_slot with its parent in order to call mark_active function to
change the boot order after a bundle installation.

Signed-off-by: Sandra Tobajas <sandra.tobajas@savoirfairelinux.com>


ejoerns:
Unfortunately, as a result of the discussion in #360 we would have to refuse this PR.

jluebbe:
Closing as discussed in #360.

PR #365, opened 2018-11-15
src/config_file: fix: always overwrite temporary system_variant

ejoerns:
We use the 'system_variant' key as a temporary storage for the filename
the key should be derived from.
No matter if we succeed with that or not, we must assure the key is
replaced afterwards.
With the current handling, in case for example get_variant_from_file()
failed and returned NULL, we do not overwrite 'system_variant' which
leads to having the variant set to the path to the filename we wanted to
read it from. This is clearly wrong.

We can simple fix this by setting the context 'system_variant'
unconditionally. Then it gets either the value 'NULL' (unset) or the
right value.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -212,19 +212,17 @@ static void r_context_configure(void)
 		if (!compatible) {
 			g_warning("Failed to read dtb compatible: %s", error->message);
 			g_clear_error(&error);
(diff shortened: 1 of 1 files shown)


ejoerns:
Fixed!

PR #366, opened 2018-11-15
src/config_file: add support for getting variant over executable

ejoerns:
This introduces the new system config key 'variant-bin' which acts
similar to the existing 'variant-name', 'variant-file', etc.

With 'variant-bin' you can specify the path to a script or binary
(executable) that prints out the variant of the board.

A possible use-case is to have the variant information encoded in
hardware bits and an application that may read these and transform to a
human-readable string.


jluebbe:
@ejoerns What was the reason to not collect this via the system-info handler?

ejoerns:
@jluebbe ouhmm.. did you read the code for `system-info` handler?... :)

This is for variant support which is already available for different cases (fixed name, dtb, file content). Thus I assumed this to be consistent.

jluebbe:
> @jluebbe ouhmm.. did you read the code for `system-info` handler?... :)
> 
> This is for variant support which is already available for different cases (fixed name, dtb, file content). Thus I assumed this to be consistent.

There current options for loading the variant are simple file reads and no process execution. I'd like to avoid executing many single-value helpers during startup and would prefer combine that into the system-info handler instead.

jluebbe:
This was implemented in a different way in #402.

PR #370, opened 2018-11-24
add example how to mount the correct data partition for redundant layout

FabianKnapp:
This is one possible solution how to select the correct data partition when using redundant data slots.

As talked with ejoerns at #rauc.

Feel free to modify.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -77,12 +77,15 @@ Release 1.2 (development)
 
 .. rubric:: Documentation
 
+* Added hints for creating ``/dev/data`` symlink to mount the right data
+  partition in dual data partition setups. (by Fabian Knapp)
 * Extended manpage to cover 'rauc status' subcommands. (by Michael Heimpold)
 * Fixed various typos.
 
-Contributions from: Bastian Krause, Ellie Reeves, Enrico Jörns, Gaël PORTAY,
-Jan Lübbe, Leif Middelschulte, Michael Heimpold , Stephan Michaelsen , Thomas
-Hämmerle, Thorsten Scherer, Tobias Junghans, Uwe Kleine-König
+Contributions from: Bastian Krause, Ellie Reeves, Enrico Jörns, Fabian Knapp,
+Gaël PORTAY, Jan Lübbe, Leif Middelschulte, Michael Heimpold , Stephan
+Michaelsen , Thomas Hämmerle, Thorsten Scherer, Tobias Junghans, Uwe
+Kleine-König
 
 Release 1.1 (released Jun 5, 2019)
 ----------------------------------
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -379,30 +379,34 @@ are listed below:
 | Fallback  | tricky (reconvert data?) | easy (old data!)          |
 +-----------+--------------------------+---------------------------+
 
+Managing a ``/dev/data`` Symbolic Link
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-For redundant data partitions the active rootfs slot has to mount the correct data partition dynamically.
-A udev ruleset can be used for this::
+For redundant data partitions the active rootfs slot has to mount the correct
+data partition dynamically.
+For example with ubifs, a udev ruleset can be used for this::
 
-  KERNEL=="ubi[0-9]_[0-9]", PROGRAM="/usr/bin/is-parent-active %k", ENV{isroot}="%c"
-  KERNEL=="ubi[0-9]_[0-9]", ENV{isroot}=="1", SYMLINK+="data"
+  KERNEL=="ubi[0-9]_[0-9]", PROGRAM="/usr/bin/is-parent-active %k", RESULT=="1", SYMLINK+="data"
 
-This example first determines if ubiX_Y is a data slot with an active parent rootfs slot and stores the result (%c) in the environment variable ``isroot``.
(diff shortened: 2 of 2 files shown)


FabianKnapp:
Yes, you are right, either the shell script or the udev rules are highly use-case specific:

Individual logic is always needed to find the correct rootfs slot for a given data slot.

I decided to make the udev rule more specific since I dont like shell scripting that much ;)

Will provide an example with the individual logic inside the shell script.

ejoerns:
> I decided to make the udev rule more specific since I dont like shell scripting that much ;)

You'd be the first one I know who likes udev scripting more than shell scripting, but.. ok.. :smirk:

FabianKnapp:
For my case the udev was indeed easier since I get the partition for free from udev.

Another thing:
Barebox uses ubi0:$ubiroot notation for booting from nand (which is also returned by `cat /proc/cmdline`): 

`bootargs="$bootargs root=ubi0:$ubiroot ubi.mtd=$rootfs_mtdblock"`
[line 89](https://git.pengutronix.de/cgit/barebox/tree/defaultenv/defaultenv-1/bin/boot?h=stable/v2017.11)

But udev uses ubiX_Y notation from the kernel. 

Any idea how to solve this without patching / scripting barebox?

FabianKnapp:
Ok, logic inside the shell script.

/usr/bin/is-parent-active :
```
#!/bin/bash

cmdline=$( cat /proc/cmdline )

# Modify this to get the corresponding rootfs partition from a data partition.
rootfs=$( echo $1 | sed 's/_3/_2/' )
rootfs=$( echo $rootfs | sed 's/_7/_6/' )

if [[ $cmdline == *"root=$rootfs "* ]]; then
	echo 1
else
	echo 0
fi
```

udev rules:
```
KERNEL=="ubi[0-9]_[0-9]", PROGRAM="/usr/bin/is-parent-active %k", ENV{isroot}="%c"
KERNEL=="ubi[0-9]_[0-9]", ENV{isroot}=="1", SYMLINK+="data"
KERNEL=="ubi[0-9]_[0-9]", ENV{isroot}=="0", SYMLINK+="data-inactive"
```

Should we use this for documentation?

ejoerns:
The udev rules now looks fine for me (maybe we can add another example for (more common) sdX devices, such as

```
KERNEL=="sd?[0-9]", PROGRAM="/usr/bin/is-parent-active %k", ENV{isroot}="%c"
```

but basically that's it.

The script however is a bit too specifc for an example and I also realized that using %k is not required in any case, but there's nothing bad to about having it ;)

If I remember correctly, you also have access to udev variables (e.g. `$ID_PART_TABLE_UUID`) from within the script.

One check I've used on a similar setup is

``` sh
DISK_ID=$(cat /proc/cmdline | sed -e 's/^.*root=PARTUUID=\([0-9a-z-]*\).*/\1/')

[ "$DISK_ID" == "$ID_PART_TABLE_UUID" ]  ...
```

Maybe we can sum it up to something like

``` sh
ROOTFS_DEV=<determine rootfs by using proc cmdline or mount>
TEST_DEV=<obtain expected rootfs device for processed device (%k)>

if [[ $ROOTFS_DEV == $TEST_DEV]]; then
	echo 1
else
	echo 0
fi
```

The actual handling is probably too use-case specific?

FabianKnapp:
Yep, in my opinion

`ROOTFS_DEV=<determine rootfs by using proc cmdline or mount>
TEST_DEV=<obtain expected rootfs device for processed device (%k)>

if [[ $ROOTFS_DEV == $TEST_DEV]]; then
	echo 1
else
	echo 0
fi`

is the best for documentation. Maybe, we can add one specific example.

jluebbe:
I'm wondering if this can't be done by letting udev look at the kernel cmdline:
```
IMPORT{cmdline}="root"
SUBSYSTEM="ubi", ENV{root}=="/dev/ubi0:aroot", ATTRS{mtd_num}=="1", ATTRS{name}=="adata", SYMLINK+="data"
SUBSYSTEM="ubi", ENV{root}=="/dev/ubi0:broot", ATTRS{mtd_num}=="1", ATTRS{name}=="bdata", SYMLINK+="data"
```
That way you have an explicit mapping by name. I've not tried this why recently, so I'm not sure I got the details correct. 

Also often useful is:
```
SUBSYSTEM=="ubi", KERNEL=="ubi*_*", SYMLINK+="ubi_mtd%s{mtd_num}_%s{name}"
```

ejoerns:
@jluebbe is right for this case and maybe we should use this by default.

It will *not* work when dealing with PARTUUID's as I had it in my setup, there we need a solution similar to what is in @FabianKnapp's last comment.

So, we should maybe start with the simple ubifs example and shortly note that other as we already did.

@FabianKnapp would you like to?

FabianKnapp:
As always: Feel free to modify.

jluebbe:
> As always: Feel free to modify.

I've added another commit on top, which simplifies the udev rules to get rid of the temporary variable. That makes it possible to use just a single rule.
I don't have a system ready to try this, but it should work like this. :)
If you have a chance to test this successfully, I'd squash these commits into one and merge it.

FabianKnapp:
> Were you able to try the approach based on `IMPORT{cmdline}="root"` and no external tool?

Not yet, but it is on my todo's

thirtythreeforty:
We hit this problem, so we solved it with a small tool and a simple configuration file:

https://github.com/KopisMobile/bootslot-mounts

jluebbe:
I've updated the branch to current master and changed some minor wording.

Issue #357, opened 2018-11-07
Release filename vs. tag name mismatch for release candidates


mhei:
Not a major problem, but not nice in my eyes:

The filename of the released tarball has a dot before "rc1", but the tag name contains a hyphen:
https://github.com/rauc/rauc/releases/download/v1.0-rc1/rauc-1.0.rc1.tar.xz

This prevents me from using something like this:

```
# helper variable
rauc_version=1.0.rc1

# download
wget https://github.com/rauc/rauc/releases/download/v${rauc_version}/rauc-${rauc_version}.tar.xz

```

Usual releases will probably not be affected due to missing hyphen, but maybe we can fix it.

ejoerns:
If I remember correctly what @jluebbe said, the git-version-gen script failed to handle the 1.0-rc1 name because of the hyphen. This made us use the dot as a 'workaround'.

As you said, fortunately we will not run into this issue when doing the 1.0 release and it is still unclear if we continue doing release candidates as we did for 1.0 now to not break anything on such a remarkable release tag ;) (Despite the most of the important functionality is backed by testing, anyway)

PR #377, opened 2018-12-11
Issue 357: Improve version mangling

ukleinek:
by making the globs for detecting an old git stricter the check actually works also for rc releases. Together with the second change a release candidate doesn't use a . in the version string which fixes issue 357.

### DIFF
--- a/build-aux/git-version-gen
+++ b/build-aux/git-version-gen
@@ -183,9 +183,14 @@ then
             ;;
     esac
 
-    # Change the first '-' to a '.', so version-comparing tools work properly.
-    # Remove the "g" in git describe's output string, to save a byte.
-    v=`echo "$v" | sed 's/-/./;s/\(.*\)-g/\1-/'`;
+    case "$v" in
+        *-[0-9]*-g*)
+	    # Change the '-' before the number of commits since the last tag to a '.',
+	    # so version-comparing tools work properly.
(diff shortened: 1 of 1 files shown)


jluebbe:
It seems that upstream for that script has fixed the issue we see:
https://github.com/coreutils/gnulib/commits/master/build-aux/git-version-gen
Do you think that updating the the most recent one would fix #357 as well?

ejoerns:
@ukleinek Any comments on this from your side?

ukleinek:
I did:

    $ git checkout v1.0-rc1
    ...
    $ git show f63891af3567fd2f6c8e754e46b76fd7ff0a3948:build-aux/git-version-gen > git-version-gen-old
    $ wget https://raw.githubusercontent.com/coreutils/gnulib/9447bd9732613c5419611bcb2b8b304d748b62b6/build-aux/git-version-gen
    ....

Just for the record:

    $ sh git-version-gen-old .tarball-version; echo # this is the broken one
    1.0.1-rc1
    $ sh build-aux/git-version-gen .tarball-version; echo # this is the fixed up with a . instead of -

So yes, an update would also fix #357.
    1.0.rc1

With the new upstream version it seems to do the right thing:

    $ sh git-version-gen .tarball-version; echo
    1.0-rc1

ejoerns:
As #436 was merge, I assume this to be fixed, too

PR #378, opened 2018-12-11
configure.ac: add a fallback version

ukleinek:
not sure this is welcome, but given it is so easy: This uses a feature of the version generation script to provide a fallback version that is used in the absence of a .tarball-version file and git metadata.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2,7 +2,7 @@
 # Process this file with autoconf to produce a configure script.
 
 AC_PREREQ([2.69])
-AC_INIT([rauc], m4_esyscmd([./build-aux/git-version-gen .tarball-version]), [rauc@pengutronix.de])
+AC_INIT([rauc], m4_esyscmd([./build-aux/git-version-gen --fallback 1.0-rc1 .tarball-version]), [rauc@pengutronix.de])
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_MACRO_DIR([m4])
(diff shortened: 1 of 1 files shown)


jluebbe:
This covers the case where we don't have a git repo and no .tarball-version either, right? As we most likely have a modified codebase in that case (but can't be sure without git), I'd prefer to indicate that in the version. Possibly 1.0-rc1+?

ukleinek:
> This covers the case where we don't have a git repo and no .tarball-version either, right? As we most likely have a modified codebase in that case (but can't be sure without git), I'd prefer to indicate that in the version. Possibly 1.0-rc1+?

Sounds good.

jluebbe:
@ukleinek It seems that ``--fallback`` is no solution for this problem. Do you have another idea? Otherwise I'd close this.

ukleinek:
Given that the fallback only works if git is not available, the script needs the following change to do what I intended it to do:

    -elif test "x$fallback" = x || git --version >/dev/null 2>&1; then
    +elif test "x$fallback" = x; then

If that is welcome, I can prepare a patch for that.

jluebbe:
Yes, that seems like it would solve the issue independent of whether git is installed.

ejoerns:
@ukleinek Still interested in fixing this? ;)

ukleinek:
After discussing with @ejoerns and @jluebbe we agreed that my usecase can be solved differently and having to maintain the version by hand is too annoying. Will close accordingly.

PR #381, opened 2018-12-12
cms_verify() directory support.

evanedstrom:
Rewrite portion of cms_verify() to accommodate a directory containing multiple certificates files. This works instead of or in addition to the existing keyring path.

Found test/signature.c was too finicky to add to, so added a set_up() and tear_down() to give a consistent test environment. This fixes some memory leaks and reduces code duplication.


evanedstrom:
This is a first open source contribution for me. If there are things I could improve on other than the code itself; or if the code style isn't fitting into the repo I would appreciate the feedback.

ejoerns:
@evanedstrom Thanksk for your contribution!

About the code style: Yes it does not fully fit. ;)
We have an uncrustify configuration and skript in the Repo that also Travis used to Check code style conformity. This checks fails (and reports the diff) if you look into the Travis output linked here.

You should be able to run [uncrustfy.sh](https://github.com/rauc/rauc/blob/master/uncrustify.sh) (with a recent version) locally to fix this.

But, Travis also stumbles over Issues when executing the test suite. Please also have a look at this.

evanedstrom:
Okay, I've addressed the formatting issues and resolved the unit test issues. I didn't realize CI would run `make distcheck`; and hadn't updated the configure.am to use the new test keys and certs.

I'm used to squashing the fixes at the end of a PR, which is why I didn't "sign off" on them. If that is necessary to proceed I am happy to rebase now; otherwise the DCO complaint will go away when rebasing. Let me know.

jluebbe:
As I understand it, the directory&hash support in OpenSSL is primarily a performance optimization for the case where you have a large number of certificates (trust anchors). This is mainly the case when using HTTPS with public servers.
For RAUC, I'd not expect more than a few trust anchors even in unusual setups. So: what is your use-case?

evanedstrom:
Absolutely, performance for handling numerous certificates is a great reason to use directory&hash support in OpenSSL; however that is not the reason I wanted to leverage this.

We ran into the scenario where we needed to switch from a prototype test key to an official dev key, but wanted to keep both on the device for a period of time to make development easier. In the future, our development images will allow the installation of both development-signed and production-signed images. I've found tracking multiple certs in one file to be tedious and hard to version. Having a directory to put certificates in, even if it is just a couple of them, has been much easier to manage as certificates come and go throughout the development process.

xyzzy42:
It's also the case the most other systems using OpenSSL (or GnuTLS for that matter) that need to check a cert against trust anchors support the CA cert directory system.  For instance, one might use curl to download a bundle, using a directory of CA certs to verify the server's TLS cert.  But this same directory can't be used with rauc.

Package management works better when a file is owner by just one package.  E.g., it's far easier for a package to add a cert file to directory when installed, and remove the file when uninstalled, than for the package to try to append the cert to a file with other certs in it, and somehow delete its cert from the file when it's uninstalled.

evanedstrom:
Hi, just wanted to follow up and see if you had thoughts on the changes here.

ejoerns:
From my point of view, the argument of being package handling-friendly is one that cannot simply be argued away. I really appreciate the concept of those `conf.d` directories most modern system applications provide.

@jluebbe What do you think about these arguments?

jluebbe:
> Hi, just wanted to follow up and see if you had thoughts on the changes here.
Please excuse the long delay, @xyzzy42 arguments have convinced me that his is a useful feature for rauc.

jluebbe:
I've squashed the fixup commits.

jluebbe:
I've rebased onto current master.

ejoerns:
Fixed uncrustify issues.

jluebbe:
@evanedstrom Thanks a lot for your contribution (and patience)!

PR #383, opened 2018-12-12
src: fix memory leaking gchar's returned by g_data_input_stream_read_line()

ejoerns:

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -137,7 +137,7 @@ static gboolean launch_and_wait_variables_handler(gchar *handler_name, GHashTabl
 			continue;
 
 		if (g_str_has_prefix(outline, "RAUC_")) {
-			gchar **split = g_strsplit(outline, "=", 2);
+			g_auto(GStrv) split = g_strsplit(outline, "=", 2);
 
 			if (g_strv_length(split) != 2) {
 				g_free(outline);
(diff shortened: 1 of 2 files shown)


src/context.c
```diff
@@ -145,6 +145,8 @@ static gboolean launch_and_wait_variables_handler(gchar *handler_name, GHashTabl
 			g_hash_table_insert(variables, g_strdup(split[0]), g_strdup(split[1]));
 			g_strfreev(split);
 		}
+
+		g_free(outline);
```
OnkelUlla: Please additionally fix the memory leaks that occur up in [line 143's](https://github.com/rauc/rauc/blob/c441d4d2001c1413ec570bc3dc331ff85e66c23f/src/context.c#L143) `continue` statement: before continuing the do-loop a `g_strfreev(split)` and a `g_free(outline)` are missing.

src/install.c
```diff
@@ -708,8 +710,10 @@ static gboolean run_bundle_hook(RaucManifest *manifest, gchar* bundledir, const
 		outline = g_data_input_stream_read_line(datainstream, NULL, NULL, NULL);
 		if (outline) {
 			g_clear_pointer(&hookreturnmsg, g_free);
-			hookreturnmsg = outline;
+			hookreturnmsg = g_strdup(outline);
 		}
+
+		g_free(outline);
```
OnkelUlla: What does in comparison to the original code the combination of `g_strdup(outline)` in line 713 and `g_free(outline)` in line 716 change other than adding an additional copy operation?

ejoerns: You're right. I've first 'fixed' the gfree and then noticed we need to keep at least the latest content of outline. But did not notice this is handled by the freeing `hookreturnmsg` already.

src/context.c
ejoerns: I've added another commit to also fix this by using auto ptr.

PR #387, opened 2018-12-18
src/bootchooser.c: Report error on missing primary slot in uboot_get_primary

### LINKED ISSUE #386: Crash due to incomplete error reporting in uboot_get_primary
The `uboot_get_primary` function can return NULL without setting the provided `GError` object in certain cases. But all callers of `uboot_get_primary` expect the `GError` object to be valid if NULL is returned, resulting in a segfault due to NULL pointer dereference.

This can happen in two cases:

* The `BOOT_ORDER` variable is an empty string, skipping the complete for loop in `uboot_get_primary` and directly returning `primary` that is initialized to NULL. This case should no happend under normal circumstances, as it is a broken system configuration. Nevertheless RAUC should not crash because of it.
* All relevant `BOOT_x_LEFT` variable are 0 making the inner while loop never set `primary`. This can happend in a working system, and that's how I found the problem.

When the system boots with `BOOT_x_LEFT` set to 1 and U-Boot picks slot x then `BOOT_x_LEFT` is decremented to 0 until the slot is marked as good again. In the meantime it is not possible to determine the primary slot. Maybe `uboot_get_primary` should report the currently booted slot as primary if the primary slot cannot be determined otherwise.

photron:
Equivalent to how barebox_get_primary and efi_get_primary do it.

Fix bootchooser test to cover this case properly. The test covered this
case before already, but because it used R in BOOT_ORDER without also
specifying BOOT_R_LEFT the test case failed differently hiding the missing
error reporting on the absence of a primary slot.

Fixes #386.

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -666,6 +666,14 @@ static RaucSlot* uboot_get_primary(GError **error)
 			break;
 	}
 
+	if (!primary) {
+		g_set_error_literal(
+				error,
+				R_BOOTCHOOSER_ERROR,
+				R_BOOTCHOOSER_ERROR_PARSE_FAILED,
+				"Unable to find primary boot slot");
+	}
+
 	return primary;
 }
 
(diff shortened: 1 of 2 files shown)


jluebbe:
Thanks for you help @photron!

PR #390, opened 2018-12-20
doc: clarify 'use-bundle-signing-time'

mhei:
While the name already suggests that this is a boolean type, we
could also make it obvious in the description.

While at, fix a typo in CHANGES which mentioned to introduction
of this key.


ejoerns:
@mhei maybe we should do the same I did in https://github.com/rauc/rauc/commit/d3f30a21f9cb62badb2cabbe8c50ebd43cbb0de5 here, too?

Yeah, seems as if we're killed by an uncrustiy update... (Travis error)

PR #391, opened 2018-12-20
src/install: prevent g_strchomp() from being called with NULL

ejoerns:
`read_file_str()` may return NULL if reading the file failed.
Calling `g_strchomp()` with NULL however will lead to an assertion error.

This is fixed by checking the return of `read_file_str()` and directly
return NULL in this case. Another benefit of this will be that we can at
least print the error message that will be returned by `read_file_str()`
which would have been lost otherwise.


jluebbe:
The g_autofree for content was wrong, as g_strchomp modifies the passed string in place and returns it. So with g_autofree it would be freed before the function actually returned.

PR #393, opened 2019-01-03
src/signature.c: make a-month-from-now computation more robust

Villemoes:
The current code tries to handle tm_mon wrapping, but fails to
increment tm_year - also, there's no handling of, say, March 31 ->
April 31, making the resulting broken-down time invalid, in which case
timegm() need not produce a sensible result.

Instead of doing these fragile computations on a struct tm, just add
30 days to now.

Signed-off-by: Rasmus Villemoes <rasmus.villemoes@prevas.dk>

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -409,17 +409,9 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, g
 
 		for (int i = 0; i < sk_X509_num(verified_chain); i++) {
 			const ASN1_TIME *expiry_time;
-			struct tm *next_month;
-			time_t now;
 			time_t comp;
-			time(&now);
 
-			next_month = gmtime(&now);
-			next_month->tm_mon += 1;
-			if (next_month->tm_mon == 12)
-				next_month->tm_mon = 0;
-			comp = timegm(next_month);
-
+			comp = time(NULL) + 30*24*60*60;
 			expiry_time = X509_get0_notAfter(sk_X509_value(verified_chain, i));
 
 			/* Check if expiry time is within last month */


src/signature.c
```diff
@@ -409,16 +409,10 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, g
...
 			time_t now;
 			time_t comp;
 			time(&now);
-
-			next_month = gmtime(&now);
-			next_month->tm_mon += 1;
-			if (next_month->tm_mon == 12)
-				next_month->tm_mon = 0;
-			comp = timegm(next_month);
+			comp = now + 30*24*60*60;
```
OnkelUlla: As the man page of [time()](http://man7.org/linux/man-pages/man2/time.2.html#BUGS) in its BUGS section states, one should use NULL as the function's argument and work with its return value instead. Please make a simple additional simplification by removing `now` completely and computing `comp` directly via

    time_t comp = time(NULL) + 30*24*60*60;

OnkelUlla:
> Updated, though I kept the declarations separate from the assignments, since I think the expiry_time line would otherwise be a bit of a mouthful.

No problem with that - thanks for changing.

OnkelUlla:
> BTW, it's a bit sad that codecov discourages cleanups that remove lines of code, and refactorings that eliminate boilerplate (at least when said code is among the lines hit). I know nothing about the tool, but is there any way to make it track basic blocks rather than LOC?

I didn't dive deep into _codecov_ as well. Enrico, could you stand up and elaborate if you know more?

ejoerns:
@Villemoes @OnkelUlla For now, the coverage report looks fine. I cannot inspect how the result of the previous was interpreted.
Absolute coverage decreased, that is right, but as long as we have a 100% diff hit I would say this is more meaningful. However, now codecov also states that relative coverage increased and thus ignores absolute coverage decrease.

Issue #388, opened 2018-12-19
Slot already mounted detection fails with symlinks


xyzzy42:
I've configured a slot to have a device path which is a symlink, such as `/dev/disk/by-partlabel/rootfs0`.  Rauc tries to determine if this slot is already mounted, but it does this by comparing the device name, like above, with a list of mounted device from `g_unix_mounts_get()`.  But that list will use a device name like `/dev/mmcblk1p2`, since it's not created by the path which was mounted, but rather reconstructed from the major/minor of the device.  A simple string comparison of the two paths fails, even though they are the same device.

This results in `determine_slot_states()` failing to detect the existing mount and set `ext_mount_point`.  Later, `load_slot_status_locally()` will try to remount the already mounted filesystem and fail.

xyzzy42:
A solution to this would be to normalize device paths before comparing them.  This could be done by recursively following symlinks, i.e. realpath(), but that does not catch a number of cases.

The mounted device could be a file which is loopback mounted (which is already detected by rauc).  This file could have multiple hard links to it.  That would not be detected by comparing realpaths.

There is no rule that only one device node of a given type/major/minor combo can exist.  It's perfectly possible to mknod multiple nodes for the same underlying device.  This, much like multiple hardlinks to a file, it not detected by comparing realpaths.

Here's a solution which allows things that can be mounted to be compared in a way that is immune to the above and not terribly complex.

```
// Something that is, or can be, mounted onto a mount point
typedef struct {
        gboolean is_device; // vs being a loop mounted file
        dev_t dev;    // the device, or for a file, the device the file is on
        ino_t inode;  // inode of file for a non-device
} MountableObj;

// Take a device (or file) path and normalize it
static MountableObj *normalize_mountable_object(const gchar *devicepath)
{
        MountableObj *obj = g_new0(MountableObj, 1);
        GStatBuf st;

        if (g_stat(devicepath, &st) == -1) {
                /* Virtual filesystems like devpts trigger this case */
                g_debug("Can't stat '%s', must not be mountable", devicepath);
                return NULL;
        }

       if (S_ISBLK(st.st_mode)) {
               obj->is_device = TRUE;
               obj->dev = st.st_rdev;
       } else if (S_ISREG(st.st_mode)) {
               obj->is_device = FALSE;
               obj->dev = st.st_dev;
               obj->inode = st.st_ino;
       } else {
               g_debug("Device '%s' is not something which is mountable", devicepath);
[...]

PR #395, opened 2019-01-21
Fixes for external mount point detection and handling

ejoerns:
* Support mount point detection for symlinked devices in system.conf
* Propagate external mount points to the frontend (D-Bus / CLI)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -967,8 +967,8 @@ static gchar* r_status_formatter_readable(RaucStatusPrint *status)
 			g_string_append_printf(text, ", parent=%s", slot->parent->name);
 		else
 			g_string_append(text, ", parent=(none)");
-		if (slot->mount_point)
-			g_string_append_printf(text, ", mountpoint=%s", slot->mount_point);
+		if (slot->mount_point || slot->ext_mount_point)
+			g_string_append_printf(text, ", mountpoint=%s", slot->mount_point ? slot->mount_point : slot->ext_mount_point);
 		else
 			g_string_append(text, ", mountpoint=(none)");
 		if (slot->bootname)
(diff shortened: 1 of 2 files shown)


src/install.c
```diff
@@ -90,8 +90,17 @@ gboolean determine_slot_states(GError **error)
...
 		GUnixMountEntry *m = (GUnixMountEntry*)l->data;
-		g_autofree gchar *devicepath = resolve_loop_device(g_unix_mount_get_device_path(m));
-		RaucSlot *s = find_config_slot_by_device(r_context()->config,
+		g_autofree gchar *devicepath = NULL;
+		RaucSlot *s;
+
+		/* We can ignore internal devices such as tmpfs, cgroups, etc. here.
+		 * They won't start with a '/', thus we simply use this for
+		 * checking */
+		if (g_unix_mount_get_device_path(m)[0] != '/')
```
jluebbe: Can this string be empty or NULL?

ejoerns: Mh.. from the implementation I fear we cannot be sure that it will not become NULL. However, empty string should be '\0' and thus not be an issue here, right?

ejoerns: Pushed a fixup addressing this.

ejoerns:
Seems as if we should first have a look at #388 first, which popped from my mental stack (maybe because it was around Christmas..).
It adresses at least the issue I resolved with `realpath()` for now.

jluebbe:
> Seems as if we should first have a look at #388 first, which popped from my mental stack (maybe because it was around Christmas..).
> It adresses at least the issue I resolved with `realpath()` for now.

Yes, that approach seems to be better.

jluebbe:
> Should I create a new PR targeting the issues addressed by the last commit? Or just force-push with the primary two removed? (While assuming the essential of the second one will go into #388, too)

You can just force-push with only "src/main: merge mount_point and ext_mount_point for frontend ￼…" remaining.

ejoerns:
> > Should I create a new PR targeting the issues addressed by the last commit? Or just force-push with the primary two removed? (While assuming the essential of the second one will go into #388, too)
> 
> You can just force-push with only "src/main: merge mount_point and ext_mount_point for frontend ￼…" remaining.

Done.

PR #400, opened 2019-01-30
Allow to dump signer certificate without verification

ejoerns:
Especially for debugging-purpose, it is quite useful to allow dumping the signer certificate also without having to verify it before.


test/rauc.t
```diff
@@ -202,6 +202,11 @@ test_expect_success "rauc info" "
     info $SHARNESS_TEST_DIRECTORY/good-bundle.raucb
 "
 
+test_expect_success "rauc info dump-cert unverified" "
+  rauc --no-verify --dump-cert \
+    info $SHARNESS_TEST_DIRECTORY/good-bundle.raucb
```
jluebbe: Shouldn't we test this on bad bundle?

src/signature.c
```diff
@@ -522,6 +522,41 @@ gchar* print_signer_cert(STACK_OF(X509) *verified_chain)
...
+	if (signers == NULL) {
+		g_set_error_literal(
+				error,
+				R_SIGNATURE_ERROR,
+				R_SIGNATURE_ERROR_GET_SIGNER,
+				"Failed to obtain signer info");
+		return NULL;
+	}
+
+	ret = dump_cms(signers);
```
jluebbe: Should you free the signers here?

src/signature.c
```diff
@@ -522,6 +522,41 @@ gchar* print_signer_cert(STACK_OF(X509) *verified_chain)
 	return ret;
 }
 
+gchar* print_signer_cert(GBytes *sig, GError **error)
```
jluebbe: This doesn't print the cert but returns it as a human-readable string.

src/signature.c
```diff
@@ -502,7 +502,7 @@ gchar** get_pubkey_hashes(STACK_OF(X509) *verified_chain)
 	return ret;
 }
 
-gchar* print_signer_cert(STACK_OF(X509) *verified_chain)
+static gchar* dump_cms(STACK_OF(X509) *verified_chain)
```
jluebbe: The verified_chain argument could use a better name.

include/signature.h
```diff
@@ -150,4 +151,6 @@ gchar* print_cert_chain(STACK_OF(X509) *verified_chain);
  *
  * @return TRUE if succeeded, FALSE if failed
  */
-gboolean cms_get_cert_chain(CMS_ContentInfo *cms, X509_STORE *store, STACK_OF(X509) **verified_chain, GError **error);;
+gboolean cms_get_cert_chain(CMS_ContentInfo *cms, X509_STORE *store, STACK_OF(X509) **verified_chain, GError **error);
+
+gboolean cms_info_file(const gchar *filename, GBytes *sig, gsize limit, CMS_ContentInfo **cms, GError **error);
```
jluebbe: This doesn't match any function.

test/rauc.t
ejoerns: The relevant check for this patch is that --dump-cert works without verification. That verification is note performed here is assured as we do not provide a keyring.

However, despite a bit unrelated, I now added a check to assure that rauc info fails verification upon invalid signature.

jluebbe:
Looks good to me now. I've rebased onto master and squashed the fixup.

PR #403, opened 2019-01-30
Add option for automatically resizing written ext4 images

ejoerns:
This adds a `resize` slot option that will trigger a call of `resize2fs` after having written an ext4 image.

The use case for this is having an image smaller than the actual partition size where the file system should span the entire partition afterwards.


jluebbe:
I've removed the ``-f`` resize2fs option and printed the destination device instead of the image path.

PR #404, opened 2019-01-31
Introduce efi-use-bootnext option to change primary-marking behavior

ejoerns:
This adds the efi-use-bootnext option that allows disabling the default
BootNext setting for marking a slot primary as this may be inappropriate
for use cases where the BIOS already handles the slot switching
properly.

For these cases, simply sorting the slot as the first element in the
bootloader is more appropriate.


ejoerns:
Mh.. I should have submitted some kind of 'Bundle Not found' feature here... #toolate

PR #406, opened 2019-02-08
Normalize device names to find mounted slots

### LINKED ISSUE #388: Slot already mounted detection fails with symlinks

xyzzy42:
This way a mounted slot can be found it the path used to mount the
device is not the same as the path in the rauc configuration file.
E.g., the configuration file uses a link like
/dev/disk/by-partlabel/rootfs0 while the kernel's list of mounted
devices uses a name like /dev/mmcblk1p2.

Not being able to detect that the current rootfs is already mounted
causes rauc to try to mount the slot, which will fail.

Fixes #388.

Signed-off-by: Trent Piepho <tpiepho@impinj.com>


ejoerns:
:+1: Thanks for your submission!
@xyzzy42 please let uncrustify run on your code and fix the errors mentioned by Travis so we can have this PR ready for review.

xyzzy42:
There's an issue getting the tests to work.

`find_slot_by_device`/`find_config_slot_by_device` are used twice in rauc's code, both times on a running target system where we want to know if a real device corresponds to a slot.

It's also used over a dozen times in the test code.  This use doesn't work as intended because the devices, both the one used to look up a slot and the device named in the slot, do not exist. On a target, we couldn't say if they are the same or not. So the lookup fails.  In actual use this behavior seems to me like it would be correct.  The "is mounted?" and "was booted from?" checks should return false given a device that does not exist.

It looks to me like `find_config_slot_by_device()` was being used simply as away to get a handle to slot, lacking any other way. It wasn't intended to test the by device lookup code.  Maybe a better way would be to create a `find_slot_by_name()` function and use that in the test code?

src/config_file.c
```diff
@@ -541,6 +590,21 @@ RaucSlot *find_slot_by_bootname(GHashTable *slots, const gchar *bootname)
 	return slot;
 }
 
+static RaucSlot *find_slot_by_name(GHashTable *slots, const gchar *name)
```
ejoerns: There is no benefit of splitting this up if it is not used. We must not add unnecessary abstractions.

src/config_file.c
```diff
@@ -503,17 +545,24 @@ RaucSlot *find_slot_by_device(GHashTable *slots, const gchar *device)
 {
 	GHashTableIter iter;
 	RaucSlot *slot;
+	g_autofree MountableObj *obj = normalize_mountable_object(device);
+
+	if (!obj) {
```
ejoerns: remove braces here. For one-line statements we normally do not use braces (according to kernel coding style)

src/config_file.c
```diff
@@ -503,17 +545,24 @@ RaucSlot *find_slot_by_device(GHashTable *slots, const gchar *device)
...
+		return NULL;
+	}
 
 	g_return_val_if_fail(slots, NULL);
 	g_return_val_if_fail(device, NULL);
 
 	g_hash_table_iter_init(&iter, slots);
 	while (g_hash_table_iter_next(&iter, NULL, (gpointer*) &slot)) {
```
ejoerns: For compatibility reasons maybe we should not break the prior behavior of this function.

Is there any disadvantage if we simply do the former string comparison as a fallback here? This would assure that we match more cases but not lose others. Checking if the device that was listed in the config actually exists, etc. should anyway be handled somewhere else.
This would also remove the need for (`find_config_slot_by_name`, despite it might still be useful as convenience fo tests, etc.

src/config_file.c
```diff
@@ -492,6 +493,47 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
+
+// Take a device (or file) path and normalize it
+static MountableObj *normalize_mountable_object(const gchar *devicepath)
+{
+	MountableObj *obj;
+	GStatBuf st;
+
+	if (g_stat(devicepath, &st) == -1) {
+		/* Virtual filesystems like devpts trigger case */
+		g_debug("Can't stat '%s', must not be mountable", devicepath);
```
ejoerns: I can't get the meaning of the error message, what do you mean by `must not be mountable`?.

And it might also be useful that we evaluate errno here to complete information using `gstrerror(errno)`.

src/config_file.c
xyzzy42: Trying to be consistent with how find_config_slot_by_device and find_slot_by_device are split.  The benefit would be someone who sees find_slot_by_device() could look for and find find_slot_by_name.

src/config_file.c
xyzzy42: I can't think of anything which can not be stat'ed, but can be mounted.  So if the purpose is to find a mounted device or file, and the path can not be stated, then it must not be a mounted device or file.

ejoerns: I stick to my opinion that the message is misleading because one can read 'must not be mountable' as an instruction in the sense of 'It should not be a mountable path', which is quite the opposite of what you intend to say here.
What about something like: `Not a mountable path: %s: %s", devicepath, gstrerror()`

What `gstrerror(errno)` will tell you in most cases is something like 'No such file or directory', which is quite more concrete than saying that it is not mountable. If we have a concrete meaningful error message: why hiding it?

xyzzy42:
Believe I have addressed all comments.

I changed it to do the pathname compare first, device based compare second.   The path string compare is much cheaper (no syscalls), and if the paths are the same, then the device comparison must the same too (excluding races!).  So this is a faster way to get the same result as device compare first, path compare second.

xyzzy42:
Is there anything else left here?  I believe I've made all requested changes.

src/install.c
```diff
@@ -90,12 +90,19 @@ gboolean determine_slot_states(GError **error)
...
-		g_autofree gchar *devicepath = resolve_loop_device(g_unix_mount_get_device_path(m));
-		RaucSlot *s = find_config_slot_by_device(r_context()->config,
-				devicepath);
-		if (s) {
-			s->ext_mount_point = g_strdup(g_unix_mount_get_mount_path(m));
-			g_debug("Found external mountpoint for slot %s at %s", s->name, s->ext_mount_point);
+		/* Relative paths will be from something like tmpfs, cgroups,
+		 * etc. which we want to ignore rather than try to follow some
+		 * random symlink named "tmpfs" in the CWD.  */
+		if (g_unix_mount_get_device_path(m)[0] == '/') {
```
ejoerns: Following the convention in the rest of the code and to reduce nesting depth, I would prefer having a

``` c
if (g_unix_mount_get_device_path(m)[0] != '/')
        continue
```

here. Maybe then we can also remove the unreadable line break after in assignment. We do not depend that strictly on code line length.


The rest looks fine for me now! Sorry for being a bit unresponsive, but we are quite busy on all fronts 'at the moment' thus struggling sometimes to follow all updates as they would deserve :/

I've also marked this PR as 1.1 release-critical now.

obbardc:
Hi @ejoerns what is the status of this pull request, please?

ejoerns:
@obbardc the 'funny' thing is that it just hit me, too in a project today. There I still use a different method for solving the same issue and iirc we had a 3rd one somewhere around, too.

We definitely need to face and fix this. This PR should not have been open for so long

@obbardc Did you already test these patches?

obbardc:
> @obbardc the 'funny' thing is that it just hit me, too in a project today. There I still use a different method for solving the same issue and iirc we had a 3rd one somewhere around, too.
> We definitely need to face and fix this. This PR should not have been open for so long
> @obbardc Did you already test these patches?
Hardcoded device names is not great, but for now what I have works.
Not tested the patches yet; it's on my list very early next week :-)

jluebbe:
@ejoerns I've rebased this onto the current master, which had moved some functions to `src/slot.c`.

src/install.c
```diff
@@ -98,19 +98,33 @@ gboolean determine_slot_states(GError **error)
...
-			s->ext_mount_point = g_strdup(g_unix_mount_get_mount_path(m));
-			g_debug("Found external mountpoint for slot %s at %s", s->name, s->ext_mount_point);
+		g_autofree gchar *devicepath = NULL;
+		RaucSlot *s = NULL;
+
+		/* Relative paths will be from something like tmpfs, cgroups,
+		 * etc. which we want to ignore rather than try to follow some
+		 * random symlink named "tmpfs" in the CWD.
+		 */
+		if (g_unix_mount_get_device_path(m)[0] != '/')
```
jluebbe: I'm not sure this is correct for ubifs, which has device names such as `ubi0:data`.

src/slot.c
```diff
@@ -30,21 +33,74 @@ void r_slot_free_status(RaucSlotStatus *slotstatus)
...
+// Take a device (or file) path and normalize it
+static MountableObj *normalize_mountable_object(const gchar *devicepath)
+{
+	MountableObj *obj;
+	GStatBuf st;
+
+	if (g_stat(devicepath, &st) == -1) {
+		/* Virtual filesystems like devpts trigger case */
+		g_debug("Can't stat '%s', assuming unmountable: %s", devicepath, g_strerror(errno));
+		return NULL;
```
jluebbe: This will break support for UBIFS.

src/slot.c
```diff
@@ -30,21 +33,74 @@ void r_slot_free_status(RaucSlotStatus *slotstatus)
...
 RaucSlot *r_slot_find_by_device(GHashTable *slots, const gchar *device)
 {
 	GHashTableIter iter;
 	RaucSlot *slot;
+	g_autofree MountableObj *obj = NULL;
 
 	g_return_val_if_fail(slots, NULL);
 	g_return_val_if_fail(device, NULL);
 
+	obj = normalize_mountable_object(device);
```
jluebbe: How should we handle NULL here?

jluebbe:
I think we can merge the first two commits as they are, though.

jluebbe:
I've merged the first two commits and rebased this PR.

jluebbe:
The open comments need to be solved, but the general approach seems fine.

Issue #248, opened 2018-03-14
Improved D-Bus Interface


ejoerns:
I would like to start this as a discussion base for future extensions, and possible re-designs of the current. D-Bus API. Thus if you've got any need for functionality, any concern about the existing implementation, or similar, please note it here so we can collect it.

One initial discussion base we already had in #238 were the current limitation of the `Info` D-Bus method was mentioned. Our current approach to an extensible and generic interface is to have a pure string-variant-dict-based approach. Similar to what we have in the `Status` method.
A drawback of this might be that this way is "too flexible" and not really an API anymore..

Another limitation we already had issues with in the past is the curent `Install` method that does not allow to provide further options and thus required us to move the 'delayed activation' feature (https://github.com/rauc/rauc/pull/134/commits/35a13bc201ca4709ded80e856eb5a51e31fbb6a5) to a system.conf-only feature, for example.

embeddedrocks:
Extending the ```Info``` method with a ```no-verify``` attribute would be great (as you already do via command line).
Background: We are using a USB pen drive to check for update files. The drive can contain multiple files. It can take a great amount of time validating each file. So currently the files are not validated when displayed, but of course will be later on. But in order to do so the command line is used instead of DBUS interface.

ejoerns:
@embeddedrocks thanks for the hint!

To be honest, the Info 'API' is rather limited and not actively propagated by us as we know it lacks a lot of features. The command line interface does not interact with it but inspects the bundle directly. I am not fully sure if this should be changed in the future but one may provide a better Info API anyway.

ejoerns:
I've prepared an initial guess on how a new interfaced could look like some time ago. I'll just dump it here for reference. I fear it lacks at least a parameter for 'Info' ;)

```
/de/pengutronix
/de/pengutronix/Rauc1
  > Interfaces
    > de.pengutronix.Rauc1
      > Methods
        - Install(String bundle, Dict of {String, Variant} options)
        - Info (String bundle) ->
        - GetAllSlots() -> Array of[Object Path]
        - GetBootedSlot() -> Object Path
      > Properties
        - String Compatible
        - String Variant
        - Struct of (int32, String, int32) Progress
        - ? LastError
    > de.pengutronix.Rauc1.Bootloader
      > Methods
        - MarkGood(String slot)
        - MarkBad(String slot)
        - MarkActive(String slot)
/de/pengutronix/Rauc1/Slot/0
  > Interfaces
[...]

PR #408, opened 2019-02-19
src/install: Add --force-install-same option

ian-abbott:
Make 'rauc install --force-install-same' behave as though the
'force-install-same' slot configuration option is set to 'true'.  This
can be useful if you only want to reload a slot with the same contents
occasionally.

Signed-off-by: Ian Abbott <abbotti@mev.co.uk>

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -44,6 +44,8 @@ typedef struct {
 	gchar *handlerextra;
 	/* ignore compatible check */
 	gboolean ignore_compatible;
+	/* force install same option */
+	gboolean force_install_same;
 
 	/* for storing installation runtime informations */
 	RContextInstallationInfo *install_info;
--- a/src/install.c
+++ b/src/install.c
@@ -924,7 +924,8 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 		}
 
 		/* skip if slot is up-to-date */
-		if (!dest_slot->force_install_same && g_strcmp0(mfimage->checksum.digest, slot_state->checksum.digest) == 0) {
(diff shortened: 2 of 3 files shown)


ejoerns:
@ian-abbott thanks for your contribution. Unfortunately this will only work when you compile RAUC without D-Bus support (which intentionally is not the default).

The reason for that is that the install D-Bus interface does not accept arguments and thus the front-line command line interface will not be able to forward this option to the backend:
https://github.com/rauc/rauc/blob/master/src/rauc-installer.xml#L9

Also see discussion in #248 for adding a new interface or extending the current.

ian-abbott:
Thanks for the review. I'll wait until the --ignore-compatible install option is added to the D-Bus interface (if ever) and work out what to do then.

ejoerns:
Recent strategy discussion made us think of moving away from the former slot skipping mechanism. Thus depending on which kind of rework happens first (skipping mechanism or D-Bus API) this might become obsolete then.

We still leave this open for reference.

ejoerns:
> Thanks for the review. I'll wait until the --ignore-compatible install option is added to the D-Bus interface (if ever) and work out what to do then.

With https://github.com/rauc/rauc/pull/446 this is within reach now.

Please have a look at https://github.com/rauc/rauc/pull/448. IMHO with this merged there will be no need for the client argument anymore. Do you agree @ian-abbott ?

ejoerns:
I'll close this now as it was obsoleted by #448.

Forcing the installation of images with a checksum equal to the one already being on the target slot is not required anymore as this was made the default behavior now.

PR #415, opened 2019-02-25
src/install: notify about post-install operation

UVV-gh:
This change adds an operation callback to the context, thus providing
a possibility to notify DBus listeners when post-install handler is executed

Signed-off-by: Vyacheslav Yurkov <Vyacheslav.Yurkov@bruker.com>

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -10,6 +10,7 @@
 
 typedef void (*progress_callback) (gint percentage, const gchar *message,
 		gint nesting_depth);
+typedef void (*operation_callback) (const gchar *message);
 
 typedef struct {
 	/* The bundle currently mounted by RAUC */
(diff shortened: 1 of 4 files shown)


ejoerns:
At the moment I cannot see that `post-install` is very special compared to e.g. pre-install and possibly hooks.
Thus If, one would need to add something similar for all these cases to be consistent.
An then we have to be sure that this really adds a value to the handling and is backwards compatible to the prior handling and we should not replace the 'progress' information here we have already.

However, we should make the use case clear. I know that post-install is often used for handling reboots, etc. but it turned out that this is error-prone or needs some delay handling etc. Otherwise it will kill the D-Bus before you are notified.
I would prefer doing reboot etc. from the outer instance that calls rauc.
This would allow waiting reliably for the install to signal completion via D-Bus and then trigger the reboot.

Are there other requirements to know that post-install runs?

jluebbe:
@UVV-gh What is your use-case for this notifier? I'd like to understand it better, so we can decide if it needs a more general mechanism.

Issue #416, opened 2019-02-26
calling `rauc status --output-format=json` from install hook fails


mbessler:
I'm using a custom install hook, and so far (in rauc 0.4) I had been using the `rauc status` json output inside my hook without issues. But after switching to 1.0, this appears to no longer work. 
```
rauc status --output-format=json | jq -r '[.slots[][] | select(.class | contains("rootfs")) ] | unique_by(.device) | .[] |  .device' 
```
yields this in the log:
```
GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code30: already processing a different method
```
It looks like this may be due to #299. Am I not supposed to use `rauc status` during the install operation?

ejoerns:
@mbessler thanks for reporting this!

If I get it right, this is triggered by the check in https://github.com/rauc/rauc/blob/master/src/service.c#L323 which appears to be not valid to me as we do not access the context writing.

mbessler:
@ejoerns yes, commenting out the context busy check where you indicated does resolve my issue.

If that does not have negative side-effects, I can create a PR, let me know...

ejoerns:
@mbessler current implementation has a negative side-effect ;) Currently, I don't see any, thus feel free to create a patch and we check if tests stay green etc.

mbessler:
PR opened: #417

ejoerns:
@mbessler We just discussed this and came to the conclusion that it is not safe to call 'rauc status' while an installation is running.
The relevant change was the rework of status handling to be fully D-Bus-based by default. Prior to this, calling 'rauc status' resulted in collection status information from the client process itself. Now it is only some D-Bus calls into the RAUC service. There we cannot assure (because of having no locking) that you will always get a consistent state of status information.

However, calling RAUC from the hook itself should not be required as you are already in a running RAUC instance that has all information available. Thus my question would be why you need to call `rauc status` and obtain slot information? Note that almost all information relevant for the slot currently handled is already available: https://rauc.readthedocs.io/en/latest/using.html#slot-hooks

mbessler:
Hi @ejoerns 
In my case the variables provided to the hook aren't sufficient as I need access to the slot names and devices of the other rootfs slots. 

I'm using a scheme somewhat similar to casync, just with a much smaller tool I wrote. Basically casync but limited to images only, and updates in place chunk-by-chunk instead of downloading to temp. storage. The tool lives here: https://github.com/mbessler/imidj.

[...]

PR #417, opened 2019-02-28
src/service: rauc status from install hook fails

mbessler:
calling `rauc status --output-format=json` from within
an install hook fails with:
`GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code30: already processing a different method`


ejoerns:
After discussion we decided that this is not the way to go as we cannot assure at the moment that we access the context read only here. Thus we must keep locking.

The actual motivation that led to this PR however is valid but should be resolved first with the methods described in #416 and we should make a fix / extension for these instead.

PR #425, opened 2019-03-14
Add an initial version of a man page

mhei:
Every good tool should have a manpage, at least as a first pointer
to further documentation.

This is mostly in preparation for Debian packaging since we
earn a lintian warning otherwise.


jluebbe:
Thanks for the contribution, @mhei.

PR #430, opened 2019-04-23
Allow slot names in rauc.slot and document booted slot detection

ejoerns:
This will allow detection of slots that do not have a bootname set.

A valid use case for this is a A+B+recovery scenario where A+B are
handled by the boot selection mechanism while 'recovery' is selected
automatically as a last resort when normal boot selection handling
fails or does not return any bootable slot.

We fix this with the first patch and document the entire mechanism with the second one.


ejoerns:
@mhei Thanks for the review! I've fixed the typos.

ejoerns:
Looks fine for me, thx!

PR #432, opened 2019-05-06
contrib: grub: export additional variables to environment

beralt:
With recent changes to the grub integration (#384) the variables exported by grub.cfg in contrib should also include A_OK, B_OK and ORDER.


ejoerns:
I've re-triggered Travis. Now it is green.. mh..

PR #433, opened 2019-05-08
tail-call real rauc process

anguslees:
Apart from avoiding the small memory overhead of a pointless persistent process, this also ensures signals, etc are sent to the rauc process and not the shell wrapper.


jluebbe:
Thanks for your contribution, @anguslees!

PR #434, opened 2019-05-10
src/main.c: A progress bar for RAUC

ejoerns:
A little gadget I always wanted to have... And did in the train locally
because of poor german LTE performance...

It gives you a two-lined progress bar while installing with RAUC. The
first line contains the message in the same style as before, the second
line contains a hash-printing progress bar in the width of the current
terminal.

Output will be similar to this:

     50% Checking Bundle done.
    [xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx                                             ] 50%

This is only activated if we detect a tty, otherwise we use the
old-style printing format.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1,3 +1,4 @@
+#include <errno.h>
 #include <gio/gio.h>
 #include <glib.h>
 #include <glib/gstdio.h>
@@ -6,6 +7,8 @@
 #include <json-glib/json-gobject.h>
 #endif
 #include <stdio.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
 
 #include "bundle.h"
 #include "bootchooser.h"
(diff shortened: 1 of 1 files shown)


ejoerns:
@mhei Thanks for the hint, I used g_autoptr to fix it.

src/main.c
```diff
@@ -27,14 +27,42 @@ gboolean info_noverify, info_dumpcert = FALSE;
 gboolean status_detailed = FALSE;
 gchar *output_format = NULL;
 
+#include <sys/ioctl.h>
+#include <stdio.h>
+#include <unistd.h>
```
jluebbe: These should be above the other includes.

src/main.c
```diff
@@ -27,14 +27,42 @@ gboolean info_noverify, info_dumpcert = FALSE;
...
+
+static gchar* get_progress(gint percentage)
```
jluebbe: This should be called something like `make_progress_line`.

src/main.c
```diff
@@ -27,14 +27,42 @@ gboolean info_noverify, info_dumpcert = FALSE;
...
+{
+	struct winsize w;
+	GString *printbuf = NULL;
+	gint pbar_len = 0;
+	
+	g_return_val_if_fail(percentage <= 100, NULL);
+
+	/* obtain terminal window parameters */
+	ioctl(STDOUT_FILENO, TIOCGWINSZ, &w);
```
jluebbe: Doesn't this need error handling?

src/main.c
```diff
@@ -27,14 +27,42 @@ gboolean info_noverify, info_dumpcert = FALSE;
...
+	pbar_len = w.ws_col - 1 - 1 - 5;
```
jluebbe: Maybe limit this to 100? We don't have more information anyway.

src/main.c
```diff
@@ -69,11 +97,21 @@ static void on_installer_changed(GDBusProxy *proxy, GVariant *changed,
 
 	g_mutex_lock(&args->status_mutex);
 	if (g_variant_lookup(changed, "Operation", "&s", &message)) {
-		g_queue_push_tail(&args->status_messages, g_strdup(message));
+		g_queue_push_tail(&args->status_messages, g_strdup_printf("%s\n", message));
 	} else if (g_variant_lookup(changed, "Progress", "(i&si)", &percentage, &message, &depth)) {
-		g_queue_push_tail(&args->status_messages, g_strdup_printf("%3"G_GINT32_FORMAT "%% %s", percentage, message));
+		g_autofree gchar *progress = get_progress(percentage);
```
jluebbe: You should only generate the progress line when needed (move it into the if block).

src/main.c
ejoerns: Neither bitbake nor wget etc. limit the width of the progress bar. Thus I would prefer using the full width here, too.

jluebbe: Hmm, OK.

src/main.c
ejoerns: Fixed.

src/main.c
ejoerns: Fixed.

src/main.c
ejoerns: Added some.

src/main.c
ejoerns: Fixed.

ejoerns:
I've now squashed all fixups and rebased onto current master.

ejoerns:
Rebased onto current master. Ping :)

PR #435, opened 2019-05-13
Atomic MBR boot partition update

thaemmerle:
This introduces a possibility to update a bootloader on a FAT partition, for SoCs with a ROM code which boots from the first partition in the MBR partition table, by changing the partition table.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -431,6 +431,25 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 				goto free;
 			}
 			g_key_file_remove_key(key_file, groups[i], "resize", NULL);
+
+			if (g_strcmp0(slot->type, "boot-mbr-switch") == 0) {
+				slot->region_start = key_file_consume_binary_suffixed_string(key_file, groups[i],
+						"region-start", &ierror);
+				if (ierror) {
+					g_propagate_prefixed_error(error, ierror, "mandatory for boot-mbr-switch: ");
+					res = FALSE;
+					goto free;
+				}
+
+				slot->region_size = key_file_consume_binary_suffixed_string(key_file, groups[i],
+						"region-size", &ierror);
+				if (ierror) {
+					g_propagate_prefixed_error(error, ierror, "mandatory for boot-mbr-switch: ");
+					res = FALSE;
+					goto free;
+				}
+			}
+
 			g_hash_table_insert(slots, (gchar*)slot->name, slot);
 		}
 		g_strfreev(groupsplit);
--- a/src/mbr.c
+++ b/src/mbr.c
@@ -1,0 +1,444 @@
+#include <errno.h>
+#include <fcntl.h>
+#include <glib/gstdio.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+#include <sys/mount.h>
+#include <linux/hdreg.h>
+
+#include "mbr.h"
+#include "update_handler.h"
+
+/* partition entry in MBR partition table, the system boots from */
+#define BOOT_PARTITION_ENTRY		0
+#define MBR_NUMBER_OF_PARTITIONS	4
+#define MBR_MAGIC_NUMBER_L		0x55
+#define MBR_MAGIC_NUMBER_H		0xAA
+
+struct chs_entry {
+	guint8 head;
+	guint8 sector;
+	guint8 cylinder;
+};
+
+struct mbr {
+	guint8 bootstrap_code[440];
+	guint32 disk_signature;
+	guint16 unused;
+	struct partition_tbl_entry {
+		guint8 boot_indicator;
+		struct chs_entry chs_start;
+		guint8 type;
+		struct chs_entry chs_end;
+		guint32 partition_start;
+		guint32 partition_size;
+	} partition_table[MBR_NUMBER_OF_PARTITIONS];
+	guint8 magic_number[2];
+} __attribute__((packed));
+
+G_STATIC_ASSERT(sizeof(struct mbr) == 512);
+
+static guint get_sectorsize(gint fd)
+{
+	guint sector_size;
+
+	if (ioctl(fd, BLKSSZGET, &sector_size) != 0)
+		return 512;
+
+	return sector_size;
+}
+
+static gboolean get_number_of_sectors(gint fd, guint *sectors,
+		GError **error)
+{
+	g_return_val_if_fail(sectors, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (ioctl(fd, BLKGETSIZE, sectors) != 0) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"ioctl command 0x%04x failed: %s",
+				BLKGETSIZE, g_strerror(errno));
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean get_hd_geometry(gint fd, struct hd_geometry *geometry,
+		GError **error)
+{
+	g_return_val_if_fail(geometry, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (ioctl(fd, HDIO_GETGEO, geometry) != 0) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"ioctl command 0x%04x failed: %s",
+				HDIO_GETGEO, g_strerror(errno));
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean validate_region(gint fd, guint64 start, guint64 size,
+		guint sector_size, GError **error)
+{
+	gboolean res = FALSE;
+	guint number_of_sectors;
+	GError *ierror = NULL;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (start < sizeof(struct mbr) || size == 0) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"no valid configuration for region");
+		goto out;
+	}
+
+	if ((start % (2 * sector_size)) != 0 || (size % (2 * sector_size)) != 0) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Region configuration is not aligned to the double"
+				"sector-size %d", 2 * sector_size);
+		goto out;
+	}
+
+	res = get_number_of_sectors(fd, &number_of_sectors, &ierror);
+	if (!res) {
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	if ((start + size) >= (guint64)number_of_sectors * sector_size) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Region configuration is bigger than device");
+		res = FALSE;
+		goto out;
+	}
+
+out:
+	return res;
+}
+
+static gboolean read_mbr(gint fd, struct mbr *mbr, GError **error)
+{
+	g_return_val_if_fail(mbr, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (read(fd, mbr, sizeof(*mbr)) != sizeof(*mbr)) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Read: %s", g_strerror(errno));
+		return FALSE;
+	}
+
+	if (mbr->magic_number[0] != MBR_MAGIC_NUMBER_L ||
+	    mbr->magic_number[1] != MBR_MAGIC_NUMBER_H) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"No valid master boot record found");
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean is_region_free(guint64 region_start, guint64 region_size,
+		const struct partition_tbl_entry *partition_tbl, guint sector_size,
+		GError **error)
+{
+	guint64 p_start, p_end;
+	guint i;
+
+	g_return_val_if_fail(partition_tbl, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	for (i = 0; i < MBR_NUMBER_OF_PARTITIONS; i++) {
+		if (i == BOOT_PARTITION_ENTRY)
+			continue;
+
+		p_start = (guint64)partition_tbl[i].partition_start * sector_size;
+		p_end = (guint64)partition_tbl[i].partition_size * sector_size +
+		        p_start - 1;
+
+		if (region_start >= p_start && region_start <= p_end)
+		{
+			g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+					"Region start address 0x%lx is in area of "
+					"partition %d (0x%lx - 0x%lx)",
+					region_start, i, p_start, p_end);
+			break;
+		}
+
+		if (p_start >= region_start &&
+		    p_start <= region_start + region_size - 1)
+		{
+			g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+					"Region end address 0x%lx is in area of "
+					"partition %d (0x%lx - 0x%lx)",
+					region_start + region_size - 1, i, p_start,
+					p_end);
+			break;
+		}
+	}
+
+	if (i < MBR_NUMBER_OF_PARTITIONS)
+		return FALSE;
+
+	return TRUE;
+}
+
+/**
+ * Calculation of the CHS value from an LBA value
+ *
+ * The 3 CHS bytes in Partition are stored with following layout:
+ *   - 8 bits for HEAD
+ *   - upper 2 bits for CYLINDER
+ *   - 6 bits for SECTOR
+ *   - lower 8 bits for CYLINDER
+ */
+static void get_chs(struct chs_entry *chs, guint32 lba,
+		const struct hd_geometry *geometry)
+{
+	g_return_if_fail(chs);
+	g_return_if_fail(geometry);
+
+	chs->sector = lba % geometry->sectors + 1;
+
+	lba /= geometry->sectors;
+	chs->head = lba % geometry->heads;
+
+	lba /= geometry->heads;
+	chs->cylinder = lba & 0xFF;
+
+	/* Move bit 8 & 9 of cylinder to bit 6 & 7 of sector */
+	chs->sector |= (lba >> 2) & 0xC0;
+}
+
+static gboolean get_raw_partition_entry(gint fd,
+		struct partition_tbl_entry *raw_entry,
+		const struct mbr_switch_partition *partition, GError **error)
+{
+	gboolean res = FALSE;
+	guint sector_size;
+	struct hd_geometry geometry;
+	GError *ierror = NULL;
+
+	g_return_val_if_fail(raw_entry, FALSE);
+	g_return_val_if_fail(partition, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	sector_size = get_sectorsize(fd);
+
+	if (partition->start % sector_size || partition->size % sector_size) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Partition start address or size is not a multipe"
+				" of sector size %d", sector_size);
+		goto out;
+	}
+
+	res = get_hd_geometry(fd, &geometry, &ierror);
+	if (!res) {
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	raw_entry->partition_start = partition->start / sector_size;
+	raw_entry->partition_size = partition->size / sector_size;
+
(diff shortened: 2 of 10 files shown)


jluebbe:
This pull request **introduces 2 alerts** when merging 862bed953aebaf6cb9d40de09f67e4afeb5be678 into 24371a6811a2f51e1cb25b344f496de235fb3f98 - [view on LGTM.com](https://lgtm.com/projects/g/rauc/rauc/rev/pr-42948ff47f9716faedf7da6d6bf6b38095645625)

**new alerts:**

* 2 for Multiplication result converted to larger type

---

*Comment posted by [LGTM.com](https://lgtm.com)*

src/mbr.c
```diff
@@ -0,0 +1,329 @@
...
+
+	if(start2 >= start1 && start2 <= end1)
+		return TRUE;
+
+	return FALSE;
+}
+
+gboolean r_mbr_get_non_active_boot_partition(const gchar *device,
+		struct mbr_partition *partition,
+		GError **error)
```
tretter: s/non_active/inactive

I am not sure if it wouldn't be better to just return a slot id here instead of the actual addresses. This would separate the addresses from the more abstract concept of slots and simplify the update of the boot slot. However, as the caller needs the addresses anyway for writing the partition, we would need another function to get the addresses of a slot. Therefore, returning the partition address is probably fine here and makes things simpler.

src/update_handler.c
```diff
@@ -1141,6 +1142,100 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+				image->checksum.size, dest_partition.size);
+		res = FALSE;
+		goto out;
+	}
+
+	g_message("Opening slot device partition %s", dest_slot->device);
+
+	destslotfile = g_file_new_for_path(dest_slot->device);
+
+	out_fd = open(g_file_get_path(destslotfile), O_WRONLY);
```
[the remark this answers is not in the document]
thaemmerle: You're right - it is almost a copy. But the reason why we have to duplicate the code is that `open_slot_device()` opens the device with flag `O_EXCL`. Since we have to operate on the device and not only on a partition, this would fail.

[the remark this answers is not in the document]
thaemmerle: gstdio is not included in `update_handler.c`.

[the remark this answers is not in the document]
thaemmerle: out_fd is closed in `copy_raw_image`

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+	struct partition_tbl_entry {
+		guint8 boot_indicator;
+		guint8 chs_begin[3];
+		guint8 type;
+		guint8 chs_end[3];
+		guint32 partition_start;
+		guint32 partition_size;
+	} partition_table[MBR_NUMBER_OF_PARTITIONS];
+	guint8 magic_number[2];
+} __attribute__((packed));
```
jluebbe: Use G_STATIC_ASSERT to check the size at compile time.

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"ioctl command 0x%04x failed: %s",
+				HDIO_GETGEO, g_strerror(errno));
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean r_mbr_read_mbr(gint fd, struct mbr *mbr, GError **error)
```
jluebbe: The r_ prefix isn't useful for static functions.

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+	if (ioctl(fd, BLKSSZGET, sector_size) != 0)
+		return FALSE;
+
+	return TRUE;
+}
+
+static gboolean get_hd_geometry(gint fd, struct hd_geometry *geometry,
+		GError **error)
+{
+	if (ioctl(fd, HDIO_GETGEO, geometry) != 0) {
```
jluebbe: Do the usual ``g_return_val_if_fail(error == NULL || *error == NULL, FALSE);`` here and below as well. Also, pointers which are filled should be checked with g_assert_nonnull(pointer);

src/mbr.c
jluebbe: Yes, returning addresses here is fine. It's still prefer inactive over non_active.

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+	boot_part = &mbr.partition_table[BOOT_PARTITION_ENTRY];
+
+	if (boot_part->partition_start == 0) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"no boot partition found in entry %d",
+				BOOT_PARTITION_ENTRY);
+		goto out;
+	}
+
+
```
jluebbe: Redundant newline.

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+	if (boot_part->partition_start == boot_part->partition_size) {
+		start = 2 * boot_part->partition_size;
```
jluebbe: This mechanism is too implicit, see my comment below.

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+
+	size = boot_part->partition_size;
+
+	/* check if new boot partition overlaps with any other partition */
+	if(mbr_is_part_overlapping(start, size, mbr.partition_table, &ierror)) {
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	if(!get_sectorsize(fd, &sector_size))
```
jluebbe: With the suggested simplification of get_sectorsize, you can get rid of the if.

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+	if (fd >= 0)
+		g_close(fd, NULL);
+
+	return res;
+}
+
+static void r_mbr_get_chs(guint8 *chs, guint32 sector,
+		const struct hd_geometry *geometry)
+{
+	chs[2] = sector % geometry->sectors + 1;
```
jluebbe: Is the "offical" way to do this documented somewhere? I'd like a comment here, describing how you got this calculation.

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+	sector /= geometry->heads;
+	chs[0] = sector & 0xFF;
+	chs[2] |= (sector >> 2) & 0xC0;
+}
+
+static gboolean r_mbr_get_raw_partition_entry(gint fd,
+		struct partition_tbl_entry *raw_entry,
+		const struct mbr_partition *partition, GError **error)
+{
+	gboolean res = FALSE;
```
jluebbe: Check pointers and error as above.

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+		const struct mbr_partition *partition,
+		GError **error)
+{
+	gboolean res = FALSE;
+	struct mbr mbr;
+	struct partition_tbl_entry *boot_part;
+	GError *ierror = NULL;
+	gint fd;
+
+	fd = g_open(device, O_RDWR);
```
jluebbe: Hmm, I think I'd prefer to have the open FD passed to this function (otherwise we open/close the device multiple times just for one update).
I'm not sure if we want to keep the MBR around to avoid the second read, though. @ejoerns?

src/mbr.c
```diff
@@ -0,0 +1,270 @@
...
+		goto out;
+	}
+
+	if (write(fd, &mbr, sizeof(mbr)) != sizeof(mbr)) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Could not write new master boot record: %s",
+				g_strerror(errno));
+		goto out;
+	}
+
```
jluebbe: Do we need to tell the kernel to reload the partition table or is that done automatically?

jluebbe:
This pull request **introduces 1 alert** when merging 207eaa8b44e9935264a8426bd0bd0fb0279cdd4c into 49b9c6dcab383d220482bb5ef7a67d28a55857c2 - [view on LGTM.com](https://lgtm.com/projects/g/rauc/rauc/rev/pr-562e0a023689a700830dc854f7a6fa1b2091b5e3)

**new alerts:**

* 1 for Multiplication result converted to larger type

---

*Comment posted by [LGTM.com](https://lgtm.com)*

jluebbe:
This pull request **introduces 1 alert** when merging 22242d8ab8a5b74860812db2732ac417a8484cf8 into 49b9c6dcab383d220482bb5ef7a67d28a55857c2 - [view on LGTM.com](https://lgtm.com/projects/g/rauc/rauc/rev/pr-e4c0cfb26ed62a9e7f26ef343fcabfed4666586f)

**new alerts:**

* 1 for Multiplication result converted to larger type

---

*Comment posted by [LGTM.com](https://lgtm.com)*

src/mbr.c
thaemmerle: This is done automatically.

src/mbr.c
thaemmerle: Open/close the device is done in boot-emmc in: `r_emmc_read_bootpart`, `r_emmc_force_part_rw`, `clear_slot`, `open_slot_device` and `r_emmc_force_part_ro`.
So I don't see a reason why we should avoid open/close of the device here.

thaemmerle:
Tried to apply your comments - pls re-review

src/mbr.c
```diff
@@ -0,0 +1,458 @@
...
+	while (clear_count < dest_partition->size) {
+		if ((dest_partition->size - clear_count) < sizeof(zerobuf))
+			clear_size = (dest_partition->size - clear_count);
+
+		tmp_count = write(fd, zerobuf, clear_size);
+
+		if (tmp_count != clear_size)
+			break;
```
jluebbe: This could return EINTR, or less than the requested bytes. In that case, you still need to count the written bytes.

src/update_handler.c
```diff
@@ -1141,6 +1142,100 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+
+	res = r_mbr_get_non_active_boot_partition(dest_slot->device, &dest_partition, &ierror);
+	if (!res) {
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	g_message("Found inactive partition at %lu %lu", dest_partition.start, dest_partition.size);
+
+	if (dest_partition.size != image->checksum.size) {
```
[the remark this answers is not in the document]
jluebbe: OK, this is now done.

src/mbr.c
```diff
@@ -0,0 +1,458 @@
...
+	boot_part = &mbr.partition_table[BOOT_PARTITION_ENTRY];
+
+	if (boot_part->partition_start == 0) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"no boot partition found in entry %d",
+				BOOT_PARTITION_ENTRY);
+		goto out;
+	}
+
+	if ((guint32)(region.start / sector_size) ==
```
jluebbe: Shouldn't this be calculated in 64 bit?

src/update_handler.c
```diff
@@ -1141,6 +1142,123 @@ static gboolean img_to_fs_handler(RaucImage *image, RaucSlot *dest_slot, const g
...
+			&ierror);
+	if (!res) {
+		g_propagate_error(error, ierror);
+		goto out;
+	}
+
+	g_message("Found inactive boot partition at %luB, size %luB",
+			dest_partition.start, dest_partition.size);
+
+	if (dest_partition.size != image->checksum.size) {
```
jluebbe: We could easily handle image-size < partition size as well. Do you have any reason do be more strict here?

thaemmerle:
Please re-review

jluebbe:
I've made some minor style changes and added a minimal test case.
```diff
diff --git a/src/config_file.c b/src/config_file.c
index cdf68281f04d..4b3bc8fc1669 100644
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -436,7 +436,7 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 				slot->region_start = key_file_consume_binary_suffixed_string(key_file, groups[i],
 						"region-start", &ierror);
 				if (ierror) {
-					g_propagate_prefixed_error(error, ierror, "mendatory for boot-mbr-switch: ");
+					g_propagate_prefixed_error(error, ierror, "mandatory for boot-mbr-switch: ");
 					res = FALSE;
 					goto free;
 				}
@@ -444,7 +444,7 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 				slot->region_size = key_file_consume_binary_suffixed_string(key_file, groups[i],
 						"region-size", &ierror);
 				if (ierror) {
-					g_propagate_prefixed_error(error, ierror, "mendatory for boot-mbr-switch: ");
+					g_propagate_prefixed_error(error, ierror, "mandatory for boot-mbr-switch: ");
 					res = FALSE;
 					goto free;
 				}
diff --git a/src/mbr.c b/src/mbr.c
index c7fa5ba736b5..70162a5fafbb 100644
--- a/src/mbr.c
+++ b/src/mbr.c
@@ -91,14 +91,14 @@ static gboolean validate_region(gint fd, guint64 start, guint64 size,
 
 	if (start < sizeof(struct mbr) || size == 0) {
 		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
-				"no valid configuration for region");
+				"no valid region configuration");
 		goto out;
 	}
 
 	if ((start % (2 * sector_size)) != 0 || (size % (2 * sector_size)) != 0) {
 		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
-				"region configuration is aligned to the double"
-				"sector-size %d", 2 * sector_size);
+				"region configuration is not aligned to two times the sector-size (%d)",
+				2 * sector_size);
 		goto out;
 	}
 
@@ -219,8 +219,8 @@ static gboolean get_raw_partition_entry(gint fd,
 
 	if (partition->start % sector_size || partition->size % sector_size) {
 		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
-				"Partition start address or size is not a multipe"
-				" of sector size %d", sector_size);
+				"Partition start address or size is not a multiple of sector size %d",
+				sector_size);
 		goto out;
 	}
 
@@ -268,8 +268,7 @@ gboolean r_mbr_switch_get_inactive_partition(const gchar *device,
 
 	res = read_mbr(fd, &mbr, &ierror);
 	if (!res) {
-		g_propagate_prefixed_error(error, ierror,
-				"failed to read mbr:");
+		g_propagate_prefixed_error(error, ierror, "failed to read MBR:");
 		goto out;
 	}
 
@@ -294,12 +293,10 @@ gboolean r_mbr_switch_get_inactive_partition(const gchar *device,
 	if ((region_start / sector_size) ==
 	    (guint64)boot_part->partition_start) {
 		partition->start = region_start + region_size / 2;
-	}
-	else if (((region_start + region_size / 2) / sector_size) ==
-	         (guint64)boot_part->partition_start) {
+	} else if (((region_start + region_size / 2) / sector_size) ==
+	           (guint64)boot_part->partition_start) {
 		partition->start = region_start;
-	}
-	else {
+	} else {
 		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
 				"boot partition's start address does match region "
 				"configuration");
@@ -339,7 +336,7 @@ gboolean r_mbr_switch_clear_partition(const gchar *device,
 	if (lseek(fd, dest_partition->start, SEEK_SET) !=
 	    (off_t)dest_partition->start) {
 		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
-				"Failed to set file to position %lu: %s",
+				"Failed to set file to offset %lu: %s",
 				dest_partition->start, g_strerror(errno));
 		goto out;
 	}
@@ -408,7 +405,7 @@ gboolean r_mbr_switch_set_boot_partition(const gchar *device,
 
 	if (lseek(fd, 0, SEEK_SET) != 0) {
 		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
-				"failed to seek to position 0");
+				"failed to seek to offset 0");
 		goto out;
 	}
```

src/mbr.c
```diff
@@ -0,0 +1,425 @@
...
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	for (i = 0; i < MBR_NUMBER_OF_PARTITIONS; i++) {
+		if (i == BOOT_PARTITION_ENTRY)
+			continue;
+
+		entry_end = partition_tbl[i].partition_start +
+		            partition_tbl[i].partition_size - 1;
+
+		if (region_start >= partition_tbl[i].partition_start &&
```
tretter: This check is wrong, because `region_start` is in bytes, but `partition_tbl[i].partition_start` is in sectors. If `region_start` is large enough, the check will think that the region is after the last partition (but actually the region is before the first partition). If you reduce `region_start`, the check will trigger, because the region conflicts with the last partition.

I suggest to convert `partition_start` and `entry_end` to bytes and add more information about the conflict to the error message.

src/mbr.c
```diff
@@ -0,0 +1,425 @@
...
+		    region_start <= entry_end)
+			break;
+
+		if (partition_tbl[i].partition_start >= region_start &&
+		    partition_tbl[i].partition_start <=
+		    region_start + region_size - 1)
```
tretter: Same as above. If you increase `region_size`, the update shall fail, because the region overlaps with the next partition, but happily results in an invalid partition table.

src/config_file.c
```diff
@@ -431,6 +431,25 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 				goto free;
 			}
 			g_key_file_remove_key(key_file, groups[i], "resize", NULL);
+
+			if(g_strcmp0(slot->type, "boot-mbr-switch") == 0) {
```
ejoerns: Not yet checked by uncrustify but de-facto style in the code: please add a space after `if`

src/utils.c
```diff
@@ -186,3 +186,45 @@ gchar * key_file_consume_string(
...
+		const gchar *key,
+		GError **error)
+{
+	gchar *string = NULL;
+	guint64 result;
+	gchar *scale;
+	guint scale_shift = 0;
+	GError *ierror = NULL;
+
+	string = key_file_consume_string(key_file, group_name, key, &ierror);
```
ejoerns: string needs to be freed (could be autoptr)

src/mbr.c
```diff
@@ -0,0 +1,438 @@
...
+	    (guint64)boot_part->partition_start) {
+		partition->start = region_start + region_size / 2;
+	}
+	else if (((region_start + region_size / 2) / sector_size) ==
+	         (guint64)boot_part->partition_start) {
+		partition->start = region_start;
+	}
+	else {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"boot partition's start address does match region "
```
ejoerns: I assume this should be 'doesn't' ? Message should also start with a capital letter to be consistent with others.

src/mbr.c
```diff
@@ -0,0 +1,438 @@
...
+		GError **error)
+{
+	gboolean res = FALSE;
+	static gchar zerobuf[512] = {};
+	gint clear_size = sizeof(zerobuf);
+	guint clear_count = 0;
+	gint tmp_count = 0;
+	gint fd;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
ejoerns: Also validate `*device` and `*dest_partition`

src/mbr.c
```diff
@@ -0,0 +1,438 @@
...
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"opening device failed: %s",
+				g_strerror(errno));
+		goto out;
+	}
+
+	res = read_mbr(fd, &mbr, &ierror);
+	if (!res) {
+		g_propagate_prefixed_error(error, ierror,
+				"failed to read MBR:");
```
ejoerns: s/failed to read MBR:/Failed to read MBR: /

src/mbr.c
```diff
@@ -0,0 +1,438 @@
...
+
+	if (start < sizeof(struct mbr) || size == 0) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"no valid configuration for region");
+		goto out;
+	}
+
+	if ((start % (2 * sector_size)) != 0 || (size % (2 * sector_size)) != 0) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"region configuration is aligned to the double"
```
ejoerns: s/is/is not/

src/mbr.c
```diff
@@ -0,0 +1,438 @@
...
+		goto out;
+	}
+
+out:
+	return res;
+}
+
+static gboolean read_mbr(gint fd, struct mbr *mbr, GError **error)
+{
+	gboolean res = FALSE;
```
ejoerns: No need to have `res` variable here.

Simplify structure by instantly returning with `return FALSE` in if clauses and do `return TRUE` at the end of the function.

src/mbr.c
```diff
@@ -0,0 +1,438 @@
...
+	guint i;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	for (i = 0; i < MBR_NUMBER_OF_PARTITIONS; i++) {
+		if (i == BOOT_PARTITION_ENTRY)
+			continue;
+
+		p_start = (guint64)partition_tbl[i].partition_start * sector_size;
+		p_end = (guint64) partition_tbl[i].partition_size * sector_size +
```
ejoerns: spacing after `(guint64)` is inconsistent here.

jluebbe:
Thanks @thaemmerle!

PR #439, opened 2019-05-29
configure: refine D-Bus install directories

mhei:
For the D-Bus system service directory, we can rely on pkg-config
to query the target directory, instead of constructing our own
assumption.

For D-Bus policy directory, there is no such pkg-config variable.
However, D-Bus's $datadir might be different from ours, so use
D-Bus's ones.

This allows to compile/install rauc to /usr/local but place the
D-Bus config files in the directories which are considered by
distribution's D-Bus.

While at, adjust the syntax for systemd service directory part
a little bit to match the D-Bus stuff.

We also rename some internal autoconf variables to match the
configure parameter names.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -50,6 +50,7 @@ AS_IF([test "x$enable_service" != "xno"], [
 
 # Checks for libraries.
 PKG_CHECK_MODULES([GLIB], [glib-2.0 >= 2.45.8 gio-2.0 gio-unix-2.0])
+PKG_CHECK_MODULES([DBUS1], [dbus-1])
 
 AC_ARG_ENABLE([network],
        AS_HELP_STRING([--disable-network], [Disable network update mode])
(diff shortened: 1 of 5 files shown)


ejoerns:
This adds a dependency to `libdbus-1-dev`, this is worth being noted and added to README.rst

As the patch already states doing multiple things at once, this indicates that these should be split up in mutliple patches ;)

mhei:
I agree and will rework this PR as soon as I find a few free minutes...

ejoerns:
@mhei Just saw you updated the PR, thanks!

configure.ac
```diff
@@ -89,14 +89,14 @@ AM_CONDITIONAL(SYSTEMD, test -n "${with_systemdunitdir}")
 AC_ARG_WITH([dbuspolicydir],
         AS_HELP_STRING([--with-dbuspolicydir=DIR], [D-Bus policy directory]),
         [],
-        [with_dbuspolicydir=${datadir}/dbus-1/system.d])
+        [with_dbuspolicydir="$($PKG_CONFIG --variable=datadir dbus-1)/dbus-1/system.d"])
```
michaelolbrich: What happens here if dbus-1.pc is not available?

mhei: Not tried it yet, but I guess this would break the installation since $(...) expands to empty string. Do you suggest to add ${datadir} as fallback for example?

mhei: Maybe we should add a check before and fail when dbus-1 package is not available... This makes the problem more obvious to the user than using a silent fallback.

michaelolbrich: Yes, failing if dbus-1 is missing makes sense.

mhei: Added a check in first patch, rebase to latest master.

jluebbe: The build now fails at exactly this place, so the new dependency needs to be added to the CI scripts and documentation.

mhei: I guess I need help here... I added the devel package name to Dockerfile and Travis script, but I've no glue how this is pulled in so that this PR builds successfully. Sorry, I did not work with Docker before and I'm not sure that I understand the CI setup completely. I guess that the Docker image needs a rebuild before Travis pulls it here and can succesfully use it?

ejoerns: It's not that easy, because the container is build independently from master branch by docker hub and there is no other clean way I would know than a 2-step process of mergingfor adding such a dependency.

The only other solution here would be to test this manually. @jluebbe what do you think?

This is the compromise we had to accept when moving to Docker-based testing, but it's worth the drawback as testing became much more stable since this.

jluebbe: We've updated the Dockerfile, and rebuilt. So the checks succeed now.

PR #442, opened 2019-06-04
src/update_handler: collection of structure simplifications and minor fixes

ejoerns:


src/update_handler.c
```diff
@@ -33,6 +33,10 @@ static GUnixOutputStream* open_slot_device(RaucSlot *slot, int *fd, GError **err
 	GError *ierror = NULL;
 	int fd_out;
 
+	g_return_val_if_fail(slot, NULL);
+	g_return_val_if_fail(fd == NULL || *fd == NULL, NULL);
```
jluebbe: The fd pointer is optional (and you depend on that in another commit), so this check is wrong. I'll remove this in a fixup.

jluebbe:
@ejoerns I've also added missing `#includes` to fix the build.

PR #446, opened 2019-06-05
Add new InstallBundle Method to D-Bus API

ejoerns:
In the past, there were a lot of cases where we ran into the limitation of the `Install` method of the D-Bus `Installer` interface. These were all caused by the missing ability to pass additional arguments to the install call.

As we cannot break existing setups, we cannot simply change the method signature. Instead, we need to introduce an alternative method.

Here it is: ``InstallBundle sa{sv}``

Where the new generic dict of string-variant (``a{sv}``) allows passing and handling all kind of current and future parameters without changing the interface again.

This initial series includes support for one argument that was already implemented for the non-D-Bus case: ``--ignore-compatible``. It allows forcing the installation of a bundle on a per-installation-call base

Reference: #248

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -45,8 +45,6 @@ typedef struct {
 
 	/* optional custom handler extra arguments */
 	gchar *handlerextra;
-	/* ignore compatible check */
-	gboolean ignore_compatible;
 
 	/* for storing installation runtime informations */
 	RContextInstallationInfo *install_info;
--- a/src/install.c
+++ b/src/install.c
@@ -446,9 +446,9 @@ static void parse_handler_output(gchar* line)
 	}
 }
 
-static gboolean verify_compatible(RaucManifest *manifest)
+static gboolean verify_compatible(RaucInstallArgs *args, RaucManifest *manifest)
 {
-	if (r_context()->ignore_compatible) {
+	if (args->ignore_compatible) {
 		return TRUE;
 	} else if (g_strcmp0(r_context()->config->system_compatible,
 			manifest->update_compatible) == 0) {
@@ -691,7 +691,7 @@ static gboolean launch_and_wait_custom_handler(RaucInstallArgs *args, gchar* bun
 			res = FALSE;
 			goto out;
 		}
-	} else if (!verify_compatible(manifest)) {
+	} else if (!verify_compatible(args, manifest)) {
 		g_set_error_literal(error, R_INSTALL_ERROR, R_INSTALL_ERROR_COMPAT_MISMATCH,
 				"Compatible mismatch");
 		res = FALSE;
@@ -793,7 +793,7 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 			res = FALSE;
 			goto early_out;
 		}
-	} else if (!verify_compatible(manifest)) {
+	} else if (!verify_compatible(args, manifest)) {
 		res = FALSE;
 		g_set_error_literal(error, R_INSTALL_ERROR, R_INSTALL_ERROR_COMPAT_MISMATCH,
 				"Compatible mismatch");
--- a/src/main.c
+++ b/src/main.c
@@ -225,11 +225,13 @@ static gboolean install_start(int argc, char **argv)
 	args->cleanup = install_cleanup;
 	args->status_result = 2;
 
-	r_context_conf()->ignore_compatible = install_ignore_compatible;
+	args->ignore_compatible = install_ignore_compatible;
 
 	r_loop = g_main_loop_new(NULL, FALSE);
 	if (ENABLE_SERVICE) {
 		g_auto(GVariantDict) dict = G_VARIANT_DICT_INIT(NULL);
+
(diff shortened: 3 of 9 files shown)


jluebbe:
I've pushed a rebase onto master. The the conflicts were caused by the network mode removal.

ejoerns:
We also have tests for this and it remained in RFC phase for some days. Thus we can remove the RFC label now and make it a normal PR ready for being reviewed and merged.

include/context.h
```diff
@@ -124,4 +122,4 @@ void r_context_register_progress_callback(progress_callback progress_cb);
 
 RaucContext *r_context_conf(void);
 const RaucContext *r_context(void);
-void r_context_clean(void);
\ No newline at end of file
+void r_context_clean(void);
```
michaelolbrich: unrelated change.

src/service.c
```diff
@@ -70,6 +71,11 @@ static gboolean r_on_handle_install_bundle(
 	args->notify = service_install_notify;
 	args->cleanup = service_install_cleanup;
 
+	g_variant_dict_init(&dict, arg_args);
+	g_variant_dict_lookup(&dict, "ignore-compatible", "b", &args->ignore_compatible);
+	arg_args = g_variant_dict_end(&dict);
+	g_variant_unref(arg_args);
```
michaelolbrich: This is a bit strange. I'm not sure if `g_variant_dict_init()` consumes the ref for `arg_args` or not, but either way, `g_variant_dict_clear()` should be used here and maybe `g_variant_unref()` if necessary.

test/service.c
```diff
@@ -203,8 +203,17 @@ static void service_test_install(ServiceFixture *fixture, gconstpointer user_dat
...
 
 	/* Actually install bundle */
-	ret = r_installer_call_install_sync(installer, bundlepath, NULL,
-			&error);
+	if (deprecated) {
+		ret = r_installer_call_install_sync(installer, bundlepath, NULL,
+				&error);
+	} else {
+		GVariantDict dict;
+		GVariant *install_args = NULL;
```
michaelolbrich: no need to initialize `install_args`.

src/service.c
michaelolbrich: Also, you might want to change the commit order: Before this change, `arg_args` was leaked.

include/context.h
ejoerns: Appears to be a line at end-of-file change done by my editor, will remove it.

test/service.c
ejoerns: I'll just drop it

src/service.c
ejoerns: Right, `g_variant_dict_clear()` should do it here then there's no need to unref (and abuse `arg_args`).

About the `arg_args` leakage: This is passed from outside and thus I would expect the caller to free it.

src/main.c
```diff
@@ -242,12 +248,15 @@ static gboolean install_start(int argc, char **argv)
...
-		if (!r_installer_call_install_sync(installer, args->name, NULL,
+		if (!r_installer_call_install_bundle_sync(installer, args->name, arg_args, NULL,
 				&error)) {
 			g_printerr("Failed %s\n", error->message);
 			g_error_free(error);
+			g_variant_unref(arg_args);
 			goto out_loop;
 		}
+
+		g_variant_unref(arg_args);
```
michaelolbrich: I'm pretty sure `arg_args` leaks in the error case. Maybe use a cleanup handler?

src/service.c
michaelolbrich: That depends on how the API is defined. And the caller is generated gdbus code, right?

src/main.c
ejoerns: Ouh, I've missed the first error handling block, only added it to the second one.

jluebbe: I've refactored this.

src/install.c
```diff
@@ -1144,7 +1144,6 @@ void install_args_free(RaucInstallArgs *args)
 {
 	g_free(args->name);
 	g_mutex_clear(&args->status_mutex);
-	g_assert_cmpint(args->status_result, >=, 0);
```
jluebbe, line 1147: Why did this fail only now? In src/main.c, we've been freeing install args for a long time.

ejoerns: Ok, I had a closer look at this now as I knew I came across this but did not recap where actually.

The potential problem I referred to is that `install_args_new()` initializes `status_result` to `-2`.
Thus if `install_args_free()` gets called before some instance changed this value, we will end up in the assertion error. But maybe, yes, these probably are the exact cases that we wanted to find with the assertion...

The only two points I dug out where this could have been an issue are:

https://github.com/rauc/rauc/blob/master/src/service.c#L63

and

https://github.com/rauc/rauc/blob/master/src/install.c#L1159

As they potentially return without having the args touched.

BUT, as we do

https://github.com/rauc/rauc/blob/master/src/service.c#L80

(using only `g_free()` on the complex initialized type).This will never trigger.

IMHO this should call `install_args_free()` here.
Also setting `args = NULL` above seems to be a bit unsafe but should not make any issues if we properly cleared args before.

jluebbe: This should be solved now.

PR #447, opened 2019-06-05
src/install: remove network mode

ejoerns:
This removes the deprecated and very rarely used network mode.

This does not disable RAUC's network capabilities (like used for
casync), this only removes the special non-bundle mode that was not
really maintained for several years now and marked as deprecated since
RAUC 1.0.


jluebbe:
If it turns out that it's still used, we can always revert this. But without any known users we should drop it.

Issue #452, opened 2019-06-16
Cross-compiling for ARM was recently broken 


marv:
Cross-compiling RAUC for ARMv7 fails to build with the following errors (actually warnings that got treated as errors due to -Werror):

```
src/mbr.c:26:10: error: packed attribute causes inefficient alignment for 'disk_signature' [-Werror=attributes]
  guint32 disk_signature;
          ^~~~~~~~~~~~~~
src/mbr.c:27:10: error: packed attribute causes inefficient alignment for 'unused' [-Werror=attributes]
  guint16 unused;
          ^~~~~~
src/mbr.c: In function 'is_region_free':
src/mbr.c:164:32: error: format '%lx' expects argument of type 'long unsigned int', but argument 5 has type 'guint64' {aka 'long long unsigned int'} [-Werror=format=]
      "Region start address 0x%lx is in area of "
                              ~~^
                              %llx
src/mbr.c:166:6:
      region_start, i, p_start, p_end);
      ~~~~~~~~~~~~               
src/mbr.c:164:6: error: format '%lx' expects argument of type 'long unsigned int', but argument 7 has type 'guint64' {aka 'long long unsigned int'} [-Werror=format=]
      "Region start address 0x%lx is in area of "
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/mbr.c:166:23:
      region_start, i, p_start, p_end);
                       ~~~~~~~
src/mbr.c:165:25: note: format string is defined here
      "partition %d (0x%lx - 0x%lx)",
                       ~~^
                       %llx
src/mbr.c:164:6: error: format '%lx' expects argument of type 'long unsigned int', but argument 8 has type 'guint64' {aka 'long long unsigned int'} [-Werror=format=]
      "Region start address 0x%lx is in area of "
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/mbr.c:166:32:
      region_start, i, p_start, p_end);
                                ~~~~~
src/mbr.c:165:33: note: format string is defined here
      "partition %d (0x%lx - 0x%lx)",
                               ~~^
                               %llx
src/mbr.c:174:30: error: format '%lx' expects argument of type 'long unsigned int', but argument 5 has type 'guint64' {aka 'long long unsigned int'} [-Werror=format=]
      "Region end address 0x%lx is in area of "
                            ~~^
                            %llx
src/mbr.c:176:6:
[...]

PR #453, opened 2019-06-17
Fix compilation on ARM32

jluebbe:
This should fix the word size issues introduced by #435 and reported in #452.


ejoerns:
@marv you might also want to have a look at this

marv:
@ejoerns Thanks, it compiles fine again in my environment with your patches

ejoerns:
@marv good to hear, thus we can merge it. Thanks for testing! (btw. patches were done by @jluebbe ;) )

PR #454, opened 2019-06-19
suppress -Werror=address-of-packed-member warning/error

zavorka:
Recent addition of `packed` attribute to the `mbr` struct in #453 has broken
x86 builds by introducing a _perceived_ potential alignment violation to which
GCC objects by issuing these warnings:

```
src/mbr.c: In function ‘r_mbr_switch_get_inactive_partition’:
src/mbr.c:291:53: error: taking address of packed member of ‘struct mbr’ may res
ult in an unaligned pointer value [-Werror=address-of-packed-member]
  291 |  res = is_region_free(region_start, region_size, mbr.partition_table,
      |                                                  ~~~^~~~~~~~~~~~~~~~
src/mbr.c:299:14: error: taking address of packed member of ‘struct mbr’ may res
ult in an unaligned pointer value [-Werror=address-of-packed-member]
  299 |  boot_part = &mbr.partition_table[BOOT_PARTITION_ENTRY];
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/mbr.c: In function ‘r_mbr_switch_set_boot_partition’:
src/mbr.c:418:14: error: taking address of packed member of ‘struct mbr’ may res
ult in an unaligned pointer value [-Werror=address-of-packed-member]
  418 |  boot_part = &mbr.partition_table[BOOT_PARTITION_ENTRY];
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
 ```

In order to tone GCC (or Clang) down, provide assurance that the compiler is safe
to assume the MBR partition table entries are aligned on 16-bit boundaries.

### DIFF
--- a/src/mbr.c
+++ b/src/mbr.c
@@ -11,6 +11,7 @@
 
 /* partition entry in MBR partition table, the system boots from */
 #define BOOT_PARTITION_ENTRY		0
+#define TBL_ENTRY_ALIGNMENT		2
 #define MBR_NUMBER_OF_PARTITIONS	4
 #define MBR_MAGIC_NUMBER_L		0x55
 #define MBR_MAGIC_NUMBER_H		0xAA
@@ -262,7 +263,7 @@ gboolean r_mbr_switch_get_inactive_partition(const gchar *device,
 	gboolean res = FALSE;
 	struct mbr mbr;
 	GError *ierror = NULL;
-	struct partition_tbl_entry *boot_part;
+	struct partition_tbl_entry *boot_part, *partition_table;
 	guint sector_size;
 	gint fd;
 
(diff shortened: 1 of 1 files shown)


jluebbe:
The __builtin_assume_aligned seems like a big hammer, and I'm not 100% sure its correct for all architectures. I've played around a bit with the compiler explorer:
https://godbolt.org/z/HdVHTP

One problem seems to be -Wpacked, which detects if the packed attribute has no effect. We'll always have some architectures, which have the correct alignment by default. So I think we we should disable that check.

On ARM 32, that code still produces the error I tried to fix with #453. We could disable -Wattribute as well, but I'd prefer to have code that doesn't need disabling warnings. I'm not sure though how do change it to fit all compilers.

jluebbe:
@zavorka, could you check if #455 fixes your issue as well?

jluebbe:
With #455 merged, this shouldn't be needed anymore and we now have CI checks to detect something link this sooner. Please open an issue if this still happens.

PR #455, opened 2019-06-19
fix compile warnings/errors in src/mbr.c and endian issues

jluebbe:
In #453, we broke compilation with newer gccs on x86, as reported in #454.

This PR tries to fix it in a simpler way by using `#pragma pack`. Also fix access to 32-bit integers on big-endian platforms by using the correct wrappers.


jluebbe:
Force-pushed to fix formatting errors.

jluebbe:
Rebased to let travis test the other platforms.

PR #456, opened 2019-06-20
minor fixes and annotation for compilation with clang

jluebbe:
Clang has stricter warnings for some cases, so we should fix these.

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -87,7 +87,8 @@ void r_context_begin_step(const gchar *name, const gchar *description,
  * @param description that is emitted via DBus on begin/end.
  *   A printf-like format string.
  */
-void r_context_begin_step_formatted(const gchar *name, gint substeps, const gchar *description, ...);
+void r_context_begin_step_formatted(const gchar *name, gint substeps, const gchar *description, ...)
+__attribute__((__format__(__printf__, 3, 4)));
 
 /**
  * Call at the end of a relevant code block. Percentage calculation is done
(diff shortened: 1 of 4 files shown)


src/install.c
```diff
@@ -410,7 +410,7 @@ GList* get_install_images(const RaucManifest *manifest, GHashTable *target_group
 			RaucSlot *target_slot = NULL;
 
 			/* Not interested in slots of other classes */
-			if (!g_strcmp0(lookup_image->slotclass, *cls) == 0)
```
ejoerns: concerning `g_strcmp0()` personally I find it clearer to have the form *!(strings_equal(stra1, str2))* but I also agree that this is a double inversion when only having `g_strcmp0()` and might confuse individuals that are not me :).

include/context.h
```diff
@@ -87,7 +87,8 @@ void r_context_begin_step(const gchar *name, const gchar *description,
  * @param description that is emitted via DBus on begin/end.
  *   A printf-like format string.
  */
-void r_context_begin_step_formatted(const gchar *name, gint substeps, const gchar *description, ...);
+void r_context_begin_step_formatted(const gchar *name, gint substeps, const gchar *description, ...)
+	__attribute__((__format__(__printf__, 3, 4)));
```
ejoerns: Seems like uncrustify doesn't like it to be placed this way. Thus we either need to simply append it to the end of the declaration line or maybe find the right bit in uncrustify to accept this...

jluebbe: OK, I've simply accepted uncrustify's suggestion.

Issue #460, opened 2019-07-05
No way to resign when original certificate already expired


mhei:
I have a bundle which was signed with a certificate which is meanwhile expired.
I tried to resign the bundle with a new certificate using `rauc resign ...`,
however this fails with
```$ resign-update foobar.raucb
rauc-Message: 12:59:45.721: Reading bundle: foobar.raucb
rauc-Message: 12:59:45.725: Verifying bundle... 
signature verification failed: Verify error:certificate has expired
```

In my opinion, this should always be possible, at least we should provide the ability with a command line parameter. For example it would make sense to me to re-use `--no-verify` here, too.

jluebbe:
Yes, we should allow resigning of expired bundles. I'd prefer to have an option like `--allow-exired` to accept just this specific problem, instead of allowing just anything. In addition, `--no-verify` should also work for `rauc resign`.

mhei:
I had a quick look and prepared two ideas here: https://github.com/mhei/rauc/tree/issue-460-extend-noverify-resign Please note, that the second commit is totally untested. Also a proper PR should include documentation updates.

Bastian-Krause:
@mhei: Your approach looks valid to me. One could probably argue that using the time of signing for `--allow-expired` might not work in all cases (wrong time set on the build host when the bundle is created). To cover this we could set the time for verification to the start or end of the validity period, but I see that your implementation is simpler.

Would you create a PR with your changes?

mhei:
My "implementation" is simply (mis-)using existing infrastructure :smile:
I also checked openssl API but I did not find another approach to ignore this specific error but doing full verification on all other aspects. Maybe this could be a feature request for openssl.
I'm a little bit concerned that `--allow-expired` does not exactly reflect what technically is going on when using this approach but it names the use-case so this would be fine with me.
I'll create a PR from my branch as is because next three weeks I will have no time to work on this further, e.g. documentation, sorry.

ejoerns:
@mhei Yes.. it does not fully keep the promises it makes.. ;)

And.. OpenSSL is a great API!... :laughing: 

I'm not quite sure if removing a parameter works some how. Setting it to a time that will be valid, as @Bastian-Krause suggested, might work by extracting signature validity from signer cert and using these for validity time, but that sounds hacky. OpenSSL upstream will for sure have a good reason to not allow these things, but maybe one could give it a try.

If I should find some time I'll also have another look at this.

jluebbe:
[...]

PR #461, opened 2019-07-10
Remove keyring path before signing with new key; Remove message on no keyring

Bastian-Krause:
`sign_bundle()` verifies the signature after signing against the given keyring. When resigning this keyring is most likely not valid anymore. So remove it and skip signature verification after signing.

Invalidating the keyring leads to the error message being printed although the user gave the required `--keyring` option.

As the message has not much benefit anyway simply remove it altogether.

Fixes: #154, ca64249 ("src/bundle: always call check_bundle explicitly")

Note: this has nothing to do with #460.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -468,6 +468,10 @@ gboolean resign_bundle(RaucBundle *bundle, const gchar *outpath, GError **error)
 		goto out;
 	}
 
+	/* invalidate keyring to avoid post-signing verification with inappropriate
+	 * keyring */
+	g_clear_pointer(&r_context()->config->keyring_path, g_free);
+
 	res = sign_bundle(outpath, &ierror);
 	if (!res) {
 		g_propagate_error(error, ierror);
--- a/test/bundle.c
+++ b/test/bundle.c
@@ -7,6 +7,7 @@
 #include <bundle.h>
 #include <context.h>
 #include <manifest.h>
+#include <signature.h>
 #include <utils.h>
 
 #include "common.h"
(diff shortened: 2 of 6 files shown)


src/bundle.c
```diff
@@ -474,6 +474,9 @@ gboolean resign_bundle(RaucBundle *bundle, const gchar *outpath, GError **error)
 		goto out;
 	}
 
+	/* invalidate old keyring */
```
ejoerns: This does not really reflect why this is needed. Maybe


```suggestion
	/* invalidate to not trigger post-signing verification with inappropriate keyring */
```

Bastian-Krause: "invalidate keyring to avoid post-signing verification with inappropriate keyring"?

ejoerns: would accept it ;)

Bastian-Krause: Done.

ejoerns:
@Bastian-Krause I've sponsored some test adaptions ;)

Bastian-Krause:
> @Bastian-Krause I've sponsored some test adaptions ;)

Thanks!

ejoerns:
I've squashed it and rebased it onto latest master.

test/bundle.c
```diff
@@ -172,7 +172,7 @@ static void bundle_test_resign(BundleFixture *fixture,
 	 * Note we have to use r_context() here as a hack to avoid re-setting
 	 * the context's 'pending' flag which would cause a re-initialization
 	 * of context and thus overwrite content of 'config' member. */
-	r_context()->config->keyring_path = g_strdup("test/openssl-ca/rel-ca.pem");
+	r_context_conf()->keyringpath = g_strdup("test/openssl-ca/rel-ca.pem");
```
jluebbe: Why did you change r_context to r_context_conf? Doesn't the comment apply any more?

jluebbe:
Closing in favor of #498.

PR #462, opened 2019-07-10
introduce enum to tweak verification during bundle resigning

mhei:
This PR allows to give `--no-verify` and `--allow-expired` when resigning
to skip over expired certificates in this process.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -465,6 +465,7 @@ out:
 G_GNUC_UNUSED
 static gboolean resign_start(int argc, char **argv)
 {
+	CheckBundleParams check_bundle_params = CHECK_BUNDLE_DEFAULT;
 	g_autoptr(RaucBundle) bundle = NULL;
 	GError *ierror = NULL;
 	g_debug("resign start");
@@ -495,7 +496,10 @@ static gboolean resign_start(int argc, char **argv)
 		goto out;
 	}
 
-	if (!check_bundle(argv[2], &bundle, !verification_disabled, &ierror)) {
+	if (verification_disabled)
+		check_bundle_params |= CHECK_BUNDLE_NO_VERIFY;
+
+	if (!check_bundle(argv[2], &bundle, check_bundle_params, &ierror)) {
 		g_printerr("%s\n", ierror->message);
 		g_clear_error(&ierror);
 		r_exit_status = 1;
@@ -540,7 +544,7 @@ static gboolean extract_signature_start(int argc, char **argv)
 	g_debug("input bundle: %s", argv[2]);
 	g_debug("output file: %s", argv[3]);
 
-	if (!check_bundle(argv[2], &bundle, TRUE, &ierror)) {
+	if (!check_bundle(argv[2], &bundle, CHECK_BUNDLE_DEFAULT, &ierror)) {
 		g_printerr("%s\n", ierror->message);
 		g_clear_error(&ierror);
 		r_exit_status = 1;
@@ -585,7 +589,7 @@ static gboolean extract_start(int argc, char **argv)
 	g_debug("input bundle: %s", argv[2]);
 	g_debug("output dir: %s", argv[3]);
 
-	if (!check_bundle(argv[2], &bundle, TRUE, &ierror)) {
+	if (!check_bundle(argv[2], &bundle, CHECK_BUNDLE_DEFAULT, &ierror)) {
 		g_printerr("%s\n", ierror->message);
 		g_clear_error(&ierror);
 		r_exit_status = 1;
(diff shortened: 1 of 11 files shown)


Bastian-Krause:
When the `--no-verify` argument is given `rauc resign` still requires `--keyring` but does not use it, right?

src/main.c
```diff
@@ -1630,12 +1641,18 @@ GOptionEntry entries_install[] = {
...
-	{"no-verify", '\0', 0, G_OPTION_ARG_NONE, &info_noverify, "disable bundle verification", NULL},
+	{"no-verify", '\0', 0, G_OPTION_ARG_NONE, &noverify, "disable bundle verification", NULL},
 	{"output-format", '\0', 0, G_OPTION_ARG_STRING, &output_format, "output format", "FORMAT"},
 	{"dump-cert", '\0', 0, G_OPTION_ARG_NONE, &info_dumpcert, "dump certificate", NULL},
 	{0}
 };
 
+GOptionEntry entries_resign[] = {
+	{"no-verify", '\0', 0, G_OPTION_ARG_NONE, &noverify, "disable bundle verification", NULL},
+	{"allow-expired", '\0', 0, G_OPTION_ARG_NONE, &resign_allow_expired, "use bundle signing time during verification", NULL},
```
jluebbe: We need to decide what this option does, allow expired signatures or use the bundle singing time. This help message would be confusing

src/main.c
```diff
@@ -877,7 +888,7 @@ static gboolean info_start(int argc, char **argv)
 		goto out;
 	g_debug("input bundle: %s", bundlelocation);
 
-	res = check_bundle(bundlelocation, &bundle, !info_noverify, &error);
+	res = check_bundle(bundlelocation, &bundle, !noverify, &error);
```
jluebbe: Instead of passing some verification options via r_context and noverify via an argument, I'd prefer replacing the verify argument by a bit-field enum (like `TestUpdateHandlerParams` in `test/update_handler.c`).
```
V_NO_VERIFY = BIT(0),
V_NO_CHECK_TIME = BIT(1),
```
The BIT macro should go in `utils.h`

jluebbe:
I've implemented only `--no-verify` in #569. As one would need to verify that the bundle is valid out-of-band, I think that covers most use-cases.
@mhei Do you have a specific reason why you would want `--allow-expired` or `--no-check-time` as well?

mhei:
You requested this in your first comment to #460 :grinning: - at least this was what I understood at that time. The difference is that `--allow-expired` would only skip this single time check whereas all other checks should still apply. I could imagine the following use-case: OEM device vendor (A) ships firmware updates and OEM's customers (B) pass resigned updates down to end customers (C), resigning the updates with their individual keys. At this point, (B) might still want to verify that the original file was created by (A) without dealing to much out-of-band.
But this is only theoretical - not backed by a real-world scenario for me at the moment. I'm happy with `--no-verify` since I'm (A) :smile:

jluebbe:
> You requested this in your first comment to #460  - at least this was what I understood at that time.

Yes, you're right. That leaves this PR for `--allow-expired` with the open questions above.

jluebbe:
I've reworked this PR as a base for #770. To see the changes, use `git range-diff origin/master d8b37aeddf4b 64d775227be2`.
As a summary:
* Rebased onto current master branch
* I've not used `BIT(0)` to avoid interpreting `TRUE` as `CHECK_BUNDLE_NO_VERIFY` (perhaps after a rebase)
* As OpenSSL's `X509_STORE_new()` already creates a default param struct internally via `X509_VERIFY_PARAM_new()` we don't need to do that on our side.

jluebbe (approved):

jluebbe:
The build fails on Ubuntu 16.04, as `X509_STORE_get0_param` was introduced in OpenSSL 1.1.0.

jluebbe:
> The build fails on Ubuntu 16.04, as `X509_STORE_get0_param` was introduced in OpenSSL 1.1.0.

For v1.5, we have a backport patch which is applied to the Ubuntu 16.04 CI build since 2020-12-23. By now that Ubuntu release is out of normal support, so I'd prefer to drop testing of that backport patch.

jluebbe:
> For v1.5, we have a backport patch which is applied to the Ubuntu 16.04 CI build since 2020-12-23. By now that Ubuntu release is out of normal support, so I'd prefer to drop testing of that backport patch.

I've created #828 for that.

ejoerns:
I've rebased this onto master that contains the merged #828 now

ejoerns:
I've extended this by a minimal test for the new `--no-check-time` command line option. Otherwise, it looks fine to me!

ejoerns (approved):

PR #464, opened 2019-07-18
service: add InfoVerify to D-Bus allowing to optionally skip bundle verification

beralt:
As already mentioned in #248 (more specific this https://github.com/rauc/rauc/issues/248#issuecomment-429223150 one), using the Info method via D-Bus is useful to obtain bundle information prior to installing a bundle.

I had the same use-case: I have a (rather large) bundle on a (slow) USB removable storage and verification is just taking too much time. I rather skip verification at that stage.

This pull request adds a new D-Bus method and deprecates the old one.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -512,7 +512,9 @@ Methods
 ~~~~~~~
 :ref:`Install <gdbus-method-de-pengutronix-rauc-Installer.Install>` (IN  s source);
 
-:ref:`Info <gdbus-method-de-pengutronix-rauc-Installer.Info>` (IN  s bundle, s compatible, s version);
+:ref:`Info <gdbus-method-de-pengutronix-rauc-Installer.Info>` (IN  s bundle, s compatible, s version); (deprecated)
+
+:ref:`InfoVerify <gdbus-method-de-pengutronix-rauc-Installer.InfoVerify>` (IN  s bundle, IN b verify, s compatible, s version);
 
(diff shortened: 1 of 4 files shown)


jluebbe:
This would run unsqashfs on unauthenticated data, significantly increasing the attack surface, so very skeptical.
Our expectation is that info is normally not used on a target device, but rather just starting the installation of a bundle.

What's your use case for accessing the (unauthenticated) info without also installing?

In general, I'd prefer to move to the design outlined on the [mailing list ](https://www.mail-archive.com/rauc@pengutronix.de/msg00086.html), which would allow authenticating parts of the image incrementally and also improve installation performance.

jluebbe:
Is your use-case the same as desribed in https://github.com/rauc/rauc/issues/248#issuecomment-429223150? If so, it' suggest using filenames (which are unauthenticated anyway) to select the bundle to install. Only after authentication, you can trust the information from the manifest anyway.

beralt:
The use-case is to provide the user with some notion if a bundle would install, before actually installing it, because it takes a significant amount of time before the manifest becomes available if bundle verification is enabled. Consider it as a sort of pre-install check to prevent the user having to wait a few minutes before a bundle is considered incompatible during installation.

But I now understand that the way this was solved is a bad decision, given that unauthenticated data is passed to the (un)squashfs and manifest parser.

I've read the new partial bundle authentication suggestion on the mailing list. It would solve the issue that I tried to solve in the first place, while providing a verified manifest in return. So I'll close this PR and wait for the new functionality instead.

PR #465, opened 2019-07-23
Fix bundle detection on big endian

ukleinek:
The squashfs magic is contained in little endian order in the image. To
be able to detect it correctly on big endian the magic value needs its
endianess switched to properly compare with the respective value found
using read(2) from the image.

With this the test suite runs successfully on a sparc64 box apart from a single test that fails because mksquashfs dies with a bus error, which I reported on https://bugs.debian.org/932787 .

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -248,7 +248,7 @@ static gboolean input_stream_read_uint64_all(GInputStream *stream,
 	return res;
 }
 
-#define SQUASHFS_MAGIC			0x73717368
+#define SQUASHFS_MAGIC			GUINT32_TO_LE(0x73717368)
 
 /* Attempts to read and verify the squashfs magic to verify having a valid bundle */
 static gboolean input_stream_check_bundle_identifier(GInputStream *stream, GError **error)


src/bundle.c
```diff
@@ -248,7 +248,7 @@ static gboolean input_stream_read_uint64_all(GInputStream *stream,
 	return res;
 }
 
-#define SQUASHFS_MAGIC			0x73717368
+#define SQUASHFS_MAGIC			GUINT32_FROM_LE(0x73717368)
```
a3f: This should be `GUINT32_TO_LE` because you are using it to compare with a little endian `GUINT32`. Doesn't matter except for PDP Endian, but makes he intent clearer

jluebbe: Yes, @a3f, you're right. @ukleinek to you want to swap it or should I?

ukleinek: I thought a bit about that and I think `GUINT32_FROM_LE` is the right one. We're reading a 32bit little endian number from the image and interpret it as native number. So that's what should be done with the little endian constant that we're expecting to read, don't we?

Alternatively we could compare byte by byte.

a3f: If we add some types it becomes clearer:

```
__le32 guint32_from_image;
__host32 guint32_integer_literal;
__host32 GUINT32_FROM_LE(__le32);
__le32 GUINT32_TO_LE(__host32);
```

So valid combinations are either `if (GUINT32_FROM_LE(guint32_from_image) == guint32_integer_literal)` or `if (GUINT32_TO_LE(guint32_integer_literal) == guin32_from_image)`.

ukleinek: ok, convinced

jluebbe:
Thanks @ukleinek and @a3f!

ejoerns:
@elboulangero This should be the fix for at least part of the failures seen in e.g. https://buildd.debian.org/status/fetch.php?pkg=rauc&arch=powerpc&ver=1.1-1&stamp=1567715803&raw=0

elboulangero:
Ack, thanks!

PR #466, opened 2019-07-23
Add casync tmpdir option to system.conf

gportay:
This commit allows setting a different temporary directory for `casync` at extraction.

`casync` caches things in the temporary directory. The default location `/var/tmp` may be limited in some use-cases. Thus, the choice of a temporary directory is very important, and the user/maintainer must have control of it.

For example, if the things that are being updated is very big (lets says 10GB), the seeding requires lots of inodes and `casync` may run out of inodes if the filesystem is already full of files.

The example below sets the temporary directory to `/tmp`:

```
$ cat /etc/rauc/system.conf
(...)
[casync]
tmppath=/tmp
(...)
```

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -145,6 +145,14 @@ For more information about using casync support of RAUC, refer to
   By default, the chunk store path is derived from the location of the RAUC
   bundle you install.
 
+``tmppath``
+  Allows to set the path to use as temporary directory for casync.
+  The temporary directory used by casync can be specified using the TMPDIR
+  environment variable. It falls back to /var/tmp if unset.
+  If ``tmppath`` is set then RAUC runs casync with TMPDIR sets to that path.
+  By default, the temporary directory is left unset by RAUC and casync uses its
+  internal default value ``/var/tmp``.
+
 **[autoinstall] section**
 
 The auto-install feature allows to configure a path that will be checked upon
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -160,8 +160,9 @@ static gboolean copy_raw_image(RaucImage *image, GUnixOutputStream *outstream, G
 	return TRUE;
 }
 
-static gboolean casync_extract(RaucImage *image, gchar *dest, const gchar *seed, const gchar *store, GError **error)
+static gboolean casync_extract(RaucImage *image, gchar *dest, const gchar *seed, const gchar *store, const gchar *tmpdir, GError **error)
 {
+	g_autoptr(GSubprocessLauncher) launcher = NULL;
 	g_autoptr(GSubprocess) sproc = NULL;
 	GError *ierror = NULL;
 	gboolean res = FALSE;
@@ -182,9 +183,13 @@ static gboolean casync_extract(RaucImage *image, gchar *dest, const gchar *seed,
 	g_ptr_array_add(args, g_strdup(dest));
 	g_ptr_array_add(args, NULL);
 
+	launcher = g_subprocess_launcher_new(G_SUBPROCESS_FLAGS_NONE);
+	if (tmpdir)
+		g_subprocess_launcher_setenv(launcher, "TMPDIR", tmpdir, TRUE);
+
 	r_debug_subprocess(args);
-	sproc = g_subprocess_newv((const gchar * const *)args->pdata,
-			G_SUBPROCESS_FLAGS_NONE, &ierror);
+	sproc = g_subprocess_launcher_spawnv(launcher,
+			(const gchar * const *)args->pdata, &ierror);
 	if (sproc == NULL) {
 		g_propagate_prefixed_error(
 				error,
@@ -233,6 +238,7 @@ static gboolean casync_extract_image(RaucImage *image, gchar *dest, GError **err
 	RaucSlot *seedslot = NULL;
 	g_autofree gchar *seed = NULL;
 	gchar *store = NULL;
+	gchar *tmpdir = NULL;
 	gboolean seed_mounted = FALSE;
 
 	/* Prepare Seed */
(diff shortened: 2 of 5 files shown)


docs/reference.rst
```diff
@@ -145,6 +145,10 @@ For more information about using casync support of RAUC, refer to
   By default, the chunk store path is derived from the location of the RAUC
   bundle you install.
 
+``tmpdir``
```
jluebbe: For consistency with the other options, I'd prefer tmppath:
```suggestion
``tmppath``
```

docs/reference.rst
```diff
@@ -145,6 +145,10 @@ For more information about using casync support of RAUC, refer to
...
+  Allows to set the path to use as temporary directory for casync.
+  By default, the temporary directory is set to /var/tmp.
```
jluebbe: You say that /var/tmp is used by default, but this is not configured explicitly when starting casync. It seems casync would use TMPDIR if set already, and then fall back to an internal casync default.
That seems to be a sensible behavior, but should be reflected in the docs.

docs/reference.rst
gportay: Done.

However, I kept a reference named `tmpdir` to reflect the `casync` environment variable `TMPDIR`.

See function `static gboolean casync_extract(RaucImage *image, gchar *dest, const gchar *seed, const gchar *store, const gchar *tmpdir, GError **error).`

This is identical to `storepath` that become `--store`.

docs/reference.rst
gportay: Done.

Hope my reword is good enough ;)

src/update_handler.c
```diff
@@ -270,8 +276,12 @@ static gboolean casync_extract_image(RaucImage *image, gchar *dest, GError **err
 	store = r_context()->install_info->mounted_bundle->storepath;
 	g_debug("Using store path: '%s'", store);
 
+	/* Set temporary directory */
+	tmpdir = r_context()->config->tmp_path;
+	g_debug("Using tmp path: '%s'", tmpdir);
```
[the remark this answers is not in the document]
gportay: Makes sense. done locally.

src/update_handler.c
```diff
@@ -183,8 +184,12 @@ static gboolean casync_extract(RaucImage *image, gchar *dest, const gchar *seed,
 	g_ptr_array_add(args, NULL);
 
 	r_debug_subprocess(args);
-	sproc = g_subprocess_newv((const gchar * const *)args->pdata,
-			G_SUBPROCESS_FLAGS_NONE, &ierror);
+	launcher = g_subprocess_launcher_new(G_SUBPROCESS_FLAGS_NONE);
```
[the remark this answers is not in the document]
gportay: Done.

gportay:
> And, we should add at least some minimal test for assuring that this option is at least tested for existence. I know we are missing a check for storepath, too, maybe we can simply combine this for now.
> 
> The relevant test file is https://github.com/rauc/rauc/blob/master/test/config_file.c

I will do this if it is only checking the existence of those options.

To let you know, I noticed that `make check` is not working on my setup. Is that my fault?

I worked around it with, I do not know if this is the right fix.

```
diff --git a/test/install.c b/test/install.c
index 41e695b..d0543cb 100644
--- a/test/install.c
+++ b/test/install.c
@@ -277,10 +277,10 @@ filename=bootloader.img";
        g_assert_true(g_hash_table_contains(tgrp, "demofs"));
        g_assert_true(g_hash_table_contains(tgrp, "bootloader"));
        g_assert_true(g_hash_table_contains(tgrp, "prebootloader"));
-       g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "rescue"))->name, ==, "rescue.0");
-       g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "rootfs"))->name, ==, "rootfs.1");
-       g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "appfs"))->name, ==, "appfs.1");
-       g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "demofs"))->name, ==, "demofs.1");
+       g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "rescue"))->name, ==, "rescue.1");
+       g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "rootfs"))->name, ==, "rootfs.2");
+       g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "appfs"))->name, ==, "appfs.2");
+       g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "demofs"))->name, ==, "demofs.2");
        g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "bootloader"))->name, ==, "bootloader.0");
        g_assert_cmpstr(((RaucSlot*)g_hash_table_lookup(tgrp, "prebootloader"))->name, ==, "prebootloader.0");
        g_assert_cmpint(g_hash_table_size(tgrp), ==, 6);
```

ejoerns:
> I will do this if it is only checking the existence of those options.

Yes.

> To let you know, I noticed that `make check` is not working on my setup. Is that my fault?
> 
> I have workaround it with the following patch, but I do not know if this is the right fix.

I cannot see any relation between the things you changed and the fix you propose.
Maybe submitt your changes an let us see if the test job also fails. This might give some more targeted information on what to change.

gportay:
> > To let you know, I noticed that `make check` is not working on my setup. Is that my fault?
> > I have workaround it with the following patch, but I do not know if this is the right fix.
> 
> I cannot see any relation between the things you changed and the fix you propose.
> Maybe submitt your changes an let us see if the test job also fails. This might give some more targeted information on what to change.

I am sorry if I was not clear...

I got that issue before applying my changes *AND* I was not using the Dockerfile (in `test/`) and docker. So the issue I faced and my changes are *NOT* related.

Now, I am using the Dockerfile (from `test/`) and docker and I do not have any issues.

Everything is fine, testing using docker.

gportay:
> Just some minor comments, looks good otherwise!

I marked the conversations as resolved. But I wonder why it still says that "changes are requested".

I am not so familiar with GitHub review. Tell me if I missed something.

ejoerns:
@gportay that was because the review from @jluebbe was stil marked as 'changes' required. I've now dismissed it as you obviously performed the requested changes.

ejoerns:
@gportay Thanks for your contribution

PR #467, opened 2019-07-25
service: add annotations for qdbusxml2cpp

tobydox:
When generating interface classes for Qt projects using qdbusxml2cpp
custom types must be annotated.

To use the generated interface class the RaucSlotStatusArray and
RaucProgress types have to be defined and registered as Qt meta types.
Additionally QDBusArgument proper streaming operators have to be
supplied.

Signed-off-by: Tobias Junghans <tobydox@veyon.io>

### DIFF
--- a/src/rauc-installer.xml
+++ b/src/rauc-installer.xml
@@ -50,6 +50,7 @@
     -->
     <method name="GetSlotStatus">
       <arg name="slot_status_array" type="a(sa{sv})" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out0" value="RaucSlotStatusArray"/>
     </method>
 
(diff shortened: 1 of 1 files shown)


jluebbe:
From your commit message:

> Additionally QDBusArgument proper streaming operators have to be
supplied.

I've not seen the corresponding change in the commit (only for the two argument types). Is there something missing from the commit?

tobydox:
That's just an additional information for developers when using `qdbusxml2cpp` and is not related to the D-Bus interface XML itself. If it's too confusing, I can strip down the commit message.

jluebbe:
> That's just an additional information for developers when using `qdbusxml2cpp` and is not related to the D-Bus interface XML itself. If it's too confusing, I can strip down the commit message.

OK, that sounds like that would be obvious to Qt developers. :)

PR #475, opened 2019-09-08
test/install: deactivate more tests that depend on GHashTable impleme…

ukleinek:
…ntation

In the same spirit as 35c98ab18f90 ("test/install: deactivate check that
depends on GHashTable implementation") also deactivate three further
tests that depend on the interna of GHashTable. 

With this change on top of a cherry-pick of 35c98ab18f and bc25067177 this fixes building the rauc 1.1 Debian package on s390x.

PR #476, opened 2019-09-09
test/install: fix test for undefined behavior in slot selection handling

ejoerns:
For the triple-redundancy rootfs.0, rootfs.1, rootfs.2 (including their
childs), we have rootfs.0 set to 'active' so far and thus both rootfs.2
and rootfs.2 being detected as 'inactive'.

For this case, there is still no real definition of when or whether to select
rootfs.1 or rootfs.2.

Prior to this, the test checked for the selected slot being rootfs.1
which was valid for the concrete implementation, but mainly depending on
the order that was given by the GHashTable (iterator) implementation.

In recent glib version this changed resulting in rootfs.2 being selected
instead.

We adapt the test for checking correctly for both cases.

Supersedes #475

### DIFF
--- a/test/install.c
+++ b/test/install.c
@@ -257,6 +257,8 @@ filename=bootloader.img";
 
 	g_assert_true(load_manifest_file(pathname, &rm, NULL));
 
+	r_context_conf()->bootslot = g_strdup("system0");
+
 	g_assert_true(determine_slot_states(NULL));
 
 	g_assert_nonnull(r_context()->config);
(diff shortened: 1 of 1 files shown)


ejoerns:
@elboulangero This should be of interest for you, too.

ukleinek:
I didn't look closely, but I wonder if we need something like https://github.com/ejoerns/rauc/commit/9230105a66c72be46719003bf5f048d82a070aaa for this commit to be correct.

ejoerns:
@ukleinek it works without, but more accidentially than intended. And there is no guarantee that it will work in the future, too. Thus for this PR it makes sense to include the commit you mentioned to fully fix the known issues in this test.

ukleinek:
I would swap the order of the two commits, other than that it looks fine for me now

ejoerns:
@ukleinek I've fixed the typos and, while at it, I also changed the order of commits if that looks cleaner to you ;)

ejoerns:
@ukleinek really reliably fixed typos now.

PR #482, opened 2019-09-13
keep BOOT_ACK and PARTITION_ACCESS values 

s-michaelsen:
when writing number of active boot partition to eMMCs EXT_CSD_PART_CONFIG register

### DIFF
--- a/src/emmc.c
+++ b/src/emmc.c
@@ -88,6 +88,7 @@ gboolean r_emmc_write_bootpart(const gchar *device, gint bootpart_active, GError
 	gint ret;
 	gboolean res = FALSE;
 	int fd;
+	guint8 extcsd[512];
 	guint8 value = 0;
 
 	g_return_val_if_fail(bootpart_active == 0 || bootpart_active == 1, FALSE);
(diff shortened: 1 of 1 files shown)


ejoerns:
@s-michaelsen please add your Signed-of-by line here (https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin)

jluebbe:
For the PARTITION_ACCESS, this should not be needed, as the kernel is responsible for maintaining it (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a0e9576634f0bcfa5c4d8efdcf8c9b28298d7570). It would be vulnerable to races between userspace an kernel anyway. Did you see problems with BOOT_ACK only, or also with PARTITION_ACCESS?
Regarding the BOOT_ACK value, I agree. A comment in the code explaining this would be useful to avoid problems in the future.

s-michaelsen:
I saw only problems with BOOT_ACK flag. Only added PARTITION_ACCESS as well after reading register spec. I'll remove the according part and change the comment.

jluebbe:
I've squashed the commits together and reworded the comment a bit.

src/emmc.c
```diff
@@ -99,11 +100,26 @@ gboolean r_emmc_write_bootpart(const gchar *device, gint bootpart_active, GError
...
+				EXT_CSD_PART_CONFIG, device);
+		goto out;
+	}
+
+	/* Keep BOOT_ACK value as it is. Resetting this bit might prevent
+	 * proper boot process on some platforms since in some SoCs, ROM boot
+	 * loaders need this flag to be set.
+	 * The PARTITION_ACCESS value is handled inside the kernel so we do not
+	 * need to take care of it here. */
+	value = extcsd[EXT_CSD_PART_CONFIG] & ~(1 << 6);
```
ejoerns: This clears `BOOT_ACK` which is the opposite of what we intended to achieve.

Shouldn't that be:
``` suggestion
	value = extcsd[EXT_CSD_PART_CONFIG] & (1 << 6);
```

ejoerns:
I've squashed the commits and rebased onto current master.

PR #484, opened 2019-09-16
Fix leaking GDataInputStream's

ejoerns:
The GDataInputStream 'datainstream' is created from 'instream' with
g_data_input_stream_new() but never freed.

Luckily we have g_autoptr() mechanism, so that we just need to declare
'datainstream' as an auto pointer variable here to have it freed
correctly as soon as it leaves the context.

Note that the bug described above also leads to an open file descriptor
remaining.
The more often on calls, for example, 'rauc status' the more open file
descriptors will be collected until they suddenly hit the maximum number
of open file descriptors allowed (e.g. by systemd service settings)
resulting in error message:

> Too many open files

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -62,7 +62,7 @@ static gboolean barebox_state_get(const gchar* bootname, BareboxSlotState *bb_st
 	g_autoptr(GSubprocess) sub = NULL;
 	GError *ierror = NULL;
 	GInputStream *instream;
-	GDataInputStream *datainstream;
+	g_autoptr(GDataInputStream) datainstream = NULL;
 	gchar* outline;
 	guint64 result[2] = {};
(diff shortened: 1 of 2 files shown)


src/bootchooser.c
```diff
@@ -62,7 +62,7 @@ static gboolean barebox_state_get(const gchar* bootname, BareboxSlotState *bb_st
 	g_autoptr(GSubprocess) sub = NULL;
 	GError *ierror = NULL;
 	GInputStream *instream;
-	GDataInputStream *datainstream;
+	g_autoptr(GDataInputStream) datainstream;
```
a3f: Must be initialized to `NULL`. Otherwise, a non-initialized value will be destructed whenever `g_subprocess_newv` fails.

ejoerns: You're right, missed that. Thx!

ejoerns:
@ukleinek thanks for the hint, fixed that!

PR #486, opened 2019-09-25
A more fancy rauc status output

ejoerns:
Note that this is based on #477

So far, especially when having more complex slot setups, the output of
'rauc status' did not provide an easy and clear view of the slots
constellation and their respective status (booted, active, inactive).

This patch aims to fix this by giving a more tidied-up view with colors
and bold font for highlighting.

The (uncolored) variant woll look as follows:

```
=== System Info ===
Compatible:  Test Config
Variant:     myvariant
Booted from: rootfs.1 (system1)

=== Bootloader ===
Activated: rootfs.0 (system0)

=== Slot States ===
  [bootloader.0] (/tmp/boot-0, raw, inactive)
	description: 

○ [rootfs.1] (/tmp/rootfs-1, raw, booted)
	bootname: system1
	description: 
	boot status: good
    [appfs.1] (/tmp/appfs-1, raw, active)
	description: 

⏺ [rootfs.0] (/tmp/rootfs-0, raw, inactive)
	bootname: system0
	description: 
	boot status: good
    [appfs.0] (/tmp/appfs-0, raw, inactive)
	description: 


```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -973,23 +973,48 @@ static void free_status_print(RaucStatusPrint *status)
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(RaucStatusPrint, free_status_print);
 
+/* Definition list for terminal colors */
+#define KNRM  "\x1B[0m"
+#define KRED  "\x1B[31m"
+#define KGRN  "\x1B[32m"
+#define KYEL  "\x1B[33m"
+#define KBLU  "\x1B[34m"
+#define KMAG  "\x1B[35m"
+#define KCYN  "\x1B[36m"
+#define KWHT  "\x1B[37m"
+#define KBLD  "\x1B[1m"
+
 static void r_string_append_slot(GString *text, RaucSlot *slot, RaucStatusPrint *status)
 {
 	RaucSlotStatus *slot_state = slot->status;
 
-	g_string_append_printf(text, "\n  %s: class=%s, device=%s, type=%s, bootname=%s\n",
-			slot->name, slot->sclass, slot->device, slot->type, slot->bootname);
-	g_string_append_printf(text, "      state=%s, description=%s", r_slot_slotstate_to_str(slot->state), slot->description);
-	if (slot->parent)
-		g_string_append_printf(text, ", parent=%s", slot->parent->name);
-	else
-		g_string_append(text, ", parent=(none)");
-	if (slot->mount_point || slot->ext_mount_point)
-		g_string_append_printf(text, ", mountpoint=%s", slot->mount_point ? slot->mount_point : slot->ext_mount_point);
-	else
-		g_string_append(text, ", mountpoint=(none)");
+	/* bootname slots get an empty or full dot depending on whether they
+	 * are primary or not */
+	if (slot->bootname) {
+		if (utf8_supported)
+			g_string_append_unichar(text, slot == status->primary ? 0x23FA :  0x25CB);
+		else
+			g_string_append_c(text, slot == status->primary ? 'o' :  'x');
+	} else {
+		g_string_append_c(text, ' ');
+	}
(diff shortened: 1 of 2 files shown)


ejoerns:
Now rebased onto current master.

src/main.c
```diff
@@ -1007,23 +1008,41 @@ static void free_status_print(RaucStatusPrint *status)
...
+			slot->state & ST_ACTIVE ? KGRN : "",
+			slot->state == ST_BOOTED ? KGRN : "",
+			slot->name,
+			slot->device,
+			slot->type,
+			slot->state == ST_BOOTED ? KBLD : "",
+			r_slot_slotstate_to_str(slot->state));
+	if (slot->bootname)
+		g_string_append_printf(text, "\n\tbootname: "KBLU "%s"KNRM, slot->bootname);
+	if (slot->description)
```
jluebbe: In you example you have empty description lines, maybe there is a pointer to an empty string?

src/main.c
```diff
@@ -1007,23 +1008,41 @@ static void free_status_print(RaucStatusPrint *status)
...
-	else
-		g_string_append(text, ", parent=(none)");
-	if (slot->mount_point || slot->ext_mount_point)
-		g_string_append_printf(text, ", mountpoint=%s", slot->mount_point ? slot->mount_point : slot->ext_mount_point);
-	else
-		g_string_append(text, ", mountpoint=(none)");
+	setlocale(LC_ALL, "");
+
+	g_string_append_printf(text, "%lc %s%s%s"KBLD "[%s]"KNRM " (%s, %s, %s%s"KWHT ")"KNRM,
+			slot->bootname ? slot == status->primary ? L'\u23FA' :  L'\u25CB' : ' ',
```
jluebbe: I'm not sure that `%lc` and L'\23FA' is the correct approach. There is g_string_append_unichar() which could be used instead. As this is at the beginning of this printf, it should be simple to adapt.

src/main.c
```diff
@@ -1007,23 +1008,41 @@ static void free_status_print(RaucStatusPrint *status)
...
-	g_string_append_printf(text, "      state=%s, description=%s", r_slot_slotstate_to_str(slot->state), slot->description);
-	if (slot->parent)
-		g_string_append_printf(text, ", parent=%s", slot->parent->name);
-	else
-		g_string_append(text, ", parent=(none)");
-	if (slot->mount_point || slot->ext_mount_point)
-		g_string_append_printf(text, ", mountpoint=%s", slot->mount_point ? slot->mount_point : slot->ext_mount_point);
-	else
-		g_string_append(text, ", mountpoint=(none)");
+	setlocale(LC_ALL, "");
```
jluebbe: Just calling setlocale here seems strange. If we need to change the locale at all, we should do it somewhere obvious.

src/main.c
jluebbe: The i386 build also fails with
```
src/main.c:1028:34: error: format '%lc' expects argument of type 'wint_t', but argument 3 has type 'long int' [-Werror=format=]
```

src/main.c
ejoerns: Yes, in config file parsing it is set to an empty string when being unset. I think we could simply drop this as we potentially handle a NULL description everywhere.

src/main.c
ejoerns: Yes, this is probably also more obvious. I change it to use `g_string_append_unichar()`.

src/main.c
ejoerns: I've tried to make this a little more sophisticated now. We attempt to set LC_ALL to use the env values and then check if this gives us utf-8 support. If not, we now also have a non-utf8 display variant.

ejoerns:
Rebased onto master to fix conflicts with merged PRs.

PR #492, opened 2019-10-07
src/main: add help text for 'service' (if configured)

ejoerns:

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1706,6 +1706,9 @@ static void cmdline_handler(int argc, char **argv)
 			"  checksum\tUpdate a manifest with checksums (and optionally sign it)\n"
 			"  install\tInstall a bundle\n"
 			"  info\t\tShow file information\n"
+#if ENABLE_SERVICE == 1
+			"  service\tStart RAUC service\n"
+#endif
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -1706,6 +1712,7 @@ static void cmdline_handler(int argc, char **argv)
 			"  checksum\tUpdate a manifest with checksums (and optionally sign it)\n" \
 			"  install\tInstall a bundle\n" \
 			"  info\t\tShow file information\n" \
+			SERVICE_HELP_TEXT \
```
jluebbe: As this is used only in one place, why use the define above (instead of just `#if` / `#endif`)?

ejoerns: I had some troubles with the multi-line string and #if in between. If you have a working solution, I would agree that this is preferable ;)

a3f: Drop the trailing backslashes and it will work.

ejoerns: @a3f You're right. I thought it should work and tested it, but obviously did something wrong or drew the wrong conclusions.. I've now removed them all and inserted the define in-place. Looks much cleaner now :+1:

PR #498, opened 2019-10-24
Introduce --signing-keyring to fix post-signing verification for 'rauc resign'

ejoerns:
When doing post-signing verification for 'rauc resign' we must be aware
that we need two different keyrings:

a) The one to verify the *original* bundle against
b) The one to verify the *resigned* bundle against

For now we only had the option `--keyring` and the same keyring was used
for both case a) and b).
While there are cases where this is valid (if the keyring accepts both
signatures), the default is that this will fail.

The existing '--keyring' parameter is already used for passing the
systems keyring to the RAUC service etc.
Thus for post-signing verification we must introduce another parameter to
explicitly pass the 'second' keyring. This is what we do here with the
'--verify-keyring' parameter.

Note that this will 'break' setups that did use the '--keyring'
verification feature before in the way that the bundle will be generated
but not verified. These setups need to migrate to the '--verify-keyring'
feature.

To be consistent, this option is added for subcommands 'bundle',
'resign', and 'convert'.

Apart from this, the series also fixes an invalid handling within the tests revealed in the context of this issue.

This aims to replace #461 by targeting the critics mentioned there.

### DIFF
--- a/include/signature.h
+++ b/include/signature.h
@@ -73,14 +73,13 @@ GBytes *cms_sign_file(const gchar *filename, const gchar *certfile, const gchar 
  *
  * @param content content to verify against signature
  * @param sig signature used to verify
+ * @param store X509 store to use for verification
  * @param cms Return location for the CMS_ContentInfo used for verification
- * @param store Return location for the X509 store used for verification
  * @param error return location for a GError, or NULL
  *
  * @return TRUE if succeeded, FALSE if failed
  */
-gboolean cms_verify(GBytes *content, GBytes *sig, CMS_ContentInfo **cms, X509_STORE **store, GError **error);
-
+gboolean cms_verify(GBytes *content, GBytes *sig, X509_STORE *store, CMS_ContentInfo **cms, GError **error);
 
 /**
  * Verify signature for given file.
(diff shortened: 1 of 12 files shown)


src/signature.c
```diff
@@ -389,17 +389,26 @@ GBytes *cms_sign(GBytes *content, const gchar *certfile, const gchar *keyfile, g
...
+	if (r_context()->verify_keyringpath) {
+		gchar *orig_keyring_path;
 		g_autoptr(CMS_ContentInfo) vcms = NULL;
 		g_autoptr(X509_STORE) store = NULL;
 		STACK_OF(X509) *verified_chain = NULL;
 
+		/* temporary set config->keyring_path to verify_keyring_pathpath */
 		g_message("Keyring given, doing signature verification");
+		orig_keyring_path = r_context()->config->keyring_path;
+		r_context()->config->keyring_path = r_context()->verify_keyringpath;
```
jluebbe: The context and config is intended to be read-only for the lower layers.

ejoerns: I've changed this by adding new arguments to cms_sign() that allow explicitly passing non-default capath/cadir.

ejoerns:
> 'verify' is not a useful prefix for a specific keyring: Any keyring is only used for verification, etiher for existing bundles or newly created bundles.
> Also, we don't need to change the existing behavior there:
> `--keyring` could set both keyrings

This is the new (old) behavior now.

> `--A-keyring` and `--B-keyring` would be used for more specific configuration
> Which names we should use for A and B needs some more thought...

I only implemented --B-keyring first of all, it is now named `signing-keyring`. We could however add an `--A-keyring` later that only sets input verification keyring but not the output one.

test/bundle.c
```diff
@@ -157,6 +157,9 @@ static void bundle_test_extract_manifest(BundleFixture *fixture,
 	free_bundle(bundle);
 }
 
+// Hack to pull-in context for testing modification
+extern RaucContext *context;
```
jluebbe: I'm not happy about this hack, but it's probably the easiest solution for the tests.

include/signature.h
```diff
@@ -73,13 +73,15 @@ GBytes *cms_sign_file(const gchar *filename, const gchar *certfile, const gchar
  *
  * @param content content to verify against signature
  * @param sig signature used to verify
+ * @param capath path to certificate file to use for verification
+ * @param cadir path to certificate dir to use for verification
```
jluebbe: Hmm, maybe we should look at passing a preparing a store for use instead of adding more parameters?

ejoerns: Probably the better (more generic) option, yes.

test/bundle.c
ejoerns: Ok, then I would leave it as it is ;)

include/signature.h
ejoerns: Next round... After a lot of refactoring and rework in test/signature.c, here is not the approach for passing a store directly.

PR #499, opened 2019-10-24
docs/integration: some clarification on grub.cfg and grubenv placement

ejoerns:

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -745,13 +745,25 @@ To enable handling of redundant booting in GRUB, manual scripting is required.
 The GRUB bootloader interface of RAUC uses the GRUB environment variables
 ``<bootname>_OK``, ``<bootname>_TRY`` and ``ORDER``.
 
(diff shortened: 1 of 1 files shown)


docs/integration.rst
```diff
@@ -745,14 +745,26 @@ To enable handling of redundant booting in GRUB, manual scripting is required.
...
+``grub-editenv`` available on your target.
+
+By default RAUC expects the grubenv file to be located at
+``/boot/grub/grubenv``, you can specify a custom directory by passing
+``grubenv=/path/to/grubenv`` in your system.conf ``[system]`` section.
+
+Make sure that the grubenv file is located outside your redundant rootfs
+partitions as the rootfs needs to be exchangeable without affecting the
+environment content.
+A proper location would be to place it on the EFI disk, e.g. at
```
jluebbe: GRUB is also used on non-UEFI systems.
```suggestion
For UEFI systems, a proper location would be to place it on the EFI disk, e.g. at
```

PR #503, opened 2019-11-04
src/update_handler: use g_seekable_tell() to get size of written data

### LINKED ISSUE #463: Failure installing large filesystems on 32bit machines
A colleague just ran into a problem installing ext4 images > 2GB on a 32bit machine. I took a look at the code and found this: 
The reason seems to be the use of g_output_stream_splice() in update_handler.c, line 134. The documentation of that function indicates that it supports writing files larger than G_MAXSSIZE (which from what I know is limited by the actual size of size_t) but the returned value does not correspond to the actual size processed any more. As a result the size check fails even if the image was written completely.

ejoerns:
Documentation of g_output_stream_splice() says:

> Note that if the number of bytes spliced is greater than G_MAXSSIZE,
> then that will be returned, and there is no way to determine the actual
> number of bytes spliced.

As copy_raw_image() performs a sanity check if returned size matches
expected write size, it will not be possible to install an image whose
size exceeds G_MAXSSIZE.

Thus instead of letting g_output_stream_splice() close the output stream, we
leave it open to use g_seekable_tell() on it and obtain the actual size
read (written), even if it exceeds gssize.

Fixes #463.

### DIFF
--- a/src/checksum.c
+++ b/src/checksum.c
@@ -17,10 +17,10 @@ G_STATIC_ASSERT(RAUC_DEFAULT_CHECKSUM != 0);
 G_DEFINE_QUARK(r-checksum-error-quark, r_checksum_error)
 
 static gboolean
-update_from_file(GChecksum *ctx, const gchar *filename, gsize *total, GError **error)
+update_from_file(GChecksum *ctx, const gchar *filename, goffset *total, GError **error)
 {
 	g_auto(filedesc) fd = -1;
-	gsize size = 0;
+	goffset size = 0;
 	gssize r;
 	guchar buf[4096];
 
@@ -51,7 +51,7 @@ gboolean compute_checksum(RaucChecksum *checksum, const gchar *filename, GError 
 {
 	g_autoptr(GChecksum) ctx = NULL;
 	GChecksumType type = checksum->type;
-	gsize total = 0;
+	goffset total = 0;
 
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
--- a/src/signature.c
+++ b/src/signature.c
@@ -851,7 +851,7 @@ out:
 	return sig;
 }
 
-gboolean cms_verify_file(const gchar *filename, GBytes *sig, gsize limit, X509_STORE *store, CMS_ContentInfo **cms, GError **error)
+gboolean cms_verify_file(const gchar *filename, GBytes *sig, goffset limit, X509_STORE *store, CMS_ContentInfo **cms, GError **error)
 {
 	GError *ierror = NULL;
 	g_autoptr(GMappedFile) file = NULL;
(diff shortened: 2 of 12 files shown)


src/update_handler.c
```diff
@@ -132,16 +133,25 @@ static gboolean copy_raw_image(RaucImage *image, GUnixOutputStream *outstream, G
...
 	if (size == -1) {
 		g_propagate_prefixed_error(error, ierror,
 				"Failed splicing data: ");
 		return FALSE;
-	} else if (size != (gssize)image->checksum.size) {
+	}
+
+	seeksize = g_seekable_tell((GSeekable*) instream);
+
+	if (seeksize != (goffset)image->checksum.size) {
```
jluebbe: checksum.size is a gsize, which is 32bit long on 32bit architectures. So this would still fail on images >= 4GB. We should use goffset in the struct and everywhere we use gsize for file offsets/sizes as well.

ejoerns: The difference of `gsize` and `goffset` is that `goffset` is signed which is useless for values that should not become negative. Thus maybe we should move to `guint64` instead? This is the same that the mbr switching implementation already uses.

jluebbe: We've just discussed this internally, and will go with `goffset` for file sizes in general.

ejoerns:
I've updated the patch set to introduce goffset usage for both image and bundle handling.

Note that the size of bundles will still be limited because of current implementation details.

src/signature.c
```diff
@@ -871,6 +871,15 @@ gboolean cms_verify_file(const gchar *filename, GBytes *sig, gsize limit, X509_S
 	}
 	content = g_mapped_file_get_bytes(file);
 
+	if (limit > (goffset)G_MAXSIZE) {
```
jluebbe: G_MAXSIZE is 4GB on 32 bit systems, did you mean G_MAXSSIZE?
mmapping even 2GB on 32 bit systems is unlikely to work due to limited free address space.

src/checksum.c
```diff
@@ -17,7 +17,7 @@ G_STATIC_ASSERT(RAUC_DEFAULT_CHECKSUM != 0);
 G_DEFINE_QUARK(r-checksum-error-quark, r_checksum_error)
 
 static gboolean
-update_from_file(GChecksum *ctx, const gchar *filename, gsize *total, GError **error)
+update_from_file(GChecksum *ctx, const gchar *filename, goffset *total, GError **error)
 {
 	g_auto(filedesc) fd = -1;
 	gsize size = 0;
```
marv: This needs to be changed to `goffset` as well as it's used to accumulate the number of processed bytes which is later assigned to `*total`

```suggestion
	goffset size = 0;
```

With that line changed I've been able to install a >2GB bundle on an armv7 system

ejoerns: Indeed. Thanks!

src/signature.c
ejoerns: I think G_MAXSIZE is right, the commit's explanation is incorrect, should be 4GB instead of 2GB. It actually aims to catch `limit` values exceeding the (potentially) mmappable size. However, one would probably run into memory issues also for smaller bundles.

ejoerns:
> Just one minor issue.
> It looks like we need to switch to using streaming for signing/verifying to fix large bundles...

Yes, this is what my commit message hint

> This might need to be changed anyway

was silently pointing at, too. 

I'd propose opening an issue for this and target it in a different PR. What do you think @jluebbe ?

ejoerns:
I've pushed an update that fixes the remaining `gsize` and the commit message

src/signature.c
ejoerns: Now also got the point that casting G_MAXSIZE to goffset might not be such a good idea on 64bit systems... I've used (guint64) casts for both `limit` and G_MAXSIZE now.

ejoerns:
Also added a patch to enhance the debugability of test/signature.c as I ran into this..

ejoerns:
After some rounds of string format specifier fixes, I now also have Travis on my side \o/

PR #504, opened 2019-11-06
install: add support for external devices

medude:
This adds the support to use external devices. By external I mean
devices not known by RAUC (listed within the system.conf). So we are
able to use RAUC to initialise the system from a 'special' medium e.g. a
usb-stick which isn't part of the 'normal' update system. To signal it
to RAUC I introduced a cmdline flag called 'rauc.slot.external'. If the
flag is present RAUC won't try to find an active 'normal' slot.

Signed-off-by: Marco Felsch <m.felsch@pengutronix.de>

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -299,7 +299,7 @@ Identification via Kernel Commandline
 RAUC evaluates different kernel commandline parameters in the order they are
 listed below.
 
-.. rubric:: ``rauc.slot=``
+.. rubric:: ``rauc.slot=`` and ``rauc.external``
 
 This is the generic way to explicitly set information about which slot was
 booted by the bootloader.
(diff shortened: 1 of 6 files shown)


jluebbe:
I've added tests and some documentation.

docs/integration.rst
```diff
@@ -314,6 +314,12 @@ selected) you can also give the name of the slot::
 
   rauc.slot=recovery.0
 
+When booting from a source not configured your system.conf (for example from a
```
ejoerns:
```suggestion
When booting from a source not configured in your system.conf (for example from a
```

docs/integration.rst
```diff
@@ -314,6 +314,12 @@ selected) you can also give the name of the slot::
...
+USB memory stick), you can tell rauc explicitly with the flag ``
+rauc.external``.
+This means that all slots are known to be inactive and will be valid
+installation targets.
```
ejoerns: Vielleicht noch ein Hinweise à la:

"A common use case for this is to use RAUC during bootstrapping procedure to initially fill up slots."

PR #505, opened 2019-11-06
Fix eMMC boot device handling

medude:


src/update_handler.c
```diff
@@ -1273,8 +1276,20 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+	rpath = realpath(dest_slot->device, NULL);
+	if (!rpath) {
+		g_set_error(error,
+				R_UPDATE_ERROR,
+				R_UPDATE_ERROR_FAILED,
+				"Can't resolve eMMC device %s", dest_slot->device);
+		goto out;
+	}
+	real_device = g_strdup(rpath);
+	free(rpath);
```
ejoerns: Why do we need to duplicate the string here?

Wouldn't a `g_autofree gchar rpath = NULL` and using `rpath` directly (instead of `real_device`) be sufficient?

medude: As written in the comment I don't wanna mixup glib and libc types ;)

ejoerns: There is no such difference that would justify this. Anyway, the function prototype of `g_strdup()` expects a `gchar` as argument. Thus you only shift the implicit conversion.

medude: There is some sugar code for tracing within `g_malloc()`. As said I wanted to keep it clean but I can change this if you want.

jluebbe: `malloc()` must not be used with `g_free`, so this is correct. The two existing calls to realpath use a buffer on the stack, though. Maybe do the same here?

medude: Checked the other conversation with @a3f and the glib code again. Now the only difference is the possible DTRACE probe add. As I said if you want to change that I can do this.

ejoerns: @medude Yes, `g_free()` basically calls `free()`. But @jluebbe wanted to point out that using stack-allocated buffer (what I missed in the other commit) might be another option that does not require freeing.

PR #507, opened 2019-11-06
src/context: fix leaking string returned by realpath()

ejoerns:
We did not free the string returned by realpath(). This we fix by using
g_autofree now.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -28,7 +28,7 @@ static const gchar* get_cmdline_bootname(void)
 	g_autofree gchar *contents = NULL;
 	static const char *bootname = NULL;
 	gchar buf[PATH_MAX + 1];
-	gchar *realdev = NULL;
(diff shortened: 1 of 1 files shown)


ejoerns:
@a3f ah, good hint. `g_mem_is_system_malloc()` should always be true (and is set to `return TRUE` in recent glib versions anyway), but if I get it right you want to emphasize that we cannot assure `realpath()` uses the system allocator.
Thus we would need to call `free()`/`g_free()` on this explicitly!?

jluebbe:
As realpath is called here with a stack allocated buffer, the return value is a pointer to that buffer instead of a newly malloc'ed string. Accordingly it's incorrect to free it.

ejoerns:
@jluebbe ganh, you're right. Did not recognize zu `buf`. Thus nothing to do here :D

PR #512, opened 2019-11-17
remove misleading reference to ext2 / ext3 support.

ejoerns:
Various places of the RAUC documentation noted ext2 and ext3 file system
support. This was quite misleading as RAUC has an ext4 file type, but
not an ext2 / ext3 type.

This removes the notions of ext2 and ext3 and adds a section in the FAQ
why they are intentionally not supported.


jluebbe:
Perhaps "not supported" is a too strict wording, because ext2/3 images will just work fine when using the ext4 slot type. Maybe document that any filesystem mountable as ext4 should work?

Issue #511, opened 2019-11-14
casync chunks the seed chunk store


fhackenberger:
RAUC with casync seems to create a chunk index of the chunk index when installing a RAUC bundle.

```
root     29897  0.0  0.1 256164 10172 pts/0    Sl   18:36   0:00 rauc -d install /tmp/board-system-images-v3/update.casync.raucb
root     29922 85.9  0.1  48132 14444 ?        R    18:36  11:15 casync --log-level=debug --verbose extract --seed /run/rauc/rootfs.0 --store /tmp/board-system-images-v3/update.casync.castr --seed-output=no /run/rauc/bundle/rootfs.tar.gz.caidx /run/rauc/rootfs.1
```

```
ubuntu@sign-pc-18:~$ sudo du -xhcs /var/tmp/*.cased | sort -n     
3.3G    total
3.3G    /var/tmp/3a875a5021649923.cased
```
Logs from casync run by rauc:
```
Nov 12 18:48:54 sign-pc-18 rauc[29863]: Seeding var/tmp/3a875a5021649923.cased/ec82/ec828c3b0e9ca9277f2b36051edce47e8cae0e128e3fb21aeb70de8978c0ae99
Nov 12 18:48:54 sign-pc-18 rauc[29863]: Seeded var/tmp/3a875a5021649923.cased/ec82/ec828c3b0e9ca9277f2b36051edce47e8cae0e128e3fb21aeb70de8978c0ae99
```

In general it would be great, if casync could use reflinks when creating an uncompressed chunk store for a seed and rauc would set casync to use that. Not sure if this feature is available though. It would resolve a big issue with using casync (Less than 50% disk usage on slots is allowed).

fhackenberger:
The solution for the bug is quite easy:
Write a .caexclude file with the content:
`/var/tmp/*.cased`
and store it in `/` before running casync. This should happen in `casync_extract` in src/update_handler.c if a `seed` is set. Maybe there is also a way to make that path explicit when calling casync, from what I understand casync creates a new random directory for the extract run, if `--seed` is specified.

fhackenberger:
See https://github.com/rauc/rauc/pull/513 for a workaround, where a user created .caexclude file can be used to fix the initial `casync make` run.

ejoerns:
Maybe the right option is to set TMPDIR to a device not being the seeded one? Should be combined with `--exclude-submounts=yes` if I get it right.

fhackenberger:
[...]

PR #513, opened 2019-11-19
Skip using the active slot as a seed when installing bundles

fhackenberger:
As a workaround for resource intensive and slow install procedures when using casync and tar images (see https://github.com/rauc/rauc/issues/511) I created an option for the install command, which skips using the active slot as a seed when running casync.

This enables the following use case:
- Remote repository chunk store contains only chunks for most recent update
- Images have a .caexclude file included, which prevents /var/tmp/update.casync.castr from being chunked
- Device does the bind-mount & `casync make --store  /var/tmp/update.casync.castr /var/tmp/update.casync.caidx /mnt/rauc/` step in the background once.
- Device does an rsync with --delete on remote chunks store / index file / raucb
- Device installs without using slot as seed `rauc install --ignore-seed bundle.raucb`

The workaround decribed splits the long running operations into separate units, which can be run independently and frees the device to do a full `casync make` on every update. Downside is that additional storage is used permanently. But that's probably good, as it makes it clear how much space is actually left, if updates should still work.

### DIFF
--- a/contrib/cgi/src/cgi.c
+++ b/contrib/cgi/src/cgi.c
@@ -681,7 +681,7 @@ static gint cgi_handler(int argc, char **argv)
 			goto remove_bundle_on_error;
 
 		/* start rauc install */
-		if (!r_installer_call_install_sync(installer, BUNDLE_TARGET_LOCATION, NULL, &error))
+		if (!r_installer_call_install_sync(installer, BUNDLE_TARGET_LOCATION, FALSE, NULL, &error))
 			goto remove_bundle_on_error;
 
 		print_headers("200 OK", "text/plain");
--- a/include/context.h
+++ b/include/context.h
@@ -15,7 +15,7 @@ typedef struct {
 	/* The bundle currently mounted by RAUC */
 	RaucBundle *mounted_bundle;
 	/* Whether to skip the active slot as a seed */
-        gboolean ignore_slot_as_seed;
+	gboolean ignore_slot_as_seed;
 } RContextInstallationInfo;
 
 typedef struct {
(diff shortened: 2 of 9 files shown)


fhackenberger:
I'm not really familiar with your testing requirements. Can someone else contribute a test for this new feature?

fhackenberger:
@ejoerns How do you feel about the tests? Would you or someone else be able to add them? I've looked at the testing code, but I'm quite unfamiliar with the framework.

ejoerns:
I agree that it is a use case to disable seeding. Mainly because of casync's limited capabilities in seed store cleanup and for devices with really constrained temporary storage.

Thus basically introducing an option to disable seeding is something that I'd like to see.

But the way for selecting this option you chose is probably the wrong one:

* You break the API of the Install D-Bus command which will prevent all current API users from simply updating to a new RAUC release
* Deciding to not use the slot as a seed store is probably not a property of an installation but instead a property of the system.

To target the first point, we have https://github.com/rauc/rauc/pull/446 hanging around and waiting for getting merged.
But because of the second point I would prefer having a simple system config option that allows disabling seeding.

I thought that we already had some own developments towards this direction, but I was not able to find any.

@OnkelUlla could you also please have a look at this as you just cared for the casync stuff in context of https://github.com/rauc/rauc/pull/501 ?

fhackenberger:
Thanks for the feedback @ejoerns! I agree it should rather be a system setting in system.conf, having it as an argument for the install command is less than ideal. Not sure if I have the time to adapt this soon, so I'd also be happy if someone else continues this work.

PR #514, opened 2019-11-19
Add --no-signatures option

billpittman:
Remove signature checking from bundle operations.

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -45,6 +45,8 @@ typedef struct {
 	gchar *handlerextra;
 	/* ignore compatible check */
 	gboolean ignore_compatible;
+	/* skip bundle verification */
+	gboolean bundle_nosig;
 
 	/* for storing installation runtime informations */
 	RContextInstallationInfo *install_info;
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -312,6 +312,8 @@ static gboolean input_stream_read_bytes_all(GInputStream *stream,
 	return TRUE;
 }
 
+/* Appends signature to specified bundle file.
+ * Signature format: N bytes followed by uint64(N) */
 static gboolean sign_bundle(const gchar *bundlename, GError **error)
 {
 	GError *ierror = NULL;
@@ -320,22 +322,7 @@ static gboolean sign_bundle(const gchar *bundlename, GError **error)
 	g_autoptr(GFileOutputStream) bundlestream = NULL;
 	guint64 offset;
 
-	g_assert_nonnull(r_context()->certpath);
-	g_assert_nonnull(r_context()->keypath);
-
-	sig = cms_sign_file(bundlename,
-			r_context()->certpath,
-			r_context()->keypath,
-			r_context()->intermediatepaths,
-			&ierror);
-	if (sig == NULL) {
-		g_propagate_prefixed_error(
-				error,
-				ierror,
-				"failed signing bundle: ");
-		return FALSE;
-	}
-
+	/* open file and seek to end */
 	bundlefile = g_file_new_for_path(bundlename);
 	bundlestream = g_file_append_to(bundlefile, G_FILE_CREATE_NONE, NULL, &ierror);
 	if (bundlestream == NULL) {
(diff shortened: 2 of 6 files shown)


ejoerns:
Thanks for your contribution.

Please note the 4th paragraph of this chapter in the RAUC documentation: https://rauc.readthedocs.io/en/latest/basic.html#update-artifacts-bundles

> A key design decision of RAUC is that signing a bundle is mandatory. For development purpose, a self-signed certificate might be sufficient, for production the signing process should be integrated with your PKI infrastructure.

If you really really do not need any type of cryptography (because you are sitting on an undiscovered island e.g.) there is still the option of providing a simple self-signed certificate.

What is your use case for this? I really would like to avoid to soften the emphasis on security in updating..

Regards, Enrico

billpittman:
I saw the docs, thanks for pointing that out.  However for our purposes, the bundle signing really isn't necessary and having the overhead of yet another key to manage is somewhat pointless when we can just skip the signing part.

Our specific use case is that the rauc bundles are themselves bundled into an IPK which is part of a feed from a feed server.  The feed itself and each IPK is already signed, so signing a bundle inside a signed IPK on a signed server is a bit redundant.

I know we can use self signed certificates, however, they still need to be managed and since it is a key, and exposed to a customer, it has to go through the security protocols and management requirements, etc.  So it seems easier to not have to have it at all.

Thoughts ?

Cheers,
Bill

________________________________

jluebbe:
Always requiring a signature reduces the complexity of the (already complex enough) verification paths. So the additional complexity (and risk) in that cirtical code is not something i'd like to have in RAUC mainline.

Transporting RAUC bundles in ipkgs is not something I've seen before, so I'd like to understand it better. You're installing the ipkg into your current rootfs and then run rauc install on the bundle in your rootfs? Do you install other software to the rootfs via ipkgs? If so, how do you ensure that you get the exact intended system state?

billpittman:
We are using RAUC bundles to effect a system update.  The IPK contains the new version of software so the IPK is downloaded to the current rootfs and installed.  The RAUC bundle that is installed contains the bootloader, kernel, initramfs, and a fairly bare rootfs.

Do you install other software to the rootfs via IPKs ?
Yes most of the software is installed on the target via IPKs.

How to you ensure that you get the exact intended system state?
We use a package manager with associated dependencies, etc. to guarantee that the intended software is installed on the system.  We have a pay-to-play sort of philosophy on the target wrt software, so only the necessary software is installed and a fixed image, thus maximizing the space a user has available.

harisokanovic:
Hi @jluebbe , maybe it would help if I describe our systems a little more:

National Instruments manufactures a variety of embedded Linux products for test and measurement applications. Our systems are highly customizable. Typical use case for our customers involves remotely imaging these systems then installing, upgrading, or sometimes replacing various software components for their specific use cases. We cannot only distribute fixed-function images since the permutations are too numerous.

Software on our systems is therefore distributed in two parts:
(1) a base image containing enough software (kernel, drivers, network stack,
      management software, pubkeys) to remotely access the controller and
      software feeds, and
(2) a feed of additional userland software and newer base images which users
      can install/upgrade.

We like RAUC because it implements a fairly effective A/B imaging strategy for power loss safe updates, and does so atop raw UEFI instead of a secondary boot loader. We're particularly impressed with it's test boot function, which give newly installed base images a chance to "mark good" before committing to it. We use this to safely migrate user configuration from one image to another, and abort upgrades if it does not apply leaving the system unchanged.

We don't need the bundle signing functions of RAUC because our software feeds are already signed to ensure integrity. A second layer of authentication adds unnecessary complexity. It would requires us and possibly some of our more advanced users to manage a second set of keys in addition to feed keys.

The implementation Bill proposed (`--no-signatures` cli arg) is just one option. We are amenable to other implementation if needed, e.g. a conf file token.

jluebbe:
So you have A and B rootfs and no additional A/B "application" partition? The additional software components are "reinstalled" in the active rootfs after performing a RAUC update?

I still think that this is a very unusual setup compared to the use-cases RAUC is targeted at (where it is the primary update tool). So i'm not really happy to change one of the basic design decisions to support this and maintain it. If you change the patch to remove signing completely, it might be easy enough to carry locally for your case?

More generally, I think using something like ostree (or even flatpak) to manage optional software components at runtime (where no bootloader level fallback is required). This is much more robust against corruption (for example from buggy pre/postinst scripts), can be checked/repaired without reinstallation and the result is guaranteed to be independent of the sequence of updates (like skipping versions). Accordingly it reduces the possible SW combinations in the field and so the testing matrix as well. ;)

harisokanovic:
> So you have A and B rootfs and no additional A/B "application" partition? The additional software components are "reinstalled" in the active rootfs after performing a RAUC update?

The root filesystem is an overlayfs composed of a read-only squashfs base image and an initially empty, writable, folder on a separate ext4 partition. Changing the base image resets the overlay, optionally migrating select system configurations if directed by users.


> I still think that this is a very unusual setup compared to the use-cases RAUC is targeted at (where it is the primary update tool). So i'm not really happy to change one of the basic design decisions to support this and maintain it. If you change the patch to remove signing completely, it might be easy enough to carry locally for your case?

I don't follow. What basic design is violated by moving the signing process? RAUC still manages base images on our systems the same way is does on every other system.


> More generally, I think using something like ostree (or even flatpak) to manage optional software components at runtime (where no bootloader level fallback is required). This is much more robust against corruption (for example from buggy pre/postinst scripts), can be checked/repaired without reinstallation and the result is guaranteed to be independent of the sequence of updates (like skipping versions). Accordingly it reduces the possible SW combinations in the field and so the testing matrix as well. ;)

As I mentioned earlier, our users expect a high degree of software customization. We cannot simply install read-only images on our systems.

jluebbe:
I'm sorry for the long silence. I've been thinking about this for some time and won't merge it in the current form. The signing code which interfaces with OpenSSL is already complex and the most critical part for the security of RAUC, so additional complexity / code paths need to be relevant for most use cases.

There are some requests for supporting multiple required (N-of-M or specific) signatures and also ideas to support incremental hashing and decryption (via dm-verity/dm-crypt). That would require larger changes to the signature code anyway, so it might be possible to have no-signature support without additional complexity.

harisokanovic:
That's unfortunate. I hope you'll consider our use case when these signature changes happen. Let us know if we can help. As I said before, we like other functionality in RAUC and will continue using it in some form. There's no point in reproducing that functionality in another tool over such a trivial thing.

PR #519, opened 2019-12-08
bundle: fix result values for error paths

Emantor:
The res variable can be set to TRUE by one of the preceding function
calls. Explicitly set the variable to FALSE for error paths.

Signed-off-by: Rouven Czerwinski <rouven@czerwinskis.de>

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -800,6 +800,7 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 					R_SIGNATURE_ERROR,
 					R_SIGNATURE_ERROR_X509_NEW,
 					"failed to allocate new X509 store");
+			res = FALSE;
 			goto out;
 		}
 		if (!X509_STORE_load_locations(store, load_capath, load_cadir)) {
@@ -808,7 +809,7 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 					R_SIGNATURE_ERROR,
 					R_SIGNATURE_ERROR_CA_LOAD,
 					"failed to load CA file '%s' and/or directory '%s'", load_capath, load_cadir);
-
+			res = FALSE;
 			goto out;
 		}
 
--- a/test/bundle.c
+++ b/test/bundle.c
@@ -222,6 +222,18 @@ static void bundle_test_resign(BundleFixture *fixture,
 	g_clear_pointer(&bundle, free_bundle);
 }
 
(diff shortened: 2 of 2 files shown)


src/bundle.c
```diff
@@ -729,6 +729,7 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 				error,
 				ierror,
 				"Failed to seek to end of bundle: ");
+		res = FALSE;
```
OnkelUlla: The surrounding if-clause just proved that `res` is already `FALSE`, so why set it again here?

src/bundle.c
```diff
@@ -740,6 +741,7 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 				error,
 				ierror,
 				"Failed to read signature size from bundle: ");
+		res = FALSE;
```
OnkelUlla: Same as above.

src/bundle.c
```diff
@@ -775,6 +777,7 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 				error,
 				ierror,
 				"Failed to seek to start of bundle signature: ");
+		res = FALSE;
```
OnkelUlla: Same as above.

src/bundle.c
```diff
@@ -785,6 +788,7 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 				error,
 				ierror,
 				"Failed to read signature from bundle: ");
+		res = FALSE;
```
OnkelUlla: Same as above.

src/bundle.c
```diff
@@ -800,6 +804,7 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 					R_SIGNATURE_ERROR,
 					R_SIGNATURE_ERROR_X509_NEW,
 					"failed to allocate new X509 store");
+			res = FALSE;
```
OnkelUlla: `res` is `TRUE` here as otherwise we wouldn't have reached this line. The same is true for the following hunk - good catch!

src/bundle.c
Emantor: Indeed, I'll amend.

src/bundle.c
Emantor: Thanks. Updated the PR, please take a look again.

PR #522, opened 2020-01-06
Fix and improve error handling in network code

ejoerns:
This fixes error handling in transfer() function and improves (simplifies) reporting bundle downloads that exceed the maximum size.

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -18,6 +18,8 @@ typedef struct {
 
 	size_t pos;
 	size_t limit;
+
+	gchar *err;
 } RaucTransfer;
 
 gboolean network_init(GError **error)
@@ -42,8 +44,10 @@ static size_t write_cb(char *ptr, size_t size, size_t nmemb, void *userdata)
 
 	/* check transfer limit */
 	if (xfer->limit) {
-		if ((xfer->pos + size*nmemb) > xfer->limit)
+		if ((xfer->pos + size*nmemb) > xfer->limit) {
+			xfer->err = g_strdup("Maximum bundle download size exceeded. Download aborted.");
 			return 0;
+		}
 	}
 
 	res = fwrite(ptr, size, nmemb, xfer->dl);
@@ -59,10 +63,11 @@ static int xfer_cb(void *clientp, curl_off_t dltotal, curl_off_t dlnow,
 
 	/* check transfer limit */
 	if (xfer->limit) {
-		if (dltotal > (curl_off_t)xfer->limit)
+		if ((dlnow > (curl_off_t)xfer->limit)
+		    || (dltotal > (curl_off_t)xfer->limit)) {
+			xfer->err = g_strdup("Maximum bundle download size exceeded. Download aborted.");
 			return 1;
-		if (dlnow > (curl_off_t)xfer->limit)
-			return 1;
+		}
 	}
 
 	return 0;
@@ -111,10 +116,14 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 		goto out;
 	} else if (r != CURLE_OK) {
 		size_t len = strlen(errbuf);
-		if (len)
+		if (xfer->err) {
+			g_set_error(error, G_IO_ERROR, G_IO_ERROR_FAILED, "Transfer failed: %s", xfer->err);
+			g_clear_pointer(&xfer->err, &g_free);
+		} else if (len) {
 			g_set_error(error, G_IO_ERROR, G_IO_ERROR_FAILED, "Transfer failed: %s%s", errbuf, ((errbuf[len - 1] != '\n') ? "\n" : ""));
-		else
+		} else {
 			g_set_error(error, G_IO_ERROR, G_IO_ERROR_FAILED, "Transfer failed: %s", curl_easy_strerror(r));
+		}
 		goto out;
 	}
 	res = TRUE;
--- a/src/network.c
+++ b/src/network.c
@@ -7,6 +7,9 @@
 
 #include "network.h"
 
+/* We need to make sure that we can support large bundles. */
+G_STATIC_ASSERT(sizeof(curl_off_t) == 8);
+
 typedef struct {
 	const gchar *url;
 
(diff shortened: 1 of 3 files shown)


src/network.c
```diff
@@ -87,12 +54,9 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 	//curl_easy_setopt(curl, CURLOPT_MAX_RECV_SPEED_LARGE, 1048576L); /* bytes per second */
 	//curl_easy_setopt(curl,  CURLOPT_LOW_SPEED_LIMIT, 1024L);
 	//curl_easy_setopt(curl,  CURLOPT_LOW_SPEED_TIME, 60L);
-	curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_cb);
-	curl_easy_setopt(curl, CURLOPT_WRITEDATA, xfer);
-	curl_easy_setopt(curl, CURLOPT_XFERINFOFUNCTION, xfer_cb);
-	curl_easy_setopt(curl, CURLOPT_XFERINFODATA, xfer);
 	curl_easy_setopt(curl, CURLOPT_FAILONERROR, xfer);
 	curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, errbuf);
+	curl_easy_setopt(curl, CURLOPT_MAXFILESIZE, xfer->limit);
```
jluebbe: See https://curl.haxx.se/libcurl/c/CURLOPT_MAXFILESIZE.html:
> The file size is not always known prior to download, and for such files this option has no effect even if the file transfer ends up being larger than this given limit. This concerns both FTP and HTTP transfers.

So we still need to track the actual received file size, but just report errors explicitly, maybe via a flag in xfer.

`CURLOPT_MAXFILESIZE` (or `CURLOPT_MAXFILESIZE_LARGE` since curl 7.11 in 2004) would still be useful for early aborts if the server reports the size.

ejoerns: I've pushed an update that does not remove xfer callbacks anymore, but adds basic error message reporting instead.

jluebbe: We should use https://curl.haxx.se/libcurl/c/CURLOPT_MAXFILESIZE_LARGE.html here and change the type of xfer->limit accordingly. This would also get rid of some casts in the following patch as well removing the limit on 32bit systems.

ejoerns:
I've added required rework to let network code to remove 32 bit limits.

src/network.c
```diff
@@ -93,6 +93,7 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 	curl_easy_setopt(curl, CURLOPT_XFERINFODATA, xfer);
 	curl_easy_setopt(curl, CURLOPT_FAILONERROR, xfer);
 	curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, errbuf);
+	curl_easy_setopt(curl, CURLOPT_MAXFILESIZE_LARGE, xfer->limit);
```
jluebbe: Do we need to cast to `curl_off_t` here?

src/network.c
```diff
@@ -16,8 +16,8 @@ typedef struct {
 	FILE *dl;
 	size_t dl_size;
 
-	size_t pos;
-	size_t limit;
+	off_t pos;
+	off_t limit;
```
jluebbe: Why do you used off_t here (instead of goffset or curl_off_t)? And why not for dl_size as well?

ejoerns: Mainly because we had size_t before. For `limit` which is passed to `curl_easy_setopt()` probably curl_off_t is the better one. Than we could use this for pos, too, to keep it consistent.

src/network.c
ejoerns: If it was not `curl_off_t` before, we could do it for making it explicit. If it was `curl_off_t` before, than not ;)

src/network.c
ejoerns: `dl_size` is only passed to `open_memstream()` which accepts a `size_t` argument...

ejoerns: maybe the right point to drop `download_mem()` which is not used anywhere else than in tests?

jluebbe: > maybe the right point to drop `download_mem()` which is not used anywhere else than in tests?

Probably yes, well likely need more control for downloading to memory when/if we implement dm-verity-based bundles, but that might be separate anyway.

ejoerns: I've now pushed a version with `curl_off_t` and `download_mem()` removed. The second force push is only a rebase onto current master.

ejoerns:
I don't like testing...

ejoerns:
> I don't like testing...

I've pushed an update to fix casting error on 32 bit.

ejoerns:
And now rebased onto master to fix conflict with just-merged #523

ejoerns:
> Because I did not have a 32-bit system at hand my suggestion is not compile tested. :-/
> If you do not follow the suggestion and leave `pos` and `limit` defined as `curl_off_t`, then you should remove the ineffective type cast to `curl_off_t` in the comparisons in `xfer_cb()`.

I'd go for removing the cast as we decided to consistently use 'offset' types for all variables that handle file sizes / offsets or similar things.

ejoerns:
I've now removed the `curl_off_t` casts.

src/network.c
```diff
@@ -7,17 +7,18 @@
 
 #include "network.h"
 
+/* We need to make sure that we can support large bundles. */
+G_STATIC_ASSERT(sizeof(curl_off_t) == 8);
```
jluebbe: @ejoerns Does this look good to you? I wanted to make extra sure that we don't break things with the new cases.

jluebbe:
@ejoerns I've also removed `download_file_checksum` (which is unused as well) and move the download error test cases to download_file instead of removing them.

src/network.c
ejoerns: Only a bit paranoid but it should not make things worse and in case this ever ever changes, we will get an early compile-time abort, this is ok.

ejoerns:
Ok, this is what the 'update branch' button does. I still prefer a rebase here that I just pushed.

ejoerns:
As I cannot self-approve my PR I just note here that I approve your changes @jluebbe and will probably force-merge it as soon as Travis has completed.

jluebbe:
> Ok, this is what the 'update branch' button does. I still prefer a rebase here that I just pushed.
Same. GitLab supports a automatic rebase.

PR #523, opened 2020-01-08
A handful of fixes for src/network.c

OnkelUlla:
These lines caught my eyes while reviewing pull request #522.

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -24,7 +24,7 @@ gboolean network_init(GError **error)
 {
 	CURLcode res;
 
-	g_return_val_if_fail(error == FALSE || *error == NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
 	res = curl_global_init(CURL_GLOBAL_ALL);
 	if (res != CURLE_OK) {
--- a/src/network.c
+++ b/src/network.c
@@ -75,6 +75,9 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 	char errbuf[CURL_ERROR_SIZE];
 	gboolean res = FALSE;
 
+	g_return_val_if_fail(xfer, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
 	curl = curl_easy_init();
 	if (curl == NULL)
 		goto out;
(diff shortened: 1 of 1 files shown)


src/network.c
```diff
@@ -75,6 +75,8 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 	char errbuf[CURL_ERROR_SIZE];
 	gboolean res = FALSE;
 
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
ejoerns: If we add an input check for `error`, we should add one for `xfer`, too, to remain consistent.

``` diff
+	g_return_val_if_fail(xfer, FALSE);
```

src/network.c
```diff
@@ -129,6 +131,8 @@ gboolean download_file(const gchar *target, const gchar *url, gsize limit, GErro
 	gboolean res = FALSE;
 	GError *ierror = NULL;
 
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
ejoerns: Input checks for `target` and `url` should be added, too.

src/network.c
```diff
@@ -195,6 +199,8 @@ gboolean download_mem(GBytes **data, const gchar *url, gsize limit, GError **err
 	GError *ierror = NULL;
 	char *dl_data = NULL;
 
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
ejoerns: Input checks for data and url should be added, too.

src/network.c
```diff
@@ -104,15 +104,13 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 	r = curl_easy_perform(curl);
 	if (r == CURLE_HTTP_RETURNED_ERROR) {
 		g_set_error(error, G_IO_ERROR, G_IO_ERROR_FAILED, "HTTP returned >=400");
-		res = FALSE;
```
ejoerns: While checking if we could replace each goto's with a `curl_easy_cleanup(curl)` + direct return, I stumbled across line 82 where we have a `goto out` without setting `error` before. We should fix this, too, either here or in a follow-up PR.

OnkelUlla:
While already in the act of adding the requested sanity checks I also added them for `download_file_checksum()`.

src/network.c
OnkelUlla: Done.

src/network.c
OnkelUlla: Done.

src/network.c
OnkelUlla: Done.

src/network.c
OnkelUlla: Done.

src/network.c
ejoerns: The check for `data` should be extended to:
``` c
g_return_val_if_fail(data && *data == NULL, FALSE);
```

OnkelUlla: Done - I left that out intentionally as there are no explicit guarantees defined anywhere with respect to `*data`.

ejoerns:
I've rebased onto master to fix trivial conflict with merged PR #503

PR #526, opened 2020-01-17
use glib's mount monitoring to update mount info on changes

ejoerns:
This allows us to move from a pull mechanism that requires re-reading all mounts when we need recent information to a push mechanism allowing us to read information once and only update them in case of being notified about changes.

### DIFF
--- a/include/install.h
+++ b/include/install.h
@@ -123,3 +123,8 @@ gboolean install_run(RaucInstallArgs *args);
  *         or NULL if an error occurred
  */
 GList* get_install_images(const RaucManifest *manifest, GHashTable *target_group, GError **error);
+
+/**
+ * Iterates over slots and updates their mount information
+ */
+void update_mount_info(void);
--- a/src/install.c
+++ b/src/install.c
@@ -98,8 +98,18 @@ gboolean determine_slot_states(GError **error)
 	mountlist = g_unix_mounts_get(NULL);
 	for (GList *l = mountlist; l != NULL; l = l->next) {
 		GUnixMountEntry *m = (GUnixMountEntry*)l->data;
-		g_autofree gchar *devicepath = resolve_loop_device(g_unix_mount_get_device_path(m));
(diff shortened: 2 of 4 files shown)


ejoerns:
Service tests and other fail as `determine_slot_state()` is now called during service startup and fail because the booted slot cannot be determined (which is fully valid for the incomplete test configuration).

The main issue is now that the `/service/info` test calls the `Info` D-Bus interface for getting bundle information. For this use case, it is absolutely irrelevant if the service startup fails or not. But for a normal service it does not make sense to start if we already now that we cannot even determine the booted slot.

The open question is now how to deal with these two cases. I do not see any good way of using the D-Bus interface for `Info` method on non-target systems.
Would it be a valid approach to not use the D-Bus interface upon `rauc info` calls and leave the D-Bus interface for target use cases where we are sure to have a 'configured' rauc service?

jluebbe:
> Service tests and other fail as `determine_slot_state()` is now called during service startup and fail because the booted slot cannot be determined (which is fully valid for the incomplete test configuration).

I think if we use the service in the tests, we need to provide a proper configuration (just like on the target), in the other cases, can we call the info methods directly?

ejoerns:
> > Service tests and other fail as `determine_slot_state()` is now called during service startup and fail because the booted slot cannot be determined (which is fully valid for the incomplete test configuration).
> 
> I think if we use the service in the tests, we need to provide a proper configuration (just like on the target), in the other cases, can we call the info methods directly?

It is not the question how to handle this for testing, the main point is how to handle this for D-Bus API users (and the test is one of them).

`rauc info` does *not* use the D-Bus API and also does a lot more (info, certchain dumping etc.) than the D-Bus 'API' provides.

Thus we agree on using the D-Bus API is supported only on targets, right?

ejoerns:
Rebased onto current master.

jluebbe:
We discussed this internally: We don't want to carry the additional complexity for a performance optimization that is not significant in practice.

PR #528, opened 2020-01-17
reduce rauc binary size

Villemoes:
This adds some configure knobs that allows one to disable parts of rauc that are not useful on the target device, such as bundle creation or resigning. To get the full benefit of that, add compiler and linker options to discard unused functions and data at link time. Just to get an idea of the effect, I've built each patch in turn with of `--disable-service --disable-create` (the latter of course only has an effect later in the series). That gives

```
$ size rauc.bin.*
   text    data     bss     dec     hex filename
 214582    8080    1736  224398   36c8e rauc.bin.0
 214582    8080    1736  224398   36c8e rauc.bin.1
 174336    5000    1672  181008   2c310 rauc.bin.2
 174368    5000    1736  181104   2c370 rauc.bin.3
 156204    4512    1704  162420   27a74 rauc.bin.4
 156172    4512    1704  162388   27a54 rauc.bin.5
```
so altogether this cuts a bit over a quarter of the binary's size - but of course the actual savings will vary with compiler and not least target architecture.

rauc pulls in quite a few shared library dependencies, so 62K may not seem that significant. However, 
I'm looking at a target with a rootfs budget of about 50M, so I need to cut where I can. Also, we can add more knobs in the future which would allow removing even more code.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -49,6 +49,15 @@ AS_IF([test "x$enable_service" != "xno"], [
 	AC_DEFINE([ENABLE_SERVICE], [0])
 ])
 
+AC_ARG_ENABLE([create],
+	AS_HELP_STRING([--disable-create], [Disable bundle creation and modification commands])
+)
+AS_IF([test "x$enable_create" != "xno"], [
+	AC_DEFINE([ENABLE_CREATE], [1], [Define to 1 to enable bundle, resign and convert commands])
+], [
+	AC_DEFINE([ENABLE_CREATE], [0])
+])
+
 # Checks for libraries.
 PKG_CHECK_MODULES([GLIB], [glib-2.0 >= 2.45.8 gio-2.0 gio-unix-2.0])
 
--- a/src/main.c
+++ b/src/main.c
@@ -282,6 +282,7 @@ out:
 	return TRUE;
 }
 
+G_GNUC_UNUSED
 static gboolean bundle_start(int argc, char **argv)
 {
 	GError *ierror = NULL;
@@ -445,6 +446,7 @@ out:
 	return TRUE;
 }
 
+G_GNUC_UNUSED
 static gboolean resign_start(int argc, char **argv)
 {
 	g_autoptr(RaucBundle) bundle = NULL;
@@ -540,6 +542,7 @@ out:
 	return TRUE;
 }
 
+G_GNUC_UNUSED
 static gboolean convert_start(int argc, char **argv)
 {
 	RaucBundle *bundle = NULL;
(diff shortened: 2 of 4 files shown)


ejoerns:
@Villemoes I like and agree with most of the changes you did and see that it is possibly not required to have all the functionality of a host RAUC compiled for the target, especially when space is that limited.
From my point of view, only the split up in such fine-grained options is a bit too much. Maybe we should use a generic option to disable this more or less 'host' features instead?

If I get your numbers right, disabling the bootloader supports only has little influence on the overall size. Thus I am not really sure if it's worth adding options for this at all.

@jluebbe What do you think about this?

include/utils.h
```diff
@@ -16,6 +16,7 @@ void close_preserve_errno(filedesc fd);
 G_DEFINE_AUTO_CLEANUP_FREE_FUNC(filedesc, close_preserve_errno, -1)
 
 #define R_LOG_DOMAIN_SUBPROCESS "rauc-subprocess"
+#define __maybe_unused __attribute__((__unused__))
```
jluebbe: GLib already has a G_GNUC_UNUSED macro, looks like it would be the same.

configure.ac
```diff
@@ -49,6 +49,33 @@ AS_IF([test "x$enable_service" != "xno"], [
 	AC_DEFINE([ENABLE_SERVICE], [0])
 ])
 
+AC_ARG_ENABLE([bundle],
+	AS_HELP_STRING([--disable-bundle], [Disable bundle command])
```
jluebbe: Just call it `--disable-create Disable bundle creation and modification` and control all three subcomands together.

configure.ac
```diff
@@ -76,6 +76,42 @@ AS_IF([test "x$enable_convert" != "xno"], [
 	AC_DEFINE([ENABLE_CONVERT], [0])
 ])
 
+AC_ARG_ENABLE([barebox],
+	AS_HELP_STRING([--disable-barebox], [Disable support for barebox bootloader])
```
jluebbe: The useful cases are probably:

- all bootloader
- a specific bootloader
- no bootloader

Can we handle that nicely in configure? If it's not obvious, maybe drop this change from this PR, so we can look at it separately.

src/main.c
```diff
@@ -1812,7 +1813,9 @@ static void cmdline_handler(int argc, char **argv)
 #if ENABLE_CONVERT == 1
 		{CONVERT, "convert", "convert <INBUNDLE> <OUTBUNDLE>", "Convert to casync index bundle and store", convert_start, convert_group, FALSE},
 #endif
+#if ENABLE_CHECKSUM == 1
 		{CHECKSUM, "checksum", "checksum <DIRECTORY>", "Deprecated", checksum_start, NULL, FALSE},
```
jluebbe: This has been marked as deprecated since 2017. We should probably just remove it completely.

@ejoerns?

ejoerns: -@jluebbe Could not find out when we actually decided to mark this deprecated, but it is neither documented nor required and was only useful for the already-removed network mode I guess..

So just remove it.

include/utils.h
Villemoes: Thanks, I'll see if I can just use that instead.

configure.ac
Villemoes: OK.

configure.ac
Villemoes: I'm not very familiar with autoconf, but perhaps there's some standard idiom that would allow a `--enable-bootloaders=barebox,u-boot`, with `all` a keyword with the obvious meaning and being default. That would at least reduce the ./configure UI complexity. I'll take a closer look and drop it for the time being if there's no reasonably simple solution.

src/main.c
Villemoes: Yeah, I also tried to figure out when and why it was deprecated. If no-one beats me to it, I'll just send a preliminary PR that just removes it, so this one doesn't have to deal with it at all.

configure.ac
jluebbe: @michaelolbrich Do you know something that would help here?

michaelolbrich: `--enable-bootloaders=` is probably the best from a user POV, but I'm not aware of any helpers to implement this.
And I'd also accept `yes`/`no` for all or nothing as we..

jluebbe: `no` would only be useful for the host, and we don't really need to save space there. So maybe drop it from this PR to get the rest merged, @Villemoes?

Villemoes: Yeah, that was what I was planning to do anyway in order not to conflict with the "custom bootloader backend" proposal. I updated my branch locally last night, will rebase and push a new one once the 'rm-checksum' PR is merged shortly.

PR #530, opened 2020-01-29
src/bundle: fix usage of g_strndup with strlen()

ejoerns:
strlen() can potentially return values < 6 for emtpy or really short strings.
If we now substract the value of 6 without further checking, we end up
with a wrap of the unsigned size_t that effectively results in a very
very large value in size_t that will likely cause the memory allocation
in g_strndup() to fail.

Fix this by checking for string size being > 6 before and otherwise
always return 0.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -498,16 +498,18 @@ static gboolean convert_to_casync_bundle(RaucBundle *bundle, const gchar *outbun
 	g_autofree gchar *contentdir = NULL;
 	g_autofree gchar *mfpath = NULL;
 	g_autofree gchar *storepath = NULL;
-	gchar *basepath = NULL;
 	g_autoptr(RaucManifest) manifest = NULL;
 
 	g_return_val_if_fail(bundle, FALSE);
 	g_return_val_if_fail(outbundle, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
-	basepath = g_strndup(outbundle, strlen(outbundle) - 6);
-	storepath = g_strconcat(basepath, ".castr", NULL);
-	g_free(basepath);
+	if (g_str_has_suffix(outbundle, ".raucb")) {
+		g_autofree gchar *basepath = g_strndup(outbundle, strlen(outbundle) - 6);
+		storepath = g_strconcat(basepath, ".castr", NULL);
+	} else {
+		storepath = g_strconcat(outbundle, ".castr", NULL);
+	}
 
 	/* Assure bundle destination path doe not already exist */
 	if (g_file_test(outbundle, G_FILE_TEST_EXISTS)) {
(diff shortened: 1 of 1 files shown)


src/bundle.c
```diff
@@ -505,7 +505,7 @@ static gboolean convert_to_casync_bundle(RaucBundle *bundle, const gchar *outbun
 	g_return_val_if_fail(outbundle, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
-	basepath = g_strndup(outbundle, strlen(outbundle) - 6);
+	basepath = g_strndup(outbundle, strlen(outbundle) > 6 ? strlen(outbundle) - 6 : 0);
```
OnkelUlla: strlen()'s second argument can be rearranged to be `strlen(outbundle) - 6 > 0 ? strlen(outbundle) - 6 : 0` revealing that it can be rearranged again to be `MAX(strlen(outbundle) - 6, 0)` which is readable more easily.
The number `6` presumably is the result of `strlen(".raucb")`, right? Hence I suggest to finally use the expression `MAX(strlen(outbundle) - strlen(".raucb"), 0)` to further enhance the code's understandability.
By the way: there is no check anywhere that `outbundle` really ends in ".raucb" which presumably is why you started this pull request. What about ensuring that instead of the current approach?

src/bundle.c
```diff
@@ -685,9 +685,9 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 		gchar *strprfx;
 
 		if (ibundle->origpath)
-			strprfx = g_strndup(ibundle->origpath, strlen(ibundle->origpath) - 6);
+			strprfx = g_strndup(ibundle->origpath, strlen(ibundle->origpath) > 6 ? strlen(ibundle->origpath) - 6 : 0);
```
OnkelUlla: See above.

src/bundle.c
```diff
@@ -685,9 +685,9 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
...
 		else
-			strprfx = g_strndup(ibundle->path, strlen(ibundle->path) - 6);
+			strprfx = g_strndup(ibundle->path, strlen(ibundle->path) > 6 ? strlen(ibundle->path) - 6 : 0);
```
OnkelUlla: See above.

src/bundle.c
```diff
@@ -505,7 +505,7 @@ static gboolean convert_to_casync_bundle(RaucBundle *bundle, const gchar *outbun
 	g_return_val_if_fail(outbundle, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
-	basepath = g_strndup(outbundle, strlen(outbundle) - 6);
+	basepath = g_strndup(outbundle, strlen(outbundle) > 6 ? strlen(outbundle) - 6 : 0);
 	storepath = g_strconcat(basepath, ".castr", NULL);
```
a3f: This can lead to non-intuitive results:

`outbundle = "bundle"` ⇒ `basepath = ""` ⇒ `storepath = ".castr"`
`outbundle = "bundle12"`  ⇒ `basepath = "bu"` ⇒ `storepath = "bu.castr"`

src/bundle.c
ejoerns: Thanks for your suggestions. It is not a requirement anymore that bundles have to end with `.raucb`. Thus we cannot check for this anymore.
Maybe we need a cleaner solution than my hack. It is a result of testing and I mainly did not want it to get lost ;)

src/bundle.c
jluebbe: So we probably want to use if (g_str_has_suffix(outbundle, ".raucb")) and only then cut of the suffix.

ejoerns: yes, that sounds pretty much to what I already had in mind. Did not check if it works in the entire logic, but should be close to the way to go.

ejoerns: I've now reworked it as suggested.

src/bundle.c
```diff
@@ -682,15 +684,20 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 	if (r_context()->config->store_path) {
 		ibundle->storepath = r_context()->config->store_path;
 	} else {
-		gchar *strprfx;
+		gchar *path;
```
OnkelUlla: What about simplifying lines 687-692 via the elvis operator: `gchar *path = ibundle->origpath ?: ibundle->path;`?

ejoerns: A little less conversation, a little more action -> fixed it.

PR #532, opened 2020-02-05
explicitly expose bundle mountpoint to hooks

Villemoes:
I was trying to figure out where I could find the mounted bundle from within my post-install hook. I think it's already possible to deduce, but I also think it's better to just provide the path explicitly.


src/update_handler.c
```diff
@@ -684,6 +684,8 @@ static gboolean run_slot_hook(const gchar *hook_name, const gchar *hook_cmd, Rau
 
 		g_subprocess_launcher_setenv(launcher, "RAUC_BUNDLE_SPKI_HASHES", string, FALSE);
 		g_free(string);
+
+		g_subprocess_launcher_setenv(launcher, "RAUC_UPDATE_SOURCE", bundle->mount_point, TRUE);
```
ejoerns: As we have `RAUC_SLOT_MOUNT_POINT` already, what about using `RAUC_BUNDLE_MOUNT_POINT` here instead?
Otherwise, I endorse the approach.

Villemoes: Well, I agree that RAUC_UPDATE_SOURCE isn't the best name, and RAUC_BUNDLE_MOUNT_POINT is more obvious. But RAUC_UPDATE_SOURCE isn't my invention, that's the name used for handlers, so I thought it was best to expose the same piece of data under the same name. But if RAUC_BUNDLE_MOUNT_POINT is preferred, I guess we could use that here, and then also expose it under that name to handlers (updating the documentation, but keeping defining RAUC_UPDATE_SOURCE as well for compatibility with existing setups). Either way works for me - but ending up using different names in the handler vs. hook cases is a bad idea, IMO.

OnkelUlla: +1 to using/adding `RAUC_BUNDLE_MOUNT_POINT` while keeping `RAUC_UPDATE_SOURCE` for backwards compatibility and explicitely stating that the latter is regarded as being obsolete.

jluebbe: Yes, `RAUC_BUNDLE_MOUNT_POINT` and `RAUC_SLOT_MOUNT_POINT` is much better, let's just deprecate `RAUC_UPDATE_SOURCE`.

Villemoes: Updated and force-pushed a new version with a first patch replacing/deprecating RAUC_UPDATE_SOURCE.

jluebbe:
I've rebased it on master and changed the commit subjects a bit for consistency.

PR #535, opened 2020-02-06
remove surrounding whitespace from system variant

jluebbe:
This was always intended to be a short string without whitespace. Remove the whitespace from when preparing the context. In the manifest, the variant is part of the image section name, so intentional surrounding whitespace would be very unusual.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -263,3 +263,20 @@ gchar * r_realpath(const gchar *path)
 
 	return g_strdup(rpath);
 }
+
+gboolean r_whitespace_removed(gchar *str)
+{
+	gsize len;
+
+	if (str == NULL)
+		return FALSE;
+
+	len = strlen(str);
+
+	if (len == 0)
+		return FALSE;
+
+	g_strstrip(str);
+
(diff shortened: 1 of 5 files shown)


src/utils.c
```diff
@@ -263,3 +263,21 @@ gchar * r_realpath(const gchar *path)
...
+
+	len = strlen(str);
+
+	if (len == 0)
+		return FALSE;
+
+	g_strstrip(str);
+
+	/* If the last char is now \0, some whitespace was removed. */
+	return str[len-1] == '\0';
```
OnkelUlla: `g_strstrip()` is implemented as a composition of `g_strchug()` and `g_strchomp()` and does not give any guarantees with respect to how the shortened string looks like compared to the original one. The test is not correct in general: looking at [`g_strchug()`'s current implementation](https://github.com/GNOME/glib/blob/mainline/glib/gstrfuncs.c#L2256) shows that applying `r_whitespace_removed()` to a string with more than one leading space is going to give the wrong result.

jluebbe: You're right. Premature optimization...

PR #537, opened 2020-02-13
Allow custom mksquashfs with --mksquashfs-args

### LINKED ISSUE #508: Configure compression level for squashfs
Hello,

I need to change compression level of squashfs bundle by using xz and some other options as specified from this link :
https://techoverflow.net/2013/01/11/creating-a-highly-compressed-squashfs-from-a-folder/

Currently, there is no way to specify custom arguments to mksquashfs invocation. For a simple POC, I changed g_subprocess_new call to add the required options. The newly generated bundle size is now ~300M, compared to the original ~390M. I think it may be an interesting gain for some use cases (limited storage space or download bandwidth for example). The only trick is to enabled CONFIG_SQUASHFS_XZ on the kernel.

To cover this, it could be useful to specify custom argument for mksquashfs on the manifest file. Can you please share with me your opinion on this feature ?

Psykar:
Closes #508

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -21,6 +21,7 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
 	g_autoptr(GSubprocess) sproc = NULL;
 	GError *ierror = NULL;
 	gboolean res = FALSE;
+	g_autoptr(GPtrArray) args = g_ptr_array_new_full(7, g_free);
 
 	r_context_begin_step("mksquashfs", "Creating squashfs", 0);
 
@@ -29,16 +30,35 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
 		goto out;
 	}
 
-	sproc = r_subprocess_new(G_SUBPROCESS_FLAGS_STDOUT_SILENCE,
-			&ierror, "mksquashfs",
-			contentdir,
-			bundlename,
-			"-all-root",
-			"-noappend",
-			"-no-progress",
-			"-no-xattrs",
-			NULL);
+	g_ptr_array_add(args, g_strdup("mksquashfs"));
+	g_ptr_array_add(args, g_strdup(contentdir));
+	g_ptr_array_add(args, g_strdup(bundlename));
+	g_ptr_array_add(args, g_strdup("-all-root"));
+	g_ptr_array_add(args, g_strdup("-noappend"));
+	g_ptr_array_add(args, g_strdup("-no-progress"));
+	g_ptr_array_add(args, g_strdup("-no-xattrs"));
+
+
+	if (r_context()->mksquashfs_args != NULL) {
+		g_auto(GStrv) mksquashfs_argvp = NULL;
+		res = g_shell_parse_argv(r_context()->mksquashfs_args, NULL, &mksquashfs_argvp, &ierror);
+		if (!res) {
+			g_propagate_prefixed_error(
+					error,
+					ierror,
+					"Failed to parse mksquashfs extra args: ");
+			goto out;
+		}
+		for (gchar **mksquashfs_args = mksquashfs_argvp; *mksquashfs_args != NULL; mksquashfs_args++) {
+			g_ptr_array_add(args, g_strdup(*mksquashfs_args));
+		}
+	}
+	g_ptr_array_add(args, NULL);
+
+	sproc = r_subprocess_newv(args, G_SUBPROCESS_FLAGS_STDOUT_SILENCE,
+			&ierror);
 	if (sproc == NULL) {
+		res = FALSE;
 		g_propagate_prefixed_error(
 				error,
 				ierror,
--- a/src/main.c
+++ b/src/main.c
@@ -30,6 +30,7 @@ gboolean info_noverify, info_dumpcert = FALSE;
 gboolean status_detailed = FALSE;
 gchar *output_format = NULL;
 gchar *signing_keyring = NULL;
+gchar *mksquashfs_args = NULL;
 
 static gchar* make_progress_line(gint percentage)
 {
@@ -1672,6 +1673,7 @@ static GOptionEntry entries_install[] = {
 
 static GOptionEntry entries_bundle[] = {
 	{"signing-keyring", '\0', 0, G_OPTION_ARG_FILENAME, &signing_keyring, "verification keyring file", "PEMFILE"},
+	{"mksquashfs-args", '\0', 0, G_OPTION_ARG_STRING, &mksquashfs_args, "mksquashfs extra args", "ARGS"},
 	{0}
 };
 
@@ -1899,6 +1901,8 @@ static void cmdline_handler(int argc, char **argv)
 			r_context_conf()->keyringpath = keyring;
 		if (signing_keyring)
 			r_context_conf()->signing_keyringpath = signing_keyring;
+		if (mksquashfs_args)
+			r_context_conf()->mksquashfs_args = mksquashfs_args;
 		if (intermediate)
 			r_context_conf()->intermediatepaths = intermediate;
 		if (mount)
--- a/include/context.h
+++ b/include/context.h
@@ -34,6 +34,7 @@ typedef struct {
 	gchar *keyringpath;
 	gchar *keyringdirectory;
 	gchar *signing_keyringpath;
+	gchar *mksquashfs_args;
 	gchar **intermediatepaths;
 	/* optional global mount prefix overwrite */
 	gchar *mountprefix;
--- a/src/context.c
+++ b/src/context.c
@@ -569,11 +569,13 @@ void r_context_clean(void)
 		g_free(context->keyringpath);
 		g_free(context->keyringdirectory);
 		g_free(context->signing_keyringpath);
+		g_free(context->mksquashfs_args);
 		context->certpath = NULL;
 		context->keypath = NULL;
 		context->keyringpath = NULL;
 		context->keyringdirectory = NULL;
 		context->signing_keyringpath = NULL;
+		context->mksquashfs_args = NULL;
 
 		if (context->config) {
 			context->config->keyring_path = NULL;
--- a/test/rauc.t
+++ b/test/rauc.t
@@ -232,6 +232,19 @@ test_expect_success "rauc bundle" "
   rm out.raucb
 "
 
+test_expect_success "rauc bundle mksquashfs extra args" "
+  rm -f out.raucb &&
+  rauc \
+    --cert $SHARNESS_TEST_DIRECTORY/openssl-ca/dev/autobuilder-1.cert.pem \
+    --key $SHARNESS_TEST_DIRECTORY/openssl-ca/dev/private/autobuilder-1.pem \
+    bundle \
+    --mksquashfs-args=\"-comp xz -info -progress\" \
+    $SHARNESS_TEST_DIRECTORY/install-content out.raucb &&
+  rauc -c $SHARNESS_TEST_DIRECTORY/test.conf info out.raucb &&
+  test -f out.raucb &&
+  rm out.raucb
+"
+
 test_expect_success PKCS11 "rauc bundle with PKCS11 (key 1)" "
   rm -f out.raucb &&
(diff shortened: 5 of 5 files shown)


src/bundle.c
```diff
@@ -29,15 +30,25 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
...
-			NULL);
+	g_ptr_array_add(args, g_strdup("mksquashfs"));
+	g_ptr_array_add(args, g_strdup(contentdir));
+	g_ptr_array_add(args, g_strdup(bundlename));
+	g_ptr_array_add(args, g_strdup("-all-root"));
+	g_ptr_array_add(args, g_strdup("-noappend"));
+	g_ptr_array_add(args, g_strdup("-no-progress"));
+	g_ptr_array_add(args, g_strdup("-no-xattrs"));
+
+	if (r_context()->squashfs_args != NULL) {
```
ejoerns: I first head to look up if the options args string is actually NULL-terminated, as I could not find any explicit mention, but this seems to be the case. For this, the rest of the RAUC code uses a more condensed way of iterating that I would propose here, too. And the if could be skipped, too.

```suggestion
      for (gchar **squashfs_args = r_context()->squashfs_args; *squashfs_args != NULL; squashfs_args++) {
              g_ptr_array_add(args, g_strdup(*squashfs_args));

```

src/main.c
```diff
@@ -1712,6 +1713,7 @@ GOptionEntry entries_install[] = {
 
 GOptionEntry entries_bundle[] = {
 	{"signing-keyring", '\0', 0, G_OPTION_ARG_FILENAME, &signing_keyring, "verification keyring file", "PEMFILE"},
+	{"squashfs-args", '\0', 0, G_OPTION_ARG_STRING_ARRAY, &squashfs_args, "squashfs extra args (repeatable)", ""},
```
ejoerns: All other options note a placeholder for the content to add, this should be done here, too for consistency reasons. Thus maybe

```suggestion
	{"squashfs-args", '\0', 0, G_OPTION_ARG_STRING_ARRAY, &squashfs_args, "squashfs extra args (repeatable)", "ARGS"},
```

jluebbe: How is this used for mksquashfs args which have an option?
`--squashfs-args "-processors 8"` or `--squashfs-args -processors --squashfs-args 8`

Often, programs expect that the shell has split their arguments into separate argv strings and fail when options values are part of the same argv string as the option itself.

glib has https://developer.gnome.org/glib/stable/glib-Shell-related-Utilities.html#g-shell-parse-argv, so an alternative would be to supply one string argument to rauc and let it split that up for the mksquashfs call. Then you could pass `--squashfs-args "-processors 8 -comp xz"` and it would work independent of clever handling in mksquashfs.

Psykar: > How is this used for mksquashfs args which have an option?
> `--squashfs-args "-processors 8"` or `--squashfs-args -processors --squashfs-args 8`
> 
> Often, programs expect that the shell has split their arguments into separate argv strings and fail when options values are part of the same argv string as the option itself.
> 
Generally: `--squashfs-args "-processors" --squashfs-args "8"` would work - I agree it's a bit ugly though.

> glib has https://developer.gnome.org/glib/stable/glib-Shell-related-Utilities.html#g-shell-parse-argv, so an alternative would be to supply one string argument to rauc and let it split that up for the mksquashfs call. Then you could pass `--squashfs-args "-processors 8 -comp xz"` and it would work independent of clever handling in mksquashfs.
I'll give it a go, agreed this would be cleaner - not particuarly familiar with glib, so thanks for pointing this out :)

Psykar:
> Good work, I assume some people waited for this as it allows, for example, changing default compressor etc (where I heard it might be a use case for some.)
Ironically in my case it's to turn off compression entirely - so we can do a binary diff between bundles and send only that, then reconstruct on the far side.
We're in an extremely bandwidth limited scenario. But yes various uses :)

jluebbe:
> Ironically in my case it's to turn off compression entirely - so we can do a binary diff between bundles and send only that, then reconstruct on the far side.
> We're in an extremely bandwidth limited scenario. But yes various uses :)

That's interesting! What do you use for binary diffs? In the medium term, I'd like to have delta-bundles supported natively by rauc. One would take two full bundles to generate a delta bundle (which would contain the per-slot deltas).

Psykar:
We've used http://www.daemonology.net/bsdiff/ with great results.

I'd love to get the diffs in natively, it's a much larger chunk of work though obviously :)

I hope to get the comments for this part of things addressed today though.

src/bundle.c
Psykar: Given the change to g_shell_parse_argv I couldn't see an easy way to get rid of this if statement after all, as it fails if the string passed is NULL.

src/main.c
Psykar: This worked a treat, thanks again. Updated PR.

Psykar:
All comments addressed. And I can actually run uncrustify locally now ;)

src/bundle.c
```diff
@@ -29,15 +32,32 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
...
+	g_ptr_array_add(args, g_strdup(contentdir));
+	g_ptr_array_add(args, g_strdup(bundlename));
+	g_ptr_array_add(args, g_strdup("-all-root"));
+	g_ptr_array_add(args, g_strdup("-noappend"));
+	g_ptr_array_add(args, g_strdup("-no-progress"));
+	g_ptr_array_add(args, g_strdup("-no-xattrs"));
+
+
+	if (r_context()->mksquashfs_args != NULL) {
+		res = g_shell_parse_argv(r_context()->mksquashfs_args, &mksquashfs_argcp, &mksquashfs_argvp, NULL);
```
OnkelUlla: You need to pass `ierror` as the last argument to `g_shell_parse_argv()` if you later on want to propagate potential errors upwards.

src/bundle.c
```diff
@@ -21,6 +21,9 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
 	g_autoptr(GSubprocess) sproc = NULL;
 	GError *ierror = NULL;
 	gboolean res = FALSE;
+	gint mksquashfs_argcp;
+	gchar **mksquashfs_argvp = NULL;
```
ejoerns: You could move this into the if body as it is only used there.

At the moment, this also leaks as it is not freed anywhere. The simplest would be to use glibs autofree helper here and replace the above by

```suggestion
	g_auto(GStrv) mksquashfs_argvp = NULL;
```

(see https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#GStrv)

src/bundle.c
```diff
@@ -29,15 +32,32 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
...
+	g_ptr_array_add(args, g_strdup("-no-xattrs"));
+
+
+	if (r_context()->mksquashfs_args != NULL) {
+		res = g_shell_parse_argv(r_context()->mksquashfs_args, &mksquashfs_argcp, &mksquashfs_argvp, NULL);
+		if (!res) {
+			g_propagate_prefixed_error(
+					error,
+					ierror,
+					"Failed parse mksquashfs args: ");
```
ejoerns: Maybe, to make it clearer
```suggestion
					"Failed parse mksquashfs extra args: ");
```

src/bundle.c
Psykar: Ah, my glib ignorance again. Interestingly I tested the error case locally and it seemed to propagate just fine, guessing it would have given a bit more context though.

src/bundle.c
Psykar: Hrm. Did this change but did notice this text
> This macro can be used to avoid having to do explicit cleanups of local variables when exiting functions. It often vastly simplifies handling of error conditions, removing the need for various tricks such as goto out or repeating of cleanup code. It is also helpful for non-error cases.

Does this mean the g_propagate_prefixed_error inside the if *shouldn't* use goto out?
Things appear to work as is, but I'm unclear on why the `goto out` is required to start with *shrug

ejoerns: It means you don't have to use `goto out` in many cases. The goto is only used for the central `r_context_end_step("mksquashfs", res);` here.  Potentially this could also be repeated in each goto and the goto then replaced by an instant return statement.

src/bundle.c
```diff
@@ -29,15 +30,34 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
...
+
+
+	if (r_context()->mksquashfs_args != NULL) {
+		gint mksquashfs_argcp;
+		g_auto(GStrv) mksquashfs_argvp = NULL;
+		res = g_shell_parse_argv(r_context()->mksquashfs_args, &mksquashfs_argcp, &mksquashfs_argvp, ierror);
+		if (!res) {
+			g_propagate_prefixed_error(
+					error,
+					&ierror,
```
OnkelUlla: The error propagation functions expect a `GError*` for the inner error, not a `GError**` (see [`g_propagate_prefixed_error()`'s documentation](https://developer.gnome.org/glib/stable/glib-Error-Reporting.html#g-propagate-prefixed-error)), so please replace `&ierror` with `ierror`.

src/bundle.c
```diff
@@ -29,15 +30,34 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
...
+					"Failed parse mksquashfs extra args: ");
```
OnkelUlla: Please replace _"Failed parse"_ with _"Failed **to** parse"_.

src/bundle.c
ejoerns: I've fixed it to shorten the loop a little.

src/bundle.c
ejoerns: Fixed.

ejoerns:
Missed that, but having `mksquashfs_argcp` was needles here and could be replaced by simply passing `NULL` as the argument. Fixed it.

src/bundle.c
```diff
@@ -29,15 +30,33 @@ static gboolean mksquashfs(const gchar *bundlename, const gchar *contentdir, GEr
...
+	g_ptr_array_add(args, g_strdup(bundlename));
+	g_ptr_array_add(args, g_strdup("-all-root"));
+	g_ptr_array_add(args, g_strdup("-noappend"));
+	g_ptr_array_add(args, g_strdup("-no-progress"));
+	g_ptr_array_add(args, g_strdup("-no-xattrs"));
+
+
+	if (r_context()->mksquashfs_args != NULL) {
+		g_auto(GStrv) mksquashfs_argvp = NULL;
+		res = g_shell_parse_argv(r_context()->mksquashfs_args, NULL, &mksquashfs_argvp, &ierror);
```
ejoerns: Oups... Consider the case we have `mkquashfs_args` set and they are parsed correctly.
This sets `res` to `TRUE`.
Now, in line 59, the `r_subprocess_newv()` call returns an error for whatever reason.
This will not modify `res` (as it sets `sproc` only) but lead us directly to the `out` jump label (with ierror set)
At the `out` label, we `return res` from the function.

At this point, the execution failed, `error` is set, but the function returns 'TRUE` that suggests a successful run!

ejoerns: Adding `res = FALSE;` in the error path of `g_subprocess_newv()` should to the trick. We have similar cases in other functions in the same module.

Psykar: Yep, makes sense. Updated :)

Psykar:
So coverage check is failing now. I did have a go at doing a test case for this, but it doesn't appear straight forward without refactoring test/bundle.c quite a bit to allow args to be set before calling `bundle_fixture_set_up_bundle`

If they're required to get this to land I'll look into it though?

src/bundle.c
OnkelUlla: @ejoerns: that was a good catch!

ejoerns:
I think this should satisfy uncrustify with manageable effort.

ejoerns:
> I think this should satisfy uncrustify with manageable effort.

satisfy *codecov* I meant

PR #541, opened 2020-02-18
implement custom bootloader backend

cbs101:
Hi

We ran into a issue where we could not rely on the standard bootchoser for our given bootloader.
in this specific case, we were using U-Boot, and could not use the U-Boot environment for rauc status information.
So made an addition to make it possible to point to a script (or program), that can choose how and where to perform the storing.

Proposed API for custom backend script access:

    API for custom backend script access:
    To obtain the primary slot
    <script> get-primary
    Returns a string that can be one of the bootnames specified in /etc/rauc/system.conf
    Return value is 0 if success, false if error
    
    To set the primary slot
    <script> set-primary <slot.bootname>
    <slot.bootname> can be one of the bootnames specified in /etc/rauc/system.conf
    Returns empty output
    Return value is 0 if success, false if error
    
    To obtain the state
    <script> get-state <slot.bootname>
    <slot.bootname> can be one of the bootnames specified in /etc/rauc/system.conf
    Returns a string that can be either "good" or "bad"
    Return value is 0 if success, false if error
    
    To set the state
    <script> set-state <slot.bootname> <state>
    <slot.bootname> can be one of the bootnames specified in /etc/rauc/system.conf
    <state> can be either "good" or "bad"
    Returns empty output
    Return value is 0 if success, false if error


ejoerns:
@cbs101 I would really appreciate if this is the last PR you open for this topic!

In case you are unfamiliar with GitHub PR workflow: You can simply force-push to you branch that you used to create the PR from and the original PR will be updated with the new content.

There is really no need to open a new PR for each change.

Thanks and regards, Enrico

cbs101:
@ejoerns : 
Noted. I'll skip doing that then. Apologies!
I wanted to force a re-test of the tests after fixing the problems found and force pushing, but no tests were triggered. Do I need to do something specific when force pushing ? I just did the changes needed and included that in the commits.

Not related to the above:
I realized that I have to add tests to the new handles. On that now.

src/bootchooser.c
```diff
@@ -1375,6 +1375,206 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+	if (!g_subprocess_communicate(sub, NULL, NULL, &stdout_buf, NULL, &ierror)) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"Failed to run : %s", script_name);
+		return FALSE;
+	}
+	data = g_bytes_get_data(stdout_buf, &size);
+	// Drop terminating character
+	*ret_str = g_string_new_len(data, size - 1);
```
Villemoes: This seems to assume the output is always terminated with exactly one newline. Perhaps just trim any leading and trailing whitespace (assuming there's a simple g_* helper for that).

src/bootchooser.c
```diff
@@ -1375,6 +1375,206 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+	g_return_val_if_fail(slot, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (!custom_backend_set(cmd, slot->bootname, state, &ierror) ) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
+	return TRUE;
+}
```
Villemoes: Can one drop the local ierror variable, the sanity checking of error, and just make the last five lines `return custom_backend_set(cmd, slot->bootname, state, error)`?

src/bootchooser.c
```diff
@@ -1375,6 +1375,206 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+	return primary;
+}
+
+/* Get state of current slot */
+static gboolean custom_get_state(RaucSlot *slot, gboolean *good, GError **error)
+{
+	GError *ierror = NULL;
+	const gchar *cmd = "get-state";
+	g_autoptr(GString) ret_str = NULL;
+	*good = FALSE;
```
Villemoes: Hm, I don't think you're supposed to write to `*good` unless the function is succesful/returns true, at least judging from the other get_state implementations.

src/bootchooser.c
```diff
@@ -1375,6 +1375,206 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+
+	if (!custom_backend_get(cmd, slot->bootname, &ret_str, &ierror) ) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
+	if (g_strcmp0(ret_str->str, "good") == 0) {
+		*good = TRUE;
+	}
+
```
Villemoes: Perhaps check that the result is actually "good" or "bad", and error out if not, i.e. append something like `else if (strcmp ... "bad") { *good = FALSE; } else { complain/set error; return FALSE; }`

src/bootchooser.c
```diff
@@ -1375,6 +1375,206 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+	const gchar *cmd = "set-primary";
+
+	g_return_val_if_fail(slot, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (!custom_backend_set(cmd, slot->bootname, NULL, &ierror) ) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
```
Villemoes: As above, I think this can be mostly a `return custom_backend_set(cmd, slot->bootname, NULL, error);`.

src/config_file.c
```diff
@@ -106,6 +106,12 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 			goto free;
 		}
 		g_key_file_remove_key(key_file, "system", "efi-use-bootnext", NULL);
+	} else if (g_strcmp0(c->system_bootloader, "custom") == 0) {
+		c->custom_script = key_file_consume_string(key_file, "system", "bootloader-custom-script", NULL);
+		if (!c->custom_script) {
```
Villemoes: A bit of bikeshedding, but the custom bootloader interface need not be a script. Perhaps just `bootloader-interface` or `bootloader-handler`. And another bit of bikeshedding, perhaps the setting even belongs in the `[handlers]` section, the documentation for that does read "Handlers allow to customize RAUC by placing scripts in the system that RAUC can call for different purposes.".

src/config_file.c
```diff
@@ -106,6 +106,12 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
+			g_printf("No custom script defined. defaulting to /usr/bin/foobar as default");
+			c->custom_script = g_strdup("/usr/bin/foobar");
+		}
```
Villemoes: Why not just a hard error?

ejoerns:
> @ejoerns :
> Noted. I'll skip doing that then. Apologies!
> I wanted to force a re-test of the tests after fixing the problems found and force pushing, but no tests were triggered. Do I need to do something specific when force pushing ? I just did the changes needed and included that in the commits.

Normally, the tests should be re-executed when the branch is updated. It also seems to have worked here, anyway.

src/bootchooser.c
cbs101: yes. Included. Thanks

src/bootchooser.c
cbs101: yes. Included. Thanks

src/config_file.c
cbs101: Changed to cause an error if not defined. Thanks

src/bootchooser.c
cbs101: included. Thanks

src/config_file.c
cbs101: s/bootloader-custom-script/bootloader-custom-backend/
and
s/custom_script/bootloader_custom_backend/

Moved to handlers. Thanks

src/bootchooser.c
cbs101: I thought it made the most sense. But true; The other get_state does not write to it, unless they get proper feedback.

Removing the initial *good setting. Thanks

src/bootchooser.c
cbs101: replace by g_strchomp. Thanks

src/bootchooser.c
```diff
@@ -1375,6 +1375,206 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+	if (!g_subprocess_communicate(sub, NULL, NULL, &stdout_buf, NULL, &ierror)) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"Failed to run : %s", backend_name);
+		return FALSE;
+	}
+	data = g_bytes_get_data(stdout_buf, &size);
+	*ret_str = g_string_new(data);
```
Villemoes: Eh, is data necessarily nul-terminated? Don't you need to use g_string_new_len() using the size you got from g_bytes_get_data?

src/bootchooser.c
```diff
@@ -1375,6 +1375,206 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+	// Cleanup string for newlines
+	g_strchomp((*ret_str)->str);
```
Villemoes: Hm. You could also use g_strstrip() to also remove leading whitespace.

But how does a GString object react to having its ->str member modified without updating the ->len member? OK, the callers only seem to use the ->str member, but perhaps that means the GString container is the wrong thing to use here - maybe just take a "const char **ret_str" and create that with g_strndup() ?

cbs101: converted to use gchar instead of GString

jluebbe:
I've just force-pushed some changes to this PR:

- fix minor syntax issues
- add quoting to helper script and

jluebbe:
> Hello everyone! I am very interested in finalizing of this PR. How can I add the documentation to this PR?

Thanks! You can just create a separate PR with the documentation, as there won't be any conflicts. We'll then merge them in order.

PR #542, opened 2020-02-18
main.c: remove deprecated checksum command

Villemoes:
Exactly why and when this was deprecated is hard to determine, but the
"Deprecated" description was added 2.5 years ago (a9ac4098a321 -
src/main: add command summary). It's time to remove it completely.

Signed-off-by: Rasmus Villemoes <rasmus.villemoes@prevas.dk>


jluebbe:
The command was used for the (now removed) network mode. I'll adjust the commit message and merge this.

PR #543, opened 2020-02-19
Add rauc status subcommands to cli help

Psykar:
This has just bugged me somewhat, not being able to see what the CLI arguments are with a simple `-h`

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1763,7 +1763,13 @@ static void cmdline_handler(int argc, char **argv)
 #endif
 		{EXTRACT, "extract", "extract <BUNDLENAME> <OUTPUTDIR>", "Extract the bundle content", extract_start, NULL, FALSE},
 		{INFO, "info", "info <FILE>", "Print bundle info", info_start, info_group, FALSE},
-		{STATUS, "status", "status", "Show system status", status_start, status_group, TRUE},
+		{STATUS, "status", "status",
+		 "Show system status\n\n"
+		 "List of status commands (default slot is the currently booted slot):\n"
+		 "  mark-good [booted | other | <SLOT_NAME>] \tMark the slot as good\n"
+		 "  mark-bad [booted | other | <SLOT_NAME>] \tMark the slot as bad\n"
+		 "  mark-active [booted | other | <SLOT_NAME>] \tMark the slot as active",
+		 status_start, status_group, TRUE},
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -1778,7 +1778,14 @@ static void cmdline_handler(int argc, char **argv)
 		{CONVERT, "convert", "convert <INBUNDLE> <OUTBUNDLE>", "Convert to casync index bundle and store", convert_start, convert_group, FALSE},
 		{CHECKSUM, "checksum", "checksum <DIRECTORY>", "Deprecated", checksum_start, NULL, FALSE},
 		{INFO, "info", "info <FILE>", "Print bundle info", info_start, info_group, FALSE},
-		{STATUS, "status", "status", "Show system status", status_start, status_group, TRUE},
+		{STATUS, "status", "status",
+			"Show system status\n\n"
+			"List of status commands: (default image is the currently booted image)\n"
```
OnkelUlla: I would prefer _subcommands_ instead of _commands_ (knowing that we have to fix that at different places in the documentation as well).
The `rauc status mark-*` subcommands act on _slots_, not on _images_.

src/main.c
```diff
@@ -1778,7 +1778,14 @@ static void cmdline_handler(int argc, char **argv)
...
+			" mark-good [booted | other | <SLOT_NAME>] \tMark the image as good \n"
```
OnkelUlla: There is a whitespace that slipped in right before the linebreak.

src/main.c
```diff
@@ -1778,7 +1778,14 @@ static void cmdline_handler(int argc, char **argv)
...
+			" mark-bad [booted | other | <SLOT_NAME>] \tMark the image as bad\n"
+			" mark-active [booted | other | <SLOT_NAME>] \tMark the image as active\n",
```
ejoerns: The last line break is not required here, as this block will be printed with a line break, anyway.

src/main.c
ejoerns: I'd agree with that and would add that the colon should be at the end of the line

Psykar: Correct terminology is important I agree, oversight on my part (slots vs images)
Addressed the others as well :)

Psykar:
Comments addressed. Clearly I threw it together a *touch* too fast ;)
Thanks for the quick review :)

Psykar:
Ha, damn opinionated code styler ;) Also resolved.

ejoerns:
I've slightly adjusted the commit message to match our prefixing conventions and increased the subcommand printout indention level to two spaces instead of one (as all other indentions of help texts are two, too). Also rebased onto master to make it ready for merging.

PR #546, opened 2020-03-02
Allow custom casync arguments with --casync-args

### LINKED ISSUE #545: rauc convert should pass on chunk size parameter to casync
Currently casync bundles are built with the default casync chunk parameters, which currently is to build chunks with an average size of 64K.

This leaves us with lots of small chunks, when downloading over http with a lot rtt this can cause long delays.

Do you think this is worth passing this option through to casync?

obbardc:
I left the test as a separate commit, not sure whether it gives full coverage? Let me know if changes needed or you would like me to squash.

Thanks!

Closes #545

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -159,6 +159,21 @@ static gboolean casync_make_arch(const gchar *idxpath, const gchar *contentpath,
 		g_ptr_array_add(iargs, g_strdup("--store"));
 		g_ptr_array_add(iargs, g_strdup(store));
 	}
+
+	if (r_context()->casync_args != NULL) {
+		g_auto(GStrv) casync_argvp = NULL;
+		res = g_shell_parse_argv(r_context()->casync_args, NULL, &casync_argvp, &ierror);
+		if (!res) {
+			g_propagate_prefixed_error(
+					error,
+					ierror,
+					"Failed to parse casync extra args: ");
+			goto out;
+		}
+		for (gchar **casync_args = casync_argvp; *casync_args != NULL; casync_args++) {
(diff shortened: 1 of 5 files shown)


src/bundle.c
```diff
@@ -159,6 +159,20 @@ static gboolean casync_make_arch(const gchar *idxpath, const gchar *contentpath,
 		g_ptr_array_add(iargs, g_strdup("--store"));
 		g_ptr_array_add(iargs, g_strdup(store));
 	}
+	if (r_context()->casync_args != NULL) {
+		g_auto(GStrv) casync_argvp = NULL;
+		res = g_shell_parse_argv(r_context()->casync_args, NULL, &casync_argvp, &ierror);
```
ejoerns: Now that `res` is already set to TRUE here, it is required to explicitly set `res = FALSE` in the error path of the `r_subprocess_newv()` call below.

src/bundle.c
```diff
@@ -206,6 +220,20 @@ static gboolean casync_make_blob(const gchar *idxpath, const gchar *contentpath,
 		g_ptr_array_add(args, g_strdup("--store"));
 		g_ptr_array_add(args, g_strdup(store));
 	}
+	if (r_context()->casync_args != NULL) {
```
ejoerns: To ease identifying contexts, please leave an empty line before this if block

src/bundle.c
obbardc: good catch ; resolved !

jluebbe:
I've rebased this on master to solve the merge conflicts with the mksquashfs_args change.

PR #548, opened 2020-03-03
src/main: the 'convert' subcommand deserves --mksquashfs-args, too.

ejoerns:
In 7a390661f089f438855d5ff3c61e7a43ca1ffb99 (#537) the --mksquashfs-args
argument was added for the 'bundle' subcommand.
However, the same method is also used when invoking the 'convert'
command.
Thus, we simply enable the option for 'convert' too here.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1684,6 +1684,7 @@ static GOptionEntry entries_resign[] = {
 
 static GOptionEntry entries_convert[] = {
 	{"signing-keyring", '\0', 0, G_OPTION_ARG_FILENAME, &signing_keyring, "verification keyring file", "PEMFILE"},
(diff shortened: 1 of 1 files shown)


obbardc:
From a quick look at the code, `mksquashfs` is only used by `create_bundle` which is not used by `rauc convert`, unless I'm missing something?

ejoerns:
@obbardc `create_bundle()` is called by `convert_to_casync_bundle()`:  https://github.com/rauc/rauc/blob/master/src/bundle.c#L623

obbardc:
> @obbardc `create_bundle()` is called by `convert_to_casync_bundle()`: https://github.com/rauc/rauc/blob/master/src/bundle.c#L623

so it is, ignore my noise please :-)!

PR #551, opened 2020-03-06
install dbus interface file with the correct name

tretter:
The "D-Bus API Design Guidelines" define that interface files shall be installed into $(datadir)/dbus-1/interfaces to avoid local copies of the interface files at clients, which would go out of sync with the original interface.

Get the interfaces directory from pkg-config or via configure argument and install the de.pengutronix.rauc.Installer interface file into that directory.


Bastian-Krause:
Could you please fix the `contrib/cgi/src/rauc-installer.xml` symlink and rename it accordingly?

Bastian-Krause:
Thanks!

Issue #506, opened 2019-11-06
Avoid dbus dependency 


jluebbe:
Since 7698cadcbbd0db63877a2947b1d6b9cc7b556f58 we have a strict dependency on the dbus-1 pkg-config module. @michaelolbrich suggested to only search for it if `--with-dbussystemservicedir=DIR` is not given on the configure command-line and we actually need to do the auto-configuration. When this is fixed, the workaround in meta-rauc should be removed (see https://github.com/rauc/meta-rauc/pull/92).

jluebbe:
@michaelolbrich In https://github.com/jluebbe/rauc/commit/d1e0914e875cc10228c6c58b026934c5ce0c804b, I tried to implement this. Is seems to work (no `checking for DBUS1... yes` when the `-with-*=DIR` options are given), but if they are not, it searches for DBUS1 three times. Is there a better way?

jluebbe:
Instead of solving these minor issues with our autotools setup, I'd prefer to close this and move to Meson (#829). @ejoerns, what do you think?

ejoerns:
@jluebbe time is probably better invested in meson support as long as this does not cause any real-world pain (which it does not seem to).

I'll just close this as this was idle for quite a long time already, anyway.

PR #557, opened 2020-03-10
configure.ac: only depend on dbus if needed to auto-detect a dbus dir…

ukleinek:
…ectory

The dbus-1 pkgfile is only used if at least one of the three parameters

  --with-dbuspolicydir
  --with-dbussystemservicedir
  --with-dbusinterfacesdir

isn't specified. So only check for dbus-1 in this case which allows to
weaken build dependencies in some circumstances.

Closes: https://github.com/rauc/rauc/issues/506

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -116,17 +116,17 @@ AS_IF([$autodetect_dbuspolicydir || $autodetect_dbussystemservicedir || $autodet
 
        AS_IF([$autodetect_dbuspolicydir], [
 	      AC_MSG_CHECKING([dbus policy dir])
-	      with_dbuspolicydir="$($PKG_CONFIG --variable=datadir dbus-1)/dbus-1/system.d"
+	      PKG_CHECK_VAR([with_dbuspolicydir], [dbus-1], [datadir],, [AC_MSG_ERROR([Failed to determine dbus policy dir])])
 	      AC_MSG_RESULT([$with_dbuspolicydir])])
 
        AS_IF([$autodetect_dbussystemservicedir], [
 	      AC_MSG_CHECKING([dbus systemservice dir])
-	      with_dbussystemservicedir="$($PKG_CONFIG --variable=system_bus_services_dir dbus-1)"
+	      PKG_CHECK_VAR([with_dbussystemservicedir], [dbus-1], [system_bus_services_dir],, [AC_MSG_ERROR([Failed to determine dbus systemservice dir])])
 	      AC_MSG_RESULT([$with_dbussystemservicedir])])
 
        AS_IF([$autodetect_dbusinterfacesdir], [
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -98,23 +97,45 @@ fi
...
+	    [autodetect_dbusinterfacesdir=false],
+	    [autodetect_dbusinterfacesdir=true])
+
+AS_IF([$autodetect_dbuspolicydir || $autodetect_dbussystemservicedir || $autodetect_dbusinterfacesdir], [
+       PKG_CHECK_MODULES([DBUS], [dbus-1])
+
+       AS_IF([$autodetect_dbuspolicydir], [
+	      AC_MSG_CHECKING([dbus policy dir])
+	      with_dbuspolicydir="$($PKG_CONFIG --variable=datadir dbus-1)/dbus-1/system.d"
+	      AC_MSG_RESULT([$with_dbuspolicydir])])
```
marckleinebudde: The `pkg-config` man-page suggests to use
`PKG_CHECK_VAR(VARIABLE, MODULE, CONFIG-VARIABLE, [ACTION-IF-FOUND], [ACTION-IF-NOT-FOUND])`

ukleinek: This is IMHO a separate issue as the pkg-config call is only moved. So I'd address this in a separate change.

configure.ac
```diff
@@ -98,23 +97,45 @@ fi
...
+	    [autodetect_dbussystemservicedir=false],
+	    [autodetect_dbussystemservicedir=true])
+
+AC_ARG_WITH([dbusinterfacesdir],
+	    AS_HELP_STRING([--with-dbusinterfacesdir=DIR], [D-Bus interfaces directory]),
+	    [autodetect_dbusinterfacesdir=false],
+	    [autodetect_dbusinterfacesdir=true])
+
+AS_IF([$autodetect_dbuspolicydir || $autodetect_dbussystemservicedir || $autodetect_dbusinterfacesdir], [
+       PKG_CHECK_MODULES([DBUS], [dbus-1])
```
jluebbe: Previously, this was only checked when the service is enabled. Should the whole block be conditional?

ukleinek: right, before my patch `PKG_CHECK_MODULES(..., [dbus-1])` was only called when `--enable-service` was passed to `./configure`. However the usage of all the variables are not conditionally in `Makefile.am`, so this was buggy before.

configure.ac
ukleinek: pushed another commit on top of the previous one to convert to `PKG_CHECK_VAR`.

ukleinek:
Will rework the pull request, I just learned about better ways to solve this issue.

ejoerns:
No need to close this if you plan to rework it...

PR #558, opened 2020-03-11
Improve configure.ac to properly use pkg-config

ukleinek:
This improves the configure script on some edges. So querying the `pkg-config` database for `systemd` and `dbus-1` is simplified and still more robust as errors are catched. Also the checks are not done when they are irrelevant (i.e. with `./configure --disable-service`).

As a side effect all the `--with-...dir` options (which don't really fit the usual semantic of autoconf's --with-...) are dropped and the paths can instead be overwritten using e.g.

    ./configure DBUS_POLICYDIR=/path/to/where/you/want/it/

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -89,10 +89,9 @@ rauc_CFLAGS = $(AM_CFLAGS) $(JSON_GLIB_CFLAGS) $(CODE_COVERAGE_CFLAGS)
 rauc_LDFLAGS = $(AM_LDFLAGS) $(JSON_GLIB_LDFLAGS) $(CODE_COVERAGE_LDFLAGS)
 rauc_LDADD = $(JSON_GLIB_LIBS) $(librauc_la_LIBADD) librauc.la
 
-if SYSTEMD
-systemdunitdir=$(SYSTEMD_UNITDIR)
+if ENABLE_SERVICE
+systemdunitdir=$(SYSTEMD_SYSTEMUNITDIR)
 nodist_systemdunit_DATA = data/rauc.service
-endif
 
 dbussystemservicedir = $(DBUS_SYSTEMSERVICEDIR)
 nodist_dbussystemservice_DATA = data/de.pengutronix.rauc.service
@@ -105,6 +104,7 @@ dist_dbusinterfaces_DATA = src/de.pengutronix.rauc.Installer.xml
 
 dbuswrapperdir = $(libexecdir)
 nodist_dbuswrapper_SCRIPTS = data/rauc-service.sh
+endif
 
 EXTRA_DIST += data/rauc.service.in \
 	      data/de.pengutronix.rauc.service.in \
--- a/configure.ac
+++ b/configure.ac
@@ -44,9 +44,20 @@ AC_ARG_ENABLE([service],
 )
 AS_IF([test "x$enable_service" != "xno"], [
 	AC_DEFINE([ENABLE_SERVICE], [1], [Define to 1 to enable background service])
+	AC_ARG_VAR([DBUS_POLICYDIR], [Directory used to store dbus policy config files @<:@default=$DBUS_DATADIR/dbus-1/system.d@:>@])
+	AS_VAR_IF([DBUS_POLICYDIR],,
+		  [PKG_CHECK_VAR([DBUS_DATADIR], [dbus-1], [datadir],, [AC_MSG_ERROR([Failed to determine dbus data directory])])
+		   AS_VAR_SET([DBUS_POLICYDIR], [$DBUS_DATADIR/dbus-1/system.d])])
+
+	PKG_CHECK_VAR([DBUS_SYSTEMBUSSERVICEDIR], [dbus-1], [system_bus_services_dir],, [AC_MSG_ERROR([Failed to determine dbus system bus services directory])])
+
+	PKG_CHECK_VAR([DBUS_INTERFACESDIR], [dbus-1], [interfaces_dir],, [AC_MSG_ERROR([Failed to determine dbus interfaces directory])])
+
+	PKG_CHECK_VAR([SYSTEMD_SYSTEMUNITDIR], [systemd], [systemdsystemunitdir],, [AC_MSG_ERROR([Failed to determine systemd unit directory])])
 ], [
 	AC_DEFINE([ENABLE_SERVICE], [0])
 ])
+AM_CONDITIONAL([ENABLE_SERVICE], [test "x$enable_service" != "xno"])
 
 AC_ARG_ENABLE([create],
 	AS_HELP_STRING([--disable-create], [Disable bundle creation and modification commands])
@@ -83,27 +94,7 @@ AS_IF([test "x$enable_json" != "xno"], [
        AC_DEFINE([ENABLE_JSON], [0])
 ])
 
-
 AX_CHECK_OPENSSL([],[AC_MSG_ERROR([OpenSSL not found])])
-
-AC_ARG_WITH([systemdunitdir],
-        AC_HELP_STRING([--with-systemdunitdir=DIR], [path to systemd service directory]),
-        [],
-        [with_systemdunitdir="$($PKG_CONFIG --variable=systemdsystemunitdir systemd)"])
-if (test -n "${with_systemdunitdir}"); then
-	SYSTEMD_UNITDIR="${with_systemdunitdir}"
-	AC_SUBST(SYSTEMD_UNITDIR)
-fi
-AM_CONDITIONAL(SYSTEMD, test -n "${with_systemdunitdir}")
-
-AC_ARG_VAR([DBUS_POLICYDIR], [Directory used to store dbus policy config files @<:@default=$DBUS_DATADIR/dbus-1/system.d@:>@])
-AS_VAR_IF([DBUS_POLICYDIR],,
-	  [PKG_CHECK_VAR([DBUS_DATADIR], [dbus-1], [datadir],, [AC_MSG_ERROR([Failed to determine dbus data directory])])
-	   AS_VAR_SET([DBUS_POLICYDIR], [$DBUS_DATADIR/dbus-1/system.d])])
-
-PKG_CHECK_VAR([DBUS_SYSTEMBUSSERVICEDIR], [dbus-1], [system_bus_services_dir],, [AC_MSG_ERROR([Failed to determine dbus system bus services directory])])
-
-PKG_CHECK_VAR([DBUS_INTERFACESDIR], [dbus-1], [interfaces_dir],, [AC_MSG_ERROR([Failed to determine dbus interfaces directory])])
 
 # Checks for header files.
 
--- a/configure.ac
+++ b/configure.ac
@@ -44,7 +44,6 @@ AC_ARG_ENABLE([service],
 )
 AS_IF([test "x$enable_service" != "xno"], [
 	AC_DEFINE([ENABLE_SERVICE], [1], [Define to 1 to enable background service])
-	PKG_CHECK_MODULES([DBUS1], [dbus-1])
 ], [
 	AC_DEFINE([ENABLE_SERVICE], [0])
 ])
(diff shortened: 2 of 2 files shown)


ukleinek:
Travis CI stumbled about the part "Also this adds error handling if the systemd pkg-config query
fails" in the second commit as it doesn't ensure that the systemd pkg-config stuff is available. The details are that in master we have:

    with_systemdunitdir="$($PKG_CONFIG --variable=systemdsystemunitdir systemd)

This results in an error message and `$with_systemdunitdir` being empty. This is then checked later and the file `rauc.service` just isn't installed. See https://travis-ci.com/github/rauc/rauc/jobs/296602638#L700 for the error message on current `master`.

The options to fix this are:

 - Ensure a package is installed that provides `systemd.pc` (On Debian Buster that would be `systemd`)
 - Explicitly specify the path where to install the `rauc.service` file for configure
 - Let the error "Oh, I cannot find out where to install the service file" fall back to the behaviour of `./configure --disable-service`.

A bit orthogonal to that I suggest to add a call to

    make install DESTDIR=/path/to/some/writeable/directory

plus a check if all expected files are then present to the automatic checks.

michaelolbrich:
I think using `PKG_CHECK_VAR` is correct, but I'd keep the `--with-*` options. That's what other packages are doing.

ukleinek:
Keeping `--with-*` just to be bug-compatible with other packages? I think to implement this without surprises is hard. (What if both `--with-dbuspolicydir=/some/path` and `DBUS_POLICYDIR=/some/other/path` are specified?)

I wonder who will benefit if we keep `--with-...`. Isn't modifying these values very special as deviating from the defaults quite probably yields broken setups? So it's mostly only about reducing build dependencies for distribution packages? Debian doesn't make use of this option. PTXdist uses these options, but as the package depends on DBUS at least the dbus-related option can better be dropped? There is no relation to SYSTEMD, but I wonder if this is wrong and should be improved.
For Yocto (with my limited understanding of it) it's the same. There is a dependency on dbus, so the dbus default settings should be fine, but no dependency on systemd.

jluebbe:
@ukleinek The CI build fails with "configure: error: Failed to determine systemd unit directory".

> Keeping `--with-*` just to be bug-compatible with other packages? I think to implement this without surprises is hard. (What if both `--with-dbuspolicydir=/some/path` and `DBUS_POLICYDIR=/some/other/path` are specified?)

From my side, this would be mainly about avoiding unexpected changes for build systems. @michaelolbrich?

> I wonder who will benefit if we keep `--with-...`. Isn't modifying these values very special as deviating from the defaults quite probably yields broken setups? So it's mostly only about reducing build dependencies for distribution packages? Debian doesn't make use of this option. PTXdist uses these options, but as the package depends on DBUS at least the dbus-related option can better be dropped? There is no relation to SYSTEMD, but I wonder if this is wrong and should be improved.
> For Yocto (with my limited understanding of it) it's the same. There is a dependency on dbus, so the dbus default settings should be fine, but no dependency on systemd.

The dependency on D-Bus is *only* for this installation information, so we'd like to get rid of that dependency in ptxdist and OE. @michaelolbrich

michaelolbrich:
All generic installation paths are configured with `--*dir=`. All other installation paths are configured with `--with-*dir=`. Probably because there is no handy macro for `--*dir=`. I've not seen a package that uses a environment variable for this.
As someone who packages a lot of stuff, I really appreciate standardized options. Exactly the same options are best, but following a pattern is also important. Any custom stuff is always painful because it's not immediately obvious what needs to be done.
So please use `--with-*dir=`.

ukleinek:
> All generic installation paths are configured with `--*dir=`. All other installation paths are configured with `--with-*dir=`. Probably because there is no handy macro for `--*dir=`. I've not seen a package that uses a environment variable for this.

And all stuff overwriting pkgconfig variable is using environment variables :-)

> As someone who packages a lot of stuff, I really appreciate standardized options. Exactly the same options are best, but following a pattern is also important. Any custom stuff is always painful because it's not immediately obvious what needs to be done.

I looked for examples on codesearch.debian.org for one or the other variant. I found the following instances that use `PKG_CHECK_VAR` for paths:

These use this macro with no --with-... to also set it:
 - https://sources.debian.org/src/bind9-libs/1:9.11.17+dfsg-4/configure.ac/?hl=762#L762
 - https://sources.debian.org/src/p11-kit/0.23.20-1/configure.ac/?hl=592#L592
 - https://sources.debian.org/src/shibboleth-sp/3.0.4+dfsg1-1/configure.ac/?hl=136#L136
 - https://sources.debian.org/src/opensaml/3.0.1-1/configure.ac/?hl=81#L81
 - https://sources.debian.org/src/nbdkit/1.18.3-1/configure.ac/?hl=498#L498
 - https://sources.debian.org/src/ipset/6.38-1.2/configure.ac/?hl=68#L68
 - https://sources.debian.org/src/varnish/6.2.1-3/varnish.m4/?hl=82#L82
 - https://sources.debian.org/src/audacious-plugins/4.0-2/acinclude.m4/?hl=220#L220
 - https://sources.debian.org/src/qbittorrent/4.1.7-1/m4/qbittorrent.m4/?hl=9#L9
 - https://sources.debian.org/src/sane-frontends/1.0.14-15/configure.in/?hl=116#L116
 - https://sources.debian.org/src/jp2a/1.0.9-1/configure.ac/?hl=117#L117
 - https://sources.debian.org/src/util-vserver/0.30.216-pre3120-1.4/configure.ac/?hl=501#L501
 - https://sources.debian.org/src/varnish-modules/0.15.0-1/configure.ac/?hl=29#L29
 - https://sources.debian.org/src/guile-git/0.3.0-2/configure.ac/?hl=39#L39

... and these also provide an option to additionally overwrite it:
 - https://sources.debian.org/src/blueman/2.1.2-1/configure.ac/?hl=223#L223
 - https://sources.debian.org/src/xdg-desktop-portal/1.6.0-1/configure.ac/?hl=53#L53
 - https://sources.debian.org/src/fuse-emulator/1.5.7+dfsg1-3/configure.ac/?hl=934#L934
 - https://sources.debian.org/src/datamash/1.6+dfsg-1/configure.ac/?hl=124#L124
 - https://sources.debian.org/src/tpm2-tools/4.1.1-1/configure.ac/?hl=69#L69
 - https://sources.debian.org/src/mptp/0.2.4-1/configure.ac/?hl=55#L55
 - https://sources.debian.org/src/onedrive/2.4.0-1/configure.ac/?hl=247#L247
 - https://sources.debian.org/src/bitlbee-facebook/1.2.0-1/configure.ac/?hl=90#L90
 - https://sources.debian.org/src/bitlbee-mastodon/1.4.3-1/configure.ac/?hl=66#L66
 - https://sources.debian.org/src/gdnsd/2.4.2-1/configure.ac/?hl=246#L246

So not providing `--with-*dir` doesn't seem so unusual.

Just looking at the first providing an overwrite:

    blueman-2.1.2$ ./configure --help
    ...
    Optional Packages:
    ...
      --with-systemdsystemunitdir=PATH
                              Directory for systemd system unit files
                              [default=/usr/lib/systemd/system]
    ...
    Some influential environment variables:
    ....
      systemd_user_unit_dir
                  value of systemduserunitdir for systemd, overriding pkg-config
    ...

This is confusing at best.

To summarize: We have the following arguments against providing `--with-*dir`:

 - `PKG_CHECK_VAR` (which we agree is the right choice) already provides a way to overwrite the setting, additionally providing `--with-*dir` results in two ways to do the same thing, resulting in a confusing help output.
 - Just using `PKG_CHECK_VAR` is simpler than `PKC_CHECK_VAR` + `AC_ARG_WITH`.
 - `--with-*` is intended for enabling use of additional external packages, `systemdunitdir` isn't such a package

arguments pro providing `--with-*dir`:

 - It is (somewhat) common
 - build system maintainers will have to understand and adapt their recipes

In my (admittedly) subjective weighting of the arguments I favour not providing `--with-*dir`, still more given that the people caring for rauc recipes in ptxdist, yocto and Debian are well aware of this discussion and so the respective issue is quite minor.

michaelolbrich:
> I looked for examples on codesearch.debian.org for one or the other variant. I found the following instances that use `PKG_CHECK_VAR` for paths:
> 
> These use this macro with no --with-... to also set it:
[...]
> ... and these also provide an option to additionally overwrite it:
[...]
> So not providing `--with-*dir` doesn't seem so unusual.

There are more than 100 cases that use `PKG_CONFIG` directly for `--with-*dir`.
And close to 2000 hits for `--with-*dir` variables in general.

So it _is_ very uncommon to not use `--with-*dir`.

> Just looking at the first providing an overwrite:
> 
> ```
> blueman-2.1.2$ ./configure --help
> ...
> Optional Packages:
> ...
>   --with-systemdsystemunitdir=PATH
>                           Directory for systemd system unit files
>                           [default=/usr/lib/systemd/system]
> ...
> Some influential environment variables:
> ....
>   systemd_user_unit_dir
>               value of systemduserunitdir for systemd, overriding pkg-config
> ...
> ```
> 
> This is confusing at best.

I disagree. Just `value of systemduserunitdir for systemd, overriding pkg-config` is confusing. There is no explanation what it's actually for.

> To summarize: We have the following arguments against providing `--with-*dir`:
> 
>     * `PKG_CHECK_VAR` (which we agree is the right choice) already provides a way to overwrite the setting, additionally providing `--with-*dir` results in two ways to do the same thing, resulting in a confusing help output.

I think It's actually better with both.

>     * Just using `PKG_CHECK_VAR` is simpler than `PKC_CHECK_VAR` + `AC_ARG_WITH`.

It' not that complex and well worth the gains.

>     * `--with-*` is intended for enabling use of additional external packages, `systemdunitdir` isn't such a package

That's your opinion. The reality (close to 2000 hits in on codesearch) says otherwise.

> arguments pro providing `--with-*dir`:
> 
>     * It is (somewhat) common

It's used everywhere.

ukleinek:
> > I looked for examples on codesearch.debian.org for one or the other variant. I found the following instances that use `PKG_CHECK_VAR` for paths:
> > These use this macro with no --with-... to also set it:
> > [...]
> > ... and these also provide an option to additionally overwrite it:
> > [...]
> > So not providing `--with-*dir` doesn't seem so unusual.
> 
> There are more than 100 cases that use `PKG_CONFIG` directly for `--with-*dir`.
> And close to 2000 hits for `--with-*dir` variables in general.
> 
> So it _is_ very uncommon to not use `--with-*dir`.

And so it is also very uncommon to use `PKG_CHECK_VAR` which is the right macro to query pkg-config variables. So your statement mostly proves that people don't care much about this all.

> > Just looking at the first providing an overwrite:
> > ```
> > blueman-2.1.2$ ./configure --help
> > ...
> > Optional Packages:
> > ...
> >   --with-systemdsystemunitdir=PATH
> >                           Directory for systemd system unit files
> >                           [default=/usr/lib/systemd/system]
> > ...
> > Some influential environment variables:
> > ....
> >   systemd_user_unit_dir
> >               value of systemduserunitdir for systemd, overriding pkg-config
> > ...
> > ```
> > 
> > 
> > This is confusing at best.
> 
> I disagree. Just `value of systemduserunitdir for systemd, overriding pkg-config` is confusing. There is no explanation what it's actually for.

Yeah, I admit the help text isn't very clear.

> > ```
> > * Just using `PKG_CHECK_VAR` is simpler than `PKC_CHECK_VAR` + `AC_ARG_WITH`.
> > ```
> 
> It' not that complex and well worth the gains.

Right, it's not very hard, but I doubt the "worth the gains" part. (Of course both your and my statement is subjective)

> > ```
> > * `--with-*` is intended for enabling use of additional external packages, `systemdunitdir` isn't such a package
> > ```
> 
> That's your opinion. The reality (close to 2000 hits in on codesearch) says otherwise.

The [documentation for `AC_ARG_WITH`](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/External-Software.html#External-Software) however confirms my statement.

I claim the problem source is historic. Someone started (mis)using --with-PACKAGE for directories because it is easy. And 2000 projects doing it in a way that I consider non-optimal isn't a good incentive in my eyes to align to their behavior.

My motivation to discuss this further now dropped below zero, I'm closing this PR. Thanks

PR #559, opened 2020-03-11
Promote Central Slot Status Usage More Obviously

ejoerns:
Using a single central slot status file in RAUC has several advantages. Thus we should be more explicit in notifying the user of this feature.

This now:
* adds a check to the design checlist
* explicitly mentions benefits in documentations
* adds an info message when not using the central status

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -29,6 +29,7 @@ Example configuration:
   [system]
   compatible=FooCorp Super BarBazzer
   bootloader=barebox
+  statusfile=/data/central-status.raucs
 
   [keyring]
   path=/etc/rauc/keyring.pem
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -297,6 +297,8 @@ hierarchical separator.
   Marks the slot as existing but not updatable. May be used for sanity checking
   or informative purpose. A ``readonly`` slot cannot be a target slot.
 
+.. _install-same:
+
 ``install-same=<true/false>``
   If set to ``false``, this will tell RAUC to skip writing slots that already
   have the same content as the one that should be installed.
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -97,10 +97,19 @@ Example configuration:
 .. _statusfile:
 
 ``statusfile``
-  If this key exists, it points to a file where slot status information should
-  be stored (e.g. slot specific metadata, see :ref:`slot-status`).
-  This file should be located on a filesystem which is not overwritten during
-  updates.
+  This key should be set to point to a central file where slot status
+  information should be stored (e.g. slot-specific metadata, see
+  :ref:`slot-status`).
+  This file must be located on a non-redundant filesystem which is not
+  overwritten during updates.
+  In most cases, a central status file is preferable to per-slot status files
+  as it allows to store data also for read-only or (temporary) filesystem-less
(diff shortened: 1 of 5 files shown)


docs/reference.rst
```diff
@@ -98,6 +99,8 @@ Example configuration:
   be stored (e.g. slot specific metadata, see :ref:`slot-status`).
   This file should be located on a filesystem which is not overwritten during
   updates.
+  This method is preferable over per-slots status files as it allows to store
```
OnkelUlla: To avoid having a recurrence at the beginning of the sentence compared to the previous one I propose to start it like

```
A central status file is preferable…
```

instead. While at it, please do ``s/preferable over per-slots/preferable to per-slots/`` and ``s/per-slots/per-slot/`` as well.

docs/reference.rst
```diff
@@ -98,6 +99,8 @@ Example configuration:
...
+  data also for read-only or (temporary) filesystem-less slots.
```
OnkelUlla: Not 100% sure, but for me it feels better to have a replacement ``s/temporary/temporarily/``.

docs/reference.rst
ejoerns: Do you know about GH suggestions? ;) I'll change it as you requested.

docs/reference.rst
```diff
@@ -27,6 +27,7 @@ Example configuration:
   [system]
   compatible=FooCorp Super BarBazzer
   bootloader=barebox
+  statusfile=/data/global-status.raucs
```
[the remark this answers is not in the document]
ejoerns: Ok

docs/reference.rst
ejoerns: I don't see that one should use an adverb here.

docs/reference.rst
OnkelUlla: > Do you know about GH suggestions? ;) I'll change it as you requested.

Having a joint coffee together in the office later will offer a good possibility for introducing me to GH suggestions! ;-)

docs/reference.rst
OnkelUlla: Perhaps I got the intended meaning wrong: I thought that it was bound to ``filesystem-less``.

jluebbe:
I'm not sure about adding the fallback warning in the code, in some situations, it might be a good choice to not use a central slot status (no or redundant data partition)/

ejoerns:
> I'm not sure about adding the fallback warning in the code, in some situations, it might be a good choice to not use a central slot status (no or redundant data partition)/

It is not a warning indeed. Maybe the next consequence would be to also let the per-slot status file be selected explicitly in config? like:

``` ini
statusfile=per-slot
```

OnkelUlla:
> It is not a warning indeed. Maybe the next consequence would be to also let the per-slot status file be selected explicitly in config? like:
> 
> ```ini
> statusfile=per-slot
> ```

This is a good idea as it allows to express that one is aware of central status files and for some reason explicitely does not choose it.

ejoerns:
@jluebbe I see this as your agreement on my proposal and will rework it accordingly ;)

ejoerns:
I've now changed the PR to support explicitly setting both central and per-slot status storage strategy. I also updated some parts of the documentation and improved the debugability of config_file checks.

PR #560, opened 2020-03-12
Fixing a config_file test that suddenly broke..

ejoerns:
This collects all changes that needed to be done to reveal and fix the issue why in my setup the test `/config-file/system-serial` suddenly painted an annoying red in the green meadow of successful test runs.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -573,6 +573,11 @@ void r_context_clean(void)
 		g_clear_pointer(&context->signing_keyringpath, g_free);
 		g_clear_pointer(&context->mksquashfs_args, g_free);
 		g_clear_pointer(&context->casync_args, g_free);
+		g_clear_pointer(&context->intermediatepaths, g_strfreev);
+		g_clear_pointer(&context->mountprefix, g_free);
+		g_clear_pointer(&context->bootslot, g_free);
+		g_clear_pointer(&context->system_serial, g_free);
+		g_clear_pointer(&context->handlerextra, g_free);
 
 		if (context->config) {
 			context->config->keyring_path = NULL;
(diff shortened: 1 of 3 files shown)


src/context.c
```diff
@@ -566,20 +566,12 @@ const RaucContext *r_context(void)
...
-		context->keyringdirectory = NULL;
-		context->signing_keyringpath = NULL;
-		context->mksquashfs_args = NULL;
-		context->casync_args = NULL;
+		g_clear_pointer(&context->certpath, g_free);
+		g_clear_pointer(&context->keypath, g_free);
+		g_clear_pointer(&context->keyringpath, g_free);
+		g_clear_pointer(&context->keyringdirectory, g_free);
+		g_clear_pointer(&context->signing_keyringpath, g_free);
+		g_clear_pointer(&context->mksquashfs_args, g_free);
```
jluebbe: You lost casync_args here, I'll push a fixup myself.

src/context.c
```diff
@@ -563,6 +563,12 @@ const RaucContext *r_context(void)
 	return context;
 }
 
+void r_context_install_info_free(RContextInstallationInfo *info)
+{
+	/* contains only reference to existing bundle instance */
+	info->mounted_bundle = NULL;
+}
```
jluebbe: Don't we need to free the struct itself here?

ejoerns: Oh, yes. seems I missed that.

src/context.c
jluebbe: This is now fixed.

src/context.c
jluebbe: OK, i'll add that as well.

Issue #556, opened 2020-03-10
RAUC verification issue with  Certificate Revocation List 


djamesmec:
Hello,
I have created a bundle, signed with autobuilder-2.pem( revoked) from [test CA folder ](https://github.com/rauc/rauc/tree/master/test) .  I have installed dev-only-ca.pem as Keyring file on my target. 
 As per the script ([openssl-ca.sh](https://github.com/rauc/rauc/blob/master/test/openssl-ca.sh), the  dev-only-ca.pem must contain CRL with revoked certficate info. 
But it is observed that neither the "rauc info" cmd nor the "rauc install" cmd showed verification failure as "Certificate Revoked", with the bundle signed using revoked key.  
Here "rauc info " showed the complete certificate chain. "rauc install" installed the bundle succesfully . 
rauc versions used :   rauc-1.0 and rauc-1.2.91
Could you let me know what went wrong here .

ejoerns:
Hi @djamesmec after a short investigation I can say that the X509 store in RAUC currently does not have the CRL checking flags set. Thus, at the moment, verification process does not take CRLs into account.
It's also not that easy as simply switching on the required flags as OpenSSL then enforces the presence of CRLs in a keyring/x509 store. This will surely break existing scenarios.

Thus our approach now is to add a system.config `[keyring]` option named `check-crl` which can be used to enable CRL checking.

As an additional safety net we also plan to add a check that should detect currently used (but still ineffective CRLs) by scanning the keyring files for a CRL header and emit a warning if a CRL is deceted but `check-crl` is not enabled.

Currently only some additional tests are missing to make a PR.

Most people anyway don't seem to use CRLs but do revocations by simply deploying a new keyring file with an update. Having CRLs will probably mostly have a benefit if they are retreived independent from the normal update process.

Thanks for reporting! 

Regards, Enrico

ejoerns:
I've pushed #561 now as a proposal to fix this.

djamesmec:
Hello @ejoerns, 
I have tested these changes and it worked . Thanks for the support.

jluebbe:
@djamesmec We'd be interested to know why using revocation instead of rotating the keyring is preferable in your use-case.

djamesmec:
[...]

PR #561, opened 2020-03-12
Allow to Enable CRL handling in RAUC's Bundle Verification

ejoerns:
In order to actually use a provided CRL for checking, some flags need to
be set in OpenSSL. These were unset up to now. As a result of this, a
keyring / cert store file or directory could have contained a CRL, but
the Bundle verification process in RAUC simply ignored it.

This patch now introduces the `[keyring]` section option `check-crl` that
allows enabling CRL handling for RAUC's signature verification.

The reason why this cannot be enabled by default is that OpenSSL would
enforce having a CRL in a cert store. This will not only break existing
setups but also make the simplest case (a simple self-signed
certificate for now) more complex.

Although, we cannot enable CRLs by default, there may be setups
that probably already have a CRL in their key store.

To more visibly tell them to enable the 'check-crl' flag, we check stores for
containing files that have the standardized CRL header in it. If these are
present and check-crl is not enabled, we print a warning to inform the user
about the ineffectiveness of their CRLs.

Fixes: #556

### DIFF
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -45,6 +45,7 @@ typedef struct {
 	gchar *statusfile_path;
 	gchar *keyring_path;
 	gchar *keyring_directory;
+	gboolean keyring_check_crl;
 	gboolean use_bundle_signing_time;
 
 	gchar *autoinstall_path;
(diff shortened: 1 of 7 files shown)


jluebbe:
@djamesmec commented in https://github.com/rauc/rauc/issues/556#issuecomment-600539177 that this fixed his issue.

jluebbe:
Fixed some commit message typos.

PR #587, opened 2020-05-21
D-Bus API: clarify license for rauc-installer.xml (fixes #468)

### LINKED ISSUE #468: License question regarding D-Bus API description file
The bundled example CGI program in contrib directory shows howto use the D-Bus API
from C programs. It also uses the file `rauc-installer.xml` which is used to
generate code which is then compiled into the resulting binary.

I would like to do the same in a custom (closed-source) project, that means I would like
to include the XML file in my project and use the generated code then. While
I guess that using the generated code is not a problem,
I'm unsure about the API description file itself and its LGPL relation.

What is your opinion?
Is this a real issue?
Should allowed usage be mentioned explicitly in documentation or with a comment in the XML file?
If so, would it be OK to change license of this single file?

mhei:
As this file contains the same information you can get from
the D-Bus introspection API, the file is considered as simple data.

Add an explicit note in the file to document this.

Signed-off-by: Michael Heimpold <michael.heimpold@in-tech.com>

### DIFF
--- a/src/de.pengutronix.rauc.Installer.xml
+++ b/src/de.pengutronix.rauc.Installer.xml
@@ -1,3 +1,4 @@
+<!-- License note: This file is considered data and thus no license is needed for any usage. -->
 <node>
   <interface name="de.pengutronix.rauc.Installer">
     <!--


src/de.pengutronix.rauc.Installer.xml
```diff
@@ -1,3 +1,4 @@
+<!-- License note: This file is considered data and thus not covered by the project's main license. -->
```
jluebbe, line 1: What do you think about:
```suggestion
<!-- License note: This file is considered data and thus no license is needed for any usage. -->
```

mhei: I'd be fine with this, will update this PR in a few seconds...

jluebbe:
I've merged this manually to add my S-o-B and adjust the prefix.

PR #589, opened 2020-05-25
service: add annotation for InstallBundle method

tobydox:
An annotation is required for the recently added `InstallBundle` method as well when using qdbusxml2cpp.


tobydox:
Thanks!

PR #591, opened 2020-05-27
prevent issues around duplicate bootnames

### LINKED ISSUE #588: Setting bootname on both parent and child slots can cause confusing non-deterministic errors: invalid slot status 7
After upgrading glib and some other packages (not rauc) we started seeing this confusing error message from rauc when running `rauc status` or any installation commands:

```
(rauc:103246): rauc-ERROR **: 06:45:31.099: invalid slot status 7
```

But the error was not occurring all the time. Eventually I narrowed down the root cause to a misconfiguration in our product's /etc/rauc/system.conf. Here is the relevant parts:

``` ini
[slot.rootfs.0]
device=/dev/rootfs-slot-0
type=raw
bootname=slot_0

[slot.kernel.0]
device=/boot/slot_0.bin
type=raw
parent=rootfs.0

[slot.kernelsig.0]
device=/boot/slot_0.bin.sig
type=raw
parent=rootfs.0

[slot.blsconf.0]
device=/boot/loader/entries/slot_0.conf
type=raw
parent=rootfs.0
bootname=slot_0

[slot.cmdlinesig.0]
device=/boot/slot_0.bin.cmdline.sig
type=raw
parent=rootfs.0

[slot.rootfs.1]
device=/dev/rootfs-slot-1
type=raw
bootname=slot_1

[slot.kernel.1]
device=/boot/slot_1.bin
type=raw
parent=rootfs.1

[slot.kernelsig.1]
device=/boot/slot_1.bin.sig
type=raw
parent=rootfs.1

[slot.blsconf.1]
device=/boot/loader/entries/slot_1.conf
type=raw
parent=rootfs.1
bootname=slot_1

[slot.cmdlinesig.1]
device=/boot/slot_1.bin.cmdline.sig
type=raw
parent=rootfs.1
```

Notice that the rootfs.0 slot has `bootname=slot_0`, but its child slot blsconf.0 also has `bootname=slot_0`. According to the rauc docs about [grouping slots](https://rauc.readthedocs.io/en/latest/integration.html#grouping-slots) this configuration is clearly wrong. But actually rauc doesn't notice the misconfiguration, it happily accepts this configuration file.

And it was actually working fine in our product for several years. The reason is the `determine_slot_states()` function iterates the slots in non-deterministic hash table order. If the hash table happens to put the parent slot first, `determine_slot_states()` correctly picks the parent slot as the booted slot and everything works.

But some change in glib, or just some bad luck, means that now the hash table iteration order is different and the child slot was being picked as the booted slot incorrectly. Then this block in `determine_slot_states()` was constructing an invalid enum value for ->state:

``` c
        /* Determine active group members */
                RaucSlot *s = (RaucSlot*) g_hash_table_lookup(r_context()->config->slots, l->data);

                if (s->parent) {
                        if (s->parent->state & ST_ACTIVE) {
                                s->state |= ST_ACTIVE;
                        } else {
                                s->state |= ST_INACTIVE;
                        }
                } else {
                        if (s->state == ST_UNKNOWN)
                                s->state |= ST_INACTIVE;
                }
        }
```

`s->state` is `ST_BOOTED` but `s->parent->state` is `ST_INACTIVE`. 5 | 2 = 7.

I would like to contribute a fix for handling this misconfiguration, but I'm not sure what would be the best solution. Here are some possibilities I thought of:

* if a slot has both `parent=` and `bootname=` make the config parser reject it with an error
* if a slot has both `parent=` and `bootname=` make the config parser warn and ignore the bootname (I am going to use this as a temporary solution in our product but it does not seem ideal)
* if two slots have the same `bootname=` make the config parser reject it with an error
* avoid using |= when constructing s->state so that an invalid enum value cannot be produced
* make the logic for s->state more robust, for example if a child is booted mark its parent as active, or just add some g_warn_if_fail() to catch nonsensical states

danc86:
The first two commits ensure the config file is rejected if it does not meet the existing assumptions about how slot groups are constructed (bootname must be set only on the parent slot in a group).

The third commit just tidies up the confusing use of |= which led to the invalid enum values.

Fixes #588.


danc86:
Btw I noticed a lot of leaks in the error handling paths of config_file.c, I didn't try to fix them here but I will see about sending a separate PR for them after.

I guess they are not super important because a given rauc process typically only attempts to parse its config at most once, right?

src/install.c
```diff
@@ -152,25 +152,17 @@ gboolean determine_slot_states(GError **error)
...
-			if (s->parent->state & ST_ACTIVE) {
-				s->state |= ST_ACTIVE;
-			} else {
-				s->state |= ST_INACTIVE;
-			}
+		if (s == booted) {
+			s->state = ST_BOOTED;
+			g_debug("Found booted slot: %s on %s", s->name, s->device);
+		} else if (s->parent && s->parent == booted) {
+			s->state = ST_ACTIVE;
```
jluebbe, line 163: Hmm, looking at this, I think it already failed if the parent itself has a parent. I that case we should probably fix all parent links to be direct, in the config parsing.

danc86: Oh hmm that's interesting. So are you thinking, some code before this which will catch where a slot has a parent= which then itself has a parent=, and fix it so that all children in the group refer to the "real" parent?

Would it make more sense to just reject that configuration with an error, like these others?

src/config_file.c
```diff
@@ -35,6 +35,7 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 	gsize group_count;
 	GList *slotlist = NULL;
 	GHashTable *slots = NULL;
+	GHashTable *bootnames = NULL;
```
ejoerns, line 38: If we use autoptr here, we could eliminate the freeing below

```suggestion
	g_autoptr(GHashTable) bootnames = NULL;
```

test/config_file.c
```diff
@@ -303,7 +303,7 @@ bootname=slot0\n\
 [slot.child.0]\n\
 device=/dev/null\n\
 parent=parent.0\n\
-bootname=slot0\n";
+bootname=slotchild0\n";
```
ejoerns, line 306: Shouldn't this be part of the previous commit?

src/install.c
jluebbe: I suspect that there might be configs which have that setup. And the intent is clear, so we can just fix it at runtime (maybe print a warning when performing the fixup).

src/config_file.c
danc86: Okay, I'll have to read about g_autoptr. It's been quite a few years since I wrote GLib code and back then it didn't have autoptr.

I suppose the autoptr might be useful for fixing the other leaks in the error paths too, so I can try that in a separate PR.

test/config_file.c
danc86: Ah, it could be, but these three commits are how I wrote it in order. I thought it was kind of interesting that the second commit shows that the test case from the earlier one is no longer a valid reproducer for this error message.

src/install.c
danc86: Sure, makes sense. I will update the PR tomorrow.

danc86:
I updated the branch although I just noticed I missed signoff on the new commit, so one more update to come...

danc86:
Okay. Updated to use g_autoptr() instead of manually freeing, and added a commit which flattens out any nested parent relationships.

jluebbe:
I've added support for loop detection, updated the documentation and rebased to fix the merge conflict.

PR #592, opened 2020-05-27
src/update_handler: set RAUC_SLOT_BOOTNAME for child slots too

danc86:
Otherwise it is tempting to try and set the bootname on the child slots too, as in #588.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -643,6 +643,10 @@ The following environment variables will be passed to the hook executable:
   ``RAUC_SLOT_BOOTNAME``
     For slots with a bootname (those that can be selected by the bootloader),
     the bootname of the currently installed slot, e.g. ``"system1"``
+    For slots with a parent, the parent's bootname is used.
+    Note that in many cases, it's better to use the explicit ``RAUC_SLOT_NAME``
+    to select different behaviour in the hook, than to rely indirectly on the
+    bootname.
 
   ``RAUC_SLOT_PARENT``
     If set, the parent of the currently installed slot, e.g. ``"rootfs.1"``
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -738,7 +738,11 @@ static gboolean run_slot_hook(const gchar *hook_name, const gchar *hook_cmd, Rau
 	g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_CLASS", slot->sclass, TRUE);
 	g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_TYPE", slot->type, TRUE);
 	g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_DEVICE", slot->device, TRUE);
-	g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_BOOTNAME", slot->bootname ?: "", TRUE);
+	if (slot->parent) {
+		g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_BOOTNAME", slot->parent->bootname ?: "", TRUE);
+	} else {
+		g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_BOOTNAME", slot->bootname ?: "", TRUE);
+	}
 	g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_PARENT", slot->parent ? slot->parent->name : "", TRUE);
 	if (slot->mount_point) {
 		g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_MOUNT_POINT", slot->mount_point, TRUE);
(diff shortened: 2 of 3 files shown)


jluebbe:
This looks reasonable, but I'd like to know why you'd need to know the bootname in a child slot update handler.

danc86:
In our particular use case, one of the child slots is a [Boot Loader Specification](https://systemd.io/BOOT_LOADER_SPECIFICATION/) config file which is written to a per-slot filename in a shared /boot partition, to instruct the boot loader how to boot each slot. The bundle contains just a generic BLS config (since it doesn't know which slot it will be installed to) and then a slot-post-install hook substitutes the actual bootname into the config file after it is written out.

I'll admit it is a bit of an abuse of the idea of a "slot" (it's just a file not a block device, and the image is not the actual desired contents, more of a template) but it's what we have now and this seemed like a relatively simple change that would keep it working in light of #588.

I suppose the alternative for us would be to have the rootfs slot-post-install hook write out the complete BLS config based on a template embedded in the hook script, or use a totally custom handler or something like that.

danc86:
Seems like I broke one test case as well:

```
ERROR: test/install.test - Bail out! rauc:ERROR:test/install.c:1188:install_test_already_mounted: assertion failed (ierror == NULL): Installation error: Failed updating slot bootloader.0: failed to run slot hook: Child process exited with code 2 (g-spawn-exit-error-quark, 2)
```

ejoerns:
@danc86  It's an invalid assumption that a slot will always have a bootname set.

Consider a bootloader slot for example. This has neither any relation to another slot as it is not part of the A+B setup, but will also not be bootable in the meaning of  `ST_BOOTED`.

Thus setting `RAUC_SLOT_BOOTNAME` whenever a bootname is available sounds reasonable for me. But relying on having this always available in the tests is wrong.

Also note that, for consistency reasons, I'd vote for creating the environment variable only when a bootslot could be found.

danc86:
Hmm right, you mean this bit I added in the test hook?

```
		test -n "$RAUC_SLOT_BOOTNAME" || die_error "missing RAUC_SLOT_BOOTNAME"
```

I was being lazy there and hoping that would be an adequate way of ensuring that my patch was filling in the RAUC_SLOT_BOOTNAME for the child slots. But of course as you say, the configuration could have a slot with no bootname, and I guess that is the cause of that broken test.

The tests passed for me when I ran `make check`, does it mean the Travis job runs some extra tests?

Any suggestions about the best way to update the tests to ensure that RAUC_SLOT_BOOTNAME *is* set for child slots, without breaking the other cases where there is no bootname?

danc86:
> Also note that, for consistency reasons, I'd vote for creating the environment variable only when a bootslot could be found.

Sorry, I don't understand this part. Does a "bootslot" mean a slot with a bootname set?

ejoerns:
> Sorry, I don't understand this part. Does a "bootslot" mean a slot with a bootname set?

What I wanted to say is that you set `RAUC_SLOT_BOOTNAME` unconditionally. Thus even if neither the slot nor its parent has a `bootname` set. I wanted to note that this is inconsistent how the other environment variables are set where the slot parameter is optional. You see there are some conditionals that set environment variables only when they are met and otherwise these environment variables will never be created.

But rethinking this I noticed that `RAUC_SLOT_PARENT` is also set unconditionally. Thus not as inconsistent as I thought. The conditional setting is applied only to information not retrieved from the config file (`if (slot->mount_point) {…`, `if (image) {…`).

The other point, when rethinking this, is that it is the first place where we set an env variable to a value that is part of the configuration of this specific slot. This, for example, would crash (existing) hook scripts that check for the slot being the one with the bootname. Despite I do not have a use-case for this at hand, setting the bootname for childs would mean a loss of information. And you cannot replace it by checking for not having parents etc. as this would be valid for a bootloader slot, too.

@jluebbe What do you think about that?

danc86:
The existing code already sets `RAUC_SLOT_BOOTNAME` unconditionally, this expression: `slot->bootname ?: ""` just leaves the value as the empty string if the slot has no bootname, right?

So the only change here is for child slots, where previously `RAUC_SLOT_BOOTNAME` was the empty string, it will now be set to their parent's bootname.

However you are quite right, this is backwards incompatible with existing hooks which are *relying* on that variable being empty for child slots. I didn't think about that. Is it likely to be a concern in practice though?

jluebbe:
If the bootname is set for the parent, I don't see it as problematic to pass to child slot hooks as well. Use-cases similar to @danc86's seem reasonable. :)

That behavior should be documented, though. The hook can look at the RAUC_SLOT_CLASS or _NAME to know which slot it's currently working on.

danc86:
I can update the PR to clarify that RAUC_SLOT_BOOTNAME will be set for child slots too.

But I am not sure how I can cover this change in tests, without breaking the hook for tests which don't set a bootname at all..

ejoerns:
@jluebbe Just wanted to warn that this can potentially be a compatibility issue and that it is a bit inconsistent with the others.

I agree that one *should* use the `_CLASS` / `_NAME` information, but you know that people never only do what is sensible.. ;)

However, it will only be problematic when trying to install an older bundle that uses these hooks to a newer RAUC. I assume this to be a rarer and mostly fixable case.

Thus, with clear documentation, ok..

jluebbe:
I've rebased this on master without changes and will write some documentation.

PR #593, opened 2020-05-27
improve testing of NAND and UBIFS slots

jluebbe:
With some additional setup in qemu, we can start testing these update_handlers.

### DIFF
--- a/test/update_handler.c
+++ b/test/update_handler.c
@@ -261,9 +261,28 @@ static void test_update_handler(UpdateHandlerFixture *fixture,
 	if (!test_running_as_root())
 		return;
 
+	g_test_message("installing '%s' image to '%s' slot", test_pair->imagetype, test_pair->slottype);
+
 	/* prepare image and slot information */
 	imagename = g_strconcat("image.", test_pair->imagetype, NULL);
-	slotpath = g_build_filename(fixture->tmpdir, "rootfs-0", NULL);
+	if (g_strcmp0(test_pair->slottype, "ubivol") == 0||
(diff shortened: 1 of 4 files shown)


jluebbe:
> Thus we should either make this configurable or clearly note that `mtd-utils` are required for running the tests (documenting it should also be done if it would only be optional, there is an already-existing list in REAMDE.rst).

I've added it to the README. It was already optional (see the check for ubiattach in ./qemu-test-init).

test/update_handler.c
```diff
@@ -261,9 +261,28 @@ static void test_update_handler(UpdateHandlerFixture *fixture,
...
+		if (!slotpath) {
+			g_test_message("no UBI volume for testing found (define RAUC_TEST_MTD_UBIVOL)");
+			g_test_skip("RAUC_TEST_MTD_UBIVOL undefined");
+			return;
+		}
+	} else if (g_strcmp0(test_pair->slottype, "nand") == 0) {
+		slotpath = g_strdup(g_getenv("RAUC_TEST_MTD_NAND"));
+		if (!slotpath) {
+			g_test_message("no MTD NAND device for testing found (define RAUC_TEST_MTD_NAND)");
+			g_test_skip("RAUC_TEST_MTD_NAND undefined");
```
ejoerns, line 280: `RAUC_TEST_MTD_NAND` is set unconditionally in `qemu-test-init`, this should get a simliar check as we have for `RAUC_TEST_MTD_UBI` already. Otherwise it fails if `flash_erase` (from mtd-utils) is not present.

PR #597, opened 2020-06-05
docs/faq: document bundle format compatibility

jluebbe:
This came up on the mailing list.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -40,3 +40,28 @@ or later with::
 
 Note that even if there is only an ext4 slot type available, potentially each
 file system mountable as ext4 should work (with the filename suffix adapted).
+
+Is the RAUC bundle format forwards/backwards compatible?
+--------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -40,3 +40,28 @@ or later with::
...
+--------------------------------------------------------
+
+The basic bundle format has not changed so far (squashfs containing images and
+the manifest, with a CMS signature), which means that newer versions can
+install old bundles.
+Going forward, any issue with installing old bundles would be considered a bug.
+
+Newer RAUC versions have added features and slot types, though (such as casync,
+eMMC boot partitions, MBR/GPT partition switching).
+If you use those features, older versions of RAUC won't be able to install
```
ejoerns, line 54: Maybe we could be more precise here what 'won't be able' means?

```suggestion
If you use those features, older versions of RAUC that cannot handle them will refuse to install the bundle.
```

PR #612, opened 2020-07-15
Fix install printout for non-service configuration

ejoerns:
The CLI for the non-service configuration case was not well tested and has some printout issues this PR addresses.


jluebbe:
We discussed this internally. For the non-service case, we don't want to have the progress bar, as it's to easy to confuse it with other log messages.

ejoerns:
I've removed progress bar support from `send_progress_callback()` and disabled `--progress` argument for non-service builds.

jluebbe:
Rebased on master.

PR #613, opened 2020-07-15
Rework parts of context and config handling for more targeted error handling.

ejoerns:
The current set up of the context does not really allow handling errors gracefully. A configuration error will lead to a simple `g_error()` and core dump which is not optimal user experience.

Additionally, the context setup actually requires a configuration only in a few cases: Mainly for the service and for the install operations in a non-service build. The current handling does not enforce using a configuration but falls back to unusable defaults for these cases. This also prevents from printing more targeted error messages.

This PR aims to fix both at once.

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -167,6 +167,26 @@ RaucContext *r_context_conf(void);
 const RaucContext *r_context(void);
 
 /**
+ * Sets up global context.
+ *
+ * Removes 'pending' flag from the context object.
+ *
+ * * Loads RAUC configuration file (system.conf)
+ * * Reads basic system information like variant, system info, etc. (if
+ *   configured)
+ * * Reads 'bootname' from kernel commandline
+ * * Overrides config file values by commandline argument values where required.
+ *
+ * Note: Must not be called when context is 'busy'.
+ *
+ * @param[out] error Return location for a GError, or NULL
+ *
+ * @return TRUE if context configuration succeeded, otherwise FALSE
+ */
+gboolean r_context_configure(GError **error)
+G_GNUC_WARN_UNUSED_RESULT;
+
+/**
  * Cleans up provided RContextInstallationInfo.
  *
  * @param[in] info RContextInstallationInfo to free
--- a/include/context.h
+++ b/include/context.h
@@ -69,10 +69,6 @@ typedef struct {
 	gint last_explicit_percent;
 } RaucProgressStep;
 
-gboolean r_context_get_busy(void)
-G_GNUC_WARN_UNUSED_RESULT;
-void r_context_set_busy(gboolean busy);
-
 /**
  * Call at the beginning of a relevant code block. Provides progress
  * information via DBus when rauc service is running.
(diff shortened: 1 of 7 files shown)


src/main.c
```diff
@@ -210,6 +210,13 @@ static gboolean install_start(int argc, char **argv)
 
 	r_exit_status = 1;
 
+#if ENABLE_SERVICE == 1
```
jluebbe, line 213: Why do we only default to `/etc/rauc/system.conf` if the service enabled? It's more relevant in the other case.

src/main.c
```diff
@@ -1557,6 +1564,13 @@ static gboolean status_start(int argc, char **argv)
 	g_debug("status start");
 	r_exit_status = 0;
 
+#if ENABLE_SERVICE == 1
```
jluebbe, line 1567: Here as well. Also I'd prefer `if (!ENABLE_SERVICE)` as below for better compile coverage.

test/service.c
```diff
@@ -269,6 +269,11 @@ static void service_test_install_api(ServiceFixture *fixture, gconstpointer user
 		return;
 	}
 
+	/* needs to run as root */
+	if (!test_running_as_root())
+		return;
+
```
jluebbe, line 275: Redundant empty line.

src/context.c
```diff
@@ -219,7 +219,7 @@ gboolean r_context_configure(GError **error)
 	g_clear_pointer(&context->config, free_config);
 	if (context->configpath) {
 		if (!load_config(context->configpath, &context->config, &ierror)) {
-			g_propagate_error(error, ierror);
+			g_propagate_prefixed_error(error, ierror, "Failed to load system config (%s): ", context->configpath);
```
jluebbe, line 222: Shouldn't this be squashed into the previous commit?

jluebbe (changes requested):
Just minor things. If you agree with the suggestions, I can update the PR accordingly.

include/context.h
```diff
@@ -125,11 +125,57 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC(RaucProgressStep, r_context_free_progress_step);
 
 void r_context_register_progress_callback(progress_callback progress_cb);
 
+/**
+ * Return if context is marked 'busy'.
+ *
+ * @return TRUE if context is 'busy', otherwise FASLE.
```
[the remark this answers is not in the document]
ejoerns: Fixed, thx!

src/main.c
ejoerns: Seems as if I accidentally inverted the logic, yes.

src/main.c
ejoerns: Fine with that.

test/service.c
ejoerns: removed

src/context.c
ejoerns: indeed. -> squashed.

ejoerns:
> Just minor things. If you agree with the suggestions, I can update the PR accordingly.

Missed that part when starting to fix things, but as these were only minor changes it should not matter ;)

I've pushed an update now containing the requested changes.

jluebbe (approved):

PR #615, opened 2020-07-15
src/network,src/update_handler: Add Proxy Support

sergeysedoy97:
This PR adds support for RAUC_CURLOPT_PROXY environment to set the corresponding curl option, also it forwards the RAUC_CURLOPT_PROXY environment value to casync via CASYNC_CURLOPT_PROXY (https://github.com/systemd/casync/pull/241) environment. The proxy needs to secure an organization's internal network from external threats.

Example:
```
[Service]
Environment="RAUC_CURLOPT_PROXY=<[protocol://][user:password@]proxyhost[:port]>"
```

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -106,6 +106,10 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 	/* decode all supported Accept-Encoding headers */
 	curl_easy_setopt(curl, CURLOPT_ACCEPT_ENCODING, "");
 
+	if (getenv("RAUC_CURLOPT_PROXY") != NULL) {
+		curl_easy_setopt(curl, CURLOPT_PROXY, getenv("RAUC_CURLOPT_PROXY"));
+	}
+
 	/* set error buffer empty before performing a request */
 	errbuf[0] = 0;
 
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -266,6 +266,10 @@ static gboolean casync_extract(RaucImage *image, gchar *dest, int out_fd, const 
 	if (tmpdir)
 		g_subprocess_launcher_setenv(launcher, "TMPDIR", tmpdir, TRUE);
 
+	if (getenv("RAUC_CURLOPT_PROXY") != NULL) {
+		g_subprocess_launcher_setenv(launcher, "CASYNC_CURLOPT_PROXY", getenv("RAUC_CURLOPT_PROXY"), TRUE);
+	}
+
 	sproc = r_subprocess_launcher_spawnv(launcher, args, &ierror);
 	if (sproc == NULL) {
 		g_propagate_prefixed_error(


src/network.c
```diff
@@ -105,6 +105,13 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 	curl_easy_setopt(curl, CURLOPT_MAXFILESIZE_LARGE, xfer->limit);
 	/* decode all supported Accept-Encoding headers */
 	curl_easy_setopt(curl, CURLOPT_ACCEPT_ENCODING, "");
+	
+	if (getenv("RAUC_CURLOPT_PROXY") != NULL) {
+		curl_easy_setopt(curl, CURLOPT_PROXY, getenv("RAUC_CURLOPT_PROXY"));
+	}
+	if (getenv("RAUC_CURLOPT_PROXYTYPE") != NULL) {
+		curl_easy_setopt(curl, CURLOPT_PROXYTYPE, getenv("RAUC_CURLOPT_PROXYTYPE"));
```
ejoerns, line 113: `CURLOPT_PROXYTYPE` expects a long, not a string.

And it is recommended to directly derive the proxy type from the CURLOPT_PROY scheme anyway.

src/network.c
```diff
@@ -105,6 +105,13 @@ static gboolean transfer(RaucTransfer *xfer, GError **error)
 	curl_easy_setopt(curl, CURLOPT_MAXFILESIZE_LARGE, xfer->limit);
 	/* decode all supported Accept-Encoding headers */
 	curl_easy_setopt(curl, CURLOPT_ACCEPT_ENCODING, "");
+
```
ejoerns, line 108: Uncrustify reveals that this adds a trailing white space.

src/network.c
sergeysedoy97: Okay, removed CURLOPT_PROXYTYPE.

src/network.c
sergeysedoy97: Okay, removed trailing spaces.

ejoerns:
Thanks for the rework, but please provide a single commit for the change as we don't squash. You can simply force-push your branch. Please put the prefix change in the commit, too.

We'd also need at least a little documentation on this. For this, it would also be good to know a proper use case.

sergeysedoy97:
> Thanks for the rework, but please provide a single commit for the change as we don't squash. You can simply force-push your branch. Please put the prefix change in the commit, too.
> 
> We'd also need at least a little documentation on this. For this, it would also be good to know a proper use case.

Done. The proxy needs to secure an organization's internal network from external threats.

vkuzmichev:
> For this, it would also be good to know a proper use case.
@ejoerns 
I believe this change is required to bypass mandatory non-transparent proxies, which some companies still use.

It is also worth to mention that if rauc is running as a service (e.g. via systemd) it is the service process which should be running with RAUC_CURLOPT_PROXY environment option, for example:
```
[Service]
Environment="RAUC_CURLOPT_PROXY=<[protocol://][user:password@]proxyhost[:port]>"
```

ejoerns:
@sergeysedoy97 Ok, I would accept the PR if you provide a short note in the documentation, also about the potential use case as @vkuzmichev pointed out.

As the corresponding casync PR is not merged yet, I would suggest excluding this change from the PR.

vkuzmichev:
@ejoerns 
After deeper digging into curl/rauc/casync source code and following testing we (me and @sergeysedoy97) have found out that executing rauc service process with environment variable `all_proxy` provides necessary functionality out of the box. :)
Due to [rauc using g_subprocess_launcher_new](https://github.com/rauc/rauc/blob/v1.5.1/src/update_handler.c#L263) and [casync using fork and execv/execvp](https://github.com/systemd/casync/blob/bd8898ed92685e12022dd33a04c87786b5262344/src/caremote.c#L1039-L1042) `all_proxy` is properly propagated down to appropriate casync-*tp download helper:
```
root@localhost:~# xargs -0i echo {} </proc/`pidof casync`/environ 
LANG=C.UTF-8
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
INVOCATION_ID=7e19a19e02b549d5bb223a66011c4cc4
JOURNAL_STREAM=9:14746119
RUNTIME_DIRECTORY=/run/rauc
all_proxy=http://192.168.3.1:3128/
GIO_USE_VFS=local
root@localhost:~# xargs -0i echo {} </proc/`pidof casync-https`/environ 
LANG=C.UTF-8
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
INVOCATION_ID=7e19a19e02b549d5bb223a66011c4cc4
JOURNAL_STREAM=9:14746119
RUNTIME_DIRECTORY=/run/rauc
all_proxy=http://192.168.3.1:3128/
GIO_USE_VFS=local
root@localhost:~# 
```

According to [libcurl logic handling environment variables](https://github.com/curl/curl/blob/curl-7_76_1/lib/url.c#L2251-L2325) setting variables `http_proxy` and `https_proxy` should also work.
One can pass this proxy setting to rauc via systemd unit file:
```
[Service]
Environment="all_proxy=<[protocol://][user:password@]proxyhost[:port]>"
```
However, I think this mechanism could be improved with ability to specify/change proxy from ```rauc install``` command so that it can be passed to rauc service process via dbus and propagated to casync.

This PR can be closed.

PR #616, opened 2020-07-15
src/install: fix leaking gchar** with g_auto(GStrv)

ejoerns:
We have to use a type name for freeing and GStrv is the standard glib
typedef prepared for this purpose.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -281,7 +281,7 @@ static RaucSlot *select_inactive_slot_class_member(const gchar *rootclass)
  */
 GHashTable* determine_target_install_group(void)
 {
-	gchar **rootclasses = NULL;
+	g_autofree gchar **rootclasses = NULL;
 	GHashTable *targetgroup = NULL;
 	GHashTableIter iter;
 	RaucSlot *iterslot = NULL;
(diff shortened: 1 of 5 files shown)


src/install.c
```diff
@@ -242,7 +242,7 @@ static gchar** get_all_manifest_slot_classes(const RaucManifest *manifest)
  * Note: This function may be extended to be more sophisticated or follow a
  * certain policy for selecting an appropriate slot!
  *
- * @param rootclass name of root slot class
+ * @param rootclass name of root slot class (must be an intern string)
```
jluebbe, line 245: As we can't easily check if a string is interned, this seems a bit too fragile for the potential performance gain.

ejoerns: Yes, I also already noted that this would require being extra careful.
Thus if at all, It's probably only useful for memory allocations. I've removed the corresponding patch from the set.

ejoerns:
I've added two further fixes, one in the doc of `r_slot_get_root_classes()` and another missing `g_autofree` in `r_status_formatter_readable()`. I've also merged the documentation and code adaptions for each functions in a single commit each.

PR #617, opened 2020-07-16
minor memory fixes

jluebbe:


src/install.c
```diff
@@ -127,8 +127,9 @@ gboolean determine_slot_states(GError **error)
 
 	for (GList *l = slotlist; l != NULL; l = l->next) {
 		g_autofree gchar *realdev = NULL;
-
-		RaucSlot *s = (RaucSlot*) g_hash_table_lookup(r_context()->config->slots, l->data);
+		RaucSlot *s = g_hash_table_lookup(r_context()->config->slots, l->data);
+		if (!s)
```
ejoerns, line 131: we have a couple of places in the code where we already do this, but by using g_assert_nonnul(). I'd like to keep this consistent and would thus prefer the assert over a g_error().

PR #622, opened 2020-07-30
src/network: make the use of .netrc optional

gportay:
Hi Maintainers,

libcurl ignores the `~/.netrc` file by default.

This makes the use of the `~/.netrc` file optional; the information in the URL is preferred.

With this patch, it is possible to hide the login/passwd in the file `~/.netrc` and enables the HTTP Basic Authentication.

This is a minimalist integration; there is no option given to the command line, nor environment variable, nor attribute in the configuration that may enable or disable the use of `~/.netrc`. I can do it if this is wanted.

Regards,
Gaël


jluebbe:
Thanks for the patch! If necessary, we can still add an option for the file name to the system config.

PR #626, opened 2020-08-20
src/main: fix positional arguments in resign usage string

Bastian-Krause:
'rauc resign' expects an existing bundle as first positional argument and the output bundle filename as a second positional argument, just like 'rauc convert' does. Make the usage string reflect that.


ejoerns (approved):
:+1:

jluebbe (approved):

PR #627, opened 2020-08-24
Small documentation fixes

mhei:
I needed to re-read some documentation details and just stumbled over two minor issues.


ejoerns (approved):
We should add this to our mental check list... Thanks for fixing!

PR #628, opened 2020-08-27
Fix systemd D-Bus handling and RAUC error reporting

ejoerns:
When shutting down a system running RAUC, one can see RAUC failing regularly with an exit 1 code and a misleading error message:

``` 
rauc[197]: Connection to the bus can't be made for de.pengutronix.rauc
systemd[1]: Stopping D-Bus System Message Bus...
systemd[1]: Stopping Rauc Update Service...
[...]
systemd[1]: rauc.service: Main process exited, code=exited, status=1/FAILURE
systemd[1]: rauc.service: Failed with result 'exit-code'.
```

This PR aims to solve both the systemd behavior as well as the error message.

### DIFF
--- a/data/rauc.service.in
+++ b/data/rauc.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=Rauc Update Service
 Documentation=https://rauc.readthedocs.io
+After=dbus.service
 
 [Service]
 Type=dbus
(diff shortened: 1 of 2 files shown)


michaelolbrich (approved):
Looks good to me.

jluebbe (approved):

PR #629, opened 2020-09-04
docs: advanced: Fix Keyring move example script

Urist-McGit:
The given script moves the keyring files on the currently active slot
instead of the target slot.

Signed-off-by: Christoph Steiger <c.steiger@lemonage.de>


ejoerns (approved):
Looks valid for me as the hook should actually move the files on the just-installed slot instead of the system we currently install from.

Good catch!

jluebbe (approved):

PR #630, opened 2020-09-07
src/update_handler: Allow overriding the configured casync store path

### LINKED ISSUE #606: casync chunk store URL handling improvements
For our current project we're looking at using RAUC with casync. Since we might have to use bundle and chunk store URLs with a query I had a look at the RAUC code to see how that's handled.

It seems like the [code that works out the chunk store path](https://github.com/rauc/rauc/blob/5dba83af16e4a0567a3f9bfce5e12fc90c7e8b46/src/bundle.c#L746) always treats the paths as if they're local files, even if they're URLs. So if you have a bundle URL like `https://something.com/update.raucb?key=12345` then the chunk store will be `https://something.com/update.raucb?key=12345.castr`, which is unlikely to work.

Unfortunately we might well need to use URLs with a query part because of stupid backend systems, so we'll need to find a solution to the problem.

It looks like a workaround might be to handle the URL parsing in our own code and modify the `[casync]storepath` setting in the system configuration file. However, proper URL parsing in RAUC would be a much cleaner solution.

kevinior:
This commit adds a command-line argument `--store-path` to the install
command and the `InstallBundle()` D-Bus method.

The argument overrides the casync store path that is configured (or the
default based on the bundle filename if not configured).

Fixes #606 

Signed-off-by: Kevin O'Rourke <dev@caboose.org.uk>

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -764,6 +764,8 @@ IN a{sv} *args*:
         installation of bundles on platforms that a compatible not matching the one
         of the bundle to be installed
 
+    :STRING 'store-path', VARIANT 's': Override the configured casync store path
+
 .. _gdbus-method-de-pengutronix-rauc-Installer.Install:
 
 The Install() Method
--- a/include/context.h
+++ b/include/context.h
@@ -14,6 +14,7 @@ typedef void (*progress_callback) (gint percentage, const gchar *message,
 typedef struct {
 	/* The bundle currently mounted by RAUC */
 	RaucBundle *mounted_bundle;
+	gchar *store_path;
 } RContextInstallationInfo;
 
 typedef struct {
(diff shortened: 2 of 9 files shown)


jluebbe:
@RyuzakiKK Is something like this also relevant for you? If so, we should perhaps pick this up again, with some minor changes (i.e. `casync-store-path` instead of `store-path`).

RyuzakiKK:
> @RyuzakiKK Is something like this also relevant for you? If so, we should perhaps pick this up again, with some minor changes (i.e. `casync-store-path` instead of `store-path`).

No, at the moment it's not because we always place the store next to the rauc bundle file URL.

Anyway, that option would probably make rauc more flexible.

PR #637, opened 2020-10-02
Minor cleanups and test improvements

jluebbe:


include/bootchooser.h
```diff
@@ -20,7 +20,8 @@ GQuark r_bootchooser_error_quark(void);
  *
  * @return TRUE if successful, FALSE if failed
  */
-gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error);
+gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
+G_GNUC_WARN_UNUSED_RESULT;
```
ejoerns, line 24: I agree this is useful to prevent unintentionally ignoring failures in methods and indirectly enforce error checking a bit stricter. I would love to see a line in the commit message about the value added by these changes to give other developers a good hint. Please also note that you adapted the test functions in order to fix compile errors now.

src/install.c
```diff
@@ -410,33 +410,41 @@ GList* get_install_images(const RaucManifest *manifest, GHashTable *target_group
...
+		message = g_strdup_printf("Error: %s", joined);
+	} else if (g_strcmp0(split[1], "debug") == 0) {
+		g_autofree gchar *joined = g_strjoinv(" ", &split[2]);
+		message = g_strdup_printf("Debug: %s", joined);
 	} else {
-		g_print("Unknown command: %s\n", split[1]);
+		message = g_strdup_printf("Unknown handler output: %s", line);
 	}
+
+	g_print("Custom Handler: %s\n", message);
```
ejoerns, line 446: With this, the extra log lines will be emitted redundantly as now both the `install_args_update` as well as the `g_print` can print the same message. Is this intentional? Also, note that we use g_print() preferably in the frontend for user output while the library functions us  g_message()` and similar.

While the `g_print()` prepends a 'Custom Handler: ' the messages emitted by `install_args_update` do not give a hint about the scope of these messages. Maybe one should add a prefix there, too?
This method is used both for the full custom handler as well as (more frequently used) pre-install and post-install handler.

src/main.c
```diff
@@ -1961,7 +1972,7 @@ static void cmdline_handler(int argc, char **argv)
 		if (bootslot)
 			r_context_conf()->bootslot = bootslot;
 		if (handlerextra)
```
ejoerns, line 1974: This condition should never evaluate to true
```suggestion
		if (handler_arrgs)
```

ejoerns (changes requested):

include/bootchooser.h
jluebbe: OK.

src/main.c
jluebbe: OK.

src/install.c
jluebbe: I've removed the print. The message is shown via `install_notify()` in the frontend and via `service_install_notify()` on the service side.

ejoerns (approved):

PR #638, opened 2020-10-04
More usage-friendly and explicitly-handled error and info printouts for 'rauc status' (subcommands)

ejoerns:
Calling 'rauc status' from the command line still produced some noise and used logging methods instead of `g_print()`/`g_printerr()` as we otherwise use halfway consistently in the CLI frontend.

We also do not need to call the boot selection backend when only invoking `mark-*` subcommands.


ejoerns:
Rebased onto master to fix conflict in `src/main.c`.

jluebbe:
@ejoerns the status handler function is becoming somewhat hard to follow. Perhaps we should split printing and mark-* into separate functions?

jluebbe:
I've dropped the last commit from this PR, so we can get this merged sooner and discuss it separately.

ejoerns:
Looks fine for me! :+1:

PR #645, opened 2020-10-23
src/service: return D-Bus error if slot states cannot be determined

ejoerns:
For now the service silently ignored if it could not determine slot
states. This led to a D-Bus response with a status message that
did not have the slots "state" member set to a proper value.

The CLI trying to interpret this then ends up in the undefined state of
`r_slot_slotstate_to_str()` that emits a `g_error()`

> (rauc:1523868): rauc-ERROR **: 12:32:50.829: invalid slot status 0

Fix this by not ignoring the error but forwarding it through D-Bus so
that the CLI or other frontends do not try to interpret broken data.

For this we equip `create_slotstatus_array()` with proper `GError` handling
and slightly adapt the flow in `r_on_handle_get_slot_status()` in order to
properly handle this and forward the error as a D-Bus error using
`g_dbus_method_invocation_return_gerror()`.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -179,11 +179,11 @@ gboolean determine_slot_states(GError **error)
 			goto out;
 		}
 
-		g_set_error_literal(
+		g_set_error(
 				error,
 				R_SLOT_ERROR,
 				R_SLOT_ERROR_NO_SLOT_WITH_STATE_BOOTED,
-				"Did not find booted slot");
+				"Did not find booted slot (matching '%s')", r_context()->bootslot);
 		goto out;
 	}
 
--- a/src/install.c
+++ b/src/install.c
@@ -200,19 +200,30 @@ gboolean determine_boot_states(GError **error)
 {
 	GHashTableIter iter;
 	RaucSlot *slot;
+	gboolean had_errors;
 
 	/* get boot state */
 	g_hash_table_iter_init(&iter, r_context()->config->slots);
 	while (g_hash_table_iter_next(&iter, NULL, (gpointer*) &slot)) {
 		GError *ierror = NULL;
 
-		if (slot->bootname && !r_boot_get_state(slot, &slot->boot_good, &ierror)) {
-			g_propagate_error(error, ierror);
-			return FALSE;
+		if (!slot->bootname)
+			continue;
+
+		if (!r_boot_get_state(slot, &slot->boot_good, &ierror)) {
+			g_message("Failed to get boot state of %s: %s", slot->name, ierror->message);
+			had_errors = TRUE;
 		}
 	}
 
-	return TRUE;
+	if (had_errors)
+		g_set_error_literal(
+				error,
+				R_SLOT_ERROR,
+				R_SLOT_ERROR_NO_SLOT_WITH_STATE_BOOTED,
+				"Could not determine all boot states");
+
+	return !had_errors;
 }
 
 /* Returns NULL-teminated intern string array of all classes listed in
--- a/src/install.c
+++ b/src/install.c
@@ -152,6 +152,41 @@ gboolean determine_slot_states(GError **error)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


a3f (changes requested):

src/service.c
```diff
@@ -312,22 +312,30 @@ static GVariant* create_slotstatus_array(void)
 	GHashTableIter iter;
 	RaucSlot *slot;
 
-	g_return_val_if_fail(r_installer, NULL);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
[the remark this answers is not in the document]
ejoerns: Yep. An that despite I copy-pasted this line so carefully ;)

jluebbe (dismissed):

jluebbe:
We get a test failure:
`
GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._r_5fbootchooser_5ferror_5fquark.Code20: Failed to determine boot states: Failed to get state of rescue.0: Variable ORDER not set in grub environment (r_bootchooser_error_quark, 20)
`

jluebbe:
@ejoerns The tests show a failure and I'm not sure about the squash! commit.

src/install.c
```diff
@@ -152,21 +152,6 @@ gboolean determine_slot_states(GError **error)
...
-			g_debug("Found booted slot: %s on %s", s->name, s->device);
-		} else if (s->parent && s->parent == booted) {
-			s->state = ST_ACTIVE;
-		} else {
-			s->state = ST_INACTIVE;
-		}
-	}
-
 	if (!booted) {
 		if (g_strcmp0(r_context()->bootslot, "/dev/nfs") == 0) {
```
jluebbe, line 156: Shouldn't the slots be marked as inactive in this case? Otherwise, we won't be able to install to them. We probably need to handle an external boot differently than a failure to determine slot states.

jluebbe (changes requested):
Looks good otherwise.

src/install.c
ejoerns: Indeed, yes. We explicitly need to set them to 'invalid'. This also does not seem to be covered by any test, yet.

I'll update the PR accordingly.

ejoerns:
The functional code change I pushed is:

``` diff
diff --git a/src/install.c b/src/install.c
index 7320293a..1a43983c 100644
--- a/src/install.c
+++ b/src/install.c
@@ -153,14 +153,28 @@ gboolean determine_slot_states(GError **error)
        }
 
        if (!booted) {
+               gboolean extboot = FALSE;
+
                if (g_strcmp0(r_context()->bootslot, "/dev/nfs") == 0) {
                        g_message("Detected nfs boot, ignoring missing active slot");
-                       res = TRUE;
-                       goto out;
+                       extboot = TRUE;
                }
 
                if (g_strcmp0(r_context()->bootslot, "_external_") == 0) {
                        g_message("Detected explicit external boot, ignoring missing active slot");
+                       extboot = TRUE;
+               }
+
+               if (extboot) {
+                       /* mark all as inactive */
+                       g_debug("Marking all slots as 'inactive'");
+                       for (GList *l = slotlist; l != NULL; l = l->next) {
+                               RaucSlot *s = g_hash_table_lookup(r_context()->config->slots, l->data);
+                               g_assert_nonnull(s);
+
+                               s->state = ST_INACTIVE;
+                       }
+
                        res = TRUE;
                        goto out;
                }

```

The additional check `test_install_image_selection_boot_external` can be found in `test/install.c`

ejoerns:
This should also be part of 1.6 too IMHO, especially the first two patches as I've seen me and others running into this quite a few times.

ejoerns:
Still struggling with my broken .vimrc as it seems. Pushed uncrustify fixes.

jluebbe (approved):

PR #648, opened 2020-10-26
various cleanups

jluebbe:
This increases g_autofree and g_autoptr usage in tests to make actual leaks easier to find. Also we add some more runtime checks and print the signer name after signature verification.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,8 @@ config.h.in
 config.status
 configure
 libtool
-rauc
-rauc.map
+/rauc
+/rauc.map
 stamp-h1
 .deps
 .dirstamp
--- a/test/bundle.c
+++ b/test/bundle.c
@@ -98,9 +98,9 @@ static void bundle_fixture_tear_down_autobuilder2(BundleFixture *fixture,
 static void test_check_empty_bundle(BundleFixture *fixture,
 		gconstpointer user_data)
 {
-	gchar *bundlename;
-	RaucBundle *bundle = NULL;
-	GError *ierror = NULL;
+	g_autofree gchar *bundlename = NULL;
+	g_autoptr(RaucBundle) bundle = NULL;
+	g_autoptr(GError) ierror = NULL;
 	gboolean res = FALSE;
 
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -19,8 +19,8 @@ config.h.in
 config.status
 configure
 libtool
-rauc
-rauc.map
```
ejoerns, line 23: Did you run into any issue with this? Or just because this is the default build location for these files?

Nit: what is meant here is probably als more *'ignore rauc and rauc.map **only** in the top-level dir'* (as we still ignore other files in the top-level dir, too).

ejoerns (changes requested):
Good work!

However, I would see patches

> src/bundle: check squashfs size during check_bundle …
> src/signature: cleanup print_cert_chain and rename to format_cert_chain …
> src/signature: fix memory leaks during certificate chain string format… …
> src/signature: print signers after signature verification …

as out-of-scope for this series as they are more than simple cleanups and partly slightly change behavior. I'd tend to say to make at least one separate PR for them.

.gitignore
jluebbe: Hmm, I had a rauc directory somewhere which was ignored by mistake, so I wanted to have the .gitignore more specific.

test/bundle.c
```diff
@@ -110,16 +110,14 @@ static void test_check_empty_bundle(BundleFixture *fixture,
 	g_assert_false(res);
 	g_assert_error(ierror, G_IO_ERROR, G_IO_ERROR_PARTIAL_INPUT);
 	g_assert_null(bundle);
```
a3f, line 112: If bundle is supposed to be NULL here, why have a g_autoptr for it?_

ejoerns: I'd say for consistency

jluebbe: Yes, maybe someone will copy this test and change it.

ejoerns (approved):
Looks fine for me now! :+1:

test/bundle.c
a3f: Ok, convinced.

a3f (approved):

PR #649, opened 2020-10-27
Print signer name after verification

jluebbe:
It's useful to have the name of the bundle signer in the logs. Fix some string leaks while updating the code to use an OpenSSL memory BIO.

### DIFF
--- a/include/signature.h
+++ b/include/signature.h
@@ -158,7 +158,7 @@ G_GNUC_WARN_UNUSED_RESULT;
  * @return allocated string containing text representation of certificate chain
  *         (signer and issuer)
  */
-gchar* print_cert_chain(STACK_OF(X509) *verified_chain)
+gchar* format_cert_chain(STACK_OF(X509) *verified_chain)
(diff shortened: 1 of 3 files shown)


ejoerns (commented):
Overall looks good for me! Especially the memory leakage fixes.

Two things to note from my site:
The minor one is that it might be a bit duplicated to print this for 'rauc info' which already dumps the whole verifications chain. But we also already have other log output besides the normal printout in 'rauc info' so this would not make much difference I guess.

The other thing I thought is if the signer information is probably also (or more?) relevant for the verification failure case where we could print something like 

> Failed verifying signature by O = Test Org, CN = Test Org Release-1

to ease debugging.

jluebbe:
If there signature verification fails, the content of the signer attributes is unauthenticated. So I wouldn't want to print it without sanitizing (against format characters, newlines, ...). That should be a separate PR, though.

ejoerns:
That sounds reasonable, yes.

ejoerns (approved):

PR #650, opened 2020-10-27
check squashfs size during check_bundle

jluebbe:
The squashfs should always be padded to 4KiB, but check this explicitly to avoid possible corner cases later.


ejoerns (approved):
I got from the manual of mksquashfs: 

> The resulting filesystem will be padded to a multiple of 4 Kb: this is required for filesystems to be used on block devices.

Thus the check sounds reasonable to me.

PR #654, opened 2020-11-09
src/install: remove checksum verification for files in bundle

jluebbe:
The RAUC bundle is already verified at this point, so there is no need
to verify the checksum of each file individually. This saves time during
the installation.


ejoerns:
As this is the exclusive user for `verify_checksum()`, we should remove this function and the corresponding tests etc, too.

Or do you see anything that speaks against this?

jluebbe:
> As this is the exclusive user for `verify_checksum()`, we should remove this function and the corresponding tests etc, too.
> 
> Or do you see anything that speaks against this?

I'd like to add an optional consistency check again, so I didn't remove that code.

ejoerns (approved):
Ok

PR #661, opened 2020-11-24
Make uboot attempt settings configurable

zonque:
Make it possible to tweak the number of attempts that are set in the U-Boot environment when marking slots active or good.
    
This is helpful for systems that want to have different settings for first boots and consecutive ones.

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -20,8 +20,8 @@ GQuark r_bootchooser_error_quark(void)
 #define BAREBOX_STATE_PRIORITY_PRIMARY	20
 #define UBOOT_FWSETENV_NAME "fw_setenv"
 #define UBOOT_FWPRINTENV_NAME "fw_printenv"
-#define UBOOT_DEFAULT_ATTEMPTS		"3"
-#define UBOOT_ATTEMPTS_PRIMARY		"3"
+#define UBOOT_DEFAULT_ATTEMPTS  3
+#define UBOOT_ATTEMPTS_PRIMARY  3
 #define EFIBOOTMGR_NAME "efibootmgr"
 #define GRUB_EDITENV "grub-editenv"
 
@@ -767,6 +767,8 @@ static gboolean uboot_set_state(RaucSlot *slot, gboolean good, GError **error)
 {
 	GError *ierror = NULL;
 	g_autofree gchar *key = NULL;
+	g_autofree gchar *val = NULL;
+	gint attempts = 0;
 
 	g_return_val_if_fail(slot, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
@@ -810,7 +812,15 @@ set_left:
 
 	key = g_strdup_printf("BOOT_%s_LEFT", slot->bootname);
 
-	if (!uboot_env_set(key, good ? UBOOT_DEFAULT_ATTEMPTS : "0", &ierror)) {
+	if (good) {
+		attempts = r_context()->config->boot_default_attempts;
+		if (attempts <= 0)
+			attempts = UBOOT_DEFAULT_ATTEMPTS;
+	}
+
+	val = g_strdup_printf("%d", attempts);
+
+	if (!uboot_env_set(key, val, &ierror)) {
 		g_propagate_error(error, ierror);
 		return FALSE;
 	}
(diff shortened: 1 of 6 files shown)


zonque:
I'm not entirely certain about the name of the config file keys. Open to suggestions.

ejoerns:
@zonque  Good work, first of all, thanks!

> I'm not entirely certain about the name of the config file keys. Open to suggestions.

We already had plans to do something similar for other bootloader implementation. Thus one option I could think about is having a generic setting for default and primary attempts being valid for both implementations, e.g. `boot_reset_attempts` and `boot_primary_attempts`.

@jluebbe Do you have any opinion on that?

src/bootchooser.c
```diff
@@ -809,8 +808,9 @@ static gboolean uboot_set_state(RaucSlot *slot, gboolean good, GError **error)
 set_left:
 
 	key = g_strdup_printf("BOOT_%s_LEFT", slot->bootname);
+	val = g_strdup_printf("%d", r_context()->config->uboot_default_attempts);
```
ejoerns, line 811: Probably it is better here to move the string conversion to `uboot_env_set` and pass integer arguments instead.

src/config_file.c
```diff
@@ -12,6 +12,9 @@ G_DEFINE_QUARK(r-slot-error-quark, r_slot_error)
 
 #define RAUC_SLOT_PREFIX	"slot"
 
+#define UBOOT_DEFAULT_ATTEMPTS  3
+#define UBOOT_ATTEMPTS_PRIMARY  3
```
ejoerns, line 16: For keeping the bootloader implementation more central, I would propose letting these values reside in `bootchooser.c` and perform the fallback to default values in `uboo_env_set()`.

src/bootchooser.c
zonque: I'm not sure I'm following. `uboot_set_env()` is used for `BOOT_ORDER` as well, which necessarily is a string. Are you suggesting introducing something like `uboot_set_env_int()` or something?

src/config_file.c
zonque: Hmm, and then look at the key again in `uboot_set_env` to infer which variable is being written?

zonque:
> We already had plans to do something similar for other bootloader implementation. Thus one option I could think about is having a generic setting for default and primary attempts being valid for both implementations, e.g. `boot_reset_attempts` and `boot_primary_attempts`.

Sure, this would be nicer. I didn't know whether all boot loaders follow the idea of boot counters for multiple attempts. But of course those which don't would just ignore these settings.

src/bootchooser.c
ejoerns: Ouh, didn't notice we use this for order, too. You're right, then it would become more complicated and would possibly require splitting the function up. Thus the solution you implemented is more straight-forward, yes.

src/config_file.c
ejoerns: I noted this under my wrong assumption that we have a method for explicitly writing the attempts only. I still would prefer having the defaults in the bootloader implementation. One could apply them conditionally when setting the `val =` we had in the discussion above. What do you think?

zonque: Yes, makes sense. Did that now. PTAL!

zonque:
> Sure, this would be nicer. I didn't know whether all boot loaders follow the idea of boot counters for multiple attempts. But of course those which don't would just ignore these settings.

I renamed the config file keys to `boot-attempts` and `boot-attempts-primary` now. Currently only the U-Boot code looks at these variables, but that can be extended trivially.

zonque:
Just FTR: I believed I addressed all the things we talked about. If there is anything else, I'm happy to discuss :)

zonque:
The CI failure seems unrelated?

jluebbe (approved):

ejoerns:
@zonque Thanks for your adaptions and sorry for letting you wait a little longer. We were quite busy with the preparation for the CVE fix and the 1.5 and 1.5.1 release during the last weeks.

zonque:
> @zonque Thanks for your adaptions and sorry for letting you wait a little longer. We were quite busy with the preparation for the CVE fix and the 1.5 and 1.5.1 release during the last weeks.

No problem at all. Thanks a bunch for RAUC, it's become a really essential tool for us! :smile:

PR #662, opened 2020-11-24
various cleanups

jluebbe:


test/bundle.c
```diff
@@ -198,7 +208,18 @@ static void bundle_test_extract_manifest(BundleFixture *fixture,
...
 	g_assert_no_error(ierror);
 	g_assert_true(res);
-	g_assert_true(g_file_test(manifestpath, G_FILE_TEST_EXISTS));
+
+	filepath = g_build_filename(outputdir, "manifest.raucm", NULL);
+	g_assert_true(g_file_test(filepath, G_FILE_TEST_IS_REGULAR));
+	g_clear_pointer(&filepath, g_free);
+
+	filepath = g_build_filename(outputdir, "rootfs.ext4", NULL);
+	g_assert_false(g_file_test(filepath, G_FILE_TEST_EXISTS));
```
ejoerns, line 226: You test for `G_FILE_TEST_IS_REGULAR` in other cases. Is this different here intentionally?

jluebbe: Yes, it should also fail if there is a "non-regular" file instead (such as a link oder device).

ejoerns: Ah, I missed that you assert non-existence here. Thus yes, that should match all kinds of files probably existing under this location.

ejoerns (approved):
:+1:

PR #668, opened 2020-12-10
test/get-coverity: fix error handling

jluebbe:
A || B && C runs C if either A or B succeeds, which happens always in this case.

This fixes aa4676cacbf23e32905b0b7a66662a816f253d0f.


ejoerns (approved):
Ouh, I actually did not notice that in bash (unlike in C and other C-like programming languages) the `&&` operator does not have higher precedence than `||`.

Thanks for fixing!

PR #669, opened 2020-12-11
src/bundle: remove duplicate signature on casync bundle

jluebbe:
As convert_to_casync_bundle() already calls sign_bundle() via create_bundle(),
this call is redundant. Remove it.


ejoerns (approved):
Good catch!

PR #673, opened 2020-12-22
src/mount.c: fix build with kernel < 4.14

ffontaine:
Build with kernel headers < 4.14 fails since version 1.5.0 and https://github.com/rauc/rauc/commit/527bf2f7f746e0253f7843542e19cb0fa0c7c869:

```
src/mount.c: In function 'r_setup_loop':
src/mount.c:201:25: error: 'LOOP_SET_BLOCK_SIZE' undeclared (first use in this function)
  looprc = ioctl(loopfd, LOOP_SET_BLOCK_SIZE, 4096);
                         ^
```

Indeed, `LOOP_SET_BLOCK_SIZE` is only defined since https://github.com/torvalds/linux/commit/89e4fdecb51cf5535867026274bc97de9480ade5

Fixes:
 - http://autobuild.buildroot.org/results/829ae7ed66686c11a941ac99bd08a06f754affb4

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/mount.c
+++ b/src/mount.c
@@ -11,6 +11,10 @@
 #include "mount.h"
 #include "utils.h"
 
+#ifndef LOOP_SET_BLOCK_SIZE
+#define LOOP_SET_BLOCK_SIZE 0x4C09
+#endif
+
 gboolean r_mount_bundle(const gchar *source, const gchar *mountpoint, GError **error)
 {
 	const unsigned long flags = MS_NODEV | MS_NOSUID | MS_RDONLY;


ejoerns:
Note we have this reported in https://github.com/rauc/meta-rauc/issues/169, too.

ejoerns:
The required functionality is not implemented in kernel <v4.14. Thus even if it compiles, the ioctl will fail during runtime with `-EINVAL`. Thus it will only work if used kernel headers are older than the actually used kernel.

jacmet:
@ejoerns so now rauc requires kernel 4.14+? Really? That is pretty unfortunate for a security fix :/

jacmet:
@ejoerns why is setting the blocksize to 4K so important? The code explicitly is written to continue if it fails:

		g_debug("Failed to set loop device block size to 4096, continuing");

Is this only a requirement for the new dm-verity based format?

ejoerns:
> @ejoerns why is setting the blocksize to 4K so important? The code explicitly is written to continue if it fails:
> 
> ```
> 	g_debug("Failed to set loop device block size to 4096, continuing");
> ```

Had a second look at this. As this is actually the only code using it, it should not hurt if the ioctl fails. Thus having only the define for making it compile (as originally suggested) should work.

> Is this only a requirement for the new dm-verity based format?

No, it is a sanity check for supporting the default block size of SquashFS (which is true for both formats).

jluebbe:
Setting the blocksize is just a performance optimization, which is the reason for ignoring failure. So the fix looks good to me.

It seems the toolchains we tested with were too new.

jacmet:
> Setting the blocksize is just a performance optimization, which is the reason for ignoring failure. So the fix looks good to me.
> 
> It seems the toolchains we tested with were too new.

Ok, great, I'll pick up the patch for the upcoming Buildroot releases then, thanks.

ejoerns:
> Ok, great, I'll pick up the patch for the upcoming Buildroot releases then, thanks.

Good to hear, thanks!

As we consider this a bug, we plan to publish a bugfix version 1.5.1 in early January. The changes we did were quite invasive in some modules and despite we carefully tested this in a number of use cases internally, the changes did not get the community test coverage that a normal master commit would have had.
Thus we might have missed a few scenarios and would like to wait if others encounter compatibility issues for their use cases too and have these collected in the release then.

We'll prepare an update for the [integration](https://github.com/rauc/rauc-1.5-integration) repository with this to allow other users to pick up the patch.

We also work on extending our tests to check against older toolchains etc to prevent a bit better from running into similar issues in the future.

jluebbe (approved):

PR #675, opened 2021-01-04
configure.ac: increase minimum glib version required (2.49.3)

ejoerns:
Since we use `G_VARIANT_DICT_INIT` introduced by f2eda48dde10dded99310122d994d68ec2b5960e ("src/main.c: switch to use new client API") which was first defined in glib 2.49.3, this is now the minimum required glib version.

See: https://gitlab.gnome.org/GNOME/glib/-/blob/master/NEWS#L3180

This affects releases 1.5, 1.4, 1.3.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -59,7 +59,7 @@ AS_IF([test "x$enable_create" != "xno"], [
 ])
 
 # Checks for libraries.
-PKG_CHECK_MODULES([GLIB], [glib-2.0 >= 2.45.8 gio-2.0 gio-unix-2.0])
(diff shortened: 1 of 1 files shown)


jluebbe:
We should make it clear that this commit doesn't actually change the required version, as the referenced commit was in v1.3 already.

ejoerns:
@jluebbe I assumed having this done by the commit message and mentioning the versions back to (including) 1.3 in the PR explicitly, but maybe I can further clarify.

jluebbe (approved):

PR #676, opened 2021-01-04
src/emmc: check the return value of g_fopen

jluebbe:
This could be NULL on error, so handle that case.

### DIFF
--- a/src/emmc.c
+++ b/src/emmc.c
@@ -150,14 +150,14 @@ static gboolean r_emmc_force_part_write(const gchar *device, gchar value, GError
 	g_return_val_if_fail(value == '0' || value == '1', FALSE);
 
 	sysfs_path = g_strdup_printf("/sys/block/%s/force_ro", device_basename);
-	if (!g_file_test(sysfs_path, G_FILE_TEST_EXISTS)) {
+	f = g_fopen(sysfs_path, "w");
+	if (!f) {
+		int err = errno;
 		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
-				"Block device does not have force_ro attribute: %s does not exist",
-				sysfs_path);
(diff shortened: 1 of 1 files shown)


src/emmc.c
```diff
@@ -156,6 +156,12 @@ static gboolean r_emmc_force_part_write(const gchar *device, gchar value, GError
 	}
 
 	f = g_fopen(sysfs_path, "w");
+	if (!f) {
```
ejoerns, line 159: Shouldn't we evaluate `errno` here to provide more targeted error messages?

jluebbe: Yes, and we could merge it with the g_file_test above, which becomes redundant with this check.

src/emmc.c
```diff
@@ -148,15 +148,14 @@ static gboolean r_emmc_force_part_write(const gchar *device, gchar value, GError
...
 
 	sysfs_path = g_strdup_printf("/sys/block/%s/force_ro", device_basename);
-	if (!g_file_test(sysfs_path, G_FILE_TEST_EXISTS)) {
+	f = g_fopen(sysfs_path, "w");
+	if (!f) {
 		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
-				"Block device does not have force_ro attribute: %s does not exist",
-				sysfs_path);
+				"Could not open device attribute %s: %s",
+				sysfs_path, g_strerror(errno));
```
ejoerns, line 155: Please save `errno` with `int err = errno;` as suggested by glib documentation and as we do for other similar calls, too.

jluebbe: I've adjusted the other calls in `emmc.c` as well.

ejoerns (approved):
Looks fine for me now.

Bastian-Krause (approved):

PR #677, opened 2021-01-05
Restore Compatibility With glib 2.50 API and add API Compatibility Checks

ejoerns:
* Optionally define `g_uuid_string_random()` that required to move to glib 2.52 before
* Check for glib API versions required to prevent from unintentionally depending on newer API
* Fix fallout


jluebbe:
I've removed the glib-specific line "Since: 2.52" in the g_uuid_string_random comment and added a check if xfer.dl is set before trying to close it.

jluebbe (approved):

Issue #631, opened 2020-09-14
partitioned loopback mount breaks test suite


ukleinek:
Hello,

on `v1.4` I experienced a failing test suite after:

```console
$ curl -s https://dl.armbian.com/helios64/archive/Armbian_20.08.3_Helios64_buster_legacy_4.4.213.img.xz | xzcat > Armbian_20.08.3_Helios64_buster_legacy_4.4.213.img
$ sudo losetup -P -f Armbian_20.08.3_Helios64_buster_legacy_4.4.213.img
$ sudo mount /dev/loop0p1 /mnt
```

This results in 9 failed tests, namely the tests defined in `test/rauc.t` with the numbers 29 to 34 and 38 to 40.

After unmounting the image all tests succeed as expected again.

ejoerns:
This happens as the image is a disk image and thus the actually mounted loop device is the partition loop device `/dev/loop0p1`.

To look up the underlying file, we access `"/sys/block/<devicename>/loop/backing_file"` which is fully valid for normal loop mounts (where `<devicename>` is e.g. `/dev/loop0`), but not for partitions from a disk loop mount (where `<devicename>`is e.g. `/dev/loop0p1`). A corresponding path does not exist here.

There were some things hiding the actual cause of this issue a little that I fixed (see #681)

But I do not have a good clue on how to handle a partition loopback mount properly. One option would be to simply not match on `/dev/loopXpY` pattern when trying to resolve (currently we only check if the devices path's prefix is `/dev/loop`). Another to explicitly resolve to the corresponding disk image.

PR #681, opened 2021-01-08
Fix and Improve Error Handling for resolve_loop_device()

ejoerns:
* Provides more helpful error message when resolving fails
* adds support for GError
* fixes missing return value check before passing to `find_config_slot_by_device()`
* Improves test debugability

Fallout from debugging behavior described in #631.


ejoerns:
Rebased onto master to ensure compatibility with the recent test suite.

jluebbe (approved):

PR #684, opened 2021-01-19
update_handler: support for nor flash devices using flashcp

Emantor:
Add a handler to write images using flashcp. Can be used to update the
bootloader or other images residing in a flash partition.

TODO:
- [x] adapt qemu tests
- [x] documentation for the "nor" type

Signed-off-by: Ladislav Michl <ladis at linux-mips.org>
[r.czerwinski@pengutronix.de: adapted for current code, added tests]
Signed-off-by: Rouven Czerwinski <r.czerwinski@pengutronix.de>

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -584,7 +584,7 @@ out:
 	return res;
 }
 
-static gboolean nand_format_slot(const gchar *device, GError **error)
+static gboolean flash_format_slot(const gchar *device, GError **error)
 {
 	g_autoptr(GSubprocess) sproc = NULL;
 	GError *ierror = NULL;
@@ -1185,7 +1185,7 @@ static gboolean img_to_nand_handler(RaucImage *image, RaucSlot *dest_slot, const
 
 	/* erase */
 	g_message("erasing slot device %s", dest_slot->device);
-	res = nand_format_slot(dest_slot->device, &ierror);
+	res = flash_format_slot(dest_slot->device, &ierror);
 	if (!res) {
 		g_propagate_error(error, ierror);
 		goto out;
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -584,6 +584,40 @@ out:
 	return res;
 }
 
+static gboolean nor_write_slot(const gchar *image, const gchar *device, GError **error)
+{
+	g_autoptr(GSubprocess) sproc = NULL;
+	GError *ierror = NULL;
+	gboolean res = FALSE;
+	g_autoptr(GPtrArray) args = g_ptr_array_new_full(5, g_free);
+
+	g_ptr_array_add(args, g_strdup("flashcp"));
+	g_ptr_array_add(args, g_strdup(image));
+	g_ptr_array_add(args, g_strdup(device));
+	g_ptr_array_add(args, NULL);
+
+	sproc = r_subprocess_newv(args, G_SUBPROCESS_FLAGS_NONE, &ierror);
+	if (sproc == NULL) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"failed to start flashcp: ");
+		goto out;
+	}
+
+	res = g_subprocess_wait_check(sproc, NULL, &ierror);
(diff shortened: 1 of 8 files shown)


src/update_handler.c
```diff
@@ -584,6 +584,40 @@ static gboolean vfat_format_slot(RaucSlot *dest_slot, GError **error)
...
+{
+	g_autoptr(GSubprocess) sproc = NULL;
+	GError *ierror = NULL;
+	gboolean res = FALSE;
+	g_autoptr(GPtrArray) args = g_ptr_array_new_full(5, g_free);
+
+	g_ptr_array_add(args, g_strdup("flashcp"));
+	g_ptr_array_add(args, g_strdup(image));
+	g_ptr_array_add(args, g_strdup(device));
+	g_ptr_array_add(args, NULL);
```
jluebbe, line 597: Shoudn't this use `--erase-all`? For other slots we make sure control the full resulting contents.

src/update_handler.c
```diff
@@ -1733,6 +1802,7 @@ RaucUpdatePair updatepairs[] = {
 	{"*.ubifs", "ubivol", img_to_ubivol_handler},
 	{"*.ubifs", "ubifs", img_to_ubifs_handler},
 	{"*.img", "ext4", img_to_fs_handler},
+	{"*.img", "flash", img_to_flash_handler},
 	{"*.img", "nand", img_to_nand_handler},
```
jluebbe, line 1806: I'm not sure that `flash` is a good name, as NAND is also a type of flash. Maybe `nor`?

test/update_handler.c
```diff
@@ -280,6 +280,13 @@ static void test_update_handler(UpdateHandlerFixture *fixture,
 			g_test_skip("RAUC_TEST_MTD_NAND undefined");
 			return;
 		}
+	} else if (g_strcmp0(test_pair->slottype, "flash") == 0) {
+		slotpath = g_strdup(g_getenv("RAUC_TEST_MTD_FLASH"));
```
jluebbe, line 284: In the current qemu-test, `RAUC_TEST_MTD_FLASH` points to a nand-sim device, which is incorrect for this use-case (but will probably work, tough). Maybe this can be done with the mtdram driver and the correct module options. Or alternatively qemu might support emulating a NOR flash als on x86.

jluebbe (changes requested):

src/update_handler.c
ejoerns: Yes, this was my first thought, too. I'd also vote for 'nor'

test/update_handler.c
Emantor: >Maybe this can be done with the mtdram driver and the correct module options. Or alternatively qemu might support emulating a NOR flash als on x86.

I looked into NOR-flash emulation for qemu and this seems to be `difficult` so I'd rather use the mtdram driver.

src/update_handler.c
Emantor: Sounds good to me.

src/update_handler.c
Emantor: This already writes the slot, I wonder if I can simply reuse the `nand_format_slot` handler to erase the device before flashing the image.

test/update_handler.c
Emantor: Currently `RAUC_TEST_MTD_FLASH` points to nowhere, the variable for nand flashes is `RAUC_TEST_MTD_NAND`.

Emantor: I adapted the qemu tests to run on the mtdram device. I also added `cat /proc/mtd` to the tests, this should make debugging easier if the device order changes for whatever reason.

src/update_handler.c
Emantor: Did that, please re-review.

jluebbe (dismissed):

jluebbe:
I'll add some documentation.

ejoerns (approved):

Issue #660, opened 2020-11-23
`invalid slot status 0` when using custom bootloader


om26er:
We are trying to enable RAUC with CBoot(Nvidia Bootloader), initally we are looking to get this thing working following https://rauc.readthedocs.io/en/latest/integration.html#custom, at a later stage we do plan to support the implementation in C i.e. upstreaming a RAUC backend for CBoot.

So I wrote a simple shell script to test things, but it says "invalid slot satus 0", not sure what's causing that. Below is a naive script

```
#!/bin/sh

arg1=$1
arg2=$2

echo $@ >> file1

if [ "$arg1" = "get-primary" ]
then
        echo "APP_b"
        exit 0
fi

if [ "$arg1" = "get-state" ]
then
        echo "good"
        exit 0
fi
```

om26er:
system.cfg looks like
```
[system]
compatible=Dev
bootloader=custom

[keyring]
path=/etc/rauc/ca.cert.pem
 
[slot.rootfs.0]
device=/dev/mmcblk0p1
type=ext4
bootname=APP

[slot.rootfs.1]
device=/dev/mmcblk0p2
type=ext4
bootname=APP_b

[handlers]
bootloader-custom-backend=cboot-script
```

om26er:
Ok, I think I might have found the root of the issue here. RAUC looks in the `/proc/cmdline` to detect the currently active slot. In case of Xavier NX, there is nothing that RAUC recognizes

```
root@jetson-xavier-nx-devkit:~# cat /proc/cmdline 
console=ttyTCU0,115200 console=ttyTCU0,115200 console=tty0 fbcon=map:0 video=tegrafb no_console_suspend=1 earlycon=tegra_comb_uart,mmio32,0x0c168000 gpt tegra_fbmem=0x800000@0xa069c000 lut_mem=0x2008@0xa0696000 usbcore.old_scheme_first=1 tegraid=19.1.2.0.0 maxcpus=6 boot.slot_suffix=_b boot.ratchetvalues=0.4.2 vpr_resize sdhci_tegra.en_boot_part_access=1
```

So I think we'd need to change RAUC to "understand" the `boot.slot_suffix=_b` parameter in the cmdline on the Nvidia Jetson platform.

om26er:
This part of the docs also mentions that https://rauc.readthedocs.io/en/latest/integration.html#booted-slot-detection

So I hacked RAUC to set a hard-coded `root=` parameter using the below patch
```
[...]

PR #685, opened 2021-01-19
Add preliminary cboot support

om26er:
We have been using CBoot with RAUC in our test stack for a month and this small patch helped us get there. This is the same patch as mentioned in https://github.com/rauc/rauc/issues/660#issuecomment-733158215

Also note: We currently implement NVIDIA's CBoot support using a custom shell script. At a later stage we plan to contribute CBoot support as a "backend".

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -29,6 +29,7 @@ static const gchar* get_cmdline_bootname(void)
 	g_autofree gchar *contents = NULL;
 	g_autofree gchar *realdev = NULL;
 	const char *bootname = NULL;
+	const char *suffix = NULL;
 
 	if (context->mock.proc_cmdline)
 		contents = g_strdup(context->mock.proc_cmdline);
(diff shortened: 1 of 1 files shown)


om26er:
ping @ejoerns, could you please take a look into that ?

src/context.c
```diff
@@ -52,6 +53,20 @@ static const gchar* get_cmdline_bootname(void)
 			return bootname;
 	}
 
+	/* Support NVIDIA's CBoot bootloader.
+	*
+	*  If /proc/cmdline contains an entry for boot.slot_suffix= with an empty value,
+	*  it implies that A partition is selected. If the value is '_b' then the system
+	*  was booted from B parition. */
+	suffix = regex_match("boot\\.slot_suffix=((\\s+|\\S+))", contents);
```
ejoerns, line 61: Just a guess: If we stick to the idea of introducing a configurable key (instead of fixed `boot.slot_suffic`) here and allow it to match an empty parameter by default, wouldn't that work, too? Maybe we forward that internally as a predefined value, like '`empty`' to allow distinguishing it from an actually empty boot name variable in the further processing (which might be an error in other cases).

@jluebbe What's your opinion on that?

windelbouwman: Are you suggesting to turn the `bootname` key in `system.conf` into a regex? This would make the matching scheme more flexible I guess.

windelbouwman: Another option would be to use a similar scheme as in `bootchooser.c`: execute a shell script when bootloader is set to `custom`.

om26er:
Nvidia is getting rid of CBoot in 2022 https://forums.developer.nvidia.com/t/jetson-software-roadmap-for-2h-2021-and-2022/177725 so we'll continue to carry our patch until it's not needed.

PR #686, opened 2021-01-20
fix checksum generation in rauc convert

jluebbe:
create_bundle() calls update_manifest_checksums(), which would result in
checksums and sizes that refer to the casync index files instead of the actual
image. While this only breaks installation when using UBIFS with casync (which
needs an additional casync patch), the incorrect checksums are still confusing
in the slot status information.

As we don't need already have saved the correctly updated manifest, just call
mksquashfs() and sign_bundle() directly. Reduce the too strict payload size
check to allow casync bundles generated with RAUC <= 1.4.


OnkelUlla (changes requested):
I really like your approach of downgrading the payload check from an error to a normal message in combination with refering to a new FAQ: +1.

Testing with a casync based bundle created by RAUC v1.4 the rauc service failed while preparing the installation as there still is the [assertion](https://github.com/rauc/rauc/blob/master/src/mount.c#L114) introduced into RAUC v1.5.

And there are some minor things in commit messages:

1. In commit "src/bundle: fix casync manifest generation", please replace `s/UBIFS/UBI volumes/` and `s/As we don't need already have saved/As we already have saved/`.
2. In commit "src/bundle: fix too strict payload size check" there is a typo: `s/4kiB/4KiB/`.

ejoerns (approved):

Issue #688, opened 2021-01-21
"bundle is stored on an unkown filesystem" casync bundle failure


om26er:
Our image is ZFS and it seems rauc doesn't like that. Here is the attached log. Running the same stuff on ext4 works fine. Does RAUC have any sort of filesystem allow list ? 
```
ERROR: mc:xavier-nx:xavier-nx-bundle-1.0-r0 do_bundle: Execution of '/home/cbbuilder/.yocto/cfx-hub/tmp-glibc/work/jetson_xavier_nx_devkit-oe-linux/xavier-nx-bundle/1.0-r0/temp/run.do_bundle.2399242' failed with exit code 1:
rauc-Message: 19:59:18.208: Debug log domains: 'rauc'
(rauc:2399245): rauc-DEBUG: 19:59:18.212: bundle start
(rauc:2399245): rauc-DEBUG: 19:59:18.212: system config not found, using default values
(rauc:2399245): rauc-DEBUG: 19:59:18.213: Resolved bootname /dev/mapper/matterhorn--vg-root to /dev/dm-0
(rauc:2399245): rauc-DEBUG: 19:59:18.213: input directory: /home/cbbuilder/.yocto/cfx-hub/tmp-glibc/work/jetson_xavier_nx_devkit-oe-linux/xavier-nx-bundle/1.0-r0/bundle
(rauc:2399245): rauc-DEBUG: 19:59:18.213: output bundle: /home/cbbuilder/.yocto/cfx-hub/tmp-glibc/work/jetson_xavier_nx_devkit-oe-linux/xavier-nx-bundle/1.0-r0/build/bundle.raucb
(rauc:2399245): rauc-DEBUG: 20:00:27.166: Payload size: 1656864768 bytes.
Creating bundle in 'plain' format
rauc-Message: 20:00:36.143: No keyring given, skipping signature verification
(rauc:2399245): rauc-DEBUG: 20:00:36.183: Signature offset: 1656864768 bytes.
(rauc:2399245): rauc-DEBUG: 20:00:36.183: Bundle size: 1656866260 bytes.
rauc-Message: 20:00:36.194: Debug log domains: 'rauc'
(rauc:2401001): rauc-DEBUG: 20:00:36.197: convert start
(rauc:2401001): rauc-DEBUG: 20:00:36.197: system config not found, using default values
(rauc:2401001): rauc-DEBUG: 20:00:36.197: Resolved bootname /dev/mapper/matterhorn--vg-root to /dev/dm-0
[...]

PR #689, opened 2021-01-25
bundle: allow installing bundle from ZFS partitions

zonque:
Add `ZFS_SUPER_MAGIC` as defines in openzfs.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -79,6 +79,10 @@
 #ifndef XFS_SUPER_MAGIC
 #define XFS_SUPER_MAGIC 0x58465342
 #endif
+#ifndef ZFS_SUPER_MAGIC
+/* Taken from https://github.com/openzfs/zfs/blob/master/include/sys/fs/zfs.h#L1198 */
+#define ZFS_SUPER_MAGIC 0x2fc12fc1
+#endif
 
 GQuark
 r_bundle_error_quark(void)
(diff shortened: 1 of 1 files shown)


jluebbe (changes requested):
https://github.com/openzfs/zfs/blob/master/include/sys/fs/zfs.h#L1198 calls it LINUX_ZFS_SUPER_MAGIC. Please add it to the list in https://github.com/rauc/rauc/blob/master/src/bundle.c#L21-L81 with a comment where it was found.

zonque:
> https://github.com/openzfs/zfs/blob/master/include/sys/fs/zfs.h#L1198 calls it LINUX_ZFS_SUPER_MAGIC. Please add it to the list in https://github.com/rauc/rauc/blob/master/src/bundle.c#L21-L81 with a comment where it was found.

Ah, I missed that list. Thanks!

Also, ZFS apparently cannot be checked with the rdev check in bundle.c so we have to set `mount_checked` to TRUE immediately. Will update.

zonque:
Updated. PTAL.

jluebbe (approved):

jluebbe:
This should also help with #688.

@zonque You're actually installing from ZFS on embedded devices? Have you tried the verity bundle format?

zonque:
> @zonque You're actually installing from ZFS on embedded devices?

Yes, we're using it both on x86 and raspberrypis.

> Have you tried the verity bundle format?

No, how would that help?

jluebbe:
> > Have you tried the verity bundle format?
> 
> No, how would that help?

With plain bundles, RAUC uses these checks to protect against the TOCTOU vulnerability:
https://rauc.readthedocs.io/en/latest/advanced.html#protection-against-concurrent-bundle-modification

If it detects a verity bundle, these checks are not necessary and skipped.

PR #698, opened 2021-02-09
Added note in docs to restart the service after changing system.conf

SijmenHuizenga:
Just a little note in the docs that would have helped me a bit. Hopefully it will be useful to someone out there :smile:

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -1055,6 +1055,9 @@ You can start the RAUC service manually by executing::
 
   $ rauc service
 
+Keep in mind that rauc service reads the system.conf during startup and needs to be
+restarted for changes in the system.conf to take affect.
+
 Systemd Integration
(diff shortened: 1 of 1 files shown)


ejoerns:
@SijmenHuizenga There is actually such a hint already, but only in the reference section: https://rauc.readthedocs.io/en/latest/reference.html#system-configuration-file
But I do not see anything that speaks against picking this up in the integration section again.

Before I retrigger the CI build that probably lost an acknowledge, maybe you want to fix the commit title prefix (`Docs:`) to be lowercase (and more consistent with the other commits)?

ejoerns:
FTR: The build passed now, seems to have been a browser cache issue on my side

SijmenHuizenga:
Let me know if there is anything else I can do!

ejoerns (approved):
Looks fine for me.

PR #701, opened 2021-02-16
Make boot slot hooks more versatile

Bastian-Krause:
The i.MX28's ROM code expects an absolute offset to the raw device instead of an offset to a partition in the bootloader's header residing in a partition with a certain type. This is problematic when used in conjunction with `boot-mbr-switch`, because depending on which partition is active, the offset varies.

In order to allow such cases, run the pre install hook after the partition is determined and pass information required to manipulate the partition to the hook. Do the same for the post install hook, but run it prior to the actual atomic MBR switch.

To keep things consistent, apply these changes also to `boot-gpt-switch` and `boot-emmc`.

Please note that I only tested the `boot-mbr-switch` case, everything else is just compile-tested.

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -1495,6 +1495,7 @@ static gboolean img_to_boot_gpt_switch_handler(RaucImage *image, RaucSlot *dest_
 	g_autoptr(GUnixOutputStream) outstream = NULL;
 	GError *ierror = NULL;
 	struct boot_switch_partition dest_partition;
+	g_autoptr(GHashTable) vars = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
 
 	res = r_gpt_switch_get_inactive_partition(dest_slot->device,
 			&dest_partition, dest_slot->region_start,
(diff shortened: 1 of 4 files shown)


Bastian-Krause:
Rebased on master.

src/update_handler.c
```diff
@@ -748,14 +748,17 @@ static gboolean unpack_archive(RaucImage *image, gchar *dest, GError **error)
...
 {
 	g_autoptr(GSubprocessLauncher) launcher = NULL;
 	g_autoptr(GSubprocess) sproc = NULL;
 	g_autofree gchar* image_size = NULL;
 	GError *ierror = NULL;
 	gboolean res = FALSE;
 	RaucBundle *bundle;
+	GHashTableIter iter;
+	gchar *key = NULL;
+	gchar *value = NULL;
```
jluebbe, line 761: This could be moved to the `if (variables)` block.

jluebbe (changes requested):
Looks good otherwise.

src/update_handler.c
Bastian-Krause: Good point, will do that.

Bastian-Krause:
Force-pushed: integrated review feedback from @jluebbe, thanks!

jluebbe (approved):

PR #703, opened 2021-02-16
add RAUC_IMAGE_SIZE as hook variable

### LINKED ISSUE #700: Passing image information to run_slot_hook
Currently, the image is set to `NULL` in `img_to_nand_handler` and other functions.

https://github.com/rauc/rauc/blob/2182bcfe9df92db085570f19bbda3c84219efdae/src/update_handler.c#L1281

My scenario is, that I want to update a device on my platform that access the nand partition, so I use the `img_to_nand_handler` to write it to `/dev/mtd2`.
After the image is copied I need to set the file size of the written image to a certain part of the devices eeprom, therefore I register a post-install hook and want to access the image name: [`$RAUC_IMAGE_NAME`](https://rauc.readthedocs.io/en/latest/reference.html?highlight=barebox#term-rauc-image-name), but that one won't be set without a valid Image pointer:

https://github.com/rauc/rauc/blob/2182bcfe9df92db085570f19bbda3c84219efdae/src/update_handler.c#L786-L790

Is there a reason why it is set to `NULL` instead of `image`? (and not only in `img_to_nand_handler`, but in almost all other update handlers as well)

hellow554:
Closes #700

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -752,6 +752,7 @@ static gboolean run_slot_hook(const gchar *hook_name, const gchar *hook_cmd, Rau
 {
 	g_autoptr(GSubprocessLauncher) launcher = NULL;
 	g_autoptr(GSubprocess) sproc = NULL;
+	g_autofree gchar* image_size = NULL;
 	GError *ierror = NULL;
 	gboolean res = FALSE;
 	RaucBundle *bundle;
@@ -784,7 +785,9 @@ static gboolean run_slot_hook(const gchar *hook_name, const gchar *hook_cmd, Rau
 		g_subprocess_launcher_setenv(launcher, "RAUC_SLOT_MOUNT_POINT", slot->ext_mount_point, TRUE);
 	}
 	if (image) {
+		image_size = g_strdup_printf("%" G_GOFFSET_FORMAT, image->checksum.size);
 		g_subprocess_launcher_setenv(launcher, "RAUC_IMAGE_NAME", image->filename, TRUE);
+		g_subprocess_launcher_setenv(launcher, "RAUC_IMAGE_SIZE", image_size, TRUE);
 		g_subprocess_launcher_setenv(launcher, "RAUC_IMAGE_DIGEST", image->checksum.digest, TRUE);
 		g_subprocess_launcher_setenv(launcher, "RAUC_IMAGE_CLASS", image->slotclass, TRUE);
 	}
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -802,6 +802,10 @@ The following environment variables will be passed to the hook executable:
     If set, the file name of the image currently to be installed,
     e.g. ``"product-rootfs.img"``
 
+  ``RAUC_IMAGE_SIZE``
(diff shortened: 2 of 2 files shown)


src/update_handler.c
```diff
@@ -785,6 +785,7 @@ static gboolean run_slot_hook(const gchar *hook_name, const gchar *hook_cmd, Rau
 	}
 	if (image) {
 		g_subprocess_launcher_setenv(launcher, "RAUC_IMAGE_NAME", image->filename, TRUE);
+		g_subprocess_launcher_setenv(launcher, "RAUC_IMAGE_SIZE", g_strdup_printf("%ld", image->checksum.size), TRUE);
```
hellow554, line 788: `g_strdup_printf` was the first result I found when I googled `glib int to string`. If you prefere another way, please let me know

ejoerns: `g_strdup_printf()` is okay, but please ensure we do not leak the string (thus assigned it to a `g_autofree gchar*`) and use the define `G_GOFFSET_FORMAT` instead of "`ld`".

hellow554:
This doesn't seem to work properly.
I used `checksum.size`, because you told that would be the actual size of the file, but I get

    SIZE:  -1257597052

when I use that variable.

Any idea why?

ejoerns:
> Any idea why?

What does a `rauc info` on the bundle show? Also negative numbers or a proper value?

hellow554:
`rauc info` shows the correct value. 

It seems to be random what will be printed out:

```
1st run: SIZE:  741400
2nd run: SIZE:  -1257574047
3rd run: SIZE:  -1247021562
```

Did I do something wrong with the auto_ptr?

ejoerns:
@hellow554 If I see it right, the context of the variable should be the function, thus this should be sufficient for setting the environment. Just to be sure that it is no autofree issue, just leave out the `g_autofree` for testing.

From a quick look into this, I don't have a clue what is going wrong, yet. The positive number from your first run is the correct size?

hellow554:
Wow.. nevermind, I'm just very stupid. I edit the wrong patch file.

Its working (: thanks for you help though!

hellow554:
Hmm, what is happening with these tests? The only one I can see so far failing is

https://github.com/rauc/rauc/pull/703/checks?check_run_id=1910777178#step:7:502

but I don't know what that actually means

ejoerns:
@hellow554 This is a sporadic error we are aware of. Unrelated to your changes.

Would you squash the two commits into one? There is no benefit in keeping them separate I guess.

hellow554:
Done.
I like to keep commits seperated when discussion about them, so it's easier to track changes. But yes, before actually merging I like to squash them.

Thanks for your help here!

hellow554:
hey @ejoerns is there something missing so this could be merged?

hellow554:
ping @ejoerns

jluebbe (approved):

ejoerns:
@hellow554 sorry for letting you wait, was a bit busy the last days. Thx @jluebbe for reviewing and merging!

PR #708, opened 2021-02-22
Add support for systemd.verity_root_data= in the kernel cmdline

gportay:
Both [systemd][1] and [systemd-veritysetup-generator][2] define and use the specific systemd kernel parameter `systemd.verity_root_data=` to setup the integrity protected for the root file-system.

This looks for the kernel parameter `systemd.verity_root_data=` to guess the booted slot name if `root=` is not unset.

[1]: https://www.freedesktop.org/software/systemd/man/kernel-command-line.html#roothash=
[2]: https://www.freedesktop.org/software/systemd/man/systemd-veritysetup-generator.html#systemd.verity_root_data=

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -391,6 +391,14 @@ There is no extra slot configuration needed for this as RAUC assumes it is safe
 to update all available slots in case the currently running system comes from
 NFS.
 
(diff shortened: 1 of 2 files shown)


jluebbe (approved):
I've added a paragraph to the docs and adjusted the commit title for consistency.

PR #716, opened 2021-03-25
Fix configurable U-Boot attempt setting

pascalhuerst:
Commit ebbdb4be9d ("Make U-Boot attempt settings configurable") made
it possible to tweak the number of attempts that are set in the U-Boot
environment.

There is an issue with that because U-Boot's setexpr can only handle
hex, while in the current implementation those values are treated as
decimals.

To work around that, let's treat those numbers as hex everywhere.

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -757,7 +757,7 @@ static gboolean uboot_get_state(RaucSlot* slot, gboolean *good, GError **error)
 		g_propagate_error(error, ierror);
 		return FALSE;
 	}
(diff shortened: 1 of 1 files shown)


ejoerns (approved):
Seems as if you were testing with quite high attempt counters :smirk: 

Looks fine for me. :+1:

pascalhuerst:
> Seems as if you were testing with quite high attempt counters smirk
> 
> Looks fine for me. +1

Yes we were :)

jluebbe (approved):
Looks fine (although I find the behavior a bit surprising ;).

PR #723, opened 2021-04-13
Remove Long Deprecated Sections from Manifest Parsing.

### LINKED ISSUE #709: Obsolete test manifest file?
Hi there,

In the manifest test file: https://github.com/rauc/rauc/blob/master/test/manifest.raucm

It makes reference to some undocumented features such as [file.<slot-class>] and [keyring].

I only found reference to keyring in the system configuration file and the file.<slot-class> option seems very useful but doesn't seem to be documented.

ejoerns:
This removes two manifest config options which were parsed but not actually used anywhere in the code.
This is unexpected and violates the concept that a manifest configuration/content actually has an impact on the installation.

Removed sections are:

* The `file.<slotclass>/<filename>` sections which originates from the long deprecated bundle-free network mode (not to be confused with the current network capabilities of RAUC) we had in the early days of RAUC.
* The `keyring` section that existed since the initial concept but actually never gained any implementation in RAUC

Note this might break very rare cases where one accidentally had one of the deprecated sections in the manifest and probably requires fixing the update bundle in this case. On the other hand, this will prevent future confusion and ensure we get distinct errors for things that are not working, anyway.

Fixes #709.


ejoerns:
Note: I've updated the commit's and PR's description as I got confused about the original purpose of the `keyring` section.

jluebbe (approved):

PR #726, opened 2021-04-16
Support Omitting Filename When 'install' Slot Hook is Used

ejoerns:
The 'install' slot hook may potentially change the slots content without strictly requiring an image. Thus we should not make the filename/image mandatory here.

As we are checking (or assuming) that `image->filename` is not ` NULL` at several places in the code, this requires some changes across the install and manifest parsing module.

We also fix some minor things in logging and info printout.


src/main.c
```diff
@@ -732,11 +732,15 @@ static gchar *info_formatter_readable(RaucManifest *manifest)
...
 		g_string_append_printf(text, "  "KBLD "[%s]"KNRM "\n", img->slotclass);
-		g_string_append_printf(text, "\tFilename:  %s\n", img->filename);
-		if (img->variant)
-			g_string_append_printf(text, "\tVariant:   %s\n", img->variant);
-		g_string_append_printf(text, "\tChecksum:  %s\n", img->checksum.digest);
-		g_string_append_printf(text, "\tSize:      %"G_GOFFSET_FORMAT "\n", img->checksum.size);
+		if (img->filename) {
+			g_string_append_printf(text, "\tFilename:  %s\n", img->filename);
+			if (img->variant)
+				g_string_append_printf(text, "\tVariant:   %s\n", img->variant);
```
jluebbe, line 738: The variant still might be relevant without having an image, so it should be always visible, perhaps directly after the slotclass.

src/install.c
```diff
@@ -499,6 +499,10 @@ static void prepare_environment(GSubprocessLauncher *launcher, gchar *update_sou
 			for (GList *l = manifest->images; l != NULL; l = l->next) {
 				RaucImage *img = l->data;
 				if (g_str_equal(slot->sclass, img->slotclass)) {
+					/* If no filename is set, do not attempt to set environment variables */
+					if (!img->filename)
+						break;
```
jluebbe, line 504: Shouldn't this be `continue`? Also, the slotclass is still relevant.

src/install.c
```diff
@@ -745,6 +749,11 @@ static gboolean pre_install_checks(gchar* bundledir, GList *install_images, GHas
 		RaucImage *mfimage = l->data;
 		RaucSlot *dest_slot = g_hash_table_lookup(target_group, mfimage->slotclass);
 
+
```
jluebbe, line 752: Redundant empty line.

docs/using.rst
```diff
@@ -447,6 +448,17 @@ migrating configuration settings.
...
   hooks=install
 
+or, without ``filename``:
+
+.. code-block:: cfg
+
+  [hooks]
+  filename=hook
+
+  [image.datafs]]
```
jluebbe, line 458:
```suggestion
  [image.datafs]
```

jluebbe (changes requested):

src/install.c
ejoerns: Yes, but today's me says: There should be no break/continue at all. We have several other places in the code that unconditionally set these env variables but can deal with C variables being NULL. Doing it the same way here seems to be the most consistent way for me.

src/main.c
ejoerns: You're right. Will move it above the outer if condition

ejoerns:
Rebased onto master for better diffability.

ejoerns:
Fixed the things pointed out in the review comments. Also added a minor patch for the example hook.sh

jluebbe (approved):

PR #732, opened 2021-05-03
docs: advanced: clarify how boot-mbr-switch works

ejoerns:
The former description had formulations like "switching between two partitions" which can be quite misleading as we do not actually switch between two partitions but switch a partition between two regions.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -803,22 +803,22 @@ The initial MBR must define a boot partition at either the first or the second
 half of the configured region.
 
 Consider the following example layout of a storage medium with a boot partition size
-of 33 MiB:
+of 32 MiB:
 
-+--------------+----------------+-----------------------------------------------+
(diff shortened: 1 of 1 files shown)


jluebbe:
In the code and existing documentation, the 'region' covers both parts. In the GPT case we use 'lower and upper halves of a region'.

ejoerns:
@jluebbe yes, this should be consistent. I've adapted the text accordingly.

ejoerns:
I've reworked the whole chapter now to fix the remaining inconsistencies. And tried to be more verbose for the example.

jluebbe:
After realizing that upper/lower region is confusing as well, I've now updated it to use first/second half. Also, I've adjusted the boot-gpt-switch section to match and aligned the partitions in the example layout.

jluebbe (dismissed):

jluebbe (approved):

PR #740, opened 2021-06-08
src/bundle: fix error handling for small squashfs sizes

### LINKED ISSUE #739: Failed to create bundle: cannot sign inconsistent manifest: Missing size for verity bundle
Hi, when attempting to sign a bundle (using latest master branch build) I'm getting this error:
`Failed to create bundle: cannot sign inconsistent manifest: Missing size for verity bundle`

I'm using the following to create the bundle:
`rauc --cert=/keys/dev-rauc-kms.crt   --key=pkcs11:token=mytoken   --debug bundle /tmp/rauc  /tmp/test.raucb`

For this test, I'm just using a file that's 127 bytes (it's not a real image, just a text file my /etc/hosts file).

My bundle dir looks like this:
```
ls -lh /tmp/rauc
total 8.0K
-rw-r--r-- 1 root root 172 Jun  8 15:29 manifest.raucm
-rw-r--r-- 1 root root 176 Jun  8 15:22 rootfs.img
```

My manifest looks like this:
```
[update]
compatible=test123

[bundle]
format=verity

[image.rootfs]
sha256=f0ecbb34e69b6fda41be5c7cd7d37c3a6bbeb3bb3295a7b8fd6d7645d184d9d3
size=176
filename=rootfs.img
```

My original bundle did not contain the size or sha256 lines, they were added automatically by RAUC.

Debug mode does not seem to yield anything more informative:
```
rauc-Message: 15:29:21.325: Debug log domains: 'rauc'
(rauc:206): rauc-DEBUG: 15:29:21.326: bundle start
(rauc:206): rauc-DEBUG: 15:29:21.326: system config not found, using default values
(rauc:206): rauc-DEBUG: 15:29:21.326: input directory: /tmp/rauc
(rauc:206): rauc-DEBUG: 15:29:21.326: output bundle: /test.raucb
(rauc:206): rauc-DEBUG: 15:29:21.352: Payload size: 4096 bytes.
Creating bundle in 'verity' format
(rauc:206): rauc-DEBUG: 15:29:21.352: Hash creation SHA256, data blocks 1.
(rauc:206): rauc-DEBUG: 15:29:21.352: Using 0 hash levels.
(rauc:206): rauc-DEBUG: 15:29:21.352: Data size: 4096 bytes.
(rauc:206): rauc-DEBUG: 15:29:21.352: Hashed size: 4096 bytes.
Failed to create bundle: cannot sign inconsistent manifest: Missing size for verity bundle
```

Any idea what the problem could be?

ejoerns:
When having a squashfs that is only 4096 bytes in size, we cannot
generate a hash table.

To catch these issues, an assert was places in the code.
However, the assert must check for '>' instead of '>=' as otherwise it
would allow 'combined_size*4096' equal to 'offset' which results in a
'verity_size' of 0.

As the size of 'offset' actually depends on user input (i.e. bundle
content size), we should not use g_assert() as primary check instance
here, but add a GError check that allows to emit a user-readable error.

Fixes #739

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -528,6 +528,13 @@ static gboolean sign_bundle(const gchar *bundlename, RaucManifest *manifest, GEr
 					"squashfs size (%"G_GUINT64_FORMAT ") is not a multiple of 4096 bytes", offset);
 			return FALSE;
 		}
+		if (offset <= 4096) {
+			g_set_error(error,
+					R_SIGNATURE_ERROR,
+					R_SIGNATURE_ERROR_CREATE_SIG,
+					"squashfs size (%"G_GUINT64_FORMAT ") must be larger than 4096 bytes", offset);
+			return FALSE;
+		}
 		if (verity_create_or_verify_hash(0, bundlefd, offset/4096, &combined_size, hash, salt) != 0) {
 			g_set_error(error,
 					R_SIGNATURE_ERROR,
@@ -536,7 +543,7 @@ static gboolean sign_bundle(const gchar *bundlename, RaucManifest *manifest, GEr
 			return FALSE;
 		}
 		/* for a squashfs <= 4096 bytes, we don't have a hash table */
-		g_assert(combined_size*4096 >= (off_t)offset);
+		g_assert(combined_size*4096 > (off_t)offset);
 		verity_size = combined_size*4096 - offset;
 		g_assert(verity_size % 4096 == 0);
 


ejoerns:
Update:

* check for squashfs being <= 4096 (instead of < 8192) in error path as this is the actual limitation (8192 is only the next possible size according to the moduly check before). Adapt error message accordingly.

jluebbe (approved):

PR #744, opened 2021-06-17
Replaced deprecated Install with InstallBundle method

rforro:
According to [this ](https://rauc.readthedocs.io/en/latest/reference.html#the-install-method) the used `Install` method in `cgi.c` has been deprecated. Therefor it was replaced by `InstallBundle`. Currently no arguments are passed to installation. If you have an idea how to pass them (custom header? query param?) I can integrate it as well.


Bastian-Krause:
Thanks for the PR!

Could you update the link to the install method in `contrib/cgi/README.rst` accordingly?
I think it's not a good idea to allow passing options such as `ignore-compatible` via HTTP.

rforro:
README is updated!

OK, I don't use the `ignore-compatible` parameter anyway.

jluebbe (approved):

PR #749, opened 2021-06-30
feature: extract signatue: add option to extract signature from bundle

Jarsop:
Many industrialization procedure requires to sign any artifact in a
dedicated secure room with restricted access
(as Public Key Infrastructure aka PKI).

This purpose add a `extract-signature` command to extract the signature
and allow resigning the signature content externally.

Signed-off-by: Jean-Pierre Geslin <jarsop@outlook.com>

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -1625,6 +1625,43 @@ out:
 	return res;
 }
 
+gboolean extract_signature(RaucBundle *bundle, const gchar *outputsig, GError **error)
+{
+	GError *ierror = NULL;
+	g_autoptr(GFile) sigfile = NULL;
+	g_autoptr(GOutputStream) sigoutstream = NULL;
+	gboolean res = FALSE;
+
+	g_return_val_if_fail(bundle != NULL, FALSE);
+	g_return_val_if_fail(outputsig != NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	r_context_begin_step("extract_signature", "Extracting bundle signature", 0);
+
+	sigfile = g_file_new_for_path(outputsig);
+	sigoutstream = (GOutputStream*)g_file_create(sigfile, G_FILE_CREATE_PRIVATE, NULL, &ierror);
+	if (sigoutstream == NULL) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"failed to create file to store signature: ");
+		goto out;
+	}
+
+	if (!output_stream_write_bytes_all(sigoutstream, bundle->sigdata, NULL, &ierror)) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"failed to write signature to file: ");
+		goto out;
+	}
+
+	res = TRUE;
+out:
+	r_context_end_step("extract_signature", res);
+	return res;
+}
+
 gboolean extract_bundle(RaucBundle *bundle, const gchar *outputdir, GError **error)
 {
 	GError *ierror = NULL;
--- a/src/main.c
+++ b/src/main.c
@@ -513,6 +513,51 @@ out:
 	return TRUE;
 }
 
+static gboolean extract_signature_start(int argc, char **argv)
+{
+	RaucBundle *bundle = NULL;
+	GError *ierror = NULL;
+	g_debug("extract signature start");
+
+	if (argc < 3) {
+		g_printerr("An input bundle must be provided\n");
+		r_exit_status = 1;
+		goto out;
+	}
+
+	if (argc < 4) {
+		g_printerr("An output signature file must be provided\n");
+		r_exit_status = 1;
+		goto out;
+	}
+
+	if (argc > 4) {
+		g_printerr("Excess argument: %s\n", argv[4]);
+		r_exit_status = 1;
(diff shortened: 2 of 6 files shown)


ejoerns:
@Jarsop Thanks for your contribution.

The existing intended way for this would be to use the `rauc resign` command. https://rauc.readthedocs.io/en/latest/advanced.html#resigning-bundles

Did you check this? Or could you describe in how this would not meet your requirements?

How would you ensure that the 'external' signing produces a valid signature format for RAUC?

Jarsop:
> @Jarsop Thanks for your contribution.
> 
> The existing intended way for this would be to use the `rauc resign` command. https://rauc.readthedocs.io/en/latest/advanced.html#resigning-bundles
> 
> Did you check this? Or could you describe in how this would not meet your requirements?
> 
> How would you ensure that the 'external' signing produces a valid signature format for RAUC?

In an embedded Linux project of a large company for which I am currently working (and others in the past which have the same functioning), they have a PKI which prohibits any external tool to sign artifacts (so no `rauc resign` possible).

So I had to provide them with the exact format of the desired signature and he provides it to me.

Today I created a tool that allows you to build this kind of bundle and then assemble the CMS in post process.
So that's why I think this functionality should be integrated into RAUC.

This problem is common in large companies which have very strict signature policies.

Hoping to have been clear and thanks for your quick reply.

Regards

Jarsop:
> How would you ensure that the 'external' signing produces a valid signature format for RAUC?

For this point although the signature can be contained when requested  to the signer, I think that a check at the command level which will allow to assemble the unsigned bundle and the CMS will be able to solve this problem

Jarsop:
Maybe it's better to will be a part of `resign` command ? (For bundle extraction and re-assembly)

jluebbe:
Both `rauc bundle` and `rauc resign` already support using PKCS#11 keys, so no direct access to the private key material is needed. Would that be enough for your use-case?

If not, how about this:
- a new `rauc extract-signature` command which would extract the CMS signature from a bundle and write it to a file
- a new `rauc replace-signature` command which would create a copy of an existing bundle playload with a CMS signature read from a file

I'd restrict this to verity bundles, as the CMS signature is "self-contained" in that mode (and also I'd like to discourage plain bundles).

In practice, you'd then build bundles as normal (signed with a development/CI key), test and review them. Then you'd use `rauc extract-signature` to get the CMS structure, process them with you special in-house PKI tools/process to get the release CMS signature. Those would then be used with `rauc replace-signature`.

I see some benefits:

1. No additional complexity in the normal bundle creation code.
2. Your tools only need to handle standard data formats (CMS) and the bundle format stays internal to RAUC.
3. Future changes to the bundle format (i.e. for encryption) can be handled in RAUC and would not affect your processes.
4. It would also allow multiple people to perform the review and `rauc extract-signature` step, resulting in multiple CMS files that could be compared in the PKI room to ensure no single person can sign a manipulated bundle.

We'd be happy to discuss this in https://app.element.io/#/room/#rauc:matrix.org (bridged from #rauc on libera.chat), as well.

src/bundle.c
```diff
@@ -1625,6 +1625,47 @@ gboolean check_bundle_payload(RaucBundle *bundle, GError **error)
...
+
+	if (sigstream == NULL) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"failed to create file to store signature: ");
+		goto out;
+	}
+
+	sigoutstream = g_io_stream_get_output_stream(G_IO_STREAM(sigstream));
```
jluebbe, line 1653: You could use `g_file_create()` instead, which directly produces a GFileOutputStream (which can be casted to  a `GOutputStream`).

jluebbe (changes requested):
Just a minor suggestion for simplification. I'll likely do some rewording on the commit message and documentation before merging if that's fine with you.

src/bundle.c
Jarsop: Yes for sure ! My English is not always very good

Jarsop: The second part `rauc replace-signature` will coming soon 😄

Jarsop:
Any ideas why uncrustify failed? Maybe because you requested changes?

jluebbe:
I've made some small changes and force-pushed the PR. You can check my changes here: https://github.com/rauc/rauc/compare/45c6972431e92d8c062c6e54c9afdebbf2384ede..7b044a64a39541e95ee8a23da0b614973b83b6db

jluebbe (approved):

src/main.c
```diff
@@ -513,6 +513,51 @@ static gboolean resign_start(int argc, char **argv)
...
+	if (argc > 4) {
+		g_printerr("Excess argument: %s\n", argv[4]);
+		r_exit_status = 1;
+		goto out;
+	}
+
+	g_debug("input bundle: %s", argv[2]);
+	g_debug("output file: %s", argv[3]);
+
+	if (!check_bundle(argv[2], &bundle, TRUE, &ierror)) {
```
Jarsop, line 543: @jluebbe Should it be configurable with `--no-verify` option ?

jluebbe: Hmm, probably yes, although I'm not a fan of `--no-verify`.

PR #752, opened 2021-07-13
src/install: remove trailing whitespace from RAUC_SLOTS and RAUC_TARGET_SLOTS

jluebbe:
The whitespace is not relevant for shell loops, but can be confusing
nevertheless. Remove it.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -466,8 +466,8 @@ static void prepare_environment(GSubprocessLauncher *launcher, gchar *update_sou
 	GHashTableIter iter;
 	RaucSlot *slot;
 	gint slotcnt = 0;
-	g_autofree gchar *targetlist = NULL;
-	g_autofree gchar *slotlist = NULL;
+	g_autoptr(GString) slots = g_string_sized_new(128);
+	g_autoptr(GString) target_slots = g_string_sized_new(128);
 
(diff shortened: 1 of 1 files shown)


src/install.c
```diff
@@ -545,8 +545,8 @@ static void prepare_environment(GSubprocessLauncher *launcher, gchar *update_sou
 		g_clear_pointer(&varname, g_free);
 	}
 
-	g_subprocess_launcher_setenv(launcher, "RAUC_SLOTS", slotlist, TRUE);
-	g_subprocess_launcher_setenv(launcher, "RAUC_TARGET_SLOTS", targetlist, TRUE);
+	g_subprocess_launcher_setenv(launcher, "RAUC_SLOTS", g_strchomp(slotlist), TRUE);
+	g_subprocess_launcher_setenv(launcher, "RAUC_TARGET_SLOTS", g_strchomp(targetlist), TRUE);
```
ejoerns, line 549: Wouldn't it be more straightforward to remove the trailing whitespaces in the corresponding `g_strdup_printf()` statements?

jluebbe: We need the spaces between the values, so how would that work when removing them from the `g_strdup_printf()`? I've pushed a version which uses GString.

ejoerns: True. I have ignored the while loop :)

I like the solution, looks much cleaner now than the original one with the tmp string. :+1:

ejoerns (approved):
:+1:

zyga:
Looks good

PR #754, opened 2021-07-22
README.rst: add CII Best Practices badge

jluebbe:


ejoerns (approved):
From the context, it could also be placed before the Documentation badge (together with other checker badges) but it's ok here too.

PR #758, opened 2021-07-26
src/main: add rauc mount command

jluebbe:
This allows easy inspection of bundle contents. As this uses squashfs in the
kernel (as `rauc install` does), this also doesn't allow --no-verify.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -670,6 +670,7 @@ Command Line Tool
     extract               Extract the bundle content
     install               Install a bundle
     info                  Show file information
+    mount                 Mount a bundle (for development purposes)
     service               Start RAUC service
     status                Show status
     write-slot            Write image to slot and bypass all update logic
--- a/src/main.c
+++ b/src/main.c
@@ -1683,6 +1683,59 @@ static gboolean service_start(int argc, char **argv)
 	return TRUE;
 }
 
+static gboolean mount_start(int argc, char **argv)
+{
+	g_autofree gchar *bundlelocation = NULL;
+	g_autoptr(RaucBundle) bundle = NULL;
+	GError *error = NULL;
+	gboolean res = FALSE;
+
+	if (argc < 3) {
+		g_printerr("A file name must be provided\n");
+		goto out;
+	}
+
+	if (argc > 3) {
+		g_printerr("Excess argument: %s\n", argv[3]);
+		goto out;
+	}
+
+	bundlelocation = resolve_bundle_path(argv[2]);
+	if (bundlelocation == NULL)
+		goto out; /* an error message was already printed by resolve_bundle_path */
+	g_debug("input bundle: %s", bundlelocation);
+
+	res = check_bundle(bundlelocation, &bundle, TRUE, &error);
+	if (!res) {
+		g_printerr("%s\n", error->message);
+		g_clear_error(&error);
+		goto out;
+	}
+
+	g_debug("bundle payload size: %"G_GOFFSET_FORMAT, bundle->size);
+
+	res = mount_bundle(bundle, &error);
+	if (!res) {
+		g_printerr("%s\n", error->message);
+		g_clear_error(&error);
+		goto out;
+	}
+
+	g_print("Mounted bundle at %s. Use 'umount %s' to unmount.\n", bundle->mount_point, bundle->mount_point);
+
+	/* The device mapper target and loopback devices are configured to remove
+	 * themselves on close. They are kept active by the mounted filesystem and
+	 * are automatically cleaned up when the user performs a normal unmount.
+	 * To avoid running the g_autoptr cleanup for the bundle, we use exit(0)
+	 * here.
+	 **/
+	exit(0);
+
+out:
+	r_exit_status = 1;
+	return FALSE;
+}
+
 static gboolean unknown_start(int argc, char **argv)
 {
 	g_debug("unknown start");
@@ -1702,6 +1755,7 @@ typedef enum  {
 	INFO,
 	WRITE_SLOT,
 	SERVICE,
+	MOUNT,
 } RaucCommandType;
 
 typedef struct {
@@ -1844,6 +1898,7 @@ static void cmdline_handler(int argc, char **argv)
 #if ENABLE_SERVICE == 1
 		{SERVICE, "service", "service", "Start RAUC service", service_start, service_group, TRUE},
 #endif
+		{MOUNT, "mount", "mount <BUNDLENAME>", "Mount a bundle (for development purposes)", mount_start, NULL, TRUE},
 		{0}
 	};
 	RaucCommand *rc;
@@ -1864,6 +1919,7 @@ static void cmdline_handler(int argc, char **argv)
 			"  extract\t\tExtract the bundle content\n"
 			"  install\t\tInstall a bundle\n"
 			"  info\t\t\tShow file information\n"
+			"  mount\t\t\tMount a bundle\n"
 #if ENABLE_SERVICE == 1
 			"  service\t\tStart RAUC service\n"
 #endif
--- a/test/rauc.t
+++ b/test/rauc.t
@@ -307,6 +307,19 @@ test_expect_success PKCS11 "rauc bundle with PKCS11 (key mismatch)" "
     bundle $SHARNESS_TEST_DIRECTORY/install-content ${TEST_TMPDIR}/out.raucb
 "
 
+test_expect_success ROOT "rauc mount" "
+  test ! -f /mnt/rauc/bundle/manifest.raucm &&
+  cp -L ${SHARNESS_TEST_DIRECTORY}/good-bundle.raucb ${TEST_TMPDIR}/ &&
+  test_when_finished rm -f ${TEST_TMPDIR}/good-bundle.raucb &&
+  test_when_finished umount /mnt/rauc/bundle &&
+  ls ${TEST_TMPDIR} &&
+  rauc --keyring $SHARNESS_TEST_DIRECTORY/openssl-ca/dev-ca.pem \
+    mount ${TEST_TMPDIR}/good-bundle.raucb &&
+  mount &&
+  test -f /mnt/rauc/bundle/manifest.raucm &&
+  test -f /mnt/rauc/bundle/rootfs.img
+"
+
 test_expect_success SERVICE "rauc service double-init failure" "
   start_rauc_dbus_service \
     --conf=${SHARNESS_TEST_DIRECTORY}/test.conf \
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -627,6 +627,40 @@ It may also be worth starting the RAUC service via command line on a second
 shell to have a live view of what is going on when you invoke e.g. ``rauc
 install`` on the first shell.
 
+Inspecting Bundle Contents
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes during development, it is useful to check whether the bundle contents
+are as expected.
+While RAUC bundles could just be mounted as a squashfs, using ``rauc mount``
+also uses the same checks and mechanisms as ``rauc install``
+(device-mapper/loopback & network support).
+The bundle is mounted below the configured mount prefix (``/mnt/rauc/bundle`` by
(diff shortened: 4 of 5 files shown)


src/main.c
```diff
@@ -1683,6 +1683,56 @@ static gboolean service_start(int argc, char **argv)
...
+	}
+
+	g_message("bundle mounted at: %s", bundle->mount_point);
+
+	/* The mount keeps the block device active. */
+	exit(0);
+
+out:
+	r_exit_status = res ? 0 : 1;
+	return TRUE;
```
Jarsop, line 1733: Why using `exit` and setting `r_exit_status` with a ternary. `out` label can it not exist that in case of error right ?

src/main.c
```diff
@@ -1683,6 +1683,56 @@ static gboolean service_start(int argc, char **argv)
...
+
+	if (argc < 3) {
+		g_printerr("A file name must be provided\n");
+		r_exit_status = 1;
+		return FALSE;
+	}
+
+	if (argc > 3) {
+		g_printerr("Excess argument: %s\n", argv[3]);
+		r_exit_status = 1;
```
Jarsop, line 1701:
```suggestion
```
Redundant with `out` label logic

Jarsop:
Perhaps is silly as question :stuck_out_tongue:  but why the `mount_start` can be return `TRUE` in some error case ?

src/main.c
```diff
@@ -1683,6 +1683,56 @@ static gboolean service_start(int argc, char **argv)
...
+	res = mount_bundle(bundle, &error);
+	if (!res) {
+		g_printerr("%s\n", error->message);
+		g_clear_error(&error);
+		goto out;
+	}
+
+	g_message("bundle mounted at: %s", bundle->mount_point);
+
+	/* The mount keeps the block device active. */
```
zyga, line 1728: How is this comment related to the `exit` call?

zyga (commented):
Quick look with one question inline.

src/main.c
jluebbe: The device mapper target and loopback devices are configured to remove themselves on close. So we don't need to handle cleanup in RAUC, a simple unmount on the is enough. I'll extend the comment accordingly.

docs/reference.rst
```diff
@@ -670,6 +670,7 @@ Command Line Tool
     extract               Extract the bundle content
     install               Install a bundle
     info                  Show file information
+    mount                 Mount RAUC bundle
```
Bastian-Krause, line 673:
```suggestion
    mount                 Mount a bundle
```
This would be consistent with the other command's descriptions.

src/main.c
Bastian-Krause: True. Why go to "out" here, but return early above?

src/main.c
```diff
@@ -1844,6 +1895,7 @@ static void cmdline_handler(int argc, char **argv)
 #if ENABLE_SERVICE == 1
 		{SERVICE, "service", "service", "Start RAUC service", service_start, service_group, TRUE},
 #endif
+		{MOUNT, "mount", "mount <BUNDLENAME>", "Mount RAUC bundle", mount_start, NULL, TRUE},
```
Bastian-Krause, line 1898:
```suggestion
		{MOUNT, "mount", "mount <BUNDLENAME>", "Mount a bundle", mount_start, NULL, TRUE},
```

src/main.c
```diff
@@ -1864,6 +1916,7 @@ static void cmdline_handler(int argc, char **argv)
 			"  extract\t\tExtract the bundle content\n"
 			"  install\t\tInstall a bundle\n"
 			"  info\t\t\tShow file information\n"
+			"  mount\t\t\tMount RAUC bundle\n"
```
Bastian-Krause, line 1919:
```suggestion
			"  mount\t\t\tMount a bundle\n"
```

Bastian-Krause (changes requested):

src/main.c
jluebbe: Yes, that's inconsistent, will fix.

src/main.c
jluebbe: Changed.

src/main.c
jluebbe: Changed.

docs/reference.rst
jluebbe: Changed.

src/main.c
jluebbe: I've expanded the comment.

src/main.c
jluebbe: I've made this consistent. `r_exit_status` is now set to `1` in `out:` and we return `FALSE`.

jluebbe:
Rebased.

Bastian-Krause (dismissed):
Looks good to me.

Jarsop:
Same for me

ejoerns:
In general, I like the idea of the command.

However, some more or less notable things that I would like to mention are:

a) Maybe we should leave a note in the help that this is not a "normal operation" command as I fear that people would tend to see this as something that is totally fine with the flow of RAUC and will work around the standard install strategies even more. For those who really know what they do and why, this is fine, but I would not unnecessarily encourage people to abuse 'rauc mount'.

Thus possibly something like

``` patch
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -670,6 +670,7 @@ Command Line Tool
      info                  Show file information
-+    mount                 Mount a bundle
++    mount                 Mount a bundle for debugging and inspection purpose
      service               Start RAUC service
```

What do you think?


b) The command requires root privileges as far as I see, thus it is possibly primarily useful for the target or should be used on the host too?
It also states to be similarly to `rauc install` but ignores the system.conf file we have on the target. Thus we must provide the required arguments (keyring, mount prefix, etc.) by hand. Especially the `--mount` argument appears to be a bit unexpected/inconvenient to me (We omit the destination argument a normal mount call has, but also do not use the configured one. Thus end up in the rauc default mount path..).
The reason for doing it only this way is obviously the need to introduce a new D-Bus method otherwise. Thus it might be a compromise we need to accept, but I would like to make sure that we have thought of the use cases properly before.

c) We do not provide an 'unmount' command. Probably the command should print the instruction to unmount the bundle to shell? Like

    Mounted bundle at /mnt/rauc/bundle. Type 'umount /mnt/rauc/bundle' to unmount.

d) The command is meant to be a debugging tool by default (if I get it right). Thus I would propose to drop a note in `Debugging RAUC` section in docs/using.rst.

e) We should take more care about updating the man page (note to myself, too) when extending the CLI. Here the added command should be reflected, too.

That's all I guess ;)

jluebbe:
> In general, I like the idea of the command.
> 
> However, some more or less notable things that I would like to mention are:
> 
> a) Maybe we should leave a note in the help that this is not a "normal operation" command as I fear that people would tend to see this as something that is totally fine with the flow of RAUC and will work around the standard install strategies even more. For those who really know what they do and why, this is fine, but I would not unnecessarily encourage people to abuse 'rauc mount'.
> 
> Thus possibly something like
> 
> ```diff
> --- a/docs/reference.rst
> +++ b/docs/reference.rst
> @@ -670,6 +670,7 @@ Command Line Tool
>       info                  Show file information
> -+    mount                 Mount a bundle
> ++    mount                 Mount a bundle for debugging and inspection purpose
>       service               Start RAUC service
> ```
> 
> What do you think?

Good idea.

> b) The command requires root privileges as far as I see, thus it is possibly primarily useful for the target or should be used on the host too?

Yes, this is mainly useful on the target.

> It also states to be similarly to `rauc install` but ignores the system.conf file we have on the target. Thus we must provide the required arguments (keyring, mount prefix, etc.) by hand. Especially the `--mount` argument appears to be a bit unexpected/inconvenient to me (We omit the destination argument a normal mount call has, but also do not use the configured one. Thus end up in the rauc default mount path..).

Hmm, it should use the normal system.conf, just like rauc info or install. If this this doesn't work for you, we need to check why. Would you still prefer to have an (optional) mount point argument after the bundle?

> The reason for doing it only this way is obviously the need to introduce a new D-Bus method otherwise. Thus it might be a compromise we need to accept, but I would like to make sure that we have thought of the use cases properly before.

A mount created via the service wouldn't be visible due to the configuration of mount namespaces. Also, it would need need more coordination on unmount. As it's intended to be a development command, a DBus interface wouldn't be useful.

> c) We do not provide an 'unmount' command. Probably the command should print the instruction to unmount the bundle to shell? Like
> 
> ```
> Mounted bundle at /mnt/rauc/bundle. Type 'umount /mnt/rauc/bundle' to unmount.
> ```

Yes.

> d) The command is meant to be a debugging tool by default (if I get it right). Thus I would propose to drop a note in `Debugging RAUC` section in docs/using.rst.

Yes.

> e) We should take more care about updating the man page (note to myself, too) when extending the CLI. Here the added command should be reflected, too.

Right.

> That's all I guess ;)

ejoerns:
> > It also states to be similarly to `rauc install` but ignores the system.conf file we have on the target. Thus we must provide the required arguments (keyring, mount prefix, etc.) by hand. Especially the `--mount` argument appears to be a bit unexpected/inconvenient to me (We omit the destination argument a normal mount call has, but also do not use the configured one. Thus end up in the rauc default mount path..).
> 
> Hmm, it should use the normal system.conf, just like rauc info or install. If this this doesn't work for you, we need to check why. Would you still prefer to have an (optional) mount point argument after the bundle?

Yes and no ;)

If you provide `-c /path/to/config` or if system.conf is located at the default path, it will use it.

For `rauc info` the situation is similar. For `rauc install`, this is not necessary of course as the service executes the actual command.

We hit the host vs target tool problematic here again I guess. From my point of view, a 'host' tool should not use system.conf while a target tool should use the system.conf via service only (for D-Bus case of course). Thus, probably an (at least optional) mount point argument would be a proper solution. Then it can be called with --keyring argument (which is similar to other host tools). The fact that this is a 'host' tool that can primarily be used on the target is probably something we have to endure ;)

jluebbe:
This was just a straight rebase to make the following fixes easier to review.

src/main.c
```diff
@@ -1683,6 +1683,59 @@ static gboolean service_start(int argc, char **argv)
...
+		goto out;
+	}
+
+	if (argc > 3) {
+		g_printerr("Excess argument: %s\n", argv[3]);
+		goto out;
+	}
+
+	bundlelocation = resolve_bundle_path(argv[2]);
+	if (bundlelocation == NULL)
```
zyga, line 1704: Is there some error that can be printed in this case?

zyga (commented):
Looks good, just one small question inline.

test/rauc.t
```diff
@@ -307,6 +307,19 @@ test_expect_success PKCS11 "rauc bundle with PKCS11 (key mismatch)" "
...
 "
 
+test_expect_success ROOT "rauc mount" "
+  test ! -f /mnt/rauc/bundle/manifest.raucm &&
+  cp -L ${SHARNESS_TEST_DIRECTORY}/good-bundle.raucb ${TEST_TMPDIR}/ &&
+  test_when_finished rm -f ${TEST_TMPDIR}/good-bundle.raucb &&
+  test_when_finished umount /mnt/rauc/bundle &&
+  ls ${TEST_TMPDIR} &&
+  rauc --keyring $SHARNESS_TEST_DIRECTORY/openssl-ca/dev-ca.pem \
+    mount ${TEST_TMPDIR}/good-bundle.raucb &&
```
zyga, line 317: Do you need to unmount this or is the test scaffolding responsible for that?

src/main.c
jluebbe: ``resolve_bundle_path()`` prints an error. I've added a comment to the ``goto out;`` line.

test/rauc.t
jluebbe: That is handled by the ``test_when_finished umount /mnt/rauc/bundle`` above.

jluebbe:
I've applied most of the changes requested by @ejoerns, except changing the configuration handling. That should be done as a separate PR and ensure that the on-target commands check that the config is loaded correctly (to avoid confusion when a different config location is used).

ejoerns (approved):
Changes look fine for me, thx!

PR #759, opened 2021-07-27
docs: explicit max-bundle-download-size in bytes

vivien:
The max-bundle-download-size value must be specified in bytes, without
unit. Specifying its default value in MiB might be confusing to some,
thus specify the actual value being used (8*1024*1024).

Signed-off-by: Vivien Didelot <vdidelot@pbsc.com>

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -153,7 +153,7 @@ Example configuration:
 ``max-bundle-download-size``
   Defines the maximum downloadable bundle size in bytes, and thus must be
   a simple integer value (without unit) greater than zero.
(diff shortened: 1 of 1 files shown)


jluebbe:
We could do the opposite and support suffixes by using https://github.com/rauc/rauc/blob/master/src/utils.c#L233.
That should probably also support uppercase suffixes.

vivien:
@jluebbe that would be a nice addition indeed. In the meantime, the value expects a number of bytes, so I think it would be convenient for the doc to reflect the current behavior.

jluebbe (approved):

PR #760, opened 2021-07-31
feature: replace-signature: add command to replace bundle signature

Jarsop:
This purpose is the second part of the previous proposal `extract-signature`
This allows to replace the signature bundle with a given signature file
and store the result in a new bundle.

Use `--disable-create` configure switch for `extract-signature`

Signed-off-by: Jean-Pierre Geslin <jarsop@outlook.com>


src/bundle.c
```diff
@@ -1625,6 +1625,136 @@ gboolean check_bundle_payload(RaucBundle *bundle, GError **error)
...
+	if (size == -1) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"failed to append original bundle payload to temporary bundle: ");
+		res = FALSE;
+		goto out;
+	}
+
+	res = g_seekable_seek(G_SEEKABLE((GOutputStream*)istream),
```
jluebbe, line 1701: Doesn't this need a `g_seekable_truncate()` before seeking to the end? Otherwise you could have leftover parts from the old signature if it was larger.

src/bundle.c
```diff
@@ -1625,6 +1625,136 @@ gboolean check_bundle_payload(RaucBundle *bundle, GError **error)
...
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"failed to append signature size to temporary bundle: ");
+		goto out;
+	}
+
+	/* Necessary to release associated fd before perform check_bundle */
+	g_object_unref(istream);
+	res = check_bundle(iname, &ibundle, verify, &ierror);
```
jluebbe, line 1732: You should use the optional signing_keyring for checking by reconfiguring with `r_context_conf()->keyringpath = r_context()->signing_keyringpath` if it's provided on the command line. This allows using two independent PKIs.

src/bundle.c
```diff
@@ -1625,6 +1625,136 @@ gboolean check_bundle_payload(RaucBundle *bundle, GError **error)
...
+	tmpdir = g_dir_make_tmp("rauc-replace-XXXXXX", &ierror);
+	if (tmpdir == NULL) {
+		g_propagate_prefixed_error(error, ierror,
+				"Failed to create tmp dir: ");
+		res = FALSE;
+		goto out;
+	}
+
+	iname = g_build_filename(tmpdir, "_bundle.raucb", NULL);
+	ifile = g_file_new_for_path(iname);
```
jluebbe, line 1657: I'd use a temporary file in the same directory (perhaps by appending `.tmp`), as this allows an atomic move by using g_rename(). Similar to `mksquashfs()` you should check if the destination file already exists.

src/bundle.c
```diff
@@ -1625,6 +1625,136 @@ gboolean check_bundle_payload(RaucBundle *bundle, GError **error)
...
+	if (!res) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"failed to move temporary bundle to final bundle: ");
+		goto out;
+	}
+
+	res = TRUE;
+out:
```
jluebbe, line 1753: You should remove the temporary file if it still exists here, to handle error cases. (see `out:` in `create_bundle()`).

docs/reference.rst
```diff
@@ -586,6 +586,8 @@ You can use `openssl asn1parse` for retrieving the message digest in the CMS.
 .. note::
   The `asn1parse` method can also be used for the `verity` bundle
 
+`rauc replace-signature` can replace the bundle signature with a new one.
```
jluebbe, line 589: Could you point to upstream documentation how one would create the new signature (perhaps links to the necessary OpenSSL functions)? 

This could also use a short 'pseudo-shell' example. The command to extract the signature, then refer to the custom creation and finally the cal to `rauc replace-signature`.

jluebbe (changes requested):
Looks good overall, just some minor issues.

src/bundle.c
Jarsop: Ok but if destination already exists, should I raise an error or should I overwrite the destination to have the same behavior as `resign`

docs/reference.rst
Jarsop: Yes for sure

src/bundle.c
Jarsop: So I should use `r_context_conf()->keyringpath` to load the bundle to replace signature and `r_context_conf()->keyringpath` reasigned with `r_context()->signing_keyringpath` to check reconfiguration right?

jluebbe: Yes. Then it matches the behavior of `rauc resign`.

src/bundle.c
jluebbe: It should raise an error. Overwriting in `rauc resign` sounds like a bug. :)

Jarsop:
Hello @jluebbe,

I hope you are well. I have finished reworking the functionality but I am facing a problem for the documentation of the CMS plain type part. I can easily explain how to get the digest signed but unfortunately openssl doesn't provide a way to sign a digest as input (the digest can only be calculated at signature time). In my company, we have implemented this part.

An issue has been raised [here](https://github.com/openssl/openssl/issues/14686) and a pull request has been opened to add this functionality but it looks like it won't be available until the next release [#15348](https://github.com/openssl/openssl/pull/15348).

How do you want to handle this? Disclaimer, link to PR, etc?

jluebbe:
> I hope you are well. I have finished reworking the functionality but I am facing a problem for the documentation of the CMS plain type part. I can easily explain how to get the digest signed but unfortunately openssl doesn't provide a way to sign a digest as input (the digest can only be calculated at signature time). In my company, we have implemented this part.
> 
> An issue has been raised [here](https://github.com/openssl/openssl/issues/14686) and a pull request has been opened to add this functionality but it looks like it won't be available until the next release [#15348](https://github.com/openssl/openssl/pull/15348).
> 
> How do you want to handle this? Disclaimer, link to PR, etc?

Thanks for pushing this forward! As signing the digest (without hashing) is only a problem with the 'plain' bundle format, I'd mention this in the [documentation](https://rauc.readthedocs.io/en/latest/reference.html#external-signing-and-pki), pointing to the issue and PR, perhaps as another part of the `note::`.

As I understand it, this only affects generating the new signature, so it's independent of the implementation here.

Maybe something like this: "The OpenSSL command line tool dos not support signing a pre-existing digest, so you may need to use the PR https://github.com/openssl/openssl/pull/15348. This is not necessary for a verity bundle format, as it's CMS signature directly contains the manifest."

Jarsop:
Hello @jluebbe, a new push with the realignment on `master`, rework and updated documentation has been done. Keep me informed if it's ok for you (when you have an avalaible slot on your todo list :wink: )

I thinking about a new enhancement, wouldn't it be better if "administration" commands like `convert`, `extract-signature` etc. could be compiled only for versions of `rauc` that won't be embedded on the final targets (like products). Maybe with an encapsulation in the form of a define? What do you think about this?

Jarsop:
I don't understand why the tests fail. It's seem that the dm_verity test fails. Do you have an idea about it ?

ejoerns:
> I don't understand why the tests fail. It's seem that the dm_verity test fails. Do you have an idea about it ?

Looks like you ran against an annoying known sporadic error that we have seen in CI-tests only so far.
I've retriggered the suite, let's see.

jluebbe:
> I thinking about a new enhancement, wouldn't it be better if "administration" commands like `convert`, `extract-signature` etc. could be compiled only for versions of `rauc` that won't be embedded on the final targets (like products). Maybe with an encapsulation in the form of a define? What do you think about this?

There's the `--disable-create` configure switch ("Disable bundle creation and modification commands"). If you grep for `ENABLE_CREATE`, you see how `bundle`, `resign` and `convert` are already gated this way. `extract-signature` would fit in that category as well.

ejoerns:
> > I don't understand why the tests fail. It's seem that the dm_verity test fails. Do you have an idea about it ?
> 
> Looks like you ran against an annoying known sporadic error that we have seen in CI-tests only so far. I've retriggered the suite, let's see.

All fine now. You are running against our distcheck test failure according to plan.

Jarsop:
> > > I don't understand why the tests fail. It's seem that the dm_verity test fails. Do you have an idea about it ?
> > 
> > 
> > Looks like you ran against an annoying known sporadic error that we have seen in CI-tests only so far. I've retriggered the suite, let's see.
> 
> All fine now. You are running against our distcheck test failure according to plan.

@ejoerns @jluebbe it's really strange because the output log show `cp: cannot stat '/home/runner/work/rauc/rauc/rauc-UNKNOWN/_build/sub/test/good-verity-bundle.raucb': No such file or directory` but the bundle is inside the test directory

ejoerns:
@Jarsop I guess you have to add a builddir symlink using `AC_CONFIG_LINKS()` in configure.ac to make the file available in tests. Just have a look at the already-existing long list there.

Jarsop:
> @Jarsop I guess you have to add a builddir symlink using `AC_CONFIG_LINKS()` in configure.ac to make the file available in tests. Just have a look at the already-existing long list there.

Thanks a lot it was the cause

jluebbe:
I've made some minor docs fixes, added a few comments and moved some code to make the code a bit clearer: https://github.com/rauc/rauc/compare/f519e06f06ff4b3c0d74fe4866735adc9dbcd7ca..03acdc401f5c25b1b37acd3f3772195c70659646

jluebbe (dismissed):

Jarsop:
Hello @jluebbe / @ejoerns, when do want merge this PR ? And do you want that I realign it with `master` ?

Jarsop:
> > Hello @jluebbe / @ejoerns, when do want merge this PR ? And do you want that I realign it with `master` ?
> 
> As we've now merged the CheckBundleParams changes from #462, the `verify` parameter needs to be updated.

It's done, don't hesitate if you want I change anything ! :)

src/bundle.c
```diff
@@ -1641,6 +1641,147 @@ gboolean check_bundle_payload(RaucBundle *bundle, GError **error)
...
+	GError *ierror = NULL;
+	gboolean res = FALSE;
+	gsize sigsize;
+
+	g_return_val_if_fail(bundle != NULL, FALSE);
+	g_return_val_if_fail(outpath != NULL, FALSE);
+	g_return_val_if_fail(insig != NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (!(params & CHECK_BUNDLE_NO_VERIFY) && r_context()->signing_keyringpath) {
```
ejoerns, line 1663: `CHECK_BUNDLE_NO_VERIFY` normally targets the input bundle only. Maybe we should remove it from the condition here and make the verification of the output bundle depending only on whether `--signing-keyring` was given or not? (Would additionally require to filter out and set the `CHECK_BUNDLE_NO_VERIFY` manually for this case)

It's probably also useful to describe in a short inline comment why this 'hack' is necessary and move this closer to `check_bundle()` to make the context clearer?

src/bundle.c
```diff
@@ -1641,6 +1641,147 @@ gboolean check_bundle_payload(RaucBundle *bundle, GError **error)
...
+		keyringpath = r_context()->config->keyring_path;
+		keyringdirectory = r_context()->config->keyring_directory;
+		r_context()->config->keyring_path = r_context()->signing_keyringpath;
+		r_context()->config->keyring_directory = NULL;
+	}
+
+	r_context_begin_step("replace_signature", "Replacing bundle signature", 5);
```
ejoerns, line 1670: Setting progress step should not be required here as this will not be called in an installation context.

ejoerns (changes requested):
Thanks for the adaptation! I left a few things in the comments that I found during review.

docs/reference.rst
```diff
@@ -575,16 +575,137 @@ External Signing and PKI
 Some industrialization procedures require signing artifacts in a dedicated
 secure room with restricted access (as Public Key Infrastructure aka PKI).
 
-For this case `rauc extract-signature` can extract the bundle signature.
+For this case `rauc extract-signature` can extract the bundle signature and
```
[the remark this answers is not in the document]
jluebbe: Probably, but we shouldn't handle that in this PR.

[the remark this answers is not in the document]
ejoerns: As the surrounding text is already part of the reference page, ok. Can be a follow-up later on.

src/bundle.c
jluebbe: We have the same in extract_signature and extract_bundle.

ejoerns: Not that this is better. :smirk: 
I already have a local patch for removing them, thus add it so I can remove it...

jluebbe: It also helps with verification that we follow the error exit paths. :)

src/main.c
```diff
@@ -528,6 +528,65 @@ static gboolean resign_start(int argc, char **argv)
...
+		goto out;
+	}
+
+	g_debug("input bundle: %s", argv[2]);
+	g_debug("input signature: %s", argv[3]);
+	g_debug("output file: %s", argv[4]);
+
+	if (verification_disabled)
+		check_bundle_params |= CHECK_BUNDLE_NO_VERIFY;
+	else
```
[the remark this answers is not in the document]
jluebbe: I've added support for `--trust-environment`, as with the ``convert`` command.

src/bundle.c
jluebbe: I've changed it so that it does verification based on whether a keyring was provided.

jluebbe (approved):

ejoerns (approved):

Jarsop:
@jluebbe @ejoerns : Thanks a lot to both of you for the support. If you want me to modify anything via another PR tell me without hesitation.

jluebbe:
> @jluebbe @ejoerns : Thanks a lot to both of you for the support. If you want me to modify anything via another PR tell me without hesitation.

It would be very helpful if you could try the current master in your environment, to make sure we didn't break something for you.

Issue #679, opened 2021-01-06
casync bundle creation fails with "unable to find mounted device for bundle" when building inside container


mschwan-phytec:
Since RAUC version 1.5 creating casync bundles no longer works. This is the relevant output when the `bundle.bbclass` executes the convert command to create the casync bundle:
```
rauc-Message: 13:21:55.950: Debug log domains: 'rauc'
(rauc:34224): rauc-DEBUG: 13:21:55.951: convert start
(rauc:34224): rauc-DEBUG: 13:21:55.951: system config not found, using default values
rauc-Message: 13:21:55.951: Failed to resolve realpath for '/dev/disk/by-uuid/be5faaeb-1680-4ca2-af65-f30f06dab903'
(rauc:34224): rauc-DEBUG: 13:21:55.951: input bundle: /home/user/documents/yocto/imx6ul/dunfell/build/tmp/work/machine/headless-bundle/1.0-r0/build/bundle.raucb
(rauc:34224): rauc-DEBUG: 13:21:55.951: output bundle: /home/user/documents/yocto/imx6ul/dunfell/build/tmp/work/machine/headless-bundle/1.0-r0/build/casync-bundle.raucb
rauc-Message: 13:21:55.951: Reading bundle: /home/user/documents/yocto/imx6ul/dunfell/build/tmp/work/machine/headless-bundle/1.0-r0/build/bundle.raucb
rauc-Message: 13:21:55.952: Verifying bundle signature...
(rauc:34224): rauc-DEBUG: 13:21:55.952: initial check_bundle_access failed with: unable to find mounted device for bundle
unable to find mounted device for bundle
WARNING: exit code 1 from a shell command.
```
I am looking at the code changes that version 1.5 brings but I do not really understand yet what the exact issue is with this one. AFAIK `check_bundle_access()` checks several parameters to prevent unintended write access to the bundle. Though I am unsure which device the error message is referring to and what is wrong with it.

jluebbe:
RAUC is trying to find the device where the bundle is located, to check if it is safe against concurrent modification. This may be too strict for the convert case...

Where is your `/home/user/documents/yocto/imx6ul/dunfell/build/tmp/work/machine/headless-bundle/1.0-r0/build/bundle.raucb` mounted from (according to mount)? Which filesystem is it? Does the /dev/disk/by-uuid/be5faaeb-1680-4ca2-af65-f30f06dab903 point to the correct block device?

jluebbe:
[...]

PR #769, opened 2021-09-10
src/main: allow passing 'no-verify' to rauc convert

ejoerns:
This option allows to disable the verification of the input bundle (and thus omitting --keyring argument) when converting a conventional to a casync-based bundle.

However, this is not recommended as signature checks are there for a reason of course.

But, as we also already allow this for similar subcommands like ('info' or 'resign') and should give the responsible user the freedom to 'know what he's doing', we enable this for convert, too.

Another quite relevant point where this can be used as a workaround is when being inside an environment (e.g. a docker container) where the standard RAUC bundle access check mechanisms must fail) and trigger errors like

> initial check_bundle_access failed with: unable to find mounted device for bundle

This happens when RAUC attempts to find the underlying device of the mountpoint the bundle is located at but cannot as the device is hidden by the container.

Addresses: #679, #688

### DIFF
--- a/rauc.1
+++ b/rauc.1
@@ -157,6 +157,10 @@ Convert an existing bundle to casync index bundle and store.
 .RS 4
 
 .TP
+\fB\-\-no\-verify\fR
+disable bundle verification
+
+.TP
 \fB\-\-signing\-keyring=\fR\fIPEMFILE\fR
 verification keyring file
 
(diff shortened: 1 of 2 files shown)


ejoerns:
Note: needs to be adapted/rebased if #462 is merged before.

Issue #773, opened 2021-09-16
casync seed mount of rootfs needs to detect mount options


j-licht:
Hi,

I'm currently playing around with casync support in rauc and two redundant rootfs slots.
Rauc mounts the booted slot to be used as seed for casync.
Usually my rootfs is mounted ro and rauc tries to mount it rw.
This fails with "device already mounted".
I think the problem here is, that this remount works only if the ReadWrite Option is the same.
A solution could be to detect the current mount options of the rootfs.

ejoerns:
Note that this mount behavior was addressed with https://github.com/util-linux/util-linux/commit/11b916cdabd9f57a4cddf10fd743af2165930dcd that now instead gives something like:

    mount: /run/rauc/rootfs.0: WARNING: source write-protected, mounted read-only

However, choosing the right mount options when invoking 'mount' should still be the proper solution.

jluebbe:
The possible fix is outlined in https://github.com/rauc/rauc/pull/774#issuecomment-1263316459.

PR #774, opened 2021-09-21
casync fixes

j-licht:
Check mount options of the seed slot.
Unmount seed slot on casync failure too.


src/update_handler.c
```diff
@@ -323,6 +323,7 @@ static gboolean casync_extract_image(RaucImage *image, gchar *dest, int out_fd,
 		 * file systems, additional mounts, etc. */
 		if (!seedslot->mount_point) {
 			g_debug("Mounting %s to use as seed", seedslot->device);
+			seedslot->extra_mount_opts = r_get_slot_current_mount_options(seedslot);
```
jluebbe, line 326: We can't just replace `extra_mount_opts` here. Also, we should just use read-only mounts for seeding.

ejoerns:
I have now rebased this on #957 and reworked it to give it a chance to get merged.

Things I changed:

* added descriptions to the commit messages
* applied uncrustify changes
* removed "ext4" from "Unmounting ext4 seed slot.."
* reworked adding "ro" mount opt while preserving already-exisiting extra mount options.

src/update_handler.c
ejoerns: It is now extended instead of being replaced.

As it accepted read-write slots before and we cannot easily determine what is actually readwrite and what is not (e.g. readonly file systems) I have left this untouched for now.

jluebbe:
I've rebased it on master after the merge of #957 to make it easier to review.

src/update_handler.c
jluebbe: What I meant is that `extra_mount_opts` is configuration, so we can't change it here. Also we don't need to check if the slot is currently mounted read-only, as an additional read-only mount should always work.

jluebbe:
We've split off the simple fix as #977. When that is merged, only the seed slot handling remains.

At that point, we should probably change it to use `slot->ext_mount_point` as the source of a ro-bind-mount. That side-steps the entire issue of the mount command complaining about '<block device> already mounted on <external mount point>', when trying to create an ro-mount of an existing rw-mounted FS.

jluebbe:
> We've split off the simple fix as #977. When that is merged, only the seed slot handling remains.
> 
> At that point, we should probably change it to use `slot->ext_mount_point` as the source of a ro-bind-mount. That side-steps the entire issue of the mount command complaining about ' already mounted on ', when trying to create an ro-mount of an existing rw-mounted FS.

@j-licht Sorry for the long silence. Would you be able to run a test if we implemented this approach in a PR?

j-licht:
@jluebbe thanks for coming back. I recently tested with 1.12 and it seems it's working now without my patch.
I'm getting this warning: `mount: /run/rauc/rootfs.0: WARNING: source write-protected, mounted read-only.`
And the casync command looks like this `casync extract --seed /run/rauc/rootfs.0 --store https://update.net/castr/ --seed-output=no /run/rauc/bundle/rootfs.tar.bz2.caidx /` 

So I think we could close this PR and #773 ?!

PR #775, opened 2021-09-28
utils: Misc fixes

LeSpocky:
Minor fixes which accumulated when hacking on different things. Patches 1 and 2 are self contained. Patch 4 depends on patch 3, those two could be squashed if desired.

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -134,6 +134,8 @@ G_GNUC_WARN_UNUSED_RESULT;
 
 /**
  * Get string argument from key and remove key from key_file.
+ *
+ * @return A newly allocated string or NULL on error.
  */
 gchar * key_file_consume_string(
 		GKeyFile *key_file,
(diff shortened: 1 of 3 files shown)


LeSpocky:
The last change which just affected documentation let the code coverage test fail. I have no idea why. 🤷🏼‍♂️

ejoerns:
Hm, not sure how it concludes that total coverage in main.c decreased which is not touched at all. But this is not a blocker for merging, anyway.

include/utils.h
```diff
@@ -134,6 +134,8 @@ G_GNUC_WARN_UNUSED_RESULT;
 
 /**
  * Get string argument from key and remove key from key_file.
+ *
+ * @return A newly allocated string or NULL.
```
ejoerns, line 138: Maybe even more clearer with:

``` suggestion
 * @return A newly allocated string or NULL on error.
```

In a follow up we should then also add documentation for the parameters, but I would not see that as the blocker for this initial enhancement ;)

ejoerns (approved):
:+1:

jluebbe (approved):

PR #777, opened 2021-09-30
RFC: Support alternative ubi rootfs format

LeSpocky:
The documentation currently has this note in section *6.6.1.1. Identification via Kernel Commandline*:

> The alternative ubi rootfs format with `root=ubi0:volname` is currently unsupported.

This patch series tries to solve that issue by resolving that format through sysfs poking for both cmdline parameter detection and configuration file parsing.

Expected issues:

- lack of tests (mostly because it would need a system with actual ubi device and ubifs volumes, and I don't know how to provide that)
- few logging or error code passing in case of errors (returns NULL in error cases)
- support of `device=UUID=…` in config file completely untested and maybe not without side effects

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -369,10 +369,13 @@ RAUC supports parsing different variants for giving these device as listed below
 
 Giving the plain device name is supported, of course.
 
-.. note::
+::
 
-  The alternative ubi rootfs format with ``root=ubi0:volname`` is currently
-  unsupported.
+  root=ubi0:volname
+
+The alternative ubi rootfs format with ``ubi0:volname`` is supported.
+The volume ID and like it the plain device path is dependent on volume creation order.
+RAUC converts the value to the actual device path by resolving it through sysfs.
 
 ::
 
--- a/include/utils.h
+++ b/include/utils.h
@@ -185,6 +185,7 @@ G_GNUC_WARN_UNUSED_RESULT;
  * - PARTLABEL=mylabel -> /dev/disk/by-partlabel/mylabel
  * - PARTUUID=4f8bb419-01 -> /dev/disk/by-partuuid/4f8bb419-01
  * - UUID=9e8b0c3e-e20f-4119-b419-ec20a132aa94 -> /dev/disk/by-uuid/9e8b0c3e-e20f-4119-b419-ec20a132aa94
+ * - ubi0:rootfs1 -> /dev/ubi0_3
  *
  * @param dev "device" string
  *
--- a/include/utils.h
+++ b/include/utils.h
@@ -191,3 +191,25 @@ G_GNUC_WARN_UNUSED_RESULT;
  * @return Resolved device part (newly allocated string) or NULL in case of an error.
  */
 gchar *r_resolve_device(const gchar *dev);
+
+/**
+ * Parse UBI device name string the same way the kernel does and return
+ * sysfs path of that volume.
+ *
+ * Those names are used in kernel cmdline for mounting rootfs, e.g. like
+ * 'root=ubi0:volname', possible names are:
+ *
+ * - ubiX_Y	UBI device number X, volume Y
+ * - ubiY	UBI device number 0, volume Y
+ * - ubiX:NAME	UBI device X, volume with name NAME
+ * - ubi:NAME	UBI device 0, volume with name NAME
+ *
+ * The kernel allows '!' as an alternative separator instead of ':'
+ * (because some shells like busybox may interpret ':' as an NFS host
+ * name separator).
+ *
+ * @param name string, one of format "ubiX_Y", "ubiY", "ubiX:NAME", "ubi:NAME".
+ *
+ * @return sysfs path like "/sys/class/ubi/ubiX/ubiX_Y" (newly-allocated string) or NULL.
+ */
+gchar *r_ubi_name_to_sysfs_path(const gchar *name);
--- a/include/utils.h
+++ b/include/utils.h
@@ -177,3 +177,17 @@ guint8 *r_hex_decode(const gchar *hex, size_t len)
 G_GNUC_WARN_UNUSED_RESULT;
 gchar *r_hex_encode(const guint8 *raw, size_t len)
 G_GNUC_WARN_UNUSED_RESULT;
+
+/**
+ * Take the value of a root= cmdline argument and resolve it to a
+ * device path, for example:
+ *
+ * - PARTLABEL=mylabel -> /dev/disk/by-partlabel/mylabel
+ * - PARTUUID=4f8bb419-01 -> /dev/disk/by-partuuid/4f8bb419-01
+ * - UUID=9e8b0c3e-e20f-4119-b419-ec20a132aa94 -> /dev/disk/by-uuid/9e8b0c3e-e20f-4119-b419-ec20a132aa94
+ *
+ * @param dev "device" string
+ *
+ * @return Resolved device part (newly allocated string) or NULL in case of an error.
+ */
+gchar *r_resolve_device(const gchar *dev);
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -493,6 +493,7 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 		if (g_str_equal(groupsplit[0], RAUC_SLOT_PREFIX)) {
 			g_autoptr(RaucSlot) slot = g_new0(RaucSlot, 1);
 			gchar* value;
+			gchar *ival;
 
 			/* Assure slot strings consist of 3 parts, delimited by dots */
 			if (g_strv_length(groupsplit) != 3) {
@@ -522,14 +523,30 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 
 			slot->sclass = g_intern_string(groupsplit[1]);
 
-			value = resolve_path(filename,
-					key_file_consume_string(key_file, groups[i], "device", &ierror));
+			/* Try to resolve special device name to an actual
+			 * device path. If it is no special device name,
+			 * just use ordinary resolve path logic. */
+			value = key_file_consume_string(key_file, groups[i], "device", &ierror);
 			if (!value) {
 				g_propagate_error(error, ierror);
 				res = FALSE;
 				goto free;
 			}
-			slot->device = value;
+			ival = r_resolve_device(value);
+			if (!ival) {
+				ival = resolve_path(filename, value);
+			}
+			g_free(value);
+			if (!ival) {
+				g_set_error(
+						error,
+						R_CONFIG_ERROR,
+						R_CONFIG_ERROR_INVALID_FORMAT,
+						"Invalid device");
+				res = FALSE;
+				goto free;
+			}
+			slot->device = ival;
 
 			value = key_file_consume_string(key_file, groups[i], "type", NULL);
 			if (!value)
--- a/src/context.c
+++ b/src/context.c
@@ -29,6 +29,7 @@ static const gchar* get_cmdline_bootname(void)
 	g_autofree gchar *contents = NULL;
 	g_autofree gchar *realdev = NULL;
 	const char *bootname = NULL;
+	gchar *devpath = NULL;
 
 	if (context->mock.proc_cmdline)
 		contents = g_strdup(context->mock.proc_cmdline);
@@ -59,37 +60,10 @@ static const gchar* get_cmdline_bootname(void)
 	if (!bootname)
 		return NULL;
 
-	if (strncmp(bootname, "PARTLABEL=", 10) == 0) {
-		gchar *partlabelpath = g_build_filename(
-				"/dev/disk/by-partlabel/",
-				&bootname[10],
-				NULL);
-		if (partlabelpath) {
-			g_free((gchar*) bootname);
-			bootname = partlabelpath;
-		}
-	}
-
-	if (strncmp(bootname, "PARTUUID=", 9) == 0) {
-		gchar *partuuidpath = g_build_filename(
-				"/dev/disk/by-partuuid/",
-				&bootname[9],
-				NULL);
-		if (partuuidpath) {
-			g_free((gchar*) bootname);
-			bootname = partuuidpath;
-		}
-	}
-
-	if (strncmp(bootname, "UUID=", 5) == 0) {
-		gchar *uuidpath = g_build_filename(
-				"/dev/disk/by-uuid/",
-				&bootname[5],
-				NULL);
-		if (uuidpath) {
-			g_free((gchar*) bootname);
-			bootname = uuidpath;
-		}
+	devpath = r_resolve_device(bootname);
+	if (devpath) {
+		g_free((gchar*) bootname);
+		bootname = devpath;
 	}
 
 	realdev = r_realpath(bootname);
--- a/src/utils.c
+++ b/src/utils.c
@@ -1,3 +1,4 @@
+#include <ctype.h>
 #include <errno.h>
 #include <ftw.h>
 #include <gio/gio.h>
@@ -9,6 +10,34 @@
 #include <unistd.h>
 
 #include "utils.h"
+
+#define R_SYSFS_UBI_BASE_PATH	"/sys/class/ubi"
+
+/**
+ * Resolve UBI volume sysfs path to device path.
+ *
+ * This is done through major:minor device number.
+ *
+ * @param path sysfs path like "/sys/class/ubi/ubiX/ubiX_Y"
+ *
+ * @return device path like "/dev/ubi0_2" (newly-allocated string) or NULL.
+ */
+static gchar *r_ubi_sysfs_to_dev_path(const gchar *path);
+
+/**
+ * Search sysfs for UBI volume name.
+ *
+ * UBI volume names can be read from sysfs path
+ * '/sys/class/ubi/ubiX/ubiX_Y/name', so to get the volume id Y by
+ * name, step through all volumes of dev X, and compare name until
+ * match.
+ *
+ * @param dev UBI device number, e.g. 0 for ubi0.
+ * @param name UBI volume name, e.g. the NAME part of a 'ubiX:NAME'
+ *
+ * @return sysfs path like "/sys/class/ubi/ubiX/ubiX_Y" (newly-allocated string) or NULL.
+ */
+static gchar *r_ubi_volname_to_sysfs_path(unsigned int dev, const gchar *name);
 
 GSubprocess *r_subprocess_new(GSubprocessFlags flags, GError **error, const gchar *argv0, ...)
 {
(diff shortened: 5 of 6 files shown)


LeSpocky:
Does this overlap with #406? 🤔

jluebbe:
Could you comment on why you prefer the `root=ubi0:volname` format `root=/dev/ubi0_01`? What are the expected benefits?

jluebbe:
> > Could you comment on why you prefer the `root=ubi0:volname` format `root=/dev/ubi0_01`? What are the expected benefits?
> 
> I tried to explain in the commit messages, but when reading over it again, it seems rather scattered. So:
> 
> * volume IDs (the last part of the device name, e.g. the _2_ in `/dev/ubi0_2`) are dependent on volume creation order
> * U-Boot _distroboot_ "framework" uses volume names
> * easier for the admin because you can put meaning in names, you can not in numbers (if I read _ubi0:rootfs1_ in a config file or kernel cmdline it directly means something to me, a `/dev/ubi0_4` carries no meaning, could be any volume)
> 
> The volume creation order thing is probably the most important. Currently I have seven volumes in my UBI partition:
> 
> 1. env
> 2. env_r
> 3. boot
> 4. rootfs0
> 5. rootfs1
> 6. rootfs2
> 7. data
> 
> With the previous approach I would not only have to make sure those volumes are created in the correct order when created for the first time, but also have to make sure no one removes volumes and recreates them in a different order on the running device.

Thanks for explaining. In cases like this, I'd usually let the bootloader set `rauc.slot=<bootname>`, which avoid this whole issue by being explicit.

I assume that U-Boots *distroboot* is not doing the A/B selection, so you could add the `rauc.slot=` parameter before calling distroboot on slot A or B? Does this have downsides I'm not aware of?

> Expected issues:
> 
> * lack of tests (mostly because it would need a system with actual ubi device and ubifs volumes, and I don't know how to provide that)

Take a look around `./qemu-test shell`, it should have a ubi volume prepared.

> * few logging or error code passing in case of errors (returns NULL in error cases)

Yes. We should also log some information on how the device names in the config are resolved (when that is used).

Perhaps first try to use it as a path first and only resolve it as a device if that fails?

> * support of `device=UUID=…` in config file completely untested and maybe not without side effects

Hmm. It should work, but I'm not sure if it's a good idea to allow that. So far we've recommended to configure that once in udev and just let RAUC use that. That would also work for ubi (by generating `/dev/ubi0_root` names, perhaps even based on mtd partition names). On the other hand, allowing the same strings as in `root=` also has benefits…

LeSpocky:
> > > Could you comment on why you prefer the `root=ubi0:volname` format `root=/dev/ubi0_01`? What are the expected benefits?
> > 
> > 
> > I tried to explain in the commit messages, but when reading over it again, it seems rather scattered. So:
> > 
> > * volume IDs (the last part of the device name, e.g. the _2_ in `/dev/ubi0_2`) are dependent on volume creation order
> > * U-Boot _distroboot_ "framework" uses volume names
> > * easier for the admin because you can put meaning in names, you can not in numbers (if I read _ubi0:rootfs1_ in a config file or kernel cmdline it directly means something to me, a `/dev/ubi0_4` carries no meaning, could be any volume)
> > 
> > The volume creation order thing is probably the most important. Currently I have seven volumes in my UBI partition:
> > 
> > 1. env
> > 2. env_r
> > 3. boot
> > 4. rootfs0
> > 5. rootfs1
> > 6. rootfs2
> > 7. data
> > 
> > With the previous approach I would not only have to make sure those volumes are created in the correct order when created for the first time, but also have to make sure no one removes volumes and recreates them in a different order on the running device.
> 
> Thanks for explaining. In cases like this, I'd usually let the bootloader set `rauc.slot=<bootname>`, which avoid this whole issue by being explicit.
> 
> I assume that U-Boots _distroboot_ is not doing the A/B selection, so you could add the `rauc.slot=` parameter before calling distroboot on slot A or B? Does this have downsides I'm not aware of?

Just double checked my u-boot script and we actually set `rauc.slot=<bootname>` and append it to kernel command line. So this would just be a fallback to get `root=ubi0:volname` from kernel command line parsed if `rauc.slot=<bootname>` is missing. Thus this might even be dropped if the device path in RAUC config file thing can be solved.

> > * support of `device=UUID=…` in config file completely untested and maybe not without side effects
> 
> Hmm. It should work, but I'm not sure if it's a good idea to allow that. So far we've recommended to configure that once in udev and just let RAUC use that. 

Sure you can use a format like `/dev/disk/by-uuid/3aa1cd8a-1234-1234-1234-7c97ebc84861` then. Are those symlinks created by udev?

> That would also work for ubi (by generating `/dev/ubi0_root` names, perhaps even based on mtd partition names). On the other hand, allowing the same strings as in `root=` also has benefits…

Okay the idea is to let someone create symlinks in /dev based on the volume name pointing to the actual ubi volume device? I like that in general. mtd however does not help here. See `lsmtd` output:

```
lsmtd
DEVICE     MAJ:MIN NAME      TYPE    SIZE   
mtd0        90:0   bootstrap nand      256K
mtd1        90:2   u-boot    nand      1.3M
mtd2        90:4   reserved  nand      1.5M
mtd3        90:6   UBI       nand      509M
└─ubi0     243:0                     492.6M
  ├─ubi0_0 243:1   env       static    124K
  ├─ubi0_1 243:2   env_r     static    124K
  ├─ubi0_2 243:3   boot      dynamic  16.1M
  ├─ubi0_3 243:4   rootfs0   dynamic  42.7M
  ├─ubi0_4 243:5   rootfs1   dynamic   120M
  ├─ubi0_5 243:6   rootfs2   dynamic   120M
  └─ubi0_6 243:7   data      dynamic 183.8M
```

Here ubi0 is always on top of mtd3, that part is static and won't change. What is (or might be) unstable is the relation between UBI volumes (rootfs0, data, …) to those ubi0_x devices. 

So if we had something creating symlinks like for example `/dev/ubi/ubi0/rootfs0` pointing to `/dev/ubi0_3` we would not need to change RAUC at all. Reminds me on LVM logical volumes, where I can access a `/dev/dm-0` device node through either `/dev/mapper/vg1-lvroot` or `/dev/vg1/lvroot` and don't have to care about device node numbering.  Question is: who can or should do that? udev or the kernel itself? I would prefer a solution without udev. Will ask on linux-mtd if this would be possible.

jluebbe:
> > > > Could you comment on why you prefer the `root=ubi0:volname` format `root=/dev/ubi0_01`? What are the expected benefits?
> > Thanks for explaining. In cases like this, I'd usually let the bootloader set `rauc.slot=<bootname>`, which avoid this whole issue by being explicit.
> > I assume that U-Boots _distroboot_ is not doing the A/B selection, so you could add the `rauc.slot=` parameter before calling distroboot on slot A or B? Does this have downsides I'm not aware of?
> 
> Just double checked my u-boot script and we actually set `rauc.slot=<bootname>` and append it to kernel command line. So this would just be a fallback to get `root=ubi0:volname` from kernel command line parsed if `rauc.slot=<bootname>` is missing. Thus this might even be dropped if the device path in RAUC config file thing can be solved.

OK. The [documentation](https://rauc.readthedocs.io/en/latest/integration.html#booted-slot-detection) already strongly recommends using the explicit approach with `rauc.slot=`.

> > > * support of `device=UUID=…` in config file completely untested and maybe not without side effects
> > 
> > 
> > Hmm. It should work, but I'm not sure if it's a good idea to allow that. So far we've recommended to configure that once in udev and just let RAUC use that.
> 
> Sure you can use a format like `/dev/disk/by-uuid/3aa1cd8a-1234-1234-1234-7c97ebc84861` then. Are those symlinks created by udev?

Yes, all symlinks in /dev are created by udev. These kinds of indirection is best handled there.

> > That would also work for ubi (by generating `/dev/ubi0_root` names, perhaps even based on mtd partition names). On the other hand, allowing the same strings as in `root=` also has benefits…
> 
> Okay the idea is to let someone create symlinks in /dev based on the volume name pointing to the actual ubi volume device? I like that in general. mtd however does not help here. See `lsmtd` output:
> 
> ```
> lsmtd
> DEVICE     MAJ:MIN NAME      TYPE    SIZE   
> mtd0        90:0   bootstrap nand      256K
> mtd1        90:2   u-boot    nand      1.3M
> mtd2        90:4   reserved  nand      1.5M
> mtd3        90:6   UBI       nand      509M
> └─ubi0     243:0                     492.6M
>   ├─ubi0_0 243:1   env       static    124K
>   ├─ubi0_1 243:2   env_r     static    124K
>   ├─ubi0_2 243:3   boot      dynamic  16.1M
>   ├─ubi0_3 243:4   rootfs0   dynamic  42.7M
>   ├─ubi0_4 243:5   rootfs1   dynamic   120M
>   ├─ubi0_5 243:6   rootfs2   dynamic   120M
>   └─ubi0_6 243:7   data      dynamic 183.8M
> ```
> 
> Here ubi0 is always on top of mtd3, that part is static and won't change. What is (or might be) unstable is the relation between UBI volumes (rootfs0, data, …) to those ubi0_x devices.

Previously I've suggested to use something like ```SUBSYSTEM=="ubi", KERNEL=="ubi*_*", SYMLINK+="ubi_mtd%s{mtd_num}_%s{name}"``` (see https://github.com/rauc/rauc/pull/370#issuecomment-448969191).

I'm not sure how much udev-specific recommendations are appropriate for the RAUC docs. Perhaps as a FAQ. :)

> So if we had something creating symlinks like for example `/dev/ubi/ubi0/rootfs0` pointing to `/dev/ubi0_3` we would not need to change RAUC at all. Reminds me on LVM logical volumes, where I can access a `/dev/dm-0` device node through either `/dev/mapper/vg1-lvroot` or `/dev/vg1/lvroot` and don't have to care about device node numbering. Question is: who can or should do that? udev or the kernel itself? I would prefer a solution without udev. Will ask on linux-mtd if this would be possible.

Yes, this sort of mapping is usually done in a project-specific udev rule. Another benefit is that it also allows you to use the same path everywhere in userspace (RAUC, mount units, statistics, fs checks, ...).

LeSpocky:
> > Sure you can use a format like `/dev/disk/by-uuid/3aa1cd8a-1234-1234-1234-7c97ebc84861` then. Are those symlinks created by udev?
> 
> Yes, all symlinks in /dev are created by udev. These kinds of indirection is best handled there.
> 
> > > That would also work for ubi (by generating `/dev/ubi0_root` names, perhaps even based on mtd partition names). On the other hand, allowing the same strings as in `root=` also has benefits…
> > 
> > 
> > Okay the idea is to let someone create symlinks in /dev based on the volume name pointing to the actual ubi volume device? I like that in general. mtd however does not help here. See `lsmtd` output:
> > ```
> > lsmtd
> > DEVICE     MAJ:MIN NAME      TYPE    SIZE   
> > mtd0        90:0   bootstrap nand      256K
> > mtd1        90:2   u-boot    nand      1.3M
> > mtd2        90:4   reserved  nand      1.5M
> > mtd3        90:6   UBI       nand      509M
> > └─ubi0     243:0                     492.6M
> >   ├─ubi0_0 243:1   env       static    124K
> >   ├─ubi0_1 243:2   env_r     static    124K
> >   ├─ubi0_2 243:3   boot      dynamic  16.1M
> >   ├─ubi0_3 243:4   rootfs0   dynamic  42.7M
> >   ├─ubi0_4 243:5   rootfs1   dynamic   120M
> >   ├─ubi0_5 243:6   rootfs2   dynamic   120M
> >   └─ubi0_6 243:7   data      dynamic 183.8M
> > ```
> >   
> > Here ubi0 is always on top of mtd3, that part is static and won't change. What is (or might be) unstable is the relation between UBI volumes (rootfs0, data, …) to those ubi0_x devices.
> 
> Previously I've suggested to use something like `SUBSYSTEM=="ubi", KERNEL=="ubi*_*", SYMLINK+="ubi_mtd%s{mtd_num}_%s{name}"` (see [#370 (comment)](https://github.com/rauc/rauc/pull/370#issuecomment-448969191)).
> 
> I'm not sure how much udev-specific recommendations are appropriate for the RAUC docs. Perhaps as a FAQ. :)

Those rules just look at the kernel commandline and that has the device node, not a volume name. Plus: I don't see how this would generate matching symlinks needed for the inactive slots, which are not the currently active rootfs. 

And again: I don't see the need for mtd related things here. For ubifs (not ubi!) we just need the `/dev/ubiX_Y` and it's not relevant which is the underlying mtd. That might be interesting for updating a whole ubi partition, but that's not scope of my problem, because I aim at ubifs on ubi volumes only.

> > So if we had something creating symlinks like for example `/dev/ubi/ubi0/rootfs0` pointing to `/dev/ubi0_3` we would not need to change RAUC at all. Reminds me on LVM logical volumes, where I can access a `/dev/dm-0` device node through either `/dev/mapper/vg1-lvroot` or `/dev/vg1/lvroot` and don't have to care about device node numbering. Question is: who can or should do that? udev or the kernel itself? I would prefer a solution without udev. Will ask on linux-mtd if this would be possible.
> 
> Yes, this sort of mapping is usually done in a project-specific udev rule. Another benefit is that it also allows you to use the same path everywhere in userspace (RAUC, mount units, statistics, fs checks, ...).

This would be a huge benefit indeed. Okay then we have to ask systemd/udev how to create generic rules working with all ubi volumes. Those are not part of upstream systemd as of today.

(Besides: this udev topic is interesting and would allow to use volume names without this PR, but without udev it would still not be possible and I have space constrained systems where systemd + all the dependencies is no option.)

jluebbe:
> > > Here ubi0 is always on top of mtd3, that part is static and won't change. What is (or might be) unstable is the relation between UBI volumes (rootfs0, data, …) to those ubi0_x devices.
> > 
> > 
> > Previously I've suggested to use something like `SUBSYSTEM=="ubi", KERNEL=="ubi*_*", SYMLINK+="ubi_mtd%s{mtd_num}_%s{name}"` (see [#370 (comment)](https://github.com/rauc/rauc/pull/370#issuecomment-448969191)).
> > I'm not sure how much udev-specific recommendations are appropriate for the RAUC docs. Perhaps as a FAQ. :)
> 
> Those rules just look at the kernel commandline and that has the device node, not a volume name. Plus: I don't see how this would generate matching symlinks needed for the inactive slots, which are not the currently active rootfs.

The one I quoted here runs for every UBI volume and does not use the kernel commandline. I added the link to the discussion for some background what is possible as well. :)

> And again: I don't see the need for mtd related things here. For ubifs (not ubi!) we just need the `/dev/ubiX_Y` and it's not relevant which is the underlying mtd. That might be interesting for updating a whole ubi partition, but that's not scope of my problem, because I aim at ubifs on ubi volumes only.

By using the mtd number (or name with a different rule) as part of the link, you make the UBI volume links independent of probe, attach and partition order. The result would be links from `/dev/ubi_mtd1_rootfs2` or `/dev/ubi_mtdlinux_rootfs2` to `/dev/ubi0_1`.

> > > So if we had something creating symlinks like for example `/dev/ubi/ubi0/rootfs0` pointing to `/dev/ubi0_3` we would not need to change RAUC at all. Reminds me on LVM logical volumes, where I can access a `/dev/dm-0` device node through either `/dev/mapper/vg1-lvroot` or `/dev/vg1/lvroot` and don't have to care about device node numbering. Question is: who can or should do that? udev or the kernel itself? I would prefer a solution without udev. Will ask on linux-mtd if this would be possible.
> > 
> > 
> > Yes, this sort of mapping is usually done in a project-specific udev rule. Another benefit is that it also allows you to use the same path everywhere in userspace (RAUC, mount units, statistics, fs checks, ...).
> 
> This would be a huge benefit indeed. Okay then we have to ask systemd/udev how to create generic rules working with all ubi volumes. Those are not part of upstream systemd as of today.

I've encountered many different opinions on how these links should be named, so I've not tried to push one of the variants upstream yet. On there other hand, we often have project-specific anyway (for permissions, systemd tags, abstracting HW variants, ...), having a custom rule for UBI links is not a large maintenance burden. :)

> (Besides: this udev topic is interesting and would allow to use volume names without this PR, but without udev it would still not be possible and I have space constrained systems where systemd + all the dependencies is no option.)

I suspect that the udev alternatives have similar features for creating links. Without udev (or an alternative) you're also missing the `/dev/disk/by-uuid/` and related links, so I'd be fine with requiring `rauc.slot=` in that case.

LeSpocky:
> > > Previously I've suggested to use something like `SUBSYSTEM=="ubi", KERNEL=="ubi*_*", SYMLINK+="ubi_mtd%s{mtd_num}_%s{name}"` (see [#370 (comment)](https://github.com/rauc/rauc/pull/370#issuecomment-448969191)).
> > > I'm not sure how much udev-specific recommendations are appropriate for the RAUC docs. Perhaps as a FAQ. :)
> > 
> > 
> > Those rules just look at the kernel commandline and that has the device node, not a volume name. Plus: I don't see how this would generate matching symlinks needed for the inactive slots, which are not the currently active rootfs.
> 
> The one I quoted here runs for every UBI volume and does not use the kernel commandline. I added the link to the discussion for some background what is possible as well. :)
> 
> > And again: I don't see the need for mtd related things here. For ubifs (not ubi!) we just need the `/dev/ubiX_Y` and it's not relevant which is the underlying mtd. That might be interesting for updating a whole ubi partition, but that's not scope of my problem, because I aim at ubifs on ubi volumes only.
> 
> By using the mtd number (or name with a different rule) as part of the link, you make the UBI volume links independent of probe, attach and partition order. The result would be links from `/dev/ubi_mtd1_rootfs2` or `/dev/ubi_mtdlinux_rootfs2` to `/dev/ubi0_1`.

Okay you're trying to rule out ubi0 vs. ubi1 etc? But what about those mtd device numbers then? Those are not necessarily stable either if you have multiple mtd devices, right? Anyways, I got udev can create symlinks (knew that) and more important it gets the volume name from sysfs, which you can use in the symlink name then.

Tried your rule now, gives this here:

```
root@unit:~ ls -l /dev/ubi*
crw-------    1 root     root      243,   0 Nov 18 12:22 /dev/ubi0
crw-------    1 root     root      243,   1 Nov 18 12:22 /dev/ubi0_0
crw-------    1 root     root      243,   2 Nov 18 12:22 /dev/ubi0_1
crw-------    1 root     root      243,   3 Nov 18 12:22 /dev/ubi0_2
crw-------    1 root     root      243,   4 Nov 18 12:22 /dev/ubi0_3
crw-------    1 root     root      243,   5 Nov 18 12:22 /dev/ubi0_4
crw-------    1 root     root      243,   6 Nov 18 12:22 /dev/ubi0_5
crw-------    1 root     root      243,   7 Nov 18 12:22 /dev/ubi0_6
crw-------    1 root     root       10,  62 Nov 18 12:22 /dev/ubi_ctrl
lrwxrwxrwx    1 root     root             6 Nov 18 12:22 /dev/ubi_mtd_boot -> ubi0_2
lrwxrwxrwx    1 root     root             6 Nov 18 12:22 /dev/ubi_mtd_data -> ubi0_6
lrwxrwxrwx    1 root     root             6 Nov 18 12:22 /dev/ubi_mtd_env -> ubi0_0
lrwxrwxrwx    1 root     root             6 Nov 18 12:22 /dev/ubi_mtd_env_r -> ubi0_1
lrwxrwxrwx    1 root     root             6 Nov 18 12:22 /dev/ubi_mtd_rootfs0 -> ubi0_3
lrwxrwxrwx    1 root     root             6 Nov 18 12:22 /dev/ubi_mtd_rootfs1 -> ubi0_4
lrwxrwxrwx    1 root     root             6 Nov 18 12:22 /dev/ubi_mtd_rootfs2 -> ubi0_5
```

(Not sure why *mtd_num* is empty, `udevadm info --attribute-walk /dev/ubi0_5` lists `ATTRS{mtd_num}=="3"` for the parent device.)

> > > > So if we had something creating symlinks like for example `/dev/ubi/ubi0/rootfs0` pointing to `/dev/ubi0_3` we would not need to change RAUC at all. Reminds me on LVM logical volumes, where I can access a `/dev/dm-0` device node through either `/dev/mapper/vg1-lvroot` or `/dev/vg1/lvroot` and don't have to care about device node numbering. Question is: who can or should do that? udev or the kernel itself? I would prefer a solution without udev. Will ask on linux-mtd if this would be possible.
> > > 
> > > 
> > > Yes, this sort of mapping is usually done in a project-specific udev rule. Another benefit is that it also allows you to use the same path everywhere in userspace (RAUC, mount units, statistics, fs checks, ...).
> > 
> > 
> > This would be a huge benefit indeed. Okay then we have to ask systemd/udev how to create generic rules working with all ubi volumes. Those are not part of upstream systemd as of today.
> 
> I've encountered many different opinions on how these links should be named, so I've not tried to push one of the variants upstream yet. On there other hand, we often have project-specific anyway (for permissions, systemd tags, abstracting HW variants, ...), so having a custom rule for UBI links is not a large maintenance burden. :)

Okay.

> > (Besides: this udev topic is interesting and would allow to use volume names without this PR, but without udev it would still not be possible and I have space constrained systems where systemd + all the dependencies is no option.)
> 
> I suspect that the udev alternatives have similar features for creating links. Without udev (or an alternative) you're also missing the `/dev/disk/by-uuid/` and related links, so I'd be fine with requiring `rauc.slot=` in that case.

On at least one system we work with kernel config options `CONFIG_DEVTMPFS=y` and `CONFIG_DEVTMPFS_MOUNT=y` only. Quoting Kconfig: *“In very limited environments, it may provide a sufficient functional /dev without any further help.”* And yes, that's true.

Okay, I might explore what busybox' mdev can do in that case or I port code from this PR to a tiny standalone tool or shellscript creating symlinks for me.

> We should probably point to udev as a possibility in that docs section and give some examples on how to configure udev for UBI volumes in the FAQ.

👍🏼

jluebbe:
I've created https://github.com/rauc/rauc/pull/813 to hopefully point people in the right direction.

ejoerns:
@LeSpocky Do you see this sufficiently resolved with #813?

LeSpocky:
> @LeSpocky Do you see this sufficiently resolved with #813?

After reading _this_ PR again, I would think so, yes.

PR #783, opened 2021-10-22
Rework subcommand and option handling

ejoerns:
While it produces a bit more code, it makes subcommand-specific option handling more flexible than before as it replaces the struct by a sequence of individual g_option_* calls that would also allow future changes in argument grouping, etc.

The only 'function' change contained here is that subcommand help output changes from:

    rauc [OPTION?] install <BUNDLE>

to 

    rauc install [OPTION?] <BUNDLE>

by manipulating the program name according to the subcommand.

Note: For consistency, this bases on #779.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -331,7 +331,8 @@ static gboolean install_start(int argc, char **argv)
 
 	g_debug("install started");
 
-	context = g_option_context_new("install <BUNDLE>");
+	g_set_prgname("rauc install");
+	context = g_option_context_new("<BUNDLE>");
 	g_option_context_set_summary(context, "Install a bundle");
 	g_option_context_set_help_enabled(context, FALSE);
 	g_option_context_add_main_entries(context, entries, NULL);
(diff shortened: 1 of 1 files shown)


jluebbe:
I find the cost of the duplicated setup too high just for having the `[OPTION…]` at a different location. The options are explicitly grouped in the help output already:
```
Usage:
  rauc [OPTION…] resign <INBUNDLE> <OUTBUNDLE>

Resign an already signed bundle

Resign options:
  --no-verify                   disable bundle verification
  --no-check-time               don't check validity period of certificates against current time
  --signing-keyring=PEMFILE     verification keyring file

Application Options:
  -c, --conf=FILENAME           config file
  --cert=PEMFILE|PKCS11-URL     cert file or PKCS#11 URL
  --key=PEMFILE|PKCS11-URL      key file or PKCS#11 URL
  --keyring=PEMFILE             keyring file
  --intermediate=PEMFILE        intermediate CA file name
  --mount=PATH                  mount prefix
  -d, --debug                   enable debug output
  --version                     display version
  -h, --help                    
```

So it should be clear already which options can be used where.

PR #787, opened 2021-10-24
Rework PR 513: Skip using the active slot as a seed when installing bundles

Esco441-91:
### Summary 

This PR is addressing the changes made by @fhackenberger and the requested fixed PR https://github.com/rauc/rauc/pull/513. The original change in general is about adding the possibility to ignore target slot as seed when doing a casync extraction. I changed the logic from  being a commandline option to an option in conf file.

### The changes

I made the changes requested by @ejoerns in the PR discussions, in comment https://github.com/rauc/rauc/pull/513#issuecomment-570577754. Based on the comment, the changes I made were in short:

1. Add option to conf file ignore-slot-seed
2. Take the code by @fhackenberger and change ignore-slot-seed variable to be read from config instead of arguments
3. Add test for new conf file option is read correctly
4. Add test for casync being called accordingly based on new option.
5. Add documentation on the new option

### Test status

For some reason, the latest commit I made went through the tests without errors and without changes, even though before it always complained about an error in distcheck, on test/service.test:

`ERROR: test/service.test - Bail out! rauc:ERROR:../../test/service.c:102:on_installer_changed: assertion failed (percentage == cmp_percentage): (100 == 90)`

And twice or thrice I've gotten error on test/dm.test:

`ERROR: test/dm.test - Bail out! rauc:ERROR:test/dm.c:256:verity_hash_create: assertion failed (readable_sectors(dmfd) == dm_data->data_size - 1): (2 == 1)`

These errors also happened when I ran the CI tests with the master branch. Feel free to deny the PR based on these results. My estimation is that they are not because of my additions (due to seeing them in master), but I will not argue if You disagree.

All the other tests are fine.

### Notes

Note about the latter test. Doing the test required alot of setup to do just a few simple tests. I'm personally not particularly happy with the result, as the test could be made shorter if the structure of the code was different. I'm willing to accept the need for cancellation/rework based on the test code status, but I could use some directions on how to make the test neater, either by changing the test or some underlying code in src/update_handler.c.

Note about the documentation. I'm personally not an expert on casync seeding. I know the principle of how the seeding itself works, but was unable to write anything concrete about using or not using the option. Therefore, the documentation is more in the lines of "beating around the bushes" rather than telling why or why not to use the new option. If the documentation status is not fine, I could again use some pointers on how to make it better.

Special thanks to @teksturi for him pre-reviewing my changes before I made this PR.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -628,6 +628,10 @@ the currently booted slot group.
   After this is done it will start writing the data and fetch missing chunks
   via the network.
 
+  Seed slot usage can be disabled by using ignore-slot-seed -option in
+  the configuration. See :ref:`ignore-slot-seed <ignore-slot-seed>` for
+  more information
+
 .. _sec-variants:
 
 Handling Board Variants With a Single Bundle
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -242,6 +242,15 @@ For more information about using casync support of RAUC, refer to
   By default, the temporary directory is left unset by RAUC and casync uses its
   internal default value ``/var/tmp``.
 
+.. _ignore-slot-seed:
+
+``ignore-slot-seed``
+  Allows to disable using the current active slot as seed in casync installation.
+  Refer to --seed option in `casync author's blog post \
+  <http://0pointer.net/blog/casync-a-tool-for-distributing-file-system-images.html>`_ 
+  For more info on seeding and using or not using seeding.   
+  By default, this option is set to false.
+
 **[autoinstall] section**
 
 The auto-install feature allows to configure a path that will be checked upon


jluebbe:
@Esco441-91 I'm not sure I fully understand the use-case: The title is about ignoring the **active** slot for seeding, while the description refers to the **target** slot. Currently, we use the active slot as the seed, as this should be closest to the target contents.

If you disable seeding, casync would have to download everything? What is the benefit of using casync in this case?

In which case would you recommend using this config option?

Esco441-91:
Thank you @jluebbe for addressing this PR. I see now that the documentation and PR description indeed are missing the use-case for this, on top of the slot is being referred to in different terms. My original intention for these changes was just to contribute to this rauc in general, not that I personally would have a use-case. Just to help out close an old PR. I didn't fully take in the original use-case described in https://github.com/rauc/rauc/pull/513

I'll attempt to amend documentation and PR description inside few days by rewording the following summary:

"Using this option helps tar-based image updates have more control over the amount of indexing done by casync, decreasing the amount of (ram)disk space used during update. Must be combined with using .caexclude in image AND creating the casync store before updating."

Please let me know if there's you see a flaw in this use-case description.

---
The summary was made using the following information from issue https://github.com/rauc/rauc/issues/511 and PR https://github.com/rauc/rauc/pull/513

Original issue https://github.com/rauc/rauc/issues/511 was a request to reduce the amount of disk space when using tar images, as casync creates a local index, which can take alot of (ram)disk space. Original comparison from @fhackenberger:
```
root     29897  0.0  0.1 256164 10172 pts/0    Sl   18:36   0:00 rauc -d install /tmp/board-system-images-v3/update.casync.raucb
root     29922 85.9  0.1  48132 14444 ?        R    18:36  11:15 casync --log-level=debug --verbose extract --seed /run/rauc/rootfs.0 --store /tmp/board-system-images-v3/update.casync.castr --seed-output=no /run/rauc/bundle/rootfs.tar.gz.caidx /run/rauc/rootfs.1

ubuntu@sign-pc-18:~$ sudo du -xhcs /var/tmp/*.cased | sort -n     
3.3G    total
3.3G    /var/tmp/3a875a5021649923.cased
```
The original usage description from https://github.com/rauc/rauc/pull/513 by @fhackenberger: 

```
* Remote repository chunk store contains only chunks for most recent update
* Images have a .caexclude file included, which prevents /var/tmp/update.casync.castr from being chunked
* Device does the bind-mount & casync make --store /var/tmp/update.casync.castr /var/tmp/update.casync.caidx /mnt/rauc/ * step in the background once.
* Device does an rsync with --delete on remote chunks store / index file / raucb
* Device installs without using slot as seed rauc install --ignore-seed bundle.raucb
```

jluebbe:
I'm sorry for the long silence. By now I think the better way forward is #817, as that gives us a path to avoid the all the additional complexity we'd have with casync (managing .caexclude, generating a local store, rsyncing store contents). We'd prefer casync/desync to mostly just work, and I currently don't see a way how we'd get there with without desync.

PR #805, opened 2021-11-09
Fix coverity warning in install.c

ejoerns:
src/install: assert not expanding unset manifest filename

During manifest file parsing, we ensure that filename is set and allow an unset filename only for an install hook.
Thus having filename unset after checking this conditition is a programming error here. Check it with `g_assert_nonnull()`.

Fixes coverity warning:

> CID 1465767 (#1 of 1): Dereference after null check (FORWARD_NULL)
> 20. var_deref_model: Passing null pointer mfimage->filename to g_file_test, which dereferences it.

Fixes 8a9c9213 which added this check:

``` patch
+               /* skip source image checks if filename is not set (install hook) */
+               if (!mfimage->filename && mfimage->hooks.install)
+                       goto skip_filename_checks;
+
```

that lets coverity assume we explicitly pass here in case of `mfimage->filename` being null:

> 17. var_compare_op: Comparing mfimage->filename to null implies that mfimage->filename might be null.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -744,9 +744,16 @@ static gboolean pre_install_checks(gchar* bundledir, GList *install_images, GHas
 		RaucImage *mfimage = l->data;
 		RaucSlot *dest_slot = g_hash_table_lookup(target_group, mfimage->slotclass);
 
-		/* skip source image checks if filename is not set (install hook) */
-		if (!mfimage->filename && mfimage->hooks.install)
-			goto skip_filename_checks;
+		if (!mfimage->filename) {
+			/* having no filename is valid for install hook only */
+			if (mfimage->hooks.install)
+				goto skip_filename_checks;
+			else
+				/* Should not be reached as the pre-conditions for optional 'filename' are already
+				 * checked during manifest parsing in manifest.c: parse_image() */
+				g_assert_not_reached();
+		}
+
 
 		/* if image filename is relative, make it absolute */
 		if (!g_path_is_absolute(mfimage->filename)) {
--- a/README.rst
+++ b/README.rst
@@ -259,8 +259,8 @@ section for more details.
    :target: https://github.com/rauc/rauc/actions?query=workflow%3Atests
 .. |Codecov_branch| image:: https://codecov.io/gh/rauc/rauc/branch/master/graph/badge.svg
    :target: https://codecov.io/gh/rauc/rauc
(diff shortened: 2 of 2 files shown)


src/install.c
```diff
@@ -748,6 +748,8 @@ static gboolean pre_install_checks(gchar* bundledir, GList *install_images, GHas
 		if (!mfimage->filename && mfimage->hooks.install)
 			goto skip_filename_checks;
 
+		g_assert_nonnull(mfimage->filename);
```
a3f, line 751: That seems like a round about way to say:

```
if (!mfimage->filename)
   goto skip_filename_checks;

g_assert_nonnull(mfimage->hooks.install);

That looks cleaner IMO.

Comment (jluebbe) [reply src/install.c:751]: Or maybe just check that mfimage->hooks.install is set and raise a GError otherwise?

Comment (ejoerns) [reply src/install.c:751]: > That seems like a round about way to say:
> 
> ```
> if (!mfimage->filename)
>    goto skip_filename_checks;
> 
> skip_filename_checks:
>   g_assert_nonnull(mfimage->hooks.install);
> ```
> 
> That looks cleaner IMO.

A little cleaner maybe, but not sure if it better describes the intention. In my version, the label that states to skip the checks is used only exactly for that 'special' purpose (when no filename is set but a custom install hook used). In your approach, we jump there and then check the hook like a side note.

> Or maybe just check that mfimage->hooks.install is set and raise a GError otherwise?

At some point, we have to decide if we want GErrors for uncheckable code paths or not.  As I stated in the description of the commit, this cannot be reached by normal code flow as the config file parser already rejects an inappropriate configuration. Thus it is no user error we check here.
Despite I've also had a GError approach in the first round, I decided to only add an assert to satisfy the checker which only handles the function context.

Another option would be to say

``` c
if (!mfimage->filename) {
    if (mfimage->hooks.install)
        goto skip_filename_checks;
    else
        g_assert_not_reached();
}
```

jluebbe: Hmm, the last option doesn't look too bad. Is the checker clever enough to understand this?

We should also have a comment for the `g_assert_not_reached()` to point to the location where this precondition is checked.

ejoerns: I've updated the PR to match the last example and also added an inline note.
According to my private coverity run, it seems that the original issue is considered resolved now.

jluebbe (approved):

PR #813, opened 2021-11-18
docs/faq: add an entry on how to avoid problems with changing device numbers

jluebbe:
This seems to be a recurrent issue, so try to point users in the right direction.


LeSpocky (approved):

jluebbe:
Force-pushed to fix a typo.

ejoerns (approved):

Issue #748, opened 2021-06-29
Optional Desync Support


om26er:
Desync is fast and is actively maintained unlike casync. I have been playing with it for the last week and it seems that it's "concurrent" parameter on `desync extract` command is very fruitful especially when downloading chunks from S3.

Casync it seems downloads one chunk at a time and that results in very slow bundle install times.

It also doesn't do client side delta calculations, which means that the `caibx` of the seed block device would have to be provided separately, this in itself saves quite a bit of time depending on the size of the rootfs.

Desync can't be used as a drop in replace because
1. It doesn't support this parameter https://github.com/rauc/rauc/blob/a67c565cf7cac60b73ad4539c03d631aa41fb7b8/src/update_handler.c#L258
2. Doesn't directly take a block device with `--seed` parameter and have to do something like https://github.com/folbricht/desync/issues/178#issuecomment-869004096

What would be the "correct" approach to use desync with RAUC ?

om26er:
I think one "clean" approach would be to extend the `rauc install` command line to take an optional "casync-args" argument where we could define 
```
rauc install --casync-args="--seed=/home/om26er/imageA.caibx" https://something.com/casync-bundle.raucb
```

The could could then check if a "seed" is provided explicitly it should only use that.

om26er:
a `--casync-args=` option can also be very useful to provide desync parameters for concurrent downloads (`-n`)

jluebbe:
As the user should not need to know which image (A/B) is active or being installed to, specifying the location of the previously installed `.caibx` on the command line is not very useful.
RAUC has the necessary information during the initial installation, so it would be simplest to store the `.caibx` next to the central slot status on a shared storage. Then it could be used automatically for the next installation.

For configuring generial casync options, a `system.conf` option seems appropriate.

ejoerns:
I'd agree on what @jluebbe said as this matches my initial thoughts on this topic.

Maybe this would be the point where we should introduce some sort of generic 'rauc data dir' (instead of having only the very specific slot status file).

There are more things about to come in the future probably that need a distinct non-redundant storage location. Thus having a folder for this where we can put in all the management data for RAUC would be quite useful imho.

jluebbe:
Since https://github.com/rauc/rauc/pull/925, we now have a shared data directory where the `.caibx` for each slot could be stored (see https://github.com/rauc/rauc/pull/817#issuecomment-1158692257).

PR #817, opened 2021-12-14
Initial desync support

RyuzakiKK:
This PR partially addresses #748 and adds an initial optional support for Desync.

I updated the Debian base version of `test/Dockerfile` to the latest stable release because Desync needs at least Go 1.14 and the old-stable (buster) has Go 1.11 (with the 1.15 in buster-backports).
This change is a pre-requisite for Desync and can be merged separately, if preferred.

While Desync is mostly a drop-in replacement of Casync, some of its launch options are different.
The most notable change is that Desync expects a seed in the form of `.caibx`. This probably requires some additional thinking/discussion to decide what's the best way to handle that (where do we look for this `.caibx`? If it is not available, should we regenerate it on the fly?)

For this reason I left out the seed for now to try keeping this PR as simple as possible so that we would have an initial Desync support that could be use right away (and experiment with the additional `install-args`).

EDIT: with the latest commit revision I enabled the new Desync `DESYNC_ENABLE_PARSABLE_PROGRESS` environment variable so that Desync will print in output its progress percentage and estimated remaining time.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -303,6 +303,10 @@ For more information about using the casync support of RAUC, refer to
   By default, the temporary directory is left unset by RAUC and casync uses its
   internal default value ``/var/tmp``.
 
+``use-desync=<true/false>``
+  If this boolean value is set to ``true``, RAUC will use desync instead of
+  casync. Desync support is still experimental, use with caution.
+
 **[autoinstall] section**
 
 The auto-install feature allows to configure a path that will be checked upon
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -283,6 +283,11 @@ The ``casync`` section contains casync-related settings.
 For more information about using the casync support of RAUC, refer to
 :ref:`casync-support`.
 
+``install-args``
+  Allows to specify additional arguments that will be passed to casync when
+  installing an update. For example it can be used to include additional
+  seeds or stores.
+
 ``storepath``
   Allows to set the path to use as chunk store path for casync to a fixed one.
   This is useful if your chunk store is on a dedicated server and will be the
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -50,6 +50,7 @@ typedef struct {
 	gchar *mount_prefix;
 	gchar *store_path;
 	gchar *tmp_path;
+	gchar *casync_install_args;
 	gboolean activate_installed;
 	gchar *statusfile_path;
 	gchar *keyring_path;
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -51,6 +51,7 @@ typedef struct {
 	gchar *store_path;
 	gchar *tmp_path;
 	gchar *casync_install_args;
+	gboolean use_desync;
 	gboolean activate_installed;
 	gchar *statusfile_path;
 	gchar *keyring_path;
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -212,6 +212,8 @@ out:
 	return res;
 }
 
+static gboolean casync_make_blob(const gchar *idxpath, const gchar *contentpath, const gchar *store, GError **error);
+
 static gboolean casync_make_arch(const gchar *idxpath, const gchar *contentpath, const gchar *store, GError **error)
 {
 	g_autoptr(GSubprocess) sproc = NULL;
@@ -224,6 +226,12 @@ static gboolean casync_make_arch(const gchar *idxpath, const gchar *contentpath,
 	g_return_val_if_fail(idxpath != NULL, FALSE);
 	g_return_val_if_fail(contentpath != NULL, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (r_context()->config->use_desync) {
+		/* Desync is able to handle tar and catar archives directly, there is
+		 * no need to manually extract them. */
+		return casync_make_blob(idxpath, contentpath, store, error);
+	}
 
 	tmpdir = g_dir_make_tmp("arch-XXXXXX", &ierror);
 	if (tmpdir == NULL) {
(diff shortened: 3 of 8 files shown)


RyuzakiKK:
It seems like codecov reports a low code coverage because the tests are currently launched with the default docker RAUC image that doesn't include Desync.

ejoerns:
@RyuzakiKK Thanks for your contribution!
I would tend to say that you should split the Dockerfile adaptions and rst changes into a separate PR as these might be a good idea anyway and desync support probably requires more strategic discussions (and testing as you already noted yourself) and might conflict with other casync-related changes not merged yet.

RyuzakiKK:
> @RyuzakiKK Thanks for your contribution! I would tend to say that you should split the Dockerfile adaptions and rst changes into a separate PR as these might be a good idea anyway and desync support probably requires more strategic discussions (and testing as you already noted yourself) and might conflict with other casync-related changes not merged yet.

Yes, make sense. I opened #821 with just the Docker upgrade and documentation fixes.

RyuzakiKK:
@ejoerns To speed up and make it easier to review, do you think it would be worth to split the commit "Add casync install-args configuration key" (07e2222) to a separate PR?

Again this is per-requisite for Desync, but it's something that could be worth to have even for Casync.

RyuzakiKK:
Thanks! I should be able to rebase and fix the CI tests later this week.

ejoerns:
As this skips desync checks, having a green test suite here is good but not sufficient for desync support, yet.

Could you split out the Dockerfile adaptions that enable desync support into a separate PR, too?

This would then allow building a docker image that is fully capable of running the desync tests.

RyuzakiKK:
> As this skips desync checks, having a green test suite here is good but not sufficient for desync support, yet.
> 
> Could you split out the Dockerfile adaptions that enable desync support into a separate PR, too?
> 
> This would then allow building a docker image that is fully capable of running the desync tests.

Sure, done in https://github.com/rauc/rauc/pull/883. I also rebased this PR to reflect that change.

jluebbe:
Force-pushed to solve merge conflicts and adjust commit titles to our convention.

jluebbe:
Overall, this looks mostly ready. I only have a few questions on how we could proceed with desync support in the future.

> While Desync is mostly a drop-in replacement of Casync, some of its launch options are different. The most notable change is that Desync expects a seed in the form of `.caibx`. This probably requires some additional thinking/discussion to decide what's the best way to handle that (where do we look for this `.caibx`? If it is not available, should we regenerate it on the fly?)

In the [desync examples](https://github.com/folbricht/desync#examples) I found:
> Note, when providing a seed like --seed <file>.ext.caibx, it is assumed that <file>.ext is available next to the index file, and matches the index.

That seems cumbersome to use with block devices. One could create a pair of symlinks (one to the device, one to the stored index file), but that feels like a hack. Perhaps desync would accept a `--seed-data` (or similar) option to point it to the corresponding block device?

Also, how much space is used for the on-target chunk store? Can it be disabled or is it possible to limit it's size (i.e. a LRU cache)?

RyuzakiKK:
> Overall, this looks mostly ready. I only have a few questions on how we could proceed with desync support in the future.
>
> [...]
> 
> > Note, when providing a seed like --seed .ext.caibx, it is assumed that .ext is available next to the index file, and matches the index.
> 
> That seems cumbersome to use with block devices. One could create a pair of symlinks (one to the device, one to the stored index file), but that feels like a hack.

In our usecase we are doing pretty much that. We make the seed index in a particular directory and, next to it, we create a symlink to the device partition (the seed data).

> Perhaps desync would accept a `--seed-data` (or similar) option to point it to the corresponding block device?

I guess the reason it doesn't have such option is because `--seed` can be repeated multiple times to include more than one seed. Additionally you can also use `--seed-dir`, where Desync will automatically pick every `*.caibx` it finds as seeds.

By the way, I agree that creating symlinks is not the cleanest solution.
Thinking out loud, probably we could investigate the possibility of adding both the seed index and the seed data in the same argument? E.g. something along the line of `--seed-and-data /path/to/seed.caibx:/path/to/seed_data`?

> Also, how much space is used for the on-target chunk store? Can it be disabled or is it possible to limit it's size (i.e. a LRU cache)?

I'm not sure I understood what you mean by "on-target chunk store".
Is it the [cache option of Desync](https://github.com/folbricht/desync#caching)?

jluebbe:
I've moved this to the next release, as we want to have the streaming and encryption features in a release.

We should get this in soon after v1.7 and don't wait with v1.8 for too long. :)

jluebbe:
> > Overall, this looks mostly ready. I only have a few questions on how we could proceed with desync support in the future.
> > [...]
> > > Note, when providing a seed like --seed .ext.caibx, it is assumed that .ext is available next to the index file, and matches the index.
> > 
> > 
> > That seems cumbersome to use with block devices. One could create a pair of symlinks (one to the device, one to the stored index file), but that feels like a hack.
> 
> In our usecase we are doing pretty much that. We make the seed index in a particular directory and, next to it, we create a symlink to the device partition (the seed data).

OK. As long as there no explicit support to pass the pair of device+index to desync, I'd be fine with the approach of creating a them as symlinks in a temporary directory. While not elegant, it's very simple.

Please take a look at https://github.com/rauc/rauc/pull/925. You could use the slot's `data_directory` to keep the `.caibx` file after installation, so the next update can use it as a seed.

> > Perhaps desync would accept a `--seed-data` (or similar) option to point it to the corresponding block device?
> 
> I guess the reason it doesn't have such option is because `--seed` can be repeated multiple times to include more than one seed. Additionally you can also use `--seed-dir`, where Desync will automatically pick every `*.caibx` it finds as seeds.
> 
> By the way, I agree that creating symlinks is not the cleanest solution. Thinking out loud, probably we could investigate the possibility of adding both the seed index and the seed data in the same argument? E.g. something along the line of `--seed-and-data /path/to/seed.caibx:/path/to/seed_data`?

If desync upstream adds support for something like that (or ordered command line options?), RAUC could switch to that later.

> > Also, how much space is used for the on-target chunk store? Can it be disabled or is it possible to limit it's size (i.e. a LRU cache)?
> 
> I'm not sure I understood what you mean by "on-target chunk store". Is it the [cache option of Desync](https://github.com/folbricht/desync#caching)?

Yes. In casync, that cache (local chunk store) is not optional and can become quite large, as it contains all downloaded chunks.

Just to make sure I understand the current PR correctly: Without support for seeds, it cannot reuse the active slot or matching data in the target slot?

jluebbe:
Ah, the symlink approach was explicitly suggested by desync's maintainer: https://github.com/folbricht/desync/issues/178#issuecomment-869004096
So we should just use that.

jluebbe (approved):

RyuzakiKK:
> Please take a look at https://github.com/rauc/rauc/pull/925. You could use the slot's data_directory to keep the .caibx file after installation, so the next update can use it as a seed.

Having a place where to automatically store the `.cabix` after the update would definitely be handy. In our use case we are currently doing it as a manual step in the post-install hook.

> If desync upstream adds support for something like that (or ordered command line options?), RAUC could switch to that later.

I agree, sounds good.

> Just to make sure I understand the current PR correctly: Without support for seeds, it cannot reuse the active slot or matching data in the target slot?

Correct. It needs a seed index in order to reuse the active slot chunks (or whatever other partition/file you want as a seed).

Issue #819, opened 2022-01-04
CWE-377: Insecure tempfile creation for remote downloads


jktjkt:
I'm on rauc 1.4, but the same code is in current `master`. It appears that all remote files always end up in `$TMP/_download.raucb`, which is a problem because it tends to be a path that's often shared and writable by multiple users. A good fix would be to use unpredictable path name below the temp directory, e.g. via `mkstemp`. A quick-and-dirty workaround would be configuring the systemd unit file so that it has a private `/tmp`.

jktjkt:
I have not checked the code whether it's feasible for an attacker to swap the file for something malicious (perhaps to bypass signature verification?). What I had in mind was more like an attacker leaving a symlink in place before the update process starts, and then RAUC (running as root) accidentally overwriting some important system file.

jluebbe:
> I have not checked the code whether it's feasible for an attacker to swap the file for something malicious (perhaps to bypass signature verification?).

I meant that bypassing signature verification should be mitigated in 1.5.

> What I had in mind was more like an attacker leaving a symlink in place before the update process starts, and then RAUC (running as root) accidentally overwriting some important system file.

Ah, yes. That looks like it would be possible. I'll make a PR.

PR #820, opened 2022-01-05
src/bundle: download temporary bundle to a separate temp directory

jluebbe:
This protects against the issue described in #819.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -1411,8 +1411,15 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 	bundlescheme = g_uri_parse_scheme(bundlename);
 	if (is_remote_scheme(bundlescheme)) {
 #if ENABLE_NETWORK
+		g_autofree gchar *tmpdir = g_dir_make_tmp("rauc-XXXXXX", &ierror);
+		if (tmpdir == NULL) {
+			g_propagate_prefixed_error(error, ierror, "Failed to create tmp dir: ");
+			res = FALSE;
+			goto out;
+		}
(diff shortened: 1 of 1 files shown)


src/bundle.c
```diff
@@ -1412,7 +1412,7 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, gboolean ver
 	if (is_remote_scheme(bundlescheme)) {
 #if ENABLE_NETWORK
 		ibundle->origpath = g_strdup(bundlename);
-		ibundle->path = g_build_filename(g_get_tmp_dir(), "_download.raucb", NULL);
+		ibundle->path = g_build_filename(g_dir_make_tmp("rauc-XXXXXX", NULL), "download.raucb", NULL);
```
a3f, line 1415: Leaks.

a3f (changes requested):

src/bundle.c
a3f: And while at it, you may want to check for `NULL` and return, even if the error is unlikely.

a3f (dismissed):
LGTM. Nitpick: Code always sets `res = FALSE` in error case, even if res hasn't been changed from initial value. (See `if (verify` branch above). It might be good to follow this scheme. Makes code more robust against future refactoring.

a3f (dismissed):

ejoerns:
With this change, the tmdir is created but (unlike the bundle itself) not removed after installation/error.

ejoerns (approved):
Code changes look fine for me now.

ejoerns (approved):

PR #821, opened 2022-01-06
Update the docker base image to the latest Debian stable

RyuzakiKK:
- docs/reference: Remove glossary directive
    Since at least Sphinx 3.4.3, glossary entries are referenceable with the
`term` role.
    For this reason Sphinx doesn't allow to repeat two times the same
`term`.
- test/Dockerfile: Update the base image to the latest Debian stable

This is a per-requisite for the proposed Desync support in #817.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -687,51 +687,49 @@ Custom Handlers (Interface)
 Interaction between RAUC and custom handler shell scripts is done using shell
 variables.
 
-.. glossary::
+``RAUC_SYSTEM_CONFIG``
+  Path to the system configuration file (default path is ``/etc/rauc/system.conf``)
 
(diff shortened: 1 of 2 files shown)


ejoerns:
Made a test build that looks fine for both container generation:

https://github.com/ejoerns/rauc/actions/runs/1695424199

and running test suite from this container:

https://github.com/ejoerns/rauc/actions/runs/1695444289

As expected, also fixes the Sphinx build warning I had before (locally).

ejoerns (approved):

jluebbe (approved):

jluebbe:
I'm not sure if Coverity Scan will still work with the new gcc (10), but we'll see.

PR #823, opened 2022-01-16
docs/advanced: fix script URL

UVV-gh:
Signed-off-by: Vyacheslav Yurkov <uvv.mail@gmail.com>


UVV-gh:
It seems `openssl-ca.sh` script is present here https://github.com/rauc/meta-rauc/blob/master/scripts/openssl-ca.sh . The question is which one should be referenced in the documentation, and should they be different?

ejoerns (approved):
Good catch!

The script name was actually changed in c10e94848c464cc58eef67449fbe452affe27795 but the docs not adjusted.

For me, the entire paragraph is a bit misleading in various aspects, but that is probably something to fix in a follow-up PR :smirk:

ejoerns:
> It seems `openssl-ca.sh` script is present here https://github.com/rauc/meta-rauc/blob/master/scripts/openssl-ca.sh . The question is which one should be referenced in the documentation, and should they be different?

As RAUC itself is independent of its Yocto integration, we should not mention the `meta-rauc` script here.

The purpose is also a bit different but I am not sure if they necessarily have to be different.

PR #770, opened 2021-09-10
src: introduce --trust-environment option for 'rauc convert'

ejoerns:
In some cases it might not be possible or wanted that RAUC (used as a host tool) performs its standard bundle access checks.

For example when running 'rauc convert' from within a (docker) container, RAUC might not be able to get the underlying device of the bundle's storage location. This way it cannot determine if the mount point is owend by root and will abort with

> initial check_bundle_access failed with: unable to find mounted device for bundle

While --no-verify potentially already works around this, it is a too big hammer for most cases where you actually want to benefit from the usual signature verificaton of the input bundle and just get stopped by the exclusiv access check.

Not the --trust-environment option allows to tell RAUC that we assume we have our environmet we run RAUC in under control and it can skip the exclusive access checks.

Fixes #679, #688

Note: bases on #769

PR #824, opened 2022-01-19
src/main: add disable-security-measures switch

leiflm:
Add an option to disable security measures for the `extract` command.

Signed-off-by: Leif Middelschulte <Leif.Middelschulte@klsmartin.com>

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -29,6 +29,7 @@ int r_exit_status = 0;
 
 gboolean install_ignore_compatible, install_progressbar = FALSE;
 gboolean verification_disabled = FALSE;
+gboolean disable_security_checks = FALSE;
 gboolean info_dumpcert = FALSE;
 gboolean status_detailed = FALSE;
 gchar *output_format = NULL;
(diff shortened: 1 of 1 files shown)


jluebbe:
You're only using this on your build machines, not on the target, right? :)

leiflm:
> You're only using this on your build machines, not on the target, right? :)

I'm just using it locally, to modify prebuilt bundles. I might, as well, use some squash extraction tool though.

Would a more verbose [switch](https://github.com/Enlightenment/enlightenment/blob/8edf2fc0a388090da74db683a690e0aed8a33eb6/src/bin/e_start_main.c#L581) make it better? ;-)

src/main.c
```diff
@@ -585,7 +586,7 @@ static gboolean extract_start(int argc, char **argv)
 	g_debug("input bundle: %s", argv[2]);
 	g_debug("output dir: %s", argv[3]);
 
-	if (!check_bundle(argv[2], &bundle, TRUE, &ierror)) {
+	if (!check_bundle(argv[2], &bundle, !disable_security_checks, &ierror)) {
```
jluebbe, line 589: This is what `--no-verify` does in other commands: disable signature verification. Why do you use a different name for extract? Or is the goal to only disable the concurrent access checks (as @ejoerns does in https://github.com/rauc/rauc/pull/770)?

jluebbe (changes requested):

src/main.c
leiflm: > This is what --no-verify does in other commands: disable signature verification.

That's true. But it has further implications than skipping the signature verification. Hence a more apt name.

> Or is the goal to only disable the concurrent access checks (as @ejoerns does in #770)?

Exactly. The primary intention is to disable the concurrent access checks. I'd say that #770 seems to be the way to go. And once it's merged, it should be quite straight forward to be supported by the `extract` command to.

leiflm:
This will be revisited, once #770 is merged.

PR #825, opened 2022-01-20
src/update_handler: add support for jffs2 tar archives

h-assmann:
This patch introduces support for tar archives for jffs2 flash storage.

For jffs2 we have to use the MTD chardev in order to erase the flash
storage. The subsequent mounting call has the peculiarity of needing a
different device parameter than a device path, meaning that no path must
be passed. We take care of that with specific handling in mount.c

Signed-off-by: Holger Assmann <h.assmann@pengutronix.de>


src/mount.c
```diff
@@ -71,7 +71,17 @@ gboolean r_mount_full(const gchar *source, const gchar *mountpoint, const gchar*
...
 	}
-	g_ptr_array_add(args, g_strdup(source));
+
+	/*
+	 * jffs2 mount must be called without /dev/ path. As we have already
+	 * checked for the device name having a (/dev/-)path we can go with
+	 * get_basename here.
+	 */
+	if (g_str_equal(type, "jffs2")) {
+		g_ptr_array_add(args, g_path_get_basename(source));
```
a3f, line 81: Did you test this with busybox `mount(1)` or the standard one from `util-linux`? RAUC docs claim that [busybox `mount` is good enough](https://rauc.readthedocs.io/en/latest/integration.html#required-target-tools), but apparently, at least in the past, the format you require here wasn't compatible with busybox. See `Do I need to use mtdblock to mount my JFFS2 filesystem?` on the [Linux MTD homepage](http://www.linux-mtd.infradead.org/faq/jffs2.html).

My preference would be to drop the `/dev` prefix check and just pass whatever the user has configured to mount.

src/config_file.c
```diff
@@ -546,6 +546,17 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 				goto free;
 			}
 
+			/* check if device has an appropriate path*/
```
a3f, line 549: nitpick while we are it: trailing space before `*/`.

src/mount.c
jluebbe: The whole point of this patch is that we need the chardev path configured to erase the the device. So using `/dev/mtdblockX` in RAUC's system.conf is wrong.
Does the current busybox work correctly?

a3f: > The whole point of this patch is that we need the chardev path configured to erase the the device.

Huh, should do more reading and less writing ^^.

> So using /dev/mtdblockX in RAUC's system.conf is wrong.

I see. I hope the error message when passing a mtdblock isn't too cryptic.

> Does the current busybox work correctly?

Just tried it out: `/bin/busybox mount -t jffs2 mtd0 /mnt` works for me on busybox `v1.3.1`. So this point is moot.

ejoerns (approved):
Haven't tested on HW but the code looks fine for me, thanks!

jluebbe (approved):

PR #836, opened 2022-02-01
Fixes for context setup and error handling

ejoerns:
* always back fallback to `/etc/rauc/system.conf` when being available (but error if its broken)
* graceful context error printout for all rauc commands

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -232,8 +232,17 @@ gboolean r_context_configure(GError **error)
 			return FALSE;
 		}
 	} else {
-		/* This is a hack as we cannot get rid of config easily */
-		default_config(&context->config);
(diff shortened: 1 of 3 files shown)


tretter:
I tested `rauc info` without the `--keyring` argument, and it works again as before.

ejoerns:
As suggested by @jluebbe, I've 'slightly' reduced code code duplication by simply moving the `r_context_configure()` call before the call of the individual command handler.

ejoerns:
Did sound like a good plan, but call to `r_context_configure()` must actually happen *after* call to `r_context_conf()->configpath`. But the latter is command-specific.

ejoerns:
Deprecated by #840.

PR #840, opened 2022-02-02
explicitly handle config file requirements per command

jluebbe:
This is an alternative to #836 and avoids the code duplication in https://github.com/rauc/rauc/pull/836/commits/e867ce96c5761d3e86cc139c2456c8d40aa3afc2.

Instead, it adds a config mode enum to the context. This allows the context setup code to choose the appropriate method to load the system configuration.

Commands which are intended to be used on the target (`install`, `status`, `write-slot`, `service` and `mount`) need to know the intended security configuration and will now require a config file. If none is given on the command line, they will use ``/etc/rauc/system.conf``.

Commands which are used during development (`bundle`, `resign`, `convert`, `replace-sig`, `extract-sig`) are mainly configured via the command line. If a config is given using ``--conf=``, it's still used, though.

The remaining commands `extract` and `info` will automatically load `/etc/rauc/system.conf` if it exists, otherwise they need to be configured via the commandline.

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -16,12 +16,19 @@ typedef struct {
 	RaucBundle *mounted_bundle;
 } RContextInstallationInfo;
 
+typedef enum {
+	R_CONTEXT_CONFIG_MODE_NONE, /* use default config values */
+	R_CONTEXT_CONFIG_MODE_AUTO, /* load config file if it exists */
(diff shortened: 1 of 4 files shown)


ejoerns (approved):
Much cleaner solution :+1:

PR #846, opened 2022-02-08
src/install: ignore slots without bootname for determining booted slot

ejoerns:
If we attempt to compare against slots that do not have a bootname set, these must fall back to device comparison.
If the device is not available (for any reason), this will print a bunch of

> Failed to resolve realpath for '/dummy/path'

warnings.


ejoerns:
Noted that this was explicitly disabled by me in https://github.com/rauc/rauc/commit/bf2a10baf3b45a35d691deb05438a5f7d6ae9aa3.

Thus this has to be rethought.

PR #865, opened 2022-02-24
docs/integration: mention CONFIG_CRYPTO_SHA256 in the format migration section

jluebbe:
It seems that some kernel configurations don't have this enabled already, so
mention it here and avoid surprises.


ejoerns (approved):
This is why we (you) have added this in 3566da3c9c4502a1977926642d9fd367cbdf57fd already, but this section seems to still lack this, yes.

PR #869, opened 2022-03-02
Bundle Check Rework and Signature Fixes

ejoerns:
As preparation for encrypted bundles, replace the check for the Bundle/SuqashFS marker in payload data.

Also, fix error code and memory handling in surrounding code.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -372,40 +372,6 @@ static gboolean input_stream_read_uint64_all(GInputStream *stream,
 	return res;
 }
 
-/* Attempts to read and verify the squashfs magic to verify having a valid bundle */
-static gboolean input_stream_check_bundle_identifier(GInputStream *stream, GError **error)
-{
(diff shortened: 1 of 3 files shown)


src/bundle.c
```diff
@@ -1389,6 +1346,12 @@ static gboolean open_local_bundle(RaucBundle *bundle, GError **error)
 		goto out;
 	}
 
+	res = data_is_parsable_cms(bundle->sigdata);
```
jluebbe, line 1349: This is a duplicate of the call at the end of `open_remote_bundle()`, so we should move it out of the local/remote if/else `check_bundle()`. At that point it would be directly before the call to `cms_is_detached()` which also has to parse the CMS. So we could simplify this by adding better error reporting for invalid CMS data in `cms_is_detached()`.

jluebbe (changes requested):

jluebbe (approved):

PR #871, opened 2022-03-02
Fix Some Memory Leaks

ejoerns:
This is a collection of memory leak fixes.
Some were discovered by Coverity, some were discovered during work on other topics.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -107,7 +107,7 @@ static gboolean parse_manifest(GKeyFile *key_file, RaucManifest **manifest, GErr
 	RaucManifest *raucm = g_new0(RaucManifest, 1);
 	gboolean res = FALSE;
 	g_autofree gchar *tmp = NULL;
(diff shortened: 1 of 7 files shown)


src/manifest.c
```diff
@@ -114,6 +114,8 @@ static gboolean parse_manifest(GKeyFile *key_file, RaucManifest **manifest, GErr
 
 	g_assert_null(*manifest);
 
+	raucm = g_new0(RaucManifest, 1);
```
jluebbe, line 118: Why do we need to move the initialization here?

jluebbe (changes requested):
Looks good otherwise.

src/manifest.c
ejoerns: Hm, does not seem to be required. Maybe just followed my instinct to initialize autofree vars to NULL first. Fixed.

jluebbe (approved):

PR #874, opened 2022-03-04
(Re)Fix two Coverity-reported defects

ejoerns:
* Fixes #871 
* Fixes 59653ab7

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -7,7 +7,7 @@ void r_stats_init(struct RaucStats *stats)
 	memset(stats, 0, sizeof(*stats));
 
 	stats->min = G_MAXDOUBLE;
-	stats->min = G_MINDOUBLE;
+	stats->max = G_MINDOUBLE;
 }
 
 void r_stats_add(struct RaucStats *stats, gdouble value)
--- a/src/utils.c
+++ b/src/utils.c
@@ -155,7 +155,7 @@ gchar *resolve_path(const gchar *basefile, gchar *path)
 gboolean check_remaining_groups(GKeyFile *key_file, GError **error)
 {
 	gsize rem_num_groups;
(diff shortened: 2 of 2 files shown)


src/stats.c
```diff
@@ -6,7 +6,7 @@ void r_stats_init(struct RaucStats *stats)
 
 	memset(stats, 0, sizeof(*stats));
 
-	stats->min = G_MAXDOUBLE;
+	stats->max = G_MAXDOUBLE;
 	stats->min = G_MINDOUBLE;
```
jluebbe, line 10: This must be the other way around. If `max` is already `G_MAXDOUBLE`, it can't change anymore.

jluebbe (changes requested):

src/stats.c
ejoerns: Indeed, now both are broken. Will fix.

jluebbe (approved):

src/stats.c
```diff
@@ -7,7 +7,7 @@ void r_stats_init(struct RaucStats *stats)
 	memset(stats, 0, sizeof(*stats));
 
 	stats->min = G_MAXDOUBLE;
-	stats->min = G_MINDOUBLE;
+	stats->max = G_MINDOUBLE;
```
a3f, line 10: Is that correct? at the very least a comment explaining why min = MAX and max = MIN might be in order.

a3f: Just saw @jluebbe comment above. Still some comment in the code would be nice.

jluebbe: These are initial values for incremental comparison. If they were the reversed, they could never change.

ejoerns: There is hardly any documentation of these functions at all. But iirc I did some for me during the review.
If I find it, I'll submit :smirk:

PR #875, opened 2022-03-04
src/main: limit --cert and --key arguments to signing commands

ejoerns:
It is not necessary (but instead misleading) to have the `--cert` and `--key` arguments as global options because they apply to only a very few commands.

Introduce a purpose-driven (rather than subcommand-driven) GOptionEntry for this.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -35,6 +35,8 @@ gboolean info_dumpcert = FALSE;
 gboolean info_dumprecipients = FALSE;
 gboolean status_detailed = FALSE;
 gchar *output_format = NULL;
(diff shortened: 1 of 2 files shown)


jluebbe:
The main problem with this change is that it breaks compatibility with usage of `--key=` or `--cert=` before the subcommand.
I don't see a way to change this without causing confusion.

ejoerns:
rebased onto current master

ejoerns:
Updated to keep the old options as invisible ones (for backwards compatibility) and using the new ones to expose them in info output.

jluebbe (approved):

PR #882, opened 2022-03-17
add support for a list of optional incremental methods

jluebbe:
These methods will add extra information to the bundle, allowing RAUC to
access only the parts of an image which are not yet available locally.
Together with streaming, this reduces the amount of downloaded data.

As the full image is still available in the bundle, older RAUC versions can
ignore unsupported incremental methods and we don't need to check the list for
explicit values. Currently, no methods are implemented.

Note that because we reject any manifest with unknown keys, bundles
which use incremental methods will be rejected by versions up to v1.6. By
integrating this change early (without actually implementing any methods), we
can add new methods that will be ignored by older versions. This way, users
can generate bundles with incremental methods without compatibility problems as
soon as a version with this minimal change is deployed on all devices.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -884,6 +884,13 @@ static gchar *info_formatter_shell(RaucManifest *manifest)
 		g_free(temp_string);
 
 		g_ptr_array_unref(hooks);
+
+		if (img->incremental) {
+			temp_string = g_strjoinv(" ", (gchar**) img->incremental);
+			formatter_shell_append_n(text, "RAUC_IMAGE_INCREMENTAL", cnt, temp_string);
+			g_free(temp_string);
+		}
+
 		cnt++;
 	}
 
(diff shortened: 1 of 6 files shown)


ejoerns:
I fear that this needs a bit more context about what 'incremental methods' are. Both in commit messages and documentation.

src/main.c
```diff
@@ -1031,6 +1044,14 @@ static gchar* info_formatter_json_base(RaucManifest *manifest, gboolean pretty)
 			json_builder_add_string_value(builder, "post-install");
 		}
 		json_builder_end_array(builder);
+		json_builder_set_member_name(builder, "incremental");
+		json_builder_begin_array(builder);
+		if (img->incremental) {
+			for (gchar *m = img->incremental[0]; m != NULL; m++) {
+				json_builder_add_string_value(builder, m);
```
ejoerns, line 1051: To not dump our memory, this should be

```suggestion
			for (gchar **m = img->incremental; *m != NULL; m++) {
				json_builder_add_string_value(builder, *m);
```

Also, if this key is compatibility-related, it is maybe worth adding a test?

ejoerns (changes requested):
Commit message and hint in reference.rst are now fine for me, thx!

Please also remove the 'fixup' line from the second commit's message.

And minor typo in the first commit: s/allows use to/allows us to/

src/main.c
jluebbe: Fixed.

jluebbe:
While adding the tests, I found some issues in `test/install.c` which I had to fix first.

jluebbe:
Updated to fix uncrustify error.

ejoerns (approved):

PR #883, opened 2022-03-17
test/Dockerfile: Install Desync

RyuzakiKK:
Desync is an alternative casync implementation.

When adding Desync support in rauc, if we already have Desync installed
in the image used for the automated tests, it will be easier to validate
the proposed changes.

---

This is a pre-requisite for https://github.com/rauc/rauc/pull/817

/cc @ejoerns

### DIFF
--- a/test/Dockerfile
+++ b/test/Dockerfile
@@ -36,7 +36,8 @@ RUN apt-get update && apt-get install -y \
   mtd-utils \
   python3-aiohttp \
   nginx-light \
-  fdisk
+  fdisk \
+  golang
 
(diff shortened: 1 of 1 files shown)


ejoerns (approved):
Tested that it builds in the respective GH Action.

Thus fine from my side.
I have no idea if this is the preferred way for building/installing go applications, but at least one that works.

Note that with building against desync master (instead of fixed release) we have another moving target which I am not sure if this is good or bad.

jluebbe (approved):

PR #890, opened 2022-03-25
Make Error Message Handling OpenSSL 3.0-Compatible

ejoerns:
This replaces `ERR_get_error_line_data()` by `ERR_get_error()`

`ERR_get_error_line_data()` was deprecated with OpenSSL 3.0.

We have never used line data anyway, thus we can use `ERR_get_error()` instead and will be compatible with all supported OpenSSL versions.

This also significantly simplifies error handling code.

:bulb: Note that oe-core master already uses OpenSSL 3.0
http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-connectivity/openssl

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -17,6 +17,17 @@ GQuark r_signature_error_quark(void)
 	return g_quark_from_static_string("r_signature_error_quark");
 }
 
+static const gchar *get_openssl_err_string(void)
+{
+	unsigned long err;
+	const gchar *data;
+	int errflags;
+
+	err = ERR_get_error_line_data(NULL, NULL, &data, &errflags);
+
+	return (errflags & ERR_TXT_STRING) ? data : ERR_error_string(err, NULL);
+}
+
 /* return 0 for error, 1 for success */
 static int check_purpose_code_sign(const X509_PURPOSE *xp, const X509 *const_x, int ca)
 {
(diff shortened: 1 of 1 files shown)


src/signature.c
```diff
@@ -62,17 +62,11 @@ gboolean signature_init(GError **error)
...
-		const gchar *data;
-		int flags;
-
-		err = ERR_get_error_line_data(NULL, NULL, &data, &flags);
 		g_set_error(
 				error,
 				R_SIGNATURE_ERROR,
 				R_SIGNATURE_ERROR_CRYPTOINIT_FAILED,
-				"Failed to initialize OpenSSL crypto: %s",
-				(flags & ERR_TXT_STRING) ? data : ERR_error_string(err, NULL));
```
jluebbe, line 75: In some cases, data can contain relevant information. We should add a helper to use as `g_autofree gchar *err_str = r_get_openssl_error_string();`.

jluebbe (changes requested):

src/signature.c
```diff
@@ -17,6 +18,21 @@ GQuark r_signature_error_quark(void)
 	return g_quark_from_static_string("r_signature_error_quark");
 }
 
+static const gchar *get_openssl_err_string(void)
+{
+	unsigned long err;
+	const gchar *data;
+	int errflags;
```
jluebbe, line 25:
```suggestion
	const gchar *data = NULL;
	int errflags = 0;
```

src/signature.c
```diff
@@ -7,6 +7,7 @@
 #include <openssl/crypto.h>
 #include <openssl/engine.h>
 #include <openssl/x509.h>
+#include <openssl/macros.h>
```
jluebbe, line 10: This doesn't exist in OpenSSL 1.1.1.

src/signature.c
```diff
@@ -17,6 +18,21 @@ GQuark r_signature_error_quark(void)
 	return g_quark_from_static_string("r_signature_error_quark");
 }
 
+static const gchar *get_openssl_err_string(void)
+{
+	unsigned long err;
+	const gchar *data;
+	int errflags;
+
+#if OPENSSL_API_LEVEL < 30000
```
jluebbe, line 27: This doesn't exist in OpenSSL 1.1.1 either.

jluebbe (changes requested):

src/signature.c
ejoerns: Yes, I actually see this in the tests but did not have the time to fix it yesterday.

ejoerns: Ok, had to rediscover that nothing but ``OPENSSL_VERSION_NUMBER`` existed before 3.0...

jluebbe (approved):

PR #895, opened 2022-04-07
add build test on ubuntu 21.10

jluebbe:
This uses glib 2.68, which will allow us to test the glib version checks.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -116,8 +116,8 @@ jobs:
     strategy:
       matrix:
         release:
-        - 18.04
-        - 20.04
+        - "18.04"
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -141,7 +142,7 @@ jobs:
 
     - name: Build
       run: |
-        docker exec -i stable make
+        docker exec -i stable make TESTS= check
         #docker exec -i stable make check
```
ejoerns, line 146: I'd replace this with

```suggestion
        docker exec -i stable make check TESTS=
```

to make it cleaner and more obvious.

ejoerns (approved):

PR #896, opened 2022-04-13
docs/scenarios: remove bootname for rescue slot in additinal rescue slot

sknsean:
Avoid having bootname defined for the rescue slot, as it when updated
would result in RAUC would overwrite bootorder to prefer the rescue slot.

Signed-off-by: Sean Nyekjaer <sean.nyekjaer.ext@siemensgamesa.com>


ejoerns (approved):
Thanks for fixing!

PR #902, opened 2022-04-22
update documentation on functionality that is out-of-scope

jluebbe:


ejoerns (approved):
Always good to at least have a place in the documentation where we can point at in these cases :sunglasses:

Issue #904, opened 2022-05-03
Prerequisites @ README


atar-axis:
Is it neccessary for the kernel parameters to be "=y" instead of "=m"?
On a common ubuntu system there are some parameters, like `_DM_VERITY`,  `_DEV_NBD` and `_DM_CRYPT` which are just modules instead of builtins. Is that a problem?

jluebbe:
They just need to be enabled, either `=y` or `=m` should be fine as long your module loading is working correctly.

PR #905, opened 2022-05-04
kernel options, see #904

atar-axis:
Kernel options, see #904

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -148,22 +148,22 @@ If you intend to use json-support you also need
 Target Prerequisites
 ~~~~~~~~~~~~~~~~~~~~
 
-Required kernel options:
+Required kernel options (`y` or at least `m`)::
 
--  ``CONFIG_MD=y``
--  ``CONFIG_BLK_DEV_DM=y``
--  ``CONFIG_BLK_DEV_LOOP=y``
--  ``CONFIG_DM_VERITY=y``
--  ``CONFIG_SQUASHFS=y``
--  ``CONFIG_CRYPTO_SHA256=y``
+-  ``CONFIG_MD``
+-  ``CONFIG_BLK_DEV_DM``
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -148,22 +148,22 @@ If you intend to use json-support you also need
...
+-  ``CONFIG_DM_VERITY``
+-  ``CONFIG_SQUASHFS``
+-  ``CONFIG_CRYPTO_SHA256``
 -  ``CONFIG_BLK_DEV_NBD`` (for streaming support)
 -  ``CONFIG_DM_CRYPT`` (for encryption support)
 
 For using tar archive in RAUC bundles with Busybox tar, you have to enable the
 following Busybox feature:
 
--  ``CONFIG_FEATURE_TAR_AUTODETECT=y``
```
a3f, line 165: What does Kconfig `=m` map to in busybox? I'd just leave this as is as it's unrelated to your kernel configuration.

atar-axis: All this change does is to allow not only `y`, but also `m`. So why does it matter to what it maps on busybox?

a3f: The commit message claims it's about kernel options, which busybox configuration is not. You can write a proper commit message and say something along the lines of "While we are at it, change required busybox CONFIG_ variables for symmetry" and that would be ok.

jluebbe: Busybox doesn't use `=m`, so changing it should stay as `=y`. Anything else would just be confusing. :)

jluebbe:
Your commit is missing the DCO sign-off: https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin

You can add it and force-push the PR.

jluebbe:
Closing in favor of #923. Thanks for pointing us to this issue.

PR #912, opened 2022-05-17
Add --trust-environment for 'extract' and 'extract-signature'

RyuzakiKK:
This option was already available for other commands, like
'rauc convert'.

Sometimes a bundle needs to be extracted but it may not be possible, or
desirable, to perform the standard bundle access checks.

For example if RAUC is executed from a controlled environment where the
ownership of a bundle is not an issue. Or when obtaining the underlying
device is not possible at all, e.g. when inside an unprivileged Docker
container.

----

I didn't add `--no-verify` because that's potentially more risky, and usually `--trust-environment` should be enough to cover nearly all use cases.

### DIFF
--- a/rauc.1
+++ b/rauc.1
@@ -155,6 +155,14 @@ verification keyring file
 .RS 4
 Extract the bundle content to a directory.
 
+\fBOptions:\fR
+
+.RS 4
+
+.TP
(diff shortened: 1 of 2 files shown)


ejoerns (approved):
Looks fine for me

jluebbe (approved):

PR #914, opened 2022-05-31
src/verify_hash.c: include inttypes.h

ffontaine:
Include `inttypes.h` to avoid the following build failure:

```
In file included from /nvmedata/autobuild/instance-22/output-1/host/bin/../sparc-buildroot-linux-uclibc/sysroot/usr/include/glib-2.0/glib.h:62,
                 from src/verity_hash.c:26:
src/verity_hash.c: In function 'verify_zero':
src/verity_hash.c:69:55: error: expected ')' before 'PRIu64'
   69 |    g_message("Spare area is not zeroed at position %" PRIu64 ".",
      |                                                       ^~~~~~

```
Fixes:
 - http://autobuild.buildroot.org/results/1a093c0e194a061836884419d2f50506105db01e

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/verity_hash.c
+++ b/src/verity_hash.c
@@ -19,6 +19,7 @@
  */
 
 #include <errno.h>
+#include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>


ejoerns (approved):
Haven't built for sparc before I guess but including inttypes.h sounds appropriate, yes.

The original code in cryptsetup includes `inttypes.h` via `internal.h`.

We *could* remove `stdint.h` inclusion here as `inttypes.h` already includes this, but it might also make sense to keep it for comparability with the original code.
And we *could* move it next to the `glib.h` include for the same reason.

@jluebbe any opinion?

jluebbe:
> The original code in cryptsetup includes `inttypes.h` via `internal.h`.
> 
> We _could_ remove `stdint.h` inclusion here as `inttypes.h` already includes this, but it might also make sense to keep it for comparability with the original code. And we _could_ move it next to the `glib.h` include for the same reason.

We need `stdint.h` for `uintN_t`, at least. So I'd merge this as is.

jluebbe (approved):

PR #918, opened 2022-06-03
atomic bootloader update for SoCs with redundant copy support

emfend:
Some SoCs (like the Rockchip RK3568) contain a fixed ROM code that searches the possible boot media for valid images in a defined order. Usually this happens at two or more fixed addresses.
This branch adds support to implement a failsafe update of the bootloader for such systems.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -1107,7 +1107,8 @@ It requires RAUC to be compiled with GPT support (``./configure --enable-gpt``)
 and adds a dependency on libfdisk.
 
 The required slot type is ``boot-gpt-switch``.
-The device to be specified is expected to be the underlying block device.
+The device to be specified is expected to be the underlying block device (not a
+partition).
 The boot partitions are derived by the definition of the values ``region-start``
 and ``region-size``.
 Both values have to be set in integer decimal bytes and can be post-fixed with
@@ -1131,6 +1132,61 @@ A ``system.conf`` section could look like this:
   type=boot-gpt-switch
   region-start=1M
   region-size=64M
+
+.. _sec-raw-partition-fallback:
+
+Update Raw Boot Partition with Fallback
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some SoCs (like the Rockchip RK3568) contain a fixed ROM code that searches the
+possible boot media for valid images in a defined order.
+Usually this happens at two or more fixed addresses ("locations").
+Special headers, magic numbers, checksums, or a combination of these may be used
+to determine if a valid image exists at such an address.
+
+This behavior can be used to implement an atomic update of a bootloader.
+To do this, the bootloader, which starts with the required header, is
+installed in two locations that the ROM code searches for possible boot images.
+For example, usually only the code in the first location is used, and the second
+is redundant and ignored.
+During an update, the currently unused location is updated first, and followed
+by the other location (which was likely just booted from).
+It is important that the corresponding header is always deleted first, then the
+boot image is written and the header is only written last.
+This ensures that there is always a valid image in either location and that
+half-written images are not attempted to boot from.
+
+The required slot type is ``boot-raw-fallback``.
+The device to be specified is expected to be the underlying block device.
+The location of each copy in the boot region is derived from the values
+``region-start`` and ``region-size``.
+Both values have to be set in integer decimal bytes and can be post-fixed with
+K/M/G/T.
+
+A ``system.conf`` section could look like this:
+
+.. code-block:: cfg
+
+  [slot.bootloader.0]
+  device=/dev/mmcblk0
+  type=boot-raw-fallback
+  region-start=32k
+  region-size=4M
+
+It defines a region starting at ``0x8000`` with a size of ``4M``.
+This region will be split up into two halves of equal size by RAUC internally.
+This results in two halves, one starting at ``0x8000`` and one at ``0x208000``,
+both with a size of ``2M``.
+The first half is the location where the normally used bootloader should be
+stored and the second defines the location of the fallback location.
+The header size is currently fixed to 512 bytes.
+
+Since the implementation makes certain assumptions, it is important that the
+SoC ROM code tries to boot from the first location first.
+Note that under most circumstances the update will appear to work fine even if
+the two locations are swapped.
+However, for the update to actually be failsafe, the locations must be searched
+in order by SoC ROM code.
 
 Bootloader Update Ideas
 ~~~~~~~~~~~~~~~~~~~~~~~
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -18,7 +18,6 @@
 #include "gpt.h"
 #include "utils.h"
 
-
 #define R_SLOT_HOOK_PRE_INSTALL "slot-pre-install"
 #define R_SLOT_HOOK_POST_INSTALL "slot-post-install"
 #define R_SLOT_HOOK_INSTALL "slot-install"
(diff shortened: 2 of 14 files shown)


docs/advanced.rst
```diff
@@ -1132,6 +1132,61 @@ A ``system.conf`` section could look like this:
...
+normally started from. It is important that the corresponding header is always
+deleted first, then the boot image is written and the header is only written
+last.
+This ensures that there is always a valid image on either address and that
+half-written images are not attempted to boot from.
+
+The required slot type is ``boot-raw-fallback``.
+The device to be specified is expected to be the underlying block device.
+The boot partitions are derived by the definition of the values
+``region-start`` and ``region-size``.
```
jluebbe, line 1161: Using _partitions_ here could be confusing.
```suggestion
The location of each copy in the boot region is derived from the values
``region-start`` and ``region-size``.
```

docs/advanced.rst
```diff
@@ -1132,6 +1132,61 @@ A ``system.conf`` section could look like this:
...
   region-size=64M
 
+.. _sec-raw-partition-fallback:
+
+Update Raw Boot Partition with Fallback
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some SoCs (like the Rockchip RK3568) contain a fixed ROM code that searches the
+possible boot media for valid images in a defined order.
+Usually this happens at two or more fixed addresses.
```
jluebbe, line 1142:
```suggestion
Usually this happens at two or more fixed addresses ("locations").
```

docs/advanced.rst
```diff
@@ -1132,6 +1132,61 @@ A ``system.conf`` section could look like this:
...
+Special headers and/or magic numbers are typically used to determine if there
+is a valid image at such an address.
+
+This behavior can be used to implement a failsafe update of a bootloader.
+To do this, the bootloader, which starts with the required header, is
+installed in two places that the ROM code searches for possible boot images.
+For example, usually only the code at the first address is used, and the second
+is redundant and unused.
```
jluebbe, line 1150:
```suggestion
To do this, the bootloader, which starts with the required header, is
installed in two locations that the ROM code searches for possible boot images.
For example, usually only the code in the first location is used, and the second
is redundant and ignored.
```

docs/advanced.rst
```diff
@@ -1132,6 +1132,61 @@ A ``system.conf`` section could look like this:
...
+During an update, the unused code is updated first, and then the area that is
+normally started from. It is important that the corresponding header is always
```
jluebbe, line 1152:
```suggestion
During an update, the currently unused location is updated first, and followed by the other
location (which as likely just booted from).
It is important that the corresponding header is always
```

docs/advanced.rst
```diff
@@ -1132,6 +1132,61 @@ A ``system.conf`` section could look like this:
...
+``0x208000``, both with a size of ``2M``.
+The first region defines the area where the normally used bootloader should be
+located and the second defines the location of the fallback area.
+
+Since the implementation makes certain assumptions, it is important that the
+SoC ROM code tries to boot from the first region first.
+Note that under normal circumstances the update will work fine even if the two
+regions are swapped.
+However, for the update to actually be failsafe, the areas must match the
+actual search order of the SoC ROM code.
```
jluebbe, line 1188:
```suggestion
Note that under most circumstances the update will appear to work fine even if the two
regions are swapped.
However, for the update to actually be failsafe, the areas must be searched in order by SoC ROM code.
```

We should also have a note that the header size is currently fixed to 512 bytes.

jluebbe (changes requested):
The code looks really good, I just had some comments on the documentation.
Before merging, we'd need some test cases (normal case, simulation of interrupted updates).

src/update_handler.c
```diff
@@ -196,6 +197,21 @@ static gboolean copy_raw_image(RaucImage *image, GUnixOutputStream *outstream, G
 		return FALSE;
 	}
 
+	if (len_header_last) {
+		header = g_malloc(len_header_last);
```
jluebbe, line 201: We should check if `len_header_last` is aligned to the sector size. You could make `get_sectorsize()` from `mbr.c` a utility function for that.

docs/advanced.rst
```diff
@@ -1132,6 +1132,61 @@ A ``system.conf`` section could look like this:
...
+.. _sec-raw-partition-fallback:
+
+Update Raw Boot Partition with Fallback
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some SoCs (like the Rockchip RK3568) contain a fixed ROM code that searches the
+possible boot media for valid images in a defined order.
+Usually this happens at two or more fixed addresses.
+Special headers and/or magic numbers are typically used to determine if there
+is a valid image at such an address.
```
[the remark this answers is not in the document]
emfend: Yes, perhaps. To be honest, I'm not familiar with the exact behavior of many ROM code loaders.
However, since that could likely be the case, I'll mention that as well.

docs/advanced.rst
emfend: Instead of checking and throwing an error, it might be better to change img_to_boot_raw_fallback_handler() to use the actual sector size as the header length instead of the fixed 512 bytes.
What do you think?

src/update_handler.c
emfend: Thank you very much for your comments and the valuable improvement to the documentation.

emfend: Instead of checking and throwing an error, it might be better to change img_to_boot_raw_fallback_handler() to use the actual sector size as the header length instead of the fixed 512 bytes.
What do you think?

emfend:
What's new in the updated version:
- Addressed comments and improvements for documentation
- Actual physical sector size used for header size (instead of a hardcoded value of 512)
- Added some tests for the new slot type

emfend:
What's new in the updated version:
- Addressed comments and improvements for documentation
- Actual physical sector size used for header size (instead of a hardcoded value of 512)
- Added some tests for the new slot type

docs/advanced.rst
emfend: Now the physical sector size is used, which is also mentioned in a small note in the documentation.

src/update_handler.c
jluebbe: By using the hardware sector size as the header size, users have no chance to detect configuration issues, so I'd prefer to check and report an error on mismatch.

src/update_handler.c
```diff
@@ -1921,6 +1940,200 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+	region_size = dest_slot->region_size / 2;
+
+	if (region_size < (guint64)image->checksum.size) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Size of image (%"G_GOFFSET_FORMAT ") does not fit to slot size %"G_GUINT64_FORMAT,
+				image->checksum.size, region_size);
+		goto out;
+	}
+
+	res = get_phys_sector_size(dest_slot->device, &header_size, &ierror);
```
jluebbe, line 2062: The correct header size can't be automatically determined by using the sector size, but it depends on how the image is structured. So we should only check that what we want to do is consistent with the sector size.

Down in `copy_raw_image` we already have the FD, so we can use `get_sectorsize()` after moving it to `src/utils.c`.

src/update_handler.c
```diff
@@ -1921,6 +1940,200 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+	res = TRUE;
+
+out:
+	if (fd >= 0)
+		g_close(fd, NULL);
+
+	return res;
+}
+
+static gboolean get_phys_sector_size(const gchar *device, gsize *sector_size, GError **error)
```
jluebbe, line 2003: We don't want to use the physical sector size, as this can be 4kB or 512B depending on the device. Relevant for us is the logical sector size (as read by `get_sectorsize()` in `src/mbr.c`).

If they differ, accesses to a single logical sector are still atomic, but the device may have to perform a read-modify-write cycle, which is what we need here.

src/update_handler.c
emfend: I can change that of course, but I would better understand the reasons why.
This type of slot work exploits the fact that block writes equal to the physical block size are atomic. This should actually always be correct and should not require any further configuration.

What is the advantage if the header is assumed to be 512 and an error message is issued if it deviates from this? If the underlying block device has a sector size of e.g. 4k, then this mechanism should also work with a header size of 4k. Or am I missing something here?

test/boot_raw_fallback.c
```diff
@@ -0,0 +1,414 @@
...
+
+	if (compare)
+		cmp_buf = g_malloc(buf_size);
+
+	while (size) {
+		gsize bytes_to_process = size < buf_size ? size : buf_size;
+		gsize bytes_processed;
+
+		if (random)
+			buf = (gchar*) random_bytes(bytes_to_process, 31415);
```
jluebbe, line 75: This leaks memory on every loop. It would probably be simpler (and avoid repeating patterns) to either use g_malloc+memset or random_bytes for the full size.

test/boot_raw_fallback.c
```diff
@@ -0,0 +1,414 @@
...
+
+	image = g_new0(RaucImage, 1);
+	image->slotclass = g_strdup("rootfs");
+	image->filename = g_strdup(imagepath);
+	image->checksum.size = size;
+	image->checksum.digest = g_strdup("0xdeadbeef");
+
+	g_assert(test_prepare_dummy_file(dirname, imagename, size, "/dev/zero") == 0);
+
+	fill_file(imagepath, 0, size, 0x00, TRUE);
```
jluebbe, line 155: This could be replaced with `write_random_file()` from `test/common.c`.

test/boot_raw_fallback.c
```diff
@@ -0,0 +1,414 @@
...
+	if (data->options & OPT_EXPECT_FAIL) {
+		g_assert_error(ierror, data->err_domain, data->err_code);
+		g_assert_false(res);
+		goto out;
+	} else {
+		g_assert_no_error(ierror);
+		g_assert_true(res);
+	}
+
+	/* Check if the area before the primary partition is untouched */
```
jluebbe, line 261:
```suggestion
	/* Check if the area before the primary image is untouched */
```

test/boot_raw_fallback.c
```diff
@@ -0,0 +1,414 @@
...
+	g_assert(g_mkdir(mountprefix, 0777) == 0);
+
+	/* get handler for this */
+	handler = get_update_handler(image, targetslot, &ierror);
+	g_assert_no_error(ierror);
+	g_assert_nonnull(handler);
+
+	/* Run to perform an update */
+	res = handler(image, targetslot, hookpath, &ierror);
+
```
jluebbe, line 251: You could wrap this with some `g_test_expect_message()` before and `g_test_assert_expected_messages()` after the handler call. This way, you can check if the "Updating primary partition at 0x??? on *" and "Updating fallback partition at 0x??? on *" appear in the correct order.

test/boot_raw_fallback.c
```diff
@@ -0,0 +1,414 @@
...
+		.region_start = (0 * 1024 * 1024),
+		.region_size = (4 * 1024 * 1024),
+	};
+	g_test_add("/boot_raw_fallback/primary-header-missing",
+			BootRawFallbackFixture,
+			data,
+			boot_raw_fallback_fixture_set_up,
+			test_boot_raw_fallback,
+			boot_raw_fallback_fixture_tear_down);
+
```
jluebbe, line 348: This should have the messages report the primary image update first.

jluebbe (changes requested):

src/update_handler.c
jluebbe: Take this scenario:

Prototypes are built with an eMMC which has a physical sector size of 4KB. The firmware image as a 1KB header, which starts with some unprotected data and continues at a 512B offset with the data actually checked by the ROM code.
When deriving the header size from the sector size, this works fine on the prototypes.

If the eMMC is later replaced with another model using 512B sectors for production, the behavior silently switches to writing only the first 512B last.


By using a documented fixed header size of 512B, we have a chance to report this as soon as the user tests it on the prototypes. At that point, an explicit config option for the header size would probably need to be added.
At that point, backwards compatibility for the config is also possible, as only 512B headers were supported until then and we can default to that.

docs/advanced.rst
emfend: Since I've switched back to the fixed head size, I've also added a note on the documentation.

src/update_handler.c
emfend: ACK

test/boot_raw_fallback.c
emfend: True. But as it is important that the content meets the expectations from compare_file(), I would like to stick with fill_file()-compare_file() pairs.

test/boot_raw_fallback.c
emfend: ACK

test/boot_raw_fallback.c
emfend: Nice! That is a very good idea.
Thanks for this tip!

emfend:
I noticed that the CI check "test/build" failed for the previous branch.
In the logs I could see that for example '/dm/create_257' reported an error.
But since this test runs fine on my local machine and (at least for me) this seems unrelated to my changes, I'm not sure if I caused this regression.

test/boot_raw_fallback.c
jluebbe: As these partitions aren't multiple MB in size, one large allocation wouldn't hurt, but it's fine this way as well.

jluebbe:
I made some minor changes to the wording in the docs and simplified the sector size assignment:
``` diff
diff --git a/docs/advanced.rst b/docs/advanced.rst
index 5f80fe2a..728b8c00 100644
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -1107,7 +1107,8 @@ It requires RAUC to be compiled with GPT support (``./configure --enable-gpt``)
 and adds a dependency on libfdisk.
 
 The required slot type is ``boot-gpt-switch``.
-The device to be specified is expected to be the underlying block device.
+The device to be specified is expected to be the underlying block device (not a
+partition).
 The boot partitions are derived by the definition of the values ``region-start``
 and ``region-size``.
 Both values have to be set in integer decimal bytes and can be post-fixed with
@@ -1149,10 +1150,10 @@ installed in two locations that the ROM code searches for possible boot images.
 For example, usually only the code in the first location is used, and the second
 is redundant and ignored.
 During an update, the currently unused location is updated first, and followed
-by the other location (which as likely just booted from).
+by the other location (which was likely just booted from).
 It is important that the corresponding header is always deleted first, then the
 boot image is written and the header is only written last.
-This ensures that there is always a valid image on either location and that
+This ensures that there is always a valid image in either location and that
 half-written images are not attempted to boot from.
 
 The required slot type is ``boot-raw-fallback``.
@@ -1173,20 +1174,19 @@ A ``system.conf`` section could look like this:
   region-size=4M
 
 It defines a region starting at ``0x8000`` with a size of ``4M``.
-This region will be split up into two region halves of equal size by RAUC
-internally.
-This results in two regions, one starting at ``0x8000`` and one at
-``0x208000``, both with a size of ``2M``.
-The first region defines the area where the normally used bootloader should be
-located and the second defines the location of the fallback area.
+This region will be split up into two halves of equal size by RAUC internally.
+This results in two halves, one starting at ``0x8000`` and one at ``0x208000``,
+both with a size of ``2M``.
+The first half is the location where the normally used bootloader should be
+stored and the second defines the location of the fallback location.
 The header size is currently fixed to 512 bytes.
 
 Since the implementation makes certain assumptions, it is important that the
-SoC ROM code tries to boot from the first region first.
+SoC ROM code tries to boot from the first location first.
 Note that under most circumstances the update will appear to work fine even if
-the two regions are swapped.
-However, for the update to actually be failsafe, the areas must be searched in
-order by SoC ROM code.
+the two locations are swapped.
+However, for the update to actually be failsafe, the locations must be searched
+in order by SoC ROM code.
 
 Bootloader Update Ideas
 ~~~~~~~~~~~~~~~~~~~~~~~
diff --git a/src/update_handler.c b/src/update_handler.c
index f7d798a4..af8d416a 100644
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -196,9 +196,7 @@ static gboolean copy_raw_image(RaucImage *image, GUnixOutputStream *outstream, g
        }
 
        if (len_header_last) {
-               gsize sector_size;
-
-               sector_size = (gsize) get_sectorsize(out_fd);
+               gsize sector_size = (gsize) get_sectorsize(out_fd);
 
                if (len_header_last != sector_size) {

```

src/update_handler.c
```diff
@@ -1921,6 +1947,162 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+	} part_desc[2];
+	int first_part_desc_index = 0;
+	g_autoptr(GHashTable) vars = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
+	const gsize header_size = 512;
+
+	g_return_val_if_fail(image, FALSE);
+	g_return_val_if_fail(dest_slot, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	region_size = dest_slot->region_size / 2;
```
jluebbe, line 2028: The name `region_size` is used for two different things here. Perhaps we should use `half_size` here.

emfend: Good idea.

src/update_handler.c
```diff
@@ -1921,6 +1947,162 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
+				"Size of image (%"G_GOFFSET_FORMAT ") does not fit to slot size %"G_GUINT64_FORMAT,
+				image->checksum.size, region_size);
+		goto out;
+	}
+
+	g_hash_table_insert(vars, g_strdup("RAUC_BOOT_PARTITION_START"),
+			g_strdup_printf("%"G_GUINT64_FORMAT, dest_slot->region_start));
+	g_hash_table_insert(vars, g_strdup("RAUC_BOOT_PARTITION_SIZE"),
+			g_strdup_printf("%"G_GUINT64_FORMAT, region_size));
```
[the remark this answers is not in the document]
jluebbe: You wouldn't be able to reuse one script for both slot types anyway. So I'd prefer different variables as the meaning is diffrent. So `RAUC_BOOT_REGION_START` and RAUC_BOOT_REGION_SIZE`.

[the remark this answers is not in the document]
jluebbe: I'll handle that myself and squash your changes.

jluebbe (approved):

jluebbe:
Thanks for this feature, @emfend!

PR #920, opened 2022-06-09
src/nbd: no need to retry if we received a 404 from the server

ejoerns:
If the server does not provide the requested file, this won't likely
change on subsequent tries.

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -846,6 +846,7 @@ gboolean r_nbd_run_server(gint sock, GError **error)
 
 		while (1) {
 			CURLcode code = 0;
+			long response_code = 0;
 			int msgs_in_queue = 0;
 			struct RaucNBDTransfer *xfer = NULL;
 			struct CURLMsg *msg = curl_multi_info_read(ctx.multi, &msgs_in_queue);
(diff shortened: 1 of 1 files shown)


src/nbd.c
```diff
@@ -860,6 +861,10 @@ gboolean r_nbd_run_server(gint sock, GError **error)
 			code = curl_easy_getinfo(msg->easy_handle, CURLINFO_PRIVATE, &xfer);
 			g_assert(code == CURLE_OK);
 
+			code = curl_easy_getinfo(xfer->easy, CURLINFO_RESPONSE_CODE, &response_code);
```
jluebbe, line 864: Can't we use
```suggestion
			code = curl_easy_getinfo(msg->easy_handle, CURLINFO_RESPONSE_CODE, &response_code);
```
here?

src/nbd.c
```diff
@@ -869,8 +874,14 @@ gboolean r_nbd_run_server(gint sock, GError **error)
...
-				g_message("request failed (retrying)");
-				xfer->errors++;
+				if (response_code == 404) {
+					g_message("request failed (not found)");
+					xfer->reply.error = GUINT32_TO_BE(5); /* NBD_EIO */
+					xfer->done = TRUE;
+				} else {
+					g_message("request failed (retrying)");
+					xfer->errors++;
+				}
```
jluebbe, line 884: Couldn't this be just another `else if`, instead of being nested in the `else`?

jluebbe (changes requested):

src/nbd.c
ejoerns: Yes, I had this in mind but wasn't sure which variant to choose. Will change.

PR #928, opened 2022-06-21
src/mount: retry when kernel reports EAGAIN for ioctl(LOOP_SET_BLOCK_SIZE)

a3f:
Just like `ioctl(LOOP_SET_STATUS64)`, `ioctl(LOOP_SET_BLOCK_SIZE)` will
also drop kernel-internal caches first to avoid giving the users unexpected
data. As the kernel doesn't prevent loop disk page cache getting dirtied
afterwards, userspace is expected to check against `EAGAIN` and retry the
ioctl if needed[0].

We have been doing just that for `LOOP_SET_STATUS64`, but not for
`LOOP_SET_BLOCK_SIZE`. Fix this by also calling that `ioctl` in a loop.

For reference, losetup(1) has been calling it in a bounded (n=64)
loop since v2.38[1].  We don't need the bound here, because unlike
losetup(1), we have multiple retry loops for each operation, while
it has a single one also handling `-EBUSY`, which ought to be bounded.

Note that the kernel will print a message on each extra loop iteration:
```
  loop_set_block_size: loop0 () still has dirty pages (nrpages=1)
```
This is somewhat ugly, but is out of RAUC's hands.


a3f:
> This is somewhat ugly, but is out of RAUC's hands.

While this is true in general, RAUC could try to prevent others from using the block device. When I send a `SIGSTOP` to systemd-udevd, the kernel doesn't log the `still has dirty pages` message, but [the obvious solution of taking an exclusive advisory lock](https://systemd.io/BLOCK_DEVICE_LOCKING/) on the loop device has no effect. If I figure it out, I will open another PR, but this patch is applicable in any case.

jluebbe (approved):

PR #935, opened 2022-07-07
improve stats messages

jluebbe:
We can now set a label when we create a stats struct and a prefix when we print it.
Combine both in `r_stats_show()` and adjust the code in `nbd.c` accordingly.


jluebbe:
Yes. Not sure how that happened. I'll fix it.

PR #937, opened 2022-07-12
minor fixes

jluebbe:


src/update_handler.c
```diff
@@ -416,6 +416,11 @@ static RaucSlot *get_active_slot_class_member(gchar *slotclass)
 
 	g_return_val_if_fail(slotclass, NULL);
 
+	if (!r_context()->config->slots) {
```
ejoerns, line 419: Currently, we use `config->slots` unconditionally in many places in the code.
And there is an explicit check in `determine_slot_states()` already.
Why is this special here and is this a case we want to support?
Or should we add an assertion instead?

jluebbe: We can call this code directly from the update_handler test cases, where no slots are configured.

ejoerns: Ok, since this is for seeding and one wants to be able to test casync/incremental probably without requiring to set up a slot hash table this sounds appropriate.

ejoerns (approved):

PR #939, opened 2022-07-14
Add doctype to dbus xml specification

### LINKED ISSUE #938: Add doctype to dbus xml specification
Was using the **qdbusxml2cpp** tool from Qt and noticed that it does not like to parse the D-bus xml specification, like so:
    
    qdbusxml2cpp de.pengutronix.rauc.Installer.xml  -a interface.h
    qdbusxml2cpp: Cannot process input: 'de.pengutronix.rauc.Installer.xml'. Stop.

Tried to add a doctype tag to the xml file, and it started working. Really, the Qt tool should be more forgiving, but it does not hurt to add the doctype I presume? For better compatibility with picky tools.

morganbengtsson:
Fixes #938.


jluebbe:
Looks good, but please add a Signed-off-by line (see https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin) and force-push to update the PR.

jluebbe (approved):

ejoerns (approved):
Looks good! (Especially since even the minimal example from https://dbus.freedesktop.org/doc/dbus-specification.html has a proper DOCTYPE set)

PR #945, opened 2022-08-15
src/signature: get key passphrase from environment

marckleinebudde:
The user can add the passphrase of the private key to the environment
variable RAUC_KEY_PASSPHRASE. Pass this to PEM_read_bio_PrivateKey()
if set.

Signed-off-by: Marc Kleine-Budde <mkl@pengutronix.de>


docs/advanced.rst
```diff
@@ -150,6 +150,18 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
 Note that you can specify the ``--intermediate`` argument multiple times to
 include multiple intermediate certificates to your bundle signature.
 
+Passphrase Handling
+~~~~~~~~~~~~~~~~~~~
+
+If the signing key is protected by a passphrase it has to entered
+during signing operation. For automation purpose the passphrase can be
+set in the environment variable RAUC_KEY_PASSPHRASE.
```
jluebbe, line 158:
```suggestion
If the signing key is protected by a passphrase it has to entered
during signing operations.
For automation purposes the passphrase can be
set using the environment variable RAUC_KEY_PASSPHRASE.
```

docs/advanced.rst
```diff
@@ -150,6 +150,18 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
...
+
+.. note::
+  Since the environment of other processes is visible on certain
+  platforms (e.g. ps under certain Unix OSes) this option should be
+  used with caution.
```
jluebbe, line 163: We only support Linux.

docs/reference.rst
```diff
@@ -925,6 +925,7 @@ Command Line Tool
     write-slot            Write image to slot and bypass all update logic
 
   Environment variables:
+    RAUC_KEY_PASSPHRASE Passphrase to use for accessing keys files (signing only)
```
jluebbe, line 928:
```suggestion
    RAUC_KEY_PASSPHRASE Passphrase to use for accessing key files (signing only)
```

rauc.1
```diff
@@ -328,6 +328,10 @@ Write image to slot and bypass all update logic.
 
 .SH ENVIRONMENT
 
+.TP
+.B RAUC_KEY_PASSPHRASE
+Passphrase to use for accessing keys files (signing only)
```
jluebbe, line 333:
```suggestion
Passphrase to use for accessing key files (signing only)
```

src/signature.c
```diff
@@ -188,6 +188,7 @@ static EVP_PKEY *load_key_file(const gchar *keyfile, GError **error)
 	BIO *key = NULL;
 	unsigned long err;
 	const gchar *data;
+	gchar *passphrase;
```
jluebbe, line 191:
```suggestion
	const gchar *passphrase;
```

jluebbe (changes requested):
We should probably add a test case as well.

docs/advanced.rst
marckleinebudde: That was copied from the openssl man page :) fixed

src/signature.c
marckleinebudde:
```
src/signature.c:210:56: warning: passing argument 4 of 'PEM_read_bio_PrivateKey' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  210 |         res = PEM_read_bio_PrivateKey(key, NULL, NULL, passphrase);
      |                                                        ^~~~~~~~~~
```

marckleinebudde: Too many deprecation warnings with openssl-3 :crying_cat_face:
I've taken the `g_strdup()` + `g_free()` route...

docs/reference.rst
marckleinebudde: fixed

rauc.1
marckleinebudde: fixed

src/signature.c
jluebbe: In cases like these (where we know that OpenSSL won't write to this pointer), we've casted away const in the function call.

docs/advanced.rst
```diff
@@ -150,6 +150,18 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
 Note that you can specify the ``--intermediate`` argument multiple times to
 include multiple intermediate certificates to your bundle signature.
 
+Passphrase Handling
+~~~~~~~~~~~~~~~~~~~
+
+If the signing key is protected by a passphrase it has to entered
```
ejoerns, line 156:
```suggestion
If the signing key is protected by a passphrase it has to be entered
```

docs/advanced.rst
```diff
@@ -150,6 +150,18 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
...
+during signing operations.
+For automation purposes the passphrase can be set using the
+environment variable RAUC_KEY_PASSPHRASE.
```
ejoerns, line 159:
```suggestion
environment variable ``RAUC_KEY_PASSPHRASE``.
```

docs/advanced.rst
```diff
@@ -150,6 +150,18 @@ To do this, specify the ``--intermediate`` argument during bundle creation::
...
+
+.. note::
+  Since the environment of other processes is visible on Linux this
```
ejoerns, line 162:
```suggestion
  Since the environment of other processes is visible on Linux, this
```

ejoerns (changes requested):

ejoerns (approved):

jluebbe (approved):

jluebbe:
Thanks for your contribution!

PR #946, opened 2022-08-17
manifest: ignore meta.<label> sections in the manifest

jluebbe:
The `meta.<label>` sections are intended to provide a forwards-compatible way to
add data to the manifest which is not interpreted by RAUC in any way.
Currently, they are just ignored when reading a manifest. In future releases,
they will be accessible via `rauc info`, the D-Bus API and in hooks/handlers.

### DIFF
--- a/test/manifest.c
+++ b/test/manifest.c
@@ -334,6 +334,51 @@ incremental=invalid-method;another-invalid-method\n\
 	free_manifest(rm);
 }
 
+static void test_manifest_load_meta(void)
+{
+	gchar *tmpdir;
+	RaucManifest *rm = NULL;
+	gchar* manifestpath = NULL;
+	gboolean res;
+	GError *error = NULL;
+	RaucImage *test_img = NULL;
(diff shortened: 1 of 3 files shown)


test/manifest.c
```diff
@@ -334,6 +334,54 @@ incremental=invalid-method;another-invalid-method\n\
...
+	RaucImage *test_img = NULL;
+	const gchar *mffile = "\
+[update]\n\
+compatible=FooCorp Super BarBazzer\n\
+version=2015.04-1\n\
+\n\
+[image.rootfs]\n\
+filename=rootfs-default.ext4\n\
+incremental=invalid-method;another-invalid-method\n\
+[meta.foocorp]\n\
```
ejoerns, line 353: for consistency, I suggest adding a blank line between all sections.

ejoerns (commented):
From my side just two notes on the tests.

ejoerns (approved):

PR #949, opened 2022-08-18
test/update_handler: do not fail when sum_target_written is zero

ejoerns:
It seems to occur that `sum_target_written` is zero.
This currently causes CI checks to fail sometimes unexpectedly.

Fix this by also allowing 0 as we cannot certainly determine yet which factors influence the found chunks for the ext4 test case.

For reference: https://github.com/rauc/rauc/runs/7895842002?check_suite_focus=true

### DIFF
--- a/test/update_handler.c
+++ b/test/update_handler.c
@@ -414,7 +414,7 @@ no_image:
 			 * that have proven to be valid on all test systems.
 			 */
 			g_assert_cmpint(sum_zero, >=, IMAGE_SIZE/4096 - 29);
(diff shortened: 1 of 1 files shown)


ejoerns:
Not sure if this isn't somehow also related to:

> ERROR: test/dm.test - Bail out! rauc:ERROR:test/dm.c:405:verity_hash_create: assertion failed (readable_sectors(dmfd) == dm_data->data_size - 1): (1 == 0)

which just did hit us again here... Will retrigger..

PR #950, opened 2022-08-18
test/nginx_backend: avoid DeprecationWarning when changing application state

jluebbe:
This fixes the following warning:
```
.../rauc/test/nginx_backend.py:15: DeprecationWarning: Changing state of
started or joined application is deprecated
```

### DIFF
--- a/test/nginx_backend.py
+++ b/test/nginx_backend.py
@@ -12,8 +12,8 @@ async def index(request):
 
 @routes.get("/sporadic.raucb")
 async def sporadic_get(request):
-    request.app["sporadic_counter"] += 1
-    if request.app["sporadic_counter"] % 2 == 0:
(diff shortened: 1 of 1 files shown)


ejoerns:
Seems that my python / aiohttp skills are insufficient for directly figuring out why this is the proper fix. Could you elaborate on this?

jluebbe:
Only direct changes to the application dictionary are deprecated. The suggestion is to use a nested dictionary:
https://github.com/aio-libs/aiohttp/discussions/6804#discussioncomment-3066116

ejoerns:
Ok, seems as if my google foo needs some fixups, too. Only found other threads related to this warning.

But anyway, I find it useful to have an explanation referenced.

ejoerns (approved):

PR #952, opened 2022-08-19
rename incremental mode updates to adaptive mode updates

jluebbe:
Since the release of 1.7, it turned out that the name 'incremental' for this
functionality is confusing.
Accordingly we decided to rename it to 'adaptive' for 1.8 and accept the
downside of not being able to benefit from compatibility with 1.7.


ejoerns (approved):
Thanks for the renaming!

PR #958, opened 2022-08-26
src/context: create the data directory on startup

jluebbe:
This allows us to detect configuration errors earlier and avoids problems where
other parts of RAUC expect this to exist.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -271,6 +271,18 @@ gboolean r_context_configure(GError **error)
 			break;
 	}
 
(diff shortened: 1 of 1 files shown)


ejoerns (approved):
Since for adaptive mode the directory would be generated anyway, it sounds useful to me to create it (and possibly fail) during startup already.

ejoerns:
Re-triggered build that ran into what I prepared #961 for.

PR #961, opened 2022-08-27
test/update_handler: be less strict with sum_source

ejoerns:
When running CI from GitHub Actions, sum_source can obviously be 29 in some cases. Since this is no exact science anyway, be less strict here to prevent false-positive test suite failures.


ejoerns:
Example of a test suite failure as mentioned in the commit: https://github.com/rauc/rauc/runs/8039601549?check_suite_focus=true

jluebbe (approved):

PR #965, opened 2022-09-02
extend Info function with  array-of-string-variant parameter to set important features used in rauc streaming

stefanu21:


src/de.pengutronix.rauc.Installer.xml
```diff
@@ -36,6 +36,21 @@
...
+ 
+    <!--
+    InfoBundle: D-Bus variant of rauc info <bundle>
+    @bundle: full path to the queried bundle.
+    @args: Additional arguments to pass
+    @compatible: the compatible string from the bundle
+    @version: the version string from the bundle
+   -->
+   <method name="InfoBundle">
+      <arg name="bundle" type="s" direction="in" />
```
jluebbe, line 48: Perhaps we should rename the argument to `source`, as it has the same function as that argument in InstallBundle.

src/de.pengutronix.rauc.Installer.xml
```diff
@@ -36,6 +36,21 @@
...
+      <arg name="args" type="a{sv}" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
+      <arg name="compatible" type="s" direction="out" />
+      <arg name="version" type="s" direction="out" />
```
jluebbe, line 52: We'll likely want to expose mode details from the manifest, for example the meta-data sections (introduced by https://github.com/rauc/rauc/pull/946). So perhaps:
```suggestion
      <arg name="details" type="a{sv}" direction="in"/>
```

It could be structured similar to the current output of `rauc info --output-format=json` and would be easy to extend without breaking compatibility:
``` json
{
  "compatible": "Test Config",
  "version": "2011.03-2",
  "description": null,
  "build": null,
  "hooks": [],
  "images": {
    "rootfs": {
      "variant": null,
      "filename": "rootfs.img",
      "checksum": "de2f256064a0af797747c2b97505dc0b9f3df0de4f489eac731c23ae9ca9cc31",
      "size": 65536,
      "hooks": [],
      "adaptive": []
    }
    "appfs": {
      "variant": null,
      "filename": "appfs.img",
      "checksum": "c35020473aed1b4642cd726cad727b63fff2824ad68cedd7ffb73c7cbd890479",
      "size": 32768,
      "hooks": [],
      "adaptive": []
    }
  }
}
```

jluebbe (commented):
Perhaps `InspectBundle` would fit the current style of functions better (`InstallBundle`, `GetSlotStatus`, `GetPrimary`, `Mark`)?

src/de.pengutronix.rauc.Installer.xml
ejoerns: Yes, this has been one of the severe limitations of the 'old' API and is the main reason it was never really usable.
Also without meta data sections we already have a number of information not exposed at all.

I am not sure if calling the argument `details` would not give the impression that there is also non-detailed information which is not the case. Maybe we could name it `info` or `bundle_info`?

src/de.pengutronix.rauc.Installer.xml
ejoerns: I have no strong opinion on this, but yes it might be more consistent (at least within the set of non-deprecated API methods).

ejoerns:
> Perhaps `InspectBundle` would fit the current style of functions better (`InstallBundle`, `GetSlotStatus`, `GetPrimary`, `Mark`)?

In the sense that the method name describes an 'action', yes. So for info this could also be `GetBundleInfo` but probably `InspectBundle` is more catchy then (with the small drawback that this does not match the CLI method's name anymore)

ejoerns:
Since maintainer edits are not allowed and since I have changed significant parts of the original PR, I have created a follow-up (#1067) that addresses the review comments here.

PR #968, opened 2022-09-07
Makefile.am: Add some more test files to the dist tar.xz

ukleinek:
Compared to the git repository there are a few files missing in the 1.7 tarball. Add them to EXTRA_DIST.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -192,6 +192,8 @@ EXTRA_DIST += tap-t \
 	      qemu-test qemu-test-init qemu-test-rauc-config \
 	      test/Dockerfile \
 	      test/bin \
+	      $(wildcard test/good-casync-bundle-1.4.castr/*/*.cacnk) \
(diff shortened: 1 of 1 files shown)


jluebbe (changes requested):
Please rebase on current master and follow the existing subject convention (no capitalization after the colon).

ukleinek:
> Please rebase on current master and follow the existing subject convention (no capitalization after the colon).

In general I think rebasing is a bad idea. There are two reasons for that that are preached for Linux kernel development::

 - rebasing looses  testability
 - Basing new development on a release tends to be more stable because you don't suffer from random breakage in a development tree.
With the commit volume of rauc and the triviality of the patch in question here, I admit this isn't that relevant,

You noted (in RL) that you'd like to see the checks based on something newer than the new release. IMHO this could be accomplished by adapting the test suite to test on `master + my branch` instead of just my branch. But as I don't want to adapt the test suite, I yield.

jluebbe (approved):

PR #976, opened 2022-09-29
src/bundle: do not attempt to take ownership if running as non-root

ejoerns:
As part of the plain bundle ownership safeguard, RAUC attempts to chown bundles that are neither owned by root nor by the uid executing rauc to user root.

This cannot work when rauc runs as non-root. Thus emit a more helpful error message and abort in advance.

To not call `geteuid()` unnecessary often, introduce variable `euid` to store the result.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -1331,8 +1331,11 @@ static gboolean take_bundle_ownership(int bundle_fd, GError **error)
 	struct stat stat = {};
 	mode_t perm_orig = 0, perm_new = 0;
 	gboolean res = FALSE;
+	uid_t euid;
 
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	euid = geteuid();
 
 	if (fstat(bundle_fd, &stat)) {
 		int err = errno;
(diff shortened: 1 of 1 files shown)


src/bundle.c
```diff
@@ -1344,8 +1347,18 @@ static gboolean take_bundle_ownership(int bundle_fd, GError **error)
...
-	/* if it belongs to someone else, try to fchown */
-	if ((stat.st_uid != 0) && (stat.st_uid != geteuid())) {
+	/* if it belongs to someone else, try to fchown if we are root */
+	if ((stat.st_uid != 0) && (stat.st_uid != euid)) {
+
+		if (euid != 0) {
+			g_set_error(error,
+					G_FILE_ERROR,
+					G_FILE_ERROR_PERM,
+					"cannot chown bundle to root when running as user (%d)", euid);
```
jluebbe, line 1357:
```suggestion
					"cannot take file ownership of bundle when running as user (%d)", euid);
```

jluebbe (changes requested):

jluebbe (approved):

PR #991, opened 2022-10-11
Usability Improvements for Adaptive Updates

ejoerns:
This fixes cases where users of adaptive updates can unexpectedly run into issues that not necessarily tell what is actually wrong:

* having adaptive mode enabled but no data directory configured in system
* trying to generate block hash index from archives
* knowing if and which adaptive method was selected

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -803,7 +803,12 @@ static gboolean write_image_to_dev(RaucImage *image, RaucSlot *slot, GError **er
 	}
 
 	/* Try adaptive mode */
-	if (image->adaptive && slot->data_directory) {
+	if (image->adaptive) {
+		if (!slot->data_directory) {
+			g_message("Ignoring adaptive method since 'data-directory' is not configured");
+			goto raw_copy;
+		}
+
 		if (!copy_adaptive_image_to_dev(image, slot, &ierror)) {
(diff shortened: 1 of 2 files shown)


jluebbe (dismissed):

ejoerns:
Sorry for dismissing the review, but I thought the added patch would fit best in this PR :smirk:

src/update_handler.c
```diff
@@ -773,6 +773,8 @@ static gboolean copy_adaptive_image_to_dev(RaucImage *image, RaucSlot *slot, GEr
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
 	if (g_strv_contains((const gchar * const*)image->adaptive, "block-hash-index")) {
+		g_info("Selected adaptive update mode 'block-hash-index");
```
jluebbe, line 776: Perhaps

```suggestion
		g_info("Selected adaptive update method 'block-hash-index");
```

jluebbe (changes requested):

src/update_handler.c
ejoerns: Fun fact: I had this first but was then confused by ``No compatible adaptive mode found in '%s'`` that comes a few lines later.

We should make this consistent, I'll add another patch.

jluebbe (approved):

Issue #996, opened 2022-10-25
test/stats.test failure on i387


ukleinek:
Hello,
I packaged rauc 1.8 for Debian and the test suite failed on i386:
https://buildd.debian.org/status/fetch.php?pkg=rauc&arch=i386&ver=1.8-1&stamp=1664982505&raw=0

The failing test is:
`g_assert_cmpfloat(r_stats_get_avg(stats), ==, 62.54615384615385);`

The problem is excess precision of the i387 architecture which can be seen when adding the following printf statements before the test:

    printf("%15.50a\n", r_stats_get_avg(stats));
    printf("%15.50La\n", (long double)r_stats_get_avg(stats));

This results in the following output:

    0x1.f45e85e85e85f0000000000000000000000000000000000000p+5
    0xf.a2f42f42f42f42f00000000000000000000000000000000000p+2

The latter value is the same as:

    0x1.f45e85e85e85e8500000000000000000000000000000000000p+5

So while the first line using a `double` has 53 base-2 digits in the significand (as expected as `DBL_MANT_DIG` is 53 on i386), the value casted to `long double` has a higher precision with (at least) 64 base-2 digits.

This makes the comparison in the assert statement above wrong.

I'm unsure about the right way to fix this. The obvious options are:
 - use `g_assert_cmpfloat_with_epsilon`
 - pass `-fexcess-precision=standard` to gcc
 - drop that assert

jluebbe:
I'd prefer `g_assert_cmpfloat_with_epsilon`, as we don't care about the bit-exact value here.

PR #997, opened 2022-10-26
test/stats: make tests involving floating point calculations more lax

ukleinek:
On i387 floating point calculations can be quite surprising. The motivation for this change came from the observation that compiling and testing rauc-1.8 on the i386 Debian build machine had a failure on

	g_assert_cmpfloat(r_stats_get_avg(stats), ==, 62.54615384615385);

. In that case r_stats_get_avg(stats) (when stored in a double typed variable) returned

	0x1.f45e85e85e85fp+5

which is 0x1f45e85e85e85f00 / 2**55. However when stored in a long double typed variable it has the value

	0x1.f45e85e85e85e85p+5

This value is 0x1f45e85e85e85e85 / 2**55 and so it has a significand that cannot be stored in the 53 bits that are allocated for double typed variables. 62.54615384615385 is a plain double and so evaluates to 0x1.f45e85e85e85fp+5.

This makes the comparison implemented in g_assert_cmpfloat() (which uses long double precision) fail. To fix that, use
g_assert_cmpfloat_with_epsilon() and also adapt the other tests that involve non-trivial calculations to this approach.

Fixes: https://github.com/rauc/rauc/issues/996

### DIFF
--- a/test/stats.c
+++ b/test/stats.c
@@ -3,6 +3,23 @@
 
 #include "stats.h"
 
+/* APPROX_VALUE exists in glib >= 2.58 */
+#ifndef G_APPROX_VALUE
+#define G_APPROX_VALUE(a, b, epsilon) \
+	(((a) > (b) ? (a) - (b) : (b) - (a)) < (epsilon))
+#endif
+
+/* g_assert_cmpfloat_with_epsilon exists in glib >= 2.58 */
+#ifndef g_assert_cmpfloat_with_epsilon
+#define g_assert_cmpfloat_with_epsilon(n1, n2, epsilon) \
+	G_STMT_START { \
+		double __n1 = (n1), __n2 = (n2), __epsilon = (epsilon); \
+		if (G_APPROX_VALUE(__n1,  __n2, __epsilon)); else \
+		g_assertion_message_cmpnum(G_LOG_DOMAIN, __FILE__, __LINE__, G_STRFUNC, \
+		#n1 " == " #n2 " (+/- " #epsilon ")", __n1, "==", __n2, 'f'); \
+	} G_STMT_END
+#endif
+
 static void test_basic(void)
 {
 	g_autoptr(RaucStats) stats = NULL;
@@ -23,9 +40,14 @@ static void test_basic(void)
 	}
 
 	g_assert_cmpuint(stats->count, ==, 130);
-	g_assert_cmpfloat(r_stats_get_avg(stats), ==, 62.54615384615385);
-	g_assert_cmpfloat(r_stats_get_recent_avg(stats), ==, 95.5);
-	g_assert_cmpfloat(stats->sum, ==, 8131.0);
+
+	/*
+	 * Don't use an exact compare on non-trivial floating-point
+	 * calculations.
+	 */
+	g_assert_cmpfloat_with_epsilon(r_stats_get_avg(stats), 62.54615384615385, 1e-10);
+	g_assert_cmpfloat_with_epsilon(r_stats_get_recent_avg(stats), 95.5, 1e-10);
+	g_assert_cmpfloat_with_epsilon(stats->sum, 8131.0, 1e-10);
 	g_assert_cmpfloat(stats->min, ==, 0.0);
 	g_assert_cmpfloat(stats->max, ==, 127.0);
 }


jluebbe:
This breaks the build for glib versions < 2.58. Do we want to update the minimum version, backport the implementation of g_assert_cmpfloat_with_epsilon or implement the assert a different way?

ukleinek:
> This breaks the build for glib versions < 2.58. Do we want to update the minimum version, backport the implementation of g_assert_cmpfloat_with_epsilon or implement the assert a different way?

Not sure if the question is directed at me, but I don't have a strong opinion. With my Debian hat on: glib 2.58 isn't a problem, that version is even contained in buster (aka Debian 10 aka oldstable).

jluebbe:
glib 2.58 is included since:
- Yocto thud (2.58.0)
- Ubuntu focal (2.64.6), bionic has only 2.56.4 but is LTS until April 2023.

So the question become if we want to drop support for bionic. @ejoerns?

ejoerns:
Could be use a preprocessor switch to use the new comparison when glib is newer and the old one when glib is older?

Then we can support either 'old' platforms (glib < 2.58) or 'special' platforms (as i387). But not both at once.

So the only thing is that I would not like to bump the minimal version just because of a test (and not at least because of an added feature).

On the other hand, I would also not strongly insist on keeping the current glib version as the minimum...

Hasn't there been something we thought about breaking compatibility with ubuntu 18.04 for anyway?

ukleinek:
> glib 2.58 is included since:
> 
>     * Yocto thud (2.58.0)
> 
>     * Ubuntu focal (2.64.6), bionic has only 2.56.4 but is LTS until April 2023.

 *  ptxdist 2019.04.0 (2.60.0)

ukleinek:
> Maybe something like:
> 
> ```
> /* g_assert_cmpfloat_with_epsilon exists in glib >= 2.58 */
> #ifndef g_assert_cmpfloat_with_epsilon
> #define g_assert_cmpfloat_with_epsilon(n1, n2, epsilon) \
>     g_assert_true (abs (n1 - n2) < epsilon)
> #endif
> ```

While the documentation claims this was the equivalent expression this results in a warning:

    test/stats.c:10:24: warning: using integer absolute value function ‘abs’ when argument is of floating-point type ‘double’ [-Wabsolute-value]

There is `fabs(3)` which requires `#include <math.h>` and `-lm` though.

Alternatively:

```
/* g_assert_cmpfloat_with_epsilon exists in glib >= 2.58 */
#ifndef g_assert_cmpfloat_with_epsilon
#define g_assert_cmpfloat_with_epsilon(n1, n2, epsilon) \
    G_STMT_START { \
        double __d = n1 - n2;
        if (__d < 0)
            __d = -__d;
        g_assert_true (__d < epsilon)
    } G_STMT_END
#endif
```

ejoerns:
> 

This, or like the G_APPROX_VALUE from glib

``` c
#define G_APPROX_VALUE(a, b, epsilon) \
  (((a) > (b) ? (a) - (b) : (b) - (a)) < (epsilon))
```

The glib implementation probably has better error output, but maybe it's not worth porting the entire define.

``` c
#define g_assert_cmpfloat_with_epsilon(n1,n2,epsilon) \
    G_STMT_START { \
        double __n1 = (n1), __n2 = (n2), __epsilon = (epsilon); \
        if (G_APPROX_VALUE (__n1,  __n2, __epsilon)) ; else \
            g_assertion_message_cmpnum (G_LOG_DOMAIN, __FILE__, __LINE__, G_STRFUNC, \
              #n1 " == " #n2 " (+/- " #epsilon ")", __n1, "==", __n2, 'f'); \
    } G_STMT_END
```

jluebbe:
We've already back-ported code from glib in similar situations, so I'd prefer just picking both macros (`G_APPROX_VALUE` and `g_assert_cmpfloat_with_epsilon`) to test/common.h with a `#ifndef` protection.

ejoerns:
@ukleinek could you run the uncrustify script on this to make it match our style flavors? :smirk: 
Then I guess it's ready for being merged, thanks!

ejoerns (approved):
Thanks!

jluebbe (approved):

PR #999, opened 2022-10-27
src/utils.c: fix null reference error for g_propagate_prefixed_error

jungte:
It's my first PR. Please let me know if I did something wrong.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -370,6 +370,10 @@ gboolean r_read_exact(const int fd, guint8 *data, size_t size, GError **error)
 					"Failed to read: %s", g_strerror(err));
 			return FALSE;
 		} else if (ret == 0) { /* end of file */
+			g_set_error(error,
+					G_FILE_ERROR,
+					G_FILE_ERROR_FAILED,
+					"Unexpected end of file");
 			return FALSE;
 		} else if ((size_t)ret <= remaining) {
 			pos += ret;
@@ -426,6 +430,10 @@ gboolean r_pread_exact(const int fd, guint8 *data, size_t size, off_t offset, GE
 					"Failed to read: %s", g_strerror(err));
 			return FALSE;
 		} else if (ret == 0) { /* end of file */
+			g_set_error(error,
+					G_FILE_ERROR,
+					G_FILE_ERROR_FAILED,
+					"Unexpected end of file");
 			return FALSE;
 		} else if ((size_t)ret <= remaining) {
 			pos += ret;


ejoerns:
@jungte Thanks for the fix! Looks fine at first sight.

Please add your `Signed-off-by` line to the patch (`git commit --amend -s`): 

https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin

You can simply force-push your branch to update the PR then.

ejoerns:
The DCO check is not fully happy yet, since the mails do not match:

https://github.com/rauc/rauc/pull/999/checks?check_run_id=9133823272

Oh, still 1 PR to go...

jluebbe:
You don't need to close the PR, you can just force-push to replace the commit.

jungte:
OK, will try to do it later.

src/utils.c
```diff
@@ -370,6 +370,11 @@ gboolean r_read_exact(const int fd, guint8 *data, size_t size, GError **error)
 					"Failed to read: %s", g_strerror(err));
 			return FALSE;
 		} else if (ret == 0) { /* end of file */
+			int err = errno;
```
jluebbe, line 373: `errno` doesn't contain anything useful when pread returns 0, so we shouldn't try to add that to the error string.

src/utils.c
```diff
@@ -370,6 +370,11 @@ gboolean r_read_exact(const int fd, guint8 *data, size_t size, GError **error)
...
+			g_set_error(error,
+					G_FILE_ERROR,
+					g_file_error_from_errno(err),
+					"End of file: %s", g_strerror(err));
```
jluebbe, line 377: Perhaps
```suggestion
					"Unexpected end of file");
```

src/utils.c
```diff
@@ -426,6 +431,11 @@ gboolean r_pread_exact(const int fd, guint8 *data, size_t size, off_t offset, GE
 					"Failed to read: %s", g_strerror(err));
 			return FALSE;
 		} else if (ret == 0) { /* end of file */
+			int err = errno;
```
jluebbe, line 434: No `errno` here as well.

src/utils.c
```diff
@@ -426,6 +431,11 @@ gboolean r_pread_exact(const int fd, guint8 *data, size_t size, off_t offset, GE
...
+			g_set_error(error,
+					G_FILE_ERROR,
+					g_file_error_from_errno(err),
+					"End of file: %s", g_strerror(err));
```
jluebbe, line 438:
```suggestion
					"Unexpected end of file");
```

jluebbe (changes requested):

ejoerns:
@jungte would you like to incorporate the requested changes so we can merge this? Or should we?

jungte:
@ejoerns Sure, I can do it in next few days if you can wait. The holiday is getting over.. :)

src/utils.c
ejoerns: @jungte this still remains as an unsued variable in your patch now. If you fix this also (and the occurrence noted below) than the PR should be ready for being merged.

jungte: @ejoerns OK, how about this?

ejoerns:
I've updated the branch to resolve a code style issue reported by uncrustify

jluebbe (approved):

PR #1001, opened 2022-11-01
fix sporadic device mapper test failures

jluebbe:
Sporadically, the test cases for detecting modified data fail because we were
still able to read the original (un-modified) data from the bundle after the
backing file was manipulated. While it seems to happen mostly in GitHub
Actions, it is still reproducible in QEMU by looping over the specific test
cases.

Although we use fsync() and drop_caches, we still sometime read the unmodified data.
We tried to use sync() and modified timing in the test case, but this was not a
reliable fix.

Instead, we now open the dm devices with O_DIRECT to avoid interacting
with any caches. As reading from O_DIRECT FDs requires page-aligned buffers, we
need to use aligned_alloc() instead of placing the buffer on the stack.

### DIFF
--- a/test/dm.c
+++ b/test/dm.c
@@ -59,14 +59,6 @@ static void flip_bits_filename(gchar *filename, off_t offset, guint8 mask)
 	g_close(fd, NULL);
 }
 
-static void drop_caches(void)
-{
-	int fd = g_open("/proc/sys/vm/drop_caches", O_WRONLY|O_CLOEXEC, 0);
-	g_assert_cmpint(fd, >, 0);
-	g_assert_true(write(fd, "1", 1) == 1);
-	g_close(fd, NULL);
-}
-
 static guint readable_sectors(int fd, GBytes *original)
 {
 	g_autofree guint8 *buf = NULL;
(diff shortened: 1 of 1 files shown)


jluebbe:
The scan-build failure is not cause by this PR. I'll create a separate one to fix it.

test/dm.c
```diff
@@ -58,53 +59,61 @@ static void flip_bits_filename(gchar *filename, off_t offset, guint8 mask)
...
-
-static guint readable_sectors(int fd)
-{
-	guint8 buf[4096];
+	g_autofree guint8 *buf = NULL;
 	ssize_t r;
 	guint sectors = 0;
 
+	/* O_DIRECT needs paged-aligned memory */
+	buf = aligned_alloc(4096, 4096);
```
a3f, line 69: RAUC configures `AC_PROG_CC_C99`, but `aligned_alloc` is C11. Alternative would be `posix_memalign` (or mandate C11?).

a3f: Apparently, you need a quite recent autotools to be able to configure C11 out of the box. Given that we want to move to meson, it may not be worth it opencoding something at our side. We don't have an error using `aligned_alloc` because autotools generated config.h has `#define _GNU_SOURCE`, which loosens the requirement that headers only export symbols mandated by the standard as not to clash with user-defined names. `aligned_alloc` has been part of glibc since v2.16 released in 2012, so I guess we can just stay with the implicit assumption that everyone running the test has a recent enough glibc and when moving to Meson, this can become explicit by mandating C11.

Bastian-Krause (approved):

PR #1018, opened 2022-11-25
docs: document our approach to bundle compatibility

jluebbe:
We had some questions on how we handle forward and backward compatibility, so we should document our approach explicitly.

### DIFF
--- a/docs/basic.rst
+++ b/docs/basic.rst
@@ -67,6 +67,58 @@ Using streaming has a few requirements:
 See the :ref:`HTTP Streaming <http-streaming>` section in the Advanced chapter
 for more details.
 
+.. _sec-compatibility:
+
+Forward and Backward Compatibility
+----------------------------------
+
+Our overall goal with regards to compatibility is a good balance between the
+requirements of users and the constraints during development.
+For users, it is mainly relevant how a given version of RAUC on the target
+handles bundles produced by older (backward compatibility) and newer versions
+(forward compatibility) of RAUC.
+As developers, we want to keep the effort for supporting old versions in the
+field at a reasonable level and have the flexibility to improve RAUC with new
+versions.
+
+To ensure forward compatibility, new bundle features need to be enabled
+explicitly during bundle creation.
+So without changing the manifest, newer RAUC versions used for bundle creation
+will not require new versions on the target.
+This includes new bundle formats, new hooks, adaptive updates or additional
+metadata.
+When a new (incompatible) feature is enabled in a bundle, older RAUC versions
+will report an error during installation to ensure that the installation result
+is deterministic.
+As long as you don't enable new features during creation, our intention is that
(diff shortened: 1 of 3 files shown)


docs/basic.rst
```diff
@@ -67,6 +67,58 @@ Using streaming has a few requirements:
...
+This includes new bundle formats, new hooks, adaptive updates or additional
+metadata.
+When a new (incompatible) feature is enabled in a bundle, older RAUC versions
+will report an error during installation to ensure that the installation result
+is deterministic.
+As long as you don't enable new features during creation, our intention is that
+bundles created by newer versions will be installable by older versions and any
+such issues would be considered a bug.
+
+To ensure backward compatibility, support for older bundle features is enabled
```
ejoerns, line 97: Currently, this is valid for bundle *formats* only. Or what kind of features do you have in mind that we can disable in `system.conf`?

docs/basic.rst
```diff
@@ -67,6 +67,58 @@ Using streaming has a few requirements:
...
+by default and can be disabled explicitly in the RAUC ``system.conf`` as
+needed.
+To keep RAUC maintainable, we may need to deprecate and later remove support
+for old features over time.
+This would be done with several years between deprecation and removal so that
+at least one Yocto LTS version contains a RAUC version that warns when using
+the deprecated feature, giving users enough time to migrate away from that
+feature.
+Any issues with installing old bundles using new RAUC versions would be
```
ejoerns, line 106: Should we be explicit and talk about *old bundle formats* here? Because the age of the bundle should not matter primarily :wink:

ejoerns:
Should we also add a statement on stable releases here? (Or in a future PR?)
Just as it probably is a direct consequence of the compatibility promises made here.

Might also make sense to add a FAQ entry about our stable release policy since this is an explicit question that comes up from time to time.

docs/basic.rst
jluebbe: Formats are just a special case of features. The age should not matter, though.

jluebbe: I've reworded this a bit.

jluebbe:
> Should we also add a statement on stable releases here? (Or in a future PR?) Just as it probably is a direct consequence of the compatibility promises made here.
> 
> Might also make sense to add a FAQ entry about our stable release policy since this is an explicit question that comes up from time to time.

It'd do that in a separate PR.

docs/basic.rst
ejoerns: To me this sound more like 'explicitly enabling new features' rather than 'explicitly disabling old features' but yes, since I also cannot really find good examples for features that have been deprecated, the (quite general) doc comment may still apply.

docs/basic.rst
ejoerns: Sounds better I guess

Issue #829, opened 2022-01-26
Add meson build configuration


jluebbe:
With [Meson](https://mesonbuild.com/), we should be able to get rid of some of the complexity we have currently. By now, it is supported well enough in build systems and distributions. We can have a transition period where we support both Meson and autotools.

ejoerns:
I've prepared an initial proof of concept: https://github.com/ejoerns/rauc/commits/meson

It is capable of compiling and release-packaging (`ninja dist`) RAUC without errors (currently requires meson >= 0.51).

Missing:

* [ ] testing support

ejoerns:
I have now added the missing testing capabilities (which was, in some details, more challenging than expected) and adapted CI tests to use meson to build and run the test suite.

I have created PR #1021 for adding meson support now.

PR #1021, opened 2022-12-12
Add Meson Build Support 

ejoerns:
This adds support for building RAUC with meson build system.

It also adapts CI tests to use meson instead of autotools.

Fixes: #829

### DIFF
--- a/.github/workflows/scan-build.yml
+++ b/.github/workflows/scan-build.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Install dependencies
       run: |
         apt-get update
-        DEBIAN_FRONTEND='noninteractive' apt-get install -qy automake make libtool libglib2.0-dev libcurl3-dev libssl-dev libdbus-1-dev libjson-glib-dev libfdisk-dev libnl-genl-3-dev dbus-x11 clang-tools
+        DEBIAN_FRONTEND='noninteractive' apt-get install -qy meson libtool libglib2.0-dev libcurl3-dev libssl-dev libdbus-1-dev libjson-glib-dev libfdisk-dev libnl-genl-3-dev dbus-x11 clang-tools
 
     - name: Inspect environment
       run: |
@@ -22,14 +22,13 @@ jobs:
 
     - uses: actions/checkout@v2
 
-    - name: Run autogen.sh
+    - name: Run meson
       run: |
-        ./autogen.sh
+        meson setup build
 
     - name: Run scan-build
       run: |
-        scan-build --status-bugs ./configure --disable-dependency-tracking
-        scan-build --status-bugs -disable-checker unix.Malloc make TESTS= check
+        SCANBUILD=$(pwd)/scan-build-wrapper.sh ninja -C build scan-build
 
     - name: Upload scan-build report
       if: failure()
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,28 +14,24 @@ jobs:
       run: |
         whoami
         gcc --version
+        apt-get update
+        DEBIAN_FRONTEND='noninteractive' apt-get install -qy meson
 
     - uses: actions/checkout@v2
 
-    - name: Run autogen.sh
-      run: |
-        ./autogen.sh
-
     - name: Test with service and gpt
       run: |
-        ./configure --enable-code-coverage --enable-gpt CFLAGS=-Werror
-        make clean
-        make -j4
-        make -j4 check TESTS=
+        rm -rf build/
+        meson setup -Dservice=true -Dgpt=enabled -Dwerror=true build
+        ninja -C build
         ./qemu-test
         lcov --directory . --capture --output-file "service.info"
         
     - name: Test without service and gpt
       run: |
-        ./configure --enable-code-coverage --disable-service CFLAGS=-Werror
-        make clean
-        make -j4
-        make -j4 check TESTS=
+        rm -rf build/
+        meson setup -Dservice=false -Dgpt=enabled -Dwerror=true build
+        ninja -C build
         ./qemu-test
         lcov --directory . --capture --output-file "noservice.info"
 
@@ -46,7 +42,7 @@ jobs:
 
     - name: Build documentation
       run: |
-        make doc SPHINXOPTS=-W
+        meson compile doc -C build
 
     - name: Build CGI example
       run: |
@@ -69,10 +65,7 @@ jobs:
     - name: Show logs
       if: ${{ failure() }}
       run: |
-        cat config.log || true
-        cat test/*.log || true
-        cat test-suite.log || true
-        cat rauc-*/_build/sub/test-suite.log || true
+        cat build/meson-logs/testlog.txt || true
         dmesg | tail -n 100
 
   cross:
@@ -95,23 +88,19 @@ jobs:
         docker logs cross
         docker exec -i cross uname -a
         docker exec -i cross apt-get update
-        docker exec -e DEBIAN_FRONTEND='noninteractive' -i cross apt-get install -qy build-essential automake libtool libglib2.0-dev libcurl3-dev libssl-dev libjson-glib-dev libdbus-1-dev libfdisk-dev libnl-genl-3-dev squashfs-tools
+        docker exec -e DEBIAN_FRONTEND='noninteractive' -i cross apt-get install -qy build-essential meson libtool libglib2.0-dev libcurl3-dev libssl-dev libjson-glib-dev libdbus-1-dev libfdisk-dev libnl-genl-3-dev squashfs-tools
 
     - name: Build
       run: |
         docker exec -i cross whoami
-        docker exec -i cross ./autogen.sh
-        docker exec -i cross ./configure --enable-gpt CFLAGS=-Werror
-        docker exec -i cross make
+        docker exec -i cross meson setup -Dgpt=enabled -Dwerror=true build
+        docker exec -i cross ninja -C build
         # don't run make check here, as we don't have full access to the kernel (mount, loopback, dm)
 
     - name: Show logs
       if: ${{ failure() }}
       run: |
-        cat config.log || true
-        cat test/*.log || true
-        cat test-suite.log || true
-        cat rauc-*/_build/sub/test-suite.log || true
+        cat build/meson-logs/testlog.txt || true
 
   stable:
     runs-on: ubuntu-latest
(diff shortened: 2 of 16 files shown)


jluebbe:
@ejoerns What needs to be done before removing the draft status?

ejoerns:
> @ejoerns What needs to be done before removing the draft status?

I left it on draft to possibly get some high-level feedback about possible major things I missed but maybe making it an actual PR is more appropriate for getting feedback.

meson_options.txt
```diff
@@ -0,0 +1,67 @@
...
+option(
+  'service',
+  type : 'boolean',
+  value : 'true',
+  description : 'Enable/Disable background service')
+option(
+  'create',
+  type : 'boolean',
+  value : 'true',
+  description : 'Disable bundle creation and modification commands')
```
jluebbe, line 11:
```suggestion
  description : 'Enable/Disable bundle creation and modification commands')
```

meson_options.txt
```diff
@@ -0,0 +1,67 @@
...
+
+
+option(
+  'streaming_user',
+  type : 'string',
+  value : 'nobody',
+  description : 'unpriviledged user for the streaming subprocess')
+
+
+
```
jluebbe, line 41: Fewer empty lines?

qemu-test
```diff
@@ -45,9 +45,6 @@ check_kernel() {
 
 check_kernel || exit 1
 
-# make sure the tests are built
-make TESTS= check > /dev/null
-
```
jluebbe, line 50: Why can't we adapt this to meson as well?

jluebbe (commented):
We should probably keep a/some GitHub workflow(s) to check that autotools still work (until we actually remove it).

qemu-test
ejoerns: Tests are currently either build or not by default, depending on whether "-Dtests=true" was configured or not. Thus it is less likely to have sources built without having tests built.

One could maybe invoke meson/ninja from within the script, but then we also need to know where the build dir is which is already a bit hacky solved in `qemu-test-init`.

meson_options.txt
ejoerns: done and replaced by separating comments

ejoerns:
> We should probably keep a/some GitHub workflow(s) to check that autotools still work (until we actually remove it).

I removed the autotools workflows in the first round since I was not sure at this point if we wanted to keep autotools or not. But if we keep it we should also keep some of the workflows, yes.

qemu-test
jluebbe: Don't we need to know the build dir anyway to find the binaries?

jluebbe:
> > We should probably keep a/some GitHub workflow(s) to check that autotools still work (until we actually remove it).
> 
> I removed the autotools workflows in the first round since I was not sure at this point if we wanted to keep autotools or not. But if we keep it we should also keep some of the workflows, yes.

I think we should keep it for one release, at least.

qemu-test
jluebbe: The intention was to avoid building in qemu (which is potentially slow when running without kvm). As we hardcode `BUILD_DIR` anyway, we could continue to build the test binaries in `qemu-init`.

meson.build
```diff
@@ -84,7 +89,10 @@ c_warn_flags = '''
 # To allow building against OpenSSL 3.0, we still need to allow deprecated
 # declarations
 c_warn_flags += '-Wno-error=deprecated-declarations'
-add_project_arguments(c_warn_flags, language : 'c')
+c_flags += cc.get_supported_arguments(c_warn_flags)
+c_flags += ['-fdata-sections', '-ffunction-sections', '-fno-strict-aliasing']
+c_flags += ['-Wl,--gc-sections', '-Wl,-Map,$@.map']
```
jluebbe, line 94: These are LD flags

qemu-test-init
```diff
@@ -11,19 +11,12 @@ mount -t tracefs tracefs /sys/kernel/tracing/ || true
...
 BUILD_DIR="build/"
 
 # parse cmdline
 for x in $(cat /proc/cmdline); do
   if [ "$x" = "shell" ]; then
     SHELL=1
-  elif [ "$x" = "coverage" ]; then
-    MAKE_TARGET="check-code-coverage"
-  elif [ "$x" = "distcheck" ]; then
-    MAKE_TARGET="distcheck"
```
jluebbe, line 23: We shouldn't remove these util we drop autotools support.

.github/workflows/scan-build.yml
```diff
@@ -22,14 +22,13 @@ jobs:
 
     - uses: actions/checkout@v2
 
-    - name: Run autogen.sh
+    - name: Run meson
       run: |
-        ./autogen.sh
+        meson setup build -Dtests=true
```
jluebbe, line 27: Tests are enabled by default.

.github/workflows/tests.yml
```diff
@@ -14,28 +14,24 @@ jobs:
...
-        ./autogen.sh
-
     - name: Test with service and gpt
       run: |
-        ./configure --enable-code-coverage --enable-gpt CFLAGS=-Werror
-        make clean
-        make -j4
-        make -j4 check TESTS=
+        rm -rf build/
+        meson setup -Dservice=true -Dgpt=enabled -Dtests=true -Dwerror=true build
```
jluebbe, line 25: Tests are enabled by default. Check below as well.

.github/workflows/tests.yml
```diff
@@ -127,30 +116,34 @@ jobs:
...
-        docker exec -e DEBIAN_FRONTEND='noninteractive' -i stable apt-get install -qy build-essential automake libtool libglib2.0-dev libcurl3-dev libssl-dev libjson-glib-dev libdbus-1-dev libfdisk-dev libnl-genl-3-dev squashfs-tools
+        docker exec -e DEBIAN_FRONTEND='noninteractive' -i stable apt-get install -qy build-essential meson libtool libglib2.0-dev libcurl3-dev libssl-dev libjson-glib-dev libdbus-1-dev libfdisk-dev libnl-genl-3-dev squashfs-tools
+
+    - name: Install meson > 0.45 for ubuntu 18.04
+      if: matrix.release == '18.04'
+      run: |
+        docker exec -e DEBIAN_FRONTEND='noninteractive' -i stable apt-get install -qy software-properties-common
+        docker exec -i stable add-apt-repository ppa:team-xbmc/ppa
+        docker exec -i stable apt-get update
+        docker exec -e DEBIAN_FRONTEND='noninteractive' -i stable apt-get install -qy meson
```
jluebbe, line 127: We should just disable meson tests for 18.04.

meson.build
```diff
@@ -4,10 +4,9 @@ project(
   version : '1.8',
   meson_version : '>=0.51',
   default_options: [
-    'c_std=c99',
```
jluebbe: We could set 'c_std=gnu11' to pin the standard version.

meson.build
```diff
@@ -0,0 +1,218 @@
...
+  interface_prefix : 'de.pengutronix.rauc.',
+  namespace : 'R',
+  install_header : true,
+)
+
+config_h = configure_file(
+  output : 'config.h',
+  configuration : conf
+)
+add_project_arguments('-include' + meson.current_build_dir() / 'config.h', language : 'c')
```
jluebbe, line 164: Can't we use the file object here?

qemu-test-init
ejoerns: We do not call `make` anymore in the script, thus these are useless anyway

meson.build
ejoerns: I attempted to do so but this does not support obtaining the (absolute) file name as a string (which would be required here)

ejoerns:
I've addressed the feedback so far

README.rst
```diff
@@ -172,16 +172,26 @@ Building from Sources
 ---------------------
 
 .. note:: RAUC is intended to be built both as a host tool as well as a target
-   tool (service). Therefore it is fully prepared for `automake cross-compilation
-   <https://www.gnu.org/software/automake/manual/html_node/Cross_002dCompilation.html>`_
+   tool (service). Therefore it is fully prepared for meson.
```
jluebbe, line 175:
```suggestion
   tool (service). Therefore it is fully prepared for cross-compilation with meson.
```

README.rst
```diff
@@ -172,16 +172,26 @@ Building from Sources
...
 
     git clone https://github.com/rauc/rauc
     cd rauc
-    ./autogen.sh
-    ./configure
-    make
+    meson setup build
+    ninja -C build
+
+.. note:: At the moment also `automake cross-compilation
```
jluebbe, line 184:
```suggestion
.. note:: At the moment, also `automake cross-compilation
```

README.rst
```diff
@@ -172,16 +172,26 @@ Building from Sources
...
+   <https://www.gnu.org/software/automake/manual/html_node/Cross_002dCompilation.html>`_
+   is still supported, but the plan is to deprecate this in the new future
```
jluebbe, line 186:
```suggestion
   is still supported, but the plan is to deprecate this soon.
```

jluebbe (changes requested):

ejoerns:
I've fixed some aspects of the scan-build invocation now.

* Added missing dependency of executable target 'rauc' to `dbus_sources` to ensure having them always generated before using them
* added missing comma-separation of `-include` and `/path/to/config.h` in `add_project_arguments()` which seems to have caused ignorance of config.h in scan-build run
* use meson's built-in `scan-build` target
* add wrapper script `scan-build-wrapper.sh` to allow disabling `unix.Malloc` checker and injecting `--status-bugs` for meson scan-build target

ejoerns:
Timeout of `rauc.t` seems to have been insufficient. Increased to `300` now. Next try...

scan-build-wrapper.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+scan-build -v --status-bugs -disable-checker unix.Malloc "$@"
```
jluebbe, line 3: Why do we want to disable `unix.Malloc`?

meson.build
jluebbe: In https://github.com/rauc/rauc/compare/4478e169d2ab51e18a47b58339496585e3a16dbe..5d566674bd024218bdd06bde7ab36a44ffa2b32b#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4L7, this was removed again?

ejoerns: Oups, rebase fail. Fixed.

scan-build-wrapper.sh
ejoerns: because scan-build does not understand auto-cleanup. This has been disabled before, too (in the workflow file).

jluebbe (changes requested):
The "meson: add tests" commit is still worded as if tests were disabled by default. LGTM otherwise.

ejoerns:
> The "meson: add tests" commit is still worded as if tests were disabled by default. LGTM otherwise.

I have adapted the commit message.

The only thing missing from my perspective would be coverage support, but I could also add this in a follow-up PR.

jluebbe (approved):

PR #1026, opened 2022-12-21
manifest: parse and store metadata

jluebbe:
Since 3cc11ecd34a1e03e99c8798fd1e9bd1a976cbf94, RAUC ignores `[meta.*]` sections in the manifest. With this PR, the metadata entries are parsed and written back to the manifest during bundle creation.

Still missing is support for passing the entries to `rauc info`, the D-Bus API and hook/handlers.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -1,3 +1,5 @@
+#include <string.h>
+
 #include "checksum.h"
 #include "config_file.h"
 #include "context.h"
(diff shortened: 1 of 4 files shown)


src/manifest.c
```diff
@@ -94,6 +96,50 @@ static gboolean parse_image(GKeyFile *key_file, const gchar *group, RaucImage **
...
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	groupsplit = g_strsplit(group, ".", 2);
+	if ((g_strv_length(groupsplit) != 2) || strchr(groupsplit[1], '.')) {
+		g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_PARSE,
+				"invalid metadata section name '%s' (must contain a single '.')", group);
+		return FALSE;
+	}
+
+	groupname = g_intern_string(groupsplit[1]);
```
ejoerns, line 118: What is the purpose of using interned strings here?

To my understanding, the primary use of interned strings is for simplifying string comparison which I do not see apply here.

Is this to omit manual string allocation and freeing?
Since interned strings are allocated in global scope, the allocation overhead should increase with each bundle being opened during the process lifetime.

ejoerns: Mh, I don't see how this is different from all other places in the code where strings are handled.

I'd suggest adding at least a note in the code that intern strings are used for not having to write a free function for `RManifestMetaEntry`.

jluebbe: I've changed it to use dynamic allocations for the strings.

ejoerns (approved):

jluebbe:
The cross-build test failures are unrelated.

PR #1030, opened 2022-12-29
README.rst in building instructions autoupdate added

robang74:
In 'Building from Sources' section 'autoupdate' command before all has been added in such a way autogen.sh will be updated accordingly with the autotools installed into the system.

Signed-off-by: Roberto A. Foglietta <roberto.foglietta@gmail.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -179,6 +179,7 @@ Building from Sources
 
     git clone https://github.com/rauc/rauc
     cd rauc
+    autoupdate
     ./autogen.sh
     ./configure
     make


jluebbe:
Is there a specific issue that's fixed by this change? If not, perhaps the changes produced by `autoupdate` should done in the source?

ejoerns:
I assume this might be related to #1032 ?

ejoerns:
For now, I do not see much benefit in adding the `autoupdate` to docs and since this did not get further attention and since we move to `meson` anyway I would close this now.

Issue #1008, opened 2022-11-08
Adaptive update fails when image is too large


landerweit-phytec:
I'm currently trying out the adaptive update feature with ext4 rootfs partitions.
The update works as expected when the ext4 partition in the bundle and the rootfs slot have the same size.
But when the size differs RAUC crashes with `g_propagate_error: assertion 'src != NULL' failed` instead of a meaningful error message. 
A check for a size mismatch could probably prevent this.

```
rauc install /tmp/phytec-headless-bundle-phyboard-segin-imx6ul-7.raucb
installing
 0% Installing
 0% Determining slot states
20% Determining slot states done.
20% Checking bundle
20% Verifying signature
40% Verifying signature done.
40% Checking bundle done.
40% Checking manifest contents
60% Checking manifest contents done.
60% Determining target install group
80% Determining target install group done.
80% Updating slots
80% Checking slot rootfs.0
83% Checking slot rootfs.0 done.
83% Copying image to rootfs.0
RAUC service disappeared
D-Bus error while installing `/tmp/phytec-headless-bundle-phyboard-segin-imx6ul-7.raucb`
```
Output of journalctl -u rauc
```
Apr 28 17:57:40 phyboard-segin-imx6ul-7 systemd[1]: Starting RAUC Update Service...
Apr 28 17:57:40 phyboard-segin-imx6ul-7 rauc[363]: Using central status file /mnt/config/rauc/central.raucs
Apr 28 17:57:40 phyboard-segin-imx6ul-7 rauc[363]: Using central status file /mnt/config/rauc/central.raucs
Apr 28 17:57:40 phyboard-segin-imx6ul-7 systemd[1]: Started RAUC Update Service.
Apr 28 17:57:40 phyboard-segin-imx6ul-7 rauc[363]: input bundle: /tmp/phytec-headless-bundle-phyboard-segin-imx6ul-7.raucb
Apr 28 17:57:40 phyboard-segin-imx6ul-7 rauc[363]: Active slot bootname: system1
Apr 28 17:57:40 phyboard-segin-imx6ul-7 rauc[363]: installing /tmp/phytec-headless-bundle-phyboard-segin-imx6ul-7.raucb: started
Apr 28 17:57:40 phyboard-segin-imx6ul-7 rauc[363]: installing /tmp/phytec-headless-bundle-phyboard-segin-imx6ul-7.raucb: Checking and mounting bundle...
[...]

PR #1037, opened 2023-01-06
Check image size before starting installation

jluebbe:
In some cases, we can abort early if the image in the bundle is too large for the slot. This is not possible when installing tar archives or when writing images to slot which point to a file.

Fixes: #1008

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -30,6 +30,34 @@ GQuark r_update_error_quark(void)
 	return g_quark_from_static_string("r_update_error_quark");
 }
 
+static gboolean check_image_size(int fd, const RaucImage *image, GError **error)
+{
+	GError *ierror = NULL;
+	goffset dev_size;
+
+	g_return_val_if_fail(image, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	dev_size = get_device_size(fd, &ierror);
+	if (g_error_matches(ierror, R_UTILS_ERROR, R_UTILS_ERROR_INAPPROPRIATE_IOCTL)) {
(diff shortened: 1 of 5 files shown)


src/utils.c
```diff
@@ -504,3 +504,30 @@ guint get_sectorsize(gint fd)
 
 	return sector_size;
 }
+
+goffset get_device_size(gint fd, GError **error)
+{
+	u_int64_t size = 0;
```
ejoerns, line 510: Why not use glib type `guint64` here? An alternative would be using the original type from the `ioctl()` prototype which is `unsigned long`.

src/update_handler.c
```diff
@@ -30,6 +30,31 @@ GQuark r_update_error_quark(void)
 	return g_quark_from_static_string("r_update_error_quark");
 }
 
+static gboolean check_image_size(int fd, const RaucImage *image, GError **error)
```
ejoerns, line 33: `image` and `error` arguments should get input value checks.

test/update_handler.c
```diff
@@ -27,6 +27,7 @@ typedef enum {
 	TEST_UPDATE_HANDLER_NO_HOOK_FILE                                  = BIT(7),
 	TEST_UPDATE_HANDLER_HOOK_FAIL                                     = BIT(8),
 	TEST_UPDATE_HANDLER_INCR_BLOCK_HASH_IDX                           = BIT(9),
+	TEST_UPDATE_HANDLER_IMAGE_TOO_LARGE                            = BIT(10),
```
ejoerns, line 30: Alignment looks broken here.

ejoerns (commented):
Just some minor notes, the rest looks fine to me. :+1:

PR #1038, opened 2023-01-10
add descriptions in dbus api

mansk789:
Add support for reading the description from a rauc bundle over the D-Bus-API.
Enhanced the Info() method with the necessary parameter


jluebbe:
This change would break compatibility with existing clients. A related discussion was in https://github.com/rauc/rauc/pull/965#discussion_r983520569.

PR #1039, opened 2023-01-10
src/nbd: be lenient with libcurl without proxy support

sirhcel:
In streaming mode, RAUC currently assumes that libcurl is built with proxy suport. When configuring libcurl, it [tries to set `CURLOPT_HTTPPROXYTUNNEL`](https://github.com/rauc/rauc/blob/315d12bb37b4b0acf3ad2859cbc21816bd8a3e0a/src/nbd.c#L442) which will fail with libcurl without proxy support with `CURLE_UNKNOWN_OPTION` (48).

In a brief discussion, @jluebbe suggested to just ignore this case and give a hint in debug output.

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -403,6 +403,7 @@ static size_t header_cb(char *buffer, size_t size, size_t nitems, void *userdata
 static void prepare_curl(struct RaucNBDTransfer *xfer)
 {
 	CURLcode code = 0;
+	CURLcode tunnel_code = 0;
 	g_assert_null(xfer->easy);
 
 	xfer->easy = curl_easy_init();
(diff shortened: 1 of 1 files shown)


sirhcel:
> Is `CURLOPT_SUPPRESS_CONNECT_HEADERS` still supported by curl without enabled proxy support?

Yes, at least with curl 7.82.0. I've tested this based on RAUC 1.7 (https://github.com/chargeITmobility/rauc/commit/8d6017a477f98c4cd3ffe1219e53290e0534ca6d) on a target where setting the option above returns zero. It looks like this option is [always enabled](https://github.com/curl/curl/blob/curl-7_82_0/lib/setopt.c#L2900).

jluebbe:
I've added a commit message and changed the debug text a bit: https://github.com/jluebbe/rauc/commit/94291be55ca944c3713a646f95191aba6b19dec3

Could you take that version for this PR? I'd have pushed directly to your branch, but you have disabled that. :)

sirhcel:
> I've added a commit message and changed the debug text a bit: [jluebbe@94291be](https://github.com/jluebbe/rauc/commit/94291be55ca944c3713a646f95191aba6b19dec3)

Thank you! Updated.


> Could you take that version for this PR? I'd have pushed directly to your branch, but you have disabled that. :)

Sure thing. I wanted to enable this option but it seems not to be available for repos of an organization account.

jluebbe (approved):

ejoerns (approved):

PR #1044, opened 2023-01-13
minor cleanups

jluebbe:
Clean up a few things found during work on artifact support.


src/main.c
```diff
@@ -1036,6 +1046,8 @@ static gchar* info_formatter_json_base(RaucManifest *manifest, gboolean pretty)
 		json_builder_begin_object(builder);
 		json_builder_set_member_name(builder, img->slotclass);
 		json_builder_begin_object(builder);
+		json_builder_set_member_name(builder, "artifact");
```
ejoerns, line 1049: Is this part of the PR on purpose?

jluebbe: No.

ejoerns: new feature preview :smirk:

src/main.c
```diff
@@ -998,6 +998,16 @@ static gchar *info_formatter_readable(RaucManifest *manifest)
 	return g_string_free(text, FALSE);
 }
 
+static void strv_to_json_array(JsonBuilder *builder, GStrv strv)
```
ejoerns, line 1001: Adding input guards for `builder` would also implicitly document that `strv` can be `NULL`.

```
g_return_if_fail(JSON_IS_BUILDER(builder));
```

(rest looks fine)

jluebbe: OK, I've also added the missing `#if` guard for JSON support.

ejoerns (approved):

PR #811, opened 2021-11-12
Emit fine grained progress to dbus for most update handlers

poeschel:
This emits progess through dbus for most update handlers.
The PR begins with add this to only on `untar_image` but then inserts the developed code into more functions. In the end most handlers are covered and should emit a more fine grained progress for them.
As I see it these handlers should now be covered:
* `img_to_fs_handler`
* `img_to_raw_handler`
* `img_to_ubivol_handler`
* `img_to_ubifs_handler`
* `archive_to_ext4_handler`
* `archive_to_ubifs_handler`
* `archive_to_vfat_handler`
* `img_to_nand_handler`

I also tried on `img_to_nor_handler` but it uses `flashcp` and `flashcp` can not read from stdin.
Progress is emitted for the 'main work' that is copying files to the destination slot only. It does not emit progress for other things that happen like filesystem creation.

Dear reviewer, please be kind to me. I never used glib before. glib is really hard for me. Thank you!

PR #1051, opened 2023-01-26
Create test tar image with bz2 compression

sirhcel:
* This is a change from @poeschel factored out from #811 
* It actually provides a bzip2 compressed image to a test expecting one
    > The test_prepare_dummy_archive creates a tar image for some tests, that pretends to be a .tar.bz2 image (based on file extension) but in reality it was just a .tar image without any compression. It would be good if it actually IS a bz2 image.
* So this change is not strictly required for #811 an easier to chew if bitten off alone

### DIFF
--- a/test/update_handler.c
+++ b/test/update_handler.c
@@ -156,7 +156,7 @@ static void update_handler_fixture_tear_down(UpdateHandlerFixture *fixture,
 	}
 }
 
-static gboolean tar_image(const gchar *dest, const gchar *dir, GError **error)
+static gboolean tar_bz2_image(const gchar *dest, const gchar *dir, GError **error)
 {
 	GSubprocess *sproc = NULL;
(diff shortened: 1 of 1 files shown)


ejoerns:
@sirhcel Thanks for splitting this out, but I think it is not really needed since the method does not promise to generate a compressed archive and I do not see much benefit from this. Instead it might take even longer to run.

One could maybe change the using test code and match matrix to not refer to "tar.bz2" anymore but simply use "tar".

ejoerns:
> One could maybe change the using test code and match matrix to not refer to "tar.bz2" anymore but simply use "tar".

FTR: in #1093 I have reworked this now by using the approach I've mentioned here.

sirhcel:
Thank you very much for cleaning this up! There is no longer a need for generating a tar.bz2 archive in the tests.

PR #1052, opened 2023-01-27
Add download sub step for installation progress

sirhcel:
* This is another change in the sense of #811 from @Urist-McGit i'd like to submit on behalf of him
    > Add an additional substep during the `check_bundle` installation step, if a download is required. Update the download progress during the transfer.
    > 
    > This will improve the user experience on slow connections since the installation progress does not seem to be stuck on `Checking bundle` and instead shows `Downloading bundle`.
* This new sub step is takes effect for both, classic and casync bundles (although it matters less in the latter case)

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -1891,13 +1891,19 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, CheckBundleP
 	g_autoptr(GBytes) manifest_bytes = NULL;
 	gchar *bundlescheme = NULL;
 	gboolean detached;
+	gboolean remote_scheme;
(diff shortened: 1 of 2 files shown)


ejoerns:
I am unsure whether we should introduce a download substep here.

The download mode is actually only meant to be used with casync. This is why it is limited to a few MB by default.

For full downloads we have RAUC's HTTP streaming support which forks an unprivileged helper process handling the download. However, for streaming there is no extra download step since downloading and installation happens in parallel.

sirhcel:
That non-streaming full downloads are not officially supported is a good point. In this case its maybe better to not break other people's expectations about the progress and maintain this as an internal patch.

PR #1058, opened 2023-02-10
Documentation and usage updates around signing parameters (primarily --intermediate)

ejoerns:
* Notes that `--intermediate` supports PKCS#11
* Makes `--intermediate` command-specific
* Adds help arg text
* Adds missing note of `RAUC_KEY_PASSPHRASE` in help
* Updates man
* Updates help text
* Updates doc

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -910,30 +910,30 @@ Command Line Tool
 .. code-block:: man
 
   Usage:
-    rauc [OPTION…] <COMMAND>
+    rauc [OPTION?] <COMMAND>
 
   Options:
-    -c, --conf=FILENAME               config file
-    --cert=PEMFILE|PKCS11-URL         cert file or PKCS#11 URL
-    --key=PEMFILE|PKCS11-URL          key file or PKCS#11 URL
-    --keyring=PEMFILE                 keyring file
-    --intermediate=PEMFILE            intermediate CA file name
-    --mount=PATH                      mount prefix
-    --override-boot-slot=BOOTNAME     override auto-detection of booted slot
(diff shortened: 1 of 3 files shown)


jluebbe:
`--intermediate` only needs the certificate. Why would one store that in an HSM?

docs/reference.rst
```diff
@@ -910,33 +910,33 @@ Command Line Tool
...
-    write-slot            Write image to slot and bypass all update logic
+    bundle		Create a bundle
+    resign		Resign an already signed bundle
+    convert		Convert classic to casync bundle
+    encrypt		Encrypt a crypt bundle
+    replace-signature	Replaces the signature of an already signed bundle
+    extract-signature	Extract the bundle signature
+    extract		Extract the bundle content
+    install		Install a bundle
+    info			Show bundle information
```
jluebbe, line 935: This mixes tabs and spaces.

jluebbe (changes requested):

docs/reference.rst
ejoerns: Not sure, maybe because it's already there? It also just documents that it should work, not how it is to be used :smirk:

ejoerns:
> `--intermediate` only needs the certificate. Why would one store that in an HSM?

Not sure, maybe because it's already there? It also just documents that it should work, not how it is to be used :smirk:

docs/reference.rst
ejoerns: It's actually just a copy of the output, but I will replace the generated tabs with spaces, yes.

jluebbe (approved):

PR #1059, opened 2023-02-10
Skip target-related context setup if no config is required

### LINKED ISSUE #737: Silence failed to resolve realpath error
When creating a bundle I get this error:
`rauc-Message: 04:13:16.008: Failed to resolve realpath for '/dev/disk/by-uuid/a4af6bed-7716-4281-8865-b8ac4b283339'`

It seems irrelevant to my situation as I'm only using the bundle command not the other functionality. Is it possible to silence this error in this case? I currently don't have the --debug flag passed.

I'm running rauc with this command line:
`rauc --cert=/rauc-keys/dev.crt --key=/rauc-keys/dev.key --debug bundle /tmp/rauc-SwIFA8IWDd8BTqrun /images/output.raucb`

ejoerns:
Do not invoke any target-related context setup steps if no config is required.
This is safe since config mode is 'required' for all target-related subcommands (like 'service') that need to have specific context available (such as bootname, etc.).

With this we gain two major benefits:

* Not invoking pointless procfs, sysfs and mount inspections when running as a host tool

* Removing the misleading 'info'-Messages like:
  > rauc-Message: 13:53:56.765: Failed to resolve realpath for '/dev/mapper/Root-root'

  e.g. when running rauc 'host' subcommands from a container

Resolves #846
Fixes #737

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -219,57 +219,15 @@ static gchar* get_variant_from_file(const gchar* filename, GError **error)
 	return contents;
 }
 
-
-gboolean r_context_configure(GError **error)
+/**
+ * Configures options that are only relevant when running as update service on
+ * the target device.
+ */
+static gboolean r_context_configure_target(GError **error)
 {
 	GError *ierror = NULL;
-	RContextConfigMode configmode;
-	const gchar *configpath = NULL;
 
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
-
-	g_assert_nonnull(context);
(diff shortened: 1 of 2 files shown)


jluebbe (approved):

jluebbe:
While it's not ideal to reuse `R_CONTEXT_CONFIG_MODE_REQUIRED` for this, as it's also enable by passing an explicit config on the commandline,  this is a clear improvement.

PR #1060, opened 2023-02-12
.github: workflows: tests: (re)enable coverage in builds

ejoerns:
Coverage data is expected to be generated and has not been (re)enabled yet when transitioning from autotools to meson in CI tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Test with service and gpt
       run: |
         rm -rf build/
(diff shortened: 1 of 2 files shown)


jluebbe (commented):
`gpt.c` and `mbr.c` are no longer covered. Instead we now have coverage info for the test code, which we didn't have before, but that shouldn't hurt.

ejoerns:
Yes, I've added the `meson configure` to see that features are not unintentionally disabled, but that does not seem to be the case. Seems as if some more fine-tuning is required here :smirk:

jluebbe (approved):

PR #1061, opened 2023-02-14
add info-dbus API support to meta custom section

abdkadri:
Signed-off-by: Abdallah Elkadri <elkadri.abdallah@gmail.com>

This pull request updates the rauc info command and the Dbus Info API to handle custom meta fields in the manifest, which was not the case in previous versions.

these changes were tested in streaming mode and works perfectly. 

note that there is already 12 tests in rauc.t that already fail, and after the modifications they remain the same. 
this is a work-in-progress PR .

### DIFF
--- a/src/de.pengutronix.rauc.Installer.xml
+++ b/src/de.pengutronix.rauc.Installer.xml
@@ -35,6 +35,7 @@
       <arg name="bundle" type="s" direction="in" />
       <arg name="compatible" type="s" direction="out" />
       <arg name="version" type="s" direction="out" />
+      <arg name="meta" type="as" direction="out" />
     </method>
 
     <!--
--- a/src/main.c
+++ b/src/main.c
@@ -923,6 +923,15 @@ static gchar *info_formatter_readable(RaucManifest *manifest)
 	g_string_append_printf(text, "Description:\t'%s'\n", manifest->update_description);
 	g_string_append_printf(text, "Build:      \t'%s'\n", manifest->update_build);
 
+	if (manifest->meta != NULL) {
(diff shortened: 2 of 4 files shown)


ejoerns:
@abdkadri thanks for starting the contribution! You might want to have a look at #965 which will probably replace the current info D-Bus Api and where it was already mentioned that one might want to switch to a more generic approach to expose more manifest data.

abdkadri:
hello, @ejoerns so in this case this PR will have no meaning if  the #965 will be merged ? otherwise what we can do in case the #965 will remain open or rejected ?

ejoerns:
Adding metadata is still required, but it would be part of other manifest data to be added.
So I fear most of your code cannot be reused.

The idea would be to convert manifest data to `GVariant` similar to what we do for status so far: https://github.com/rauc/rauc/blob/master/src/service.c#L229

Since @jluebbe started working on metadata support, maybe he also already has some code for this or a concept of how to implement it?

abdkadri:
@ejoerns i am aware that my code will not be reused in this case .
also I can re push another solution using the GVariant

ejoerns:
@abdkadri I have prepared #1067 that could serve as a base for extending with meta section information.

jluebbe:
If we produce a nested GVariant structure from the manifest information, we should also be able to reuse that for the JSON output of `rauc info` by using https://gnome.pages.gitlab.gnome.org/json-glib/func.gvariant_serialize.html.

ejoerns:
This was addressed and fixed by #1094 now.

@abdkadri you may want to check if this meets your requirements and let us know if something is missing for you.

PR #1067, opened 2023-02-20
Introduce new 'InspectBundle' D-Bus API 

ejoerns:
The old 'InfoBundle' method lacked the capability of passing additional arguments, which prevented adding a lot of features in the past.

Compared to the existing one, the new method differs only in the additional array-of-string-variant parameter 'args'.

This allows passing all kinds of future additional simple arguments.

To guarantee API stability, the old method will be kept as an alternative.

This builds on top of #965 but with the review feedback addressed.

Closes #965.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -1148,6 +1148,8 @@ Methods
 
 :ref:`Info <gdbus-method-de-pengutronix-rauc-Installer.Info>` (IN  s bundle, s compatible, s version);
 
+:ref:`InspectBundle <gdbus-method-de-pengutronix-rauc-Installer.InspectBundle>` (IN  s source, IN a{sv} args, a{sv} info);
+
 :ref:`Mark <gdbus-method-de-pengutronix-rauc-Installer.Mark>` (IN  s state, IN  s slot_identifier, s slot_name, s message);
 
 :ref:`GetSlotStatus <gdbus-method-de-pengutronix-rauc-Installer.GetSlotStatus>` (a(sa{sv}) slot_status_array);
@@ -1244,6 +1246,8 @@ IN s *source*:
 The Info() Method
 ^^^^^^^^^^^^^^^^^
 
+.. note:: This method is deprecated. Use InspectBundle() instead.
+
 .. code::
 
   de.pengutronix.rauc.Installer.Info()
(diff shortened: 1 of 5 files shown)


src/de.pengutronix.rauc.Installer.xml
```diff
@@ -36,6 +36,19 @@
...
+         InspectBundle: D-Bus variant of rauc info <bundle>
+         @source: Path to the queried bundle.
+         @args: Additional arguments to pass
+         @bundle_info: string variant dict of bundle information
+    -->
+    <method name="InspectBundle">
+      <arg name="source" type="s" direction="in" />
+      <arg name="args" type="a{sv}" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
+      <arg name="bundle_info" type="a{sv}" direction="out"/>
```
jluebbe, line 50: Just use `info`, as we may also want to return HTTP headers from the server in the future.

src/service.c
```diff
@@ -133,22 +147,60 @@ static gboolean r_on_handle_install(RInstaller *interface,
...
+		g_variant_dict_insert(&dict, "compatible", "s", manifest->update_compatible);
+
+	if (manifest->update_version)
+		g_variant_dict_insert(&dict, "version", "s", manifest->update_version);
+
+	if (manifest->update_description)
+		g_variant_dict_insert(&dict, "description", "s", manifest->update_description);
+
+	if (manifest->update_build)
+		g_variant_dict_insert(&dict, "build", "s", manifest->update_build);
```
jluebbe, line 166: Perhaps these should be part of a "update" dict to match the manifest structure?

src/service.c
```diff
@@ -133,22 +147,60 @@ static gboolean r_on_handle_install(RInstaller *interface,
...
+
+	return g_variant_dict_end(&dict);
+}
+
+static gboolean r_on_handle_inspect_bundle(RInstaller *interface,
 		GDBusMethodInvocation  *invocation,
-		const gchar *arg_bundle)
+		const gchar *arg_bundle, GVariant *arg_args)
 {
+	RaucBundleAccessArgs access_args;
```
jluebbe, line 175:
```suggestion
	RaucBundleAccessArgs access_args = {0};
```

docs/reference.rst
```diff
@@ -1260,6 +1264,57 @@ s *compatible*:
...
+        certificate to authenticate the server (instead of the system wide
+        store)
+
+    :STRING 'http-headers', VARIANT 'as' <array of strings>: Add the provided
+        headers to every request (i.e. for bearer tokens)
+
+    :STRING 'tls-no-verify', VARIANT 'b' <true/false>: Ignore verification
+        errors for the server certificate
+
+a{sv} *bundle_info*:
```
jluebbe, line 1303: This needs to be adjusted as well if we change the name.

docs/using.rst
```diff
@@ -601,6 +601,12 @@ Monitor the D-Bus interface
 
   busctl monitor de.pengutronix.rauc
 
+Obtain bundle information
+
+.. code-block:: sh
+
+  busctl call de.pengutronix.rauc / de.pengutronix.rauc.Installer InspectBundle sa{sv} "/path/to/bundle" 0
```
jluebbe, line 608: Perhaps this should mention that URLs are supported if streaming mode is enabled?

docs/reference.rst
```diff
@@ -1244,6 +1246,8 @@ IN s *source*:
 The Info() Method
 ^^^^^^^^^^^^^^^^^
 
+.. note:: This method is deprecated.
+
```
jluebbe, line 1250:
```suggestion
.. note:: This method is deprecated. Use InspectBundle() instead.
```

jluebbe (changes requested):

src/service.c
ejoerns: We can do this. First of all I thougth about aligning it with the current 'rauc info' json output which is

``` json
{
  "compatible" : "Test Config",
  "version" : "2011.03-2",
  "description" : null,
  "build" : null,
  "hooks" : [],
  "images" : [
    {
      "rootfs" : {
        "variant" : null,
        "filename" : "rootfs.img",
        "checksum" : "101a4fc5c369a5c89a51a61bcbacedc9016e9510e59a4383f739ef55521f678d",
        "size" : 8192,
        "hooks" : [],
        "adaptive" : []
      }
    },
    {
      "appfs" : {
        "variant" : null,
        "filename" : "appfs.img",
        "checksum" : "f95c0891937265df18ff962869b78e32148e7e97eab53fad7341536a24242450",
        "size" : 8192,
        "hooks" : [],
        "adaptive" : []
      }
    }
  ]
}
```

But since sections tend to increase probably just reflecting the actual manifest structure is more future-proof.

jluebbe: In the example in https://github.com/rauc/rauc/pull/965#discussion_r983520569 I already addressed the "list of single value dicts" in "images", which we also should do at some point.

ejoerns: Could only be fixed by breaking existing consumers I guess. Or by duplicating the info.

docs/using.rst
ejoerns: Replaced by `<bundle-path>/<bundle-url>` now

src/service.c
jluebbe: We could also add a json-v2 format, which matches the layout of the InspectBundle method?

src/de.pengutronix.rauc.Installer.xml
```diff
@@ -25,17 +25,30 @@
...
+    -->
+    <method name="Info">
       <arg name="bundle" type="s" direction="in" />
       <arg name="compatible" type="s" direction="out" />
       <arg name="version" type="s" direction="out" />
     </method>
+ 
+    <!--
+         InspectBundle: D-Bus variant of rauc info <bundle>
+         @source: Path to the queried bundle.
```
jluebbe, line 42:
```suggestion
         @source: Path or URL to the queried bundle.
```

jluebbe (changes requested):
Looks good otherwise.

ejoerns:
@stefanu21 Would that API still fit your original use case?

jluebbe (approved):

PR #1068, opened 2023-02-21
Update slot skipping documentation

ejoerns:
* note that combination with streaming serves as download optimization
* shorten and clarify the proper use case
* replace term 'hash' by 'checksum'

### DIFF
--- a/docs/basic.rst
+++ b/docs/basic.rst
@@ -197,11 +197,10 @@ If this hash equals the hash of the image to write, RAUC can skip updating this
 slot as a configurable performance optimization
 (refer :ref:`install-same <install-same>` per-slot option).
 
-This is especially useful when having a setup with, for example, two redundant
-application file systems and two redundant root file systems. In case you
-update the application file system content much more frequently while keeping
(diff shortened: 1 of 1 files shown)


docs/basic.rst
```diff
@@ -183,25 +183,28 @@ Slot Status and Skipping Slot Updates
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 RAUC hashes each image or archive when packing it into a bundle and stores this
```
jluebbe, line 185: Hmm, in reality it *is* a hash, so I'm not sure that this change actually makes it less confusing. As we can't easily change the name in the manifest, perhaps we should add more details here?
```suggestion
RAUC hashes each image or archive with SHA 256 when packing it into a bundle and stores this
```

jluebbe (changes requested):

docs/basic.rst
ejoerns: Added

jluebbe (approved):

PR #1073, opened 2023-02-21
docs: meson.build: add input files as build dependencies

ejoerns:
Since we do not track all doc input files, we cannot say if sphinx needs to rebuild. Thus, with the configuration so far, meson would have built the docs only once and then assumed they were built.

Since sphinx can track required rebuilds itself and does not rebuild unnecessary parts, we can safely enforce always calling sphinx here by settings the target as 'stale'.

### DIFF
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -4,9 +4,28 @@ if not sphinx.found()
   subdir_done()
 endif
 
+sources_doc = files([
+  'conf.py',
+  'advanced.rst',
+  'basic.rst',
+  'changes.rst',
+  'checklist.rst',
+  'contributing.rst',
+  'examples.rst',
+  'faq.rst',
+  'index.rst',
+  'integration.rst',
+  'reference.rst',
+  'scenarios.rst',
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> Since sphinx can track required rebuilds itself and does not rebuild unnecessary parts, we can safely enforce always calling sphinx here by settings the target as 'stale'.

This will still cause ninja to be unable to perform no-op builds. You could do something like this, instead: https://github.com/qemu/qemu/blob/master/docs/sphinx/depfile.py

ejoerns:
@eli-schwartz Thanks for the hint. I chose the 'lazy' approach here since the doc target is excluded from the normal build. So is using `build_always_stale` always discouraged?

I have seen the depfile approach in other sphinx projects already.  Will check if this or en explicit list of source `.rst` files looks more promising.

eli-schwartz:
I dislike build_always_stale on principle and wish to extend all tools if possible, but that's more of a personal thing.

jluebbe (approved):

PR #1074, opened 2023-02-22
Support Configurable  Attempts for Barebox Bootloader

ejoerns:
This applies the already existing config values of `boot-attempts` and `boot-attempts-primary` to the barebox logic, too and thus allow overriding the compiled-in defaults.

Some pre-work for fixing config option handling and adding missing test cases is included.


src/config_file.c
```diff
@@ -172,20 +171,21 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 	gchar *variant_data;
 	g_autofree gchar *bundle_formats = NULL;
 
+	/* in case of an early abort, return NULL */
```
jluebbe, line 174: Perhaps add a:
``` C
 	g_return_val_if_fail(config, NULL);
```

src/config_file.c
```diff
@@ -239,8 +239,49 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+	if (c->boot_default_attempts < -1) {
+		g_set_error(
+				error,
+				R_CONFIG_ERROR,
+				R_CONFIG_ERROR_BOOTLOADER,
+				"Value for \"boot-attempts\" must not be negative");
+		return FALSE;
```
jluebbe, line 256: This would be confusing error message if the key was missing. The same applies below. We should probably set a real default value.

test/config_file.c
```diff
@@ -269,6 +269,79 @@ mountprefix=/mnt/myrauc/\n";
...
+	g_assert_true(load_config(pathname, &config, &ierror));
+	g_assert_no_error(ierror);
+
+	g_free(pathname);
+
+	pathname = write_tmp_file(fixture->tmpdir, "invalid_bootloader.conf", invalid_boot_attempts_cfg_file, NULL);
+	g_assert_nonnull(pathname);
+
+	g_assert_false(load_config(pathname, &config, &ierror));
+	g_assert_cmpstr(ierror->message, ==, "Key file contains key \342\200\234boot-attempts\342\200\235 in group \342\200\234system\342\200\235 which has a value that cannot be interpreted.");
```
jluebbe, line 321: This might depend too much on glib implementation details. Perhaps use g_assert_error with domain and code? Or check only that specific substrings are in the message?

docs/reference.rst
```diff
@@ -107,14 +107,14 @@ Example configuration:
 ``boot-attempts``
   This configures the number of boot attempts to set when a slot is marked good
   through the D-Bus API or via the command line tool.
-  This is currently only supported when ``bootloader`` is set to ``uboot`` and
-  defaults to 3 if not set.
+  This is currently only supported when ``bootloader`` is set to ``uboot`` or
+  ``barebox`` and defaults to 3 if not set.
```
jluebbe, line 111: This should mention that the value should match the default configured in the bootloader for consistency.

jluebbe (changes requested):

src/config_file.c
ejoerns: If the key isn't found, we set `-1` and we test for `< -1` to emit the error message, thus this should be safe.
It might only have an 'unexpected' result if someone writes `-1` in the config file :smirk:.

I chose this approach since the default value is currently set by bootchooser.c for each backend individually (based on a `#define`).

test/config_file.c
ejoerns: Mh, yes, might be better here indeed

docs/reference.rst
ejoerns: Ack

src/config_file.c
jluebbe: Hmm, I read this several times and always saw `<= -1`. Perhaps use 0 as "unspecified"?

ejoerns: I was not sure if there could be any valid use case or test setup that should allow setting 0 from config. But probably enforcing reset attempts to be > 0 is ok here.

jluebbe: Otherwise, I wouldn't mind moving the default to the config file parsing.

src/config_file.c
```diff
@@ -241,8 +241,49 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
+	}
+	if (c->boot_attempts_primary < 0) {
+		g_set_error(
+				error,
+				R_CONFIG_ERROR,
+				R_CONFIG_ERROR_BOOTLOADER,
+				"Value for \"boot-attempts-primary\" must not be negative");
+		return FALSE;
+	}
+	if (c->boot_default_attempts >= 0 || c->boot_attempts_primary >= 0) {
```
jluebbe: This should be `> 0` from this commit on (currently, you change it later).

Looks good otherwise.

jluebbe (changes requested):

src/config_file.c
ejoerns: Oh, good catch! Fixed.

jluebbe (approved):

PR #1075, opened 2023-02-23
Introduce Manifest Hash

ejoerns:
This introduces a manifest hash that is calculated when loading the manifest from the bundle.

Together with 'verity' (or derived) bundle formats, this will allow unambiguously identifying a bundle, since the verity root hash is also part of the manifest and thus indirectly both the payload and the meta data are part of the hash.

The information is exposed both for bundle introspection (`rauc info`) as well as system introspection (`rauc status`) and thus allows for example to see if a given bundle is exactly the one that was installed on the system.

### DIFF
--- a/include/manifest.h
+++ b/include/manifest.h
@@ -71,6 +71,8 @@ typedef struct {
 
 	/* internal marker that this was encrypted */
 	gboolean was_encrypted;
+	/* computed manifest hash */
+	gchar *hash;
 } RaucManifest;
 
 /**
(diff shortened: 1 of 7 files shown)


jluebbe (approved):
I've changed some commit message wording and fixed a copy paste error (`"bundle.build"` instead of `"bundle.hash"`).

ejoerns (commented):
Changes look fine for me :+1:

Issue #1057, opened 2023-02-08
Error getting loop backing_file for /dev/loop0p1


paletteguy:
**Describe the bug**
If a loop0 device is already used for other purpose when Rauc installer  is unable mount image from rauc file.

**Background information**
RAUC version 1.8 - Yocto kirkstone - IMX8-MP

**To Reproduce**
1. Mount a loop0 devices
2. run rauc install image.raucb

**Expected behavior**
* Installation completed and RAUC used another loop-back device for mounting.

**Logs**
```
Feb 08 13:39:45 CCB18177 rauc[357]: input bundle: /tmp/tmp.aDGIesmks4/image-update-99.47.0.84.raucb
Feb 08 13:39:45 CCB18177 rauc[357]: Active slot bootname: system1
Feb 08 13:39:45 CCB18177 rauc[357]: installing /tmp/tmp.aDGIesmks4/image-update-99.47.0.84.raucb: started
Feb 08 13:39:45 CCB18177 updated[1367]: installing
Feb 08 13:39:45 CCB18177 rauc[357]: Error getting loop backing_file for /dev/loop0p1: Failed to open file “/sys/block/loop0p1/loop/backing_file”: No such file or directory
Feb 08 13:39:45 CCB18177 rauc[357]: installing /tmp/tmp.aDGIesmks4/image-update-99.47.0.84.raucb: Error getting loop backing_file for /dev/loop0p1: Failed to open file “/sys/block/loop0p1/loop/backing_file”: No such file or directory
Feb 08 13:39:45 CCB18177 rauc[357]: installing /tmp/tmp.aDGIesmks4/image-update-99.47.0.84.raucb: finished
Feb 08 13:39:45 CCB18177 rauc[357]: installing `/tmp/tmp.aDGIesmks4/image-update-99.47.0.84.raucb` failed: 1
Feb 08 13:39:45 CCB18177 updated[1367]: 0% Installing
Feb 08 13:39:45 CCB18177 updated[1367]: 0% Determining slot states
Feb 08 13:39:45 CCB18177 updated[1367]: 20% Determining slot states failed.
Feb 08 13:39:45 CCB18177 updated[1367]: 100% Installing failed.
```

**Additional context**
Reproducible 100 %

jluebbe:
Could you add the output of `losetup -l -a` and `cat /proc/mounts`?

It seems that RAUC gets confused while checking mounted filesystems for a partitioned loop device. The backing file is probably only set for the main loop device.

[...]

PR #1077, opened 2023-02-27
Remove partition information from loop when getting backing_file #1057

paletteguy:
Fixing #1057 removing partition part of /dev/loop when getting backing file.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -53,6 +53,8 @@ static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 		return g_strdup(devicepath);
 
 	devicename = g_path_get_basename(devicepath);
+	g_string_replace(devicename, "p1", "\0", 0);
+
 	syspath = g_build_filename("/sys/block", devicename, "loop/backing_file", NULL);
 
 	content = read_file_str(syspath, &ierror);
--- a/src/install.c
+++ b/src/install.c
@@ -53,8 +53,6 @@ static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 		return g_strdup(devicepath);
 
 	devicename = g_path_get_basename(devicepath);
-	g_string_replace(devicename, "p1", "\0", 0);
-
 	syspath = g_build_filename("/sys/block", devicename, "loop/backing_file", NULL);
 
 	content = read_file_str(syspath, &ierror);
--- a/src/install.c
+++ b/src/install.c
@@ -45,6 +45,7 @@ static void install_args_update(RaucInstallArgs *args, const gchar *msg)
 static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 {
 	g_autoptr(GString) devicename = NULL;
+	g_autofree gchar *basename = NULL;
 	g_autofree gchar *syspath = NULL;
 	gchar *content = NULL;
 	GError *ierror = NULL;
@@ -52,9 +53,12 @@ static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 	if (!g_str_has_prefix(devicepath, "/dev/loop"))
 		return g_strdup(devicepath);
 
-	devicename = g_string_new(g_path_get_basename(devicepath));
-	devicename = g_string_set_size(devicename, 5);
+	basename = g_path_get_basename(devicepath);
 
+	devicename = g_string_new(basename);
+	
+	/* Cut of any partition information like p1*/
+	devicename = g_string_set_size(devicename, sizeof("loop0"));
 
 	syspath = g_build_filename("/sys/block", devicename->str, "loop/backing_file", NULL);
 
--- a/src/install.c
+++ b/src/install.c
@@ -57,7 +57,7 @@ static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 
 	devicename = g_string_new(basename);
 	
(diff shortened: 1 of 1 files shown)


jluebbe (changes requested):

src/install.c
```diff
@@ -53,6 +53,8 @@ static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 		return g_strdup(devicepath);
 
 	devicename = g_path_get_basename(devicepath);
+	g_string_replace(devicename, "p1", "\0", 0);
```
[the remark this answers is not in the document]
paletteguy: Rewrote using g_string_set_size

ejoerns (changes requested):
Thanks! I've left some comments but in principle, the solution looks plausible.

Please squash and force-push your commits so that they for a single one again.

Please add some reasoning to the commit message (and address type 'parition' in the title)

src/install.c
```diff
@@ -44,16 +44,21 @@ static void install_args_update(RaucInstallArgs *args, const gchar *msg)
...
 	g_autofree gchar *syspath = NULL;
 	gchar *content = NULL;
 	GError *ierror = NULL;
 
 	if (!g_str_has_prefix(devicepath, "/dev/loop"))
 		return g_strdup(devicepath);
 
-	devicename = g_path_get_basename(devicepath);
-	syspath = g_build_filename("/sys/block", devicename, "loop/backing_file", NULL);
+	devicename = g_string_new(g_path_get_basename(devicepath));
```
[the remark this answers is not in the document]
paletteguy: Fixed leak

src/install.c
```diff
@@ -44,16 +44,21 @@ static void install_args_update(RaucInstallArgs *args, const gchar *msg)
...
+	devicename = g_string_set_size(devicename, 5);
+
+
```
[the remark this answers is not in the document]
paletteguy: Removed line..

src/install.c
```diff
@@ -44,16 +44,25 @@ static void install_args_update(RaucInstallArgs *args, const gchar *msg)
...
 		return g_strdup(devicepath);
 
-	devicename = g_path_get_basename(devicepath);
-	syspath = g_build_filename("/sys/block", devicename, "loop/backing_file", NULL);
+	basename = g_path_get_basename(devicepath);
+
+	devicename = g_string_new(basename);
+	
+	/* Cut of any partition information like p1 */
+	devicename = g_string_set_size(devicename, sizeof("loop0"));
```
ejoerns, line 61: Just had another quick look into alternatives, maybe a simple `g_strndup()` is less overhead here.

src/install.c
```diff
@@ -44,16 +44,25 @@ static void install_args_update(RaucInstallArgs *args, const gchar *msg)
...
 	if (!g_str_has_prefix(devicepath, "/dev/loop"))
 		return g_strdup(devicepath);
 
-	devicename = g_path_get_basename(devicepath);
-	syspath = g_build_filename("/sys/block", devicename, "loop/backing_file", NULL);
+	basename = g_path_get_basename(devicepath);
+
+	devicename = g_string_new(basename);
+	
+	/* Cut of any partition information like p1 */
```
ejoerns, line 60:
```suggestion
	/* Cut off any partition information like p1 */
```

ejoerns (commented):
Do you have the ability to squash and force-push? There are more and more commits piling up. If not, I could do this, too.

src/install.c
```diff
@@ -44,16 +44,25 @@ static void install_args_update(RaucInstallArgs *args, const gchar *msg)
...
+	devicename = g_string_set_size(devicename, sizeof("loop0"));
+
+	syspath = g_build_filename("/sys/block", devicename->str, "loop/backing_file", NULL);
+
+	g_debug("Getting backing file for Loopback device: %s from %s", devicename->str, syspath);
```
a3f, line 65: Given that `devicename->str` will be a substring of `syspath`, you may want to replace it with `devicepath` instead.

src/install.c
jluebbe: The simple approach with cutting of a fixed number of characters will fail if more than 10 loop devices exist.

We need to check if it is a partition on a loop device, and in that case remove the `p\d+` part.

jluebbe (changes requested):

ejoerns:
> The simple approach with cutting of a fixed number of characters will fail if more than 10 loop devices exist.

Who would ever.. :smirk:  but you are right. If we already touch it, then we should apply a generic solution.
I wonder if / how other tools handle this but haven't found anything yet.

> We need to check if it is a partition on a loop device, and in that case remove the p\d+ part.

ejoerns:
Something like this could work:

``` diff
-       if (!g_str_has_prefix(devicepath, "/dev/loop"))
+       regex = g_regex_new("(loop\\d+)(p\\d+)?", 0, 0, NULL);
+       g_assert_nonnull(regex);
+
+       g_regex_match(regex, devicepath, 0, &match_info);
+       if (!g_match_info_matches(match_info))
                return g_strdup(devicepath);
 
-       devicename = g_path_get_basename(devicepath);
+       devicename = g_match_info_fetch(match_info, 1);
```

ejoerns:
FTR: I have now implemented the suggestion for a generic fix in #1082

PR #1078, opened 2023-02-28
Fix number format handling for U-Boot

sirhcel:
* The remaining attempts are hexadecimal numbers (due to `setexpr`) but without a prefix
* The comparisons should take this into account
    * Previously [the comparison]() failed for all values containing non-decimal digits like 'f', for example
        ```
        => setenv BOOT_A_LEFT b00d
        => if test ${BOOT_A_LEFT} -gt 0; then echo greater; else echo not greater; fi
        not greater
        ```
    * Providing the prefix when comparing with `test` gives the expected result
        ```
        => if test 0x${BOOT_A_LEFT} -gt 0; then echo greater; else echo not greater; fi
        greater
        ```
    * As `setexpr` emits the result without prefix, adding the prefix when setting the variable is not possible
*  RAUC also should write hex values to the U-Boot environment consistently
    * Previously a successful update left the new boot slot with its remaining attempts in decimal representation

### DIFF
--- a/contrib/uboot.sh
+++ b/contrib/uboot.sh
@@ -7,14 +7,14 @@ for BOOT_SLOT in "${BOOT_ORDER}"; do
   if test "x${bootargs}" != "x"; then
     # skip remaining slots
   elif test "x${BOOT_SLOT}" = "xA"; then
-    if test ${BOOT_A_LEFT} -gt 0; then
+    if test 0x${BOOT_A_LEFT} -gt 0; then
       echo "Found valid slot A, ${BOOT_A_LEFT} attempts remaining"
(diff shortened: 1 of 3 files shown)


ejoerns:
Thanks for fixing the boot logic here!

Currently, the `/bootchooser/uboot-conf-attempts/` test fails since it test for the attempts number to be `10` instead of `a`.
This should be adapted. Luckily this serves then also as a test for boot attempts being > 9.

ejoerns (approved):

jluebbe (approved):

jluebbe:
Thanks for your contribution!

PR #1084, opened 2023-03-02
meson.build: support generating version information from git

ejoerns:
This creates a `version.h` file with version information derived from `git describe` using `vcs_tag`.
The `version.h` include will be injected into rauc and librauc builds.

For `meson dist`, no git information will be available, thus `meson.project_version()` serves as a fallback.

However, the generated tarball then still produces a non-fatal message that might upset users:

> fatal: not a git repository (or any parent up to mount point /)

Note that meson does not seem to allow including generated files in dist tarballs by design, thus adding a '.tarball-version'-like file will not be possible.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -169,6 +169,8 @@ version_h = vcs_tag(
 )
 versiondep = declare_dependency(sources: version_h)
 
+meson.add_dist_script('version-gen', meson.project_version())
+
 rauc_deps = [threaddep, libcurldep, libnlgenldep, jsonglibdep, dbusdep, glibdep, giodep, giounixdep, openssldep, fdiskdep]
 
 librauc = static_library('rauc',
--- a/meson.build
+++ b/meson.build
@@ -14,8 +14,6 @@ cc = meson.get_compiler('c')
 
 conf = configuration_data()
 conf.set_quoted('PACKAGE_NAME', 'rauc')
-conf.set_quoted('PACKAGE_VERSION', meson.project_version())
-conf.set_quoted('PACKAGE_STRING', 'rauc ' + meson.project_version())
 conf.set_quoted('STREAMING_USER', get_option('streaming_user'))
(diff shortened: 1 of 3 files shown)


eli-schwartz:
> However, the generated tarball then still produces a non-fatal message that might upset users:

This happens because you unconditionally run a `command: ['git', ...]`. If you don't specify one, meson's builtin command passed to the vcstagger handler script, is `git describe --dirty=+`, ***but*** if the current project isn't a git repository then it uses a command that fails silently.

> Note that meson does not seem to allow including generated files in dist tarballs by design, thus adding a '.tarball-version'-like file will not be possible.

You're supposed to use `meson.add_dist_script()` for this, to post-process the staging directory of the dist tarball before it is tarred up.

ejoerns:
> > However, the generated tarball then still produces a non-fatal message that might upset users:
> 
> This happens because you unconditionally run a `command: ['git', ...]`. If you don't specify one, meson's builtin command passed to the vcstagger handler script, is `git describe --dirty=+`, _**but**_ if the current project isn't a git repository then it uses a command that fails silently.

@eli-schwartz Thanks again for your valuable feedback! Somehow I wasn't aware that it works with a built-in command, too. And it also just seems to fall back to the meson project version if no git is available.

I've updated my PR to use this. What is open to discuss is if we need the git-generated version in the tarball, too or if the hard-coded value is sufficient for us.

> > Note that meson does not seem to allow including generated files in dist tarballs by design, thus adding a '.tarball-version'-like file will not be possible.
> 
> You're supposed to use `meson.add_dist_script()` for this, to post-process the staging directory of the dist tarball before it is tarred up.

Good hint, haven't found this in documentation / examples so far. I've added a minimal hack for supporting autotools in getting basic version information (meson project version first of all).

jluebbe (approved):

PR #1093, opened 2023-03-11
Emitting Fine-Grained Progress over D-Bus for Most Update Handler (Rework)

ejoerns:
Based on #811, this adds support for emitting fine-grained progress.

This PR addresses the review comments of #811 and is based on current `master`.

Changes compared to #811 are:

* Pipe-based update handlers now use `splice()` instead of manual copying through RAUC. This required reworking some parts to explicitly use `GUnix{Input,Output}Streams`.
* Separate stream-based copy function for raw copy
* Still support running progress-enabled update handlers in non-progress context. This is e.g. required for supporting `write-slot`.
* Renamed some methods to better fit their purpose
* Error handling and error message fixes
* Fix tests by using .tar extension (instead of actually creating .tar.bz2)
* Original commits are squashed to reduce noise of introducing methods and instantly factoring them out. Test adaptions were moved into commits making the adaptions necessary (also to improve bisectability).


poeschel:
@ejoerns thank you very much for bringing this forward!

jluebbe:
I've rebased on master to fix the conflict introduced by merging #1092. Also I've added a loop to generate the expected progress messages.

ejoerns:
> I've rebased on master to fix the conflict introduced by merging #1092. Also I've added a loop to generate the expected progress messages.

Yes, adding a loop is really a good idea here since we now have even more steps. :+1:

src/update_handler.c
```diff
@@ -208,6 +238,106 @@ static gboolean ubifs_ioctl(RaucImage *image, int fd, GError **error)
...
+static gboolean splice_with_progress(GUnixInputStream *image_stream,
+		GUnixOutputStream *out_stream, goffset image_size, GError **error)
+{
+	ssize_t out_size = 0;
+	goffset sum_size = 0;
+	gint last_percent = -1, percent;
+
+	do {
+		int in_fd = g_unix_input_stream_get_fd(image_stream);
+		int out_fd = g_unix_output_stream_get_fd(out_stream);
```
jluebbe, line 289: Pull these out of the loop.

src/update_handler.c
```diff
@@ -208,6 +238,106 @@ static gboolean ubifs_ioctl(RaucImage *image, int fd, GError **error)
...
+			G_FILE_ATTRIBUTE_STANDARD_SIZE,
+			G_FILE_QUERY_INFO_NONE,
+			NULL,
+			&ierror);
+	if (image_info == NULL) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
+	image_size = g_file_info_get_size(image_info);
```
jluebbe, line 337: This can be done with a `fstat()` in `splice_with_progress()`, which would simplify the logic.

src/update_handler.c
jluebbe, line 1: The commit message should refer to `nandwrite` instead of `casync`.

jluebbe (changes requested):

src/update_handler.c
ejoerns: I am unsure how this simplifies the logic. With `fstat()` we still need a function call with error handling and having that inside `splice_with_progress()` will just move the code from one to another location.

jluebbe: We don't need to open the file twice, as fstat simply works directly on the FD.

jluebbe (approved):

jluebbe:
Please merge if you agree with my changes.

ejoerns:
Looks fine to me (even if `open()` actually should only return `-1` for errors :smirk:).

PR #1095, opened 2023-03-12
Add support for FTPS

sirhcel:
* This PR adds support for FTPS
* It depends on https://github.com/systemd/casync/pull/266 to do the same for casync
* libcurl already supports FTPS in its default configuration
* We just need to allow the FTPS URL schema at RAUC

* Successfully tested casync updates via FTPS (unsing vsftpd as server)


jluebbe:
This won't work with streaming, but adding the check for that case should be a separate change anyway.

jluebbe (approved):

PR #1105, opened 2023-04-03
src/install: un-break external mount point detection

ejoerns:
The memory leakage 'fix' introduced in a6c17a402 ("src/install: avoid leaking string if multpile mount points match") erroneously uses 'break' where it should have used 'continue'.
As a result, external mount point detection aborts way too early in some cases and further external mount points will not be detected then.

This is also not ideal in terms of safety, since it could lead to unintentionally overriding a still-mounted partition which RAUC refuses normally.

This affects all RAUC releases since (including) v1.2.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -132,7 +132,7 @@ gboolean determine_slot_states(GError **error)
 			 * To avoid leaking the string returned by g_unix_mount_get_mount_path() here, we skip all further matches
 			 */
 			if (s->ext_mount_point) {
-				break;
+				continue;
 			}
 			s->ext_mount_point = g_strdup(g_unix_mount_get_mount_path(m));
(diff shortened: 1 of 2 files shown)


ejoerns:
I had started with a test case actually and then threw it away first of all as it did not seem to be straightforward to implement. But I could continue where I left.
You mean a test case in sharness tests, right?

jluebbe:
> You mean a test case in sharness tests, right?

I had a C-based test in mind, but a sharness test could be easier.

ejoerns:
> > You mean a test case in sharness tests, right?
> 
> I had a C-based test in mind, but a sharness test could be easier.

Turned out that the opposite was correct. At least compared to the effort required for other tests in this category.

I have added a C-based test in `test/install.c` now.

jluebbe (approved):

PR #1106, opened 2023-04-05
Add perform-pre-check configuration option

chhitz:
Extend RAUC to allow a complete bundle consistency check before installing or mounting a bundle. The consistency check is enabled with the new "--pre-check" option.

This feature is especially useful in cases where a bundle contains images for different device variants and is only corrupt in images for on variant.


src/bundle.c
```diff
@@ -2502,6 +2502,41 @@ gboolean load_manifest_from_bundle(RaucBundle *bundle, RaucManifest **manifest,
 	return res;
 }
 
+static gboolean read_complete_dm_device(gchar *dev, GError **error)
+{
+	int checkfd = -1;
+	char checkbuf[65536];
```
jluebbe, line 2508: Hmm, that's quite a large stack allocation.

Perhaps we should use O_DIRECT anyway, to avoid confusing the page cache algorithm by now accessing the data twice instead. It might also force out currently cached data. Take a look at `readable_sectors()` in `test/dm.c`. We should still use a large read size, than in the test (64kiB seems fine).

src/bundle.c
```diff
@@ -2502,6 +2502,41 @@ gboolean load_manifest_from_bundle(RaucBundle *bundle, RaucManifest **manifest,
...
+		int err = errno;
+		g_set_error(error,
+				G_FILE_ERROR,
+				g_file_error_from_errno(err),
+				"Failed to open %s: %s", dev, g_strerror(err));
+		return FALSE;
+	}
+
+	while (ret && bytesRead > 0)
+	{
```
jluebbe, line 2523: Please use our existing style. You can use `./uncrustify.sh` to fix this automatically.

src/main.c
```diff
@@ -2182,6 +2182,11 @@ static GOptionEntry entries_encryption[] = {
 	{0}
 };
 
+static GOptionEntry entries_mount[] = {
+	{"pre-check", '\0', 0, G_OPTION_ARG_NONE, &pre_check, "check complete bundle before mounting", NULL},
```
jluebbe, line 2186: Instead of having the `pre-check` entry twice (here and in `entries_install`), you can reuse this list for `install_group` by calling `g_option_group_add_entries()`. See how it's done for `bundle_group`.

jluebbe (changes requested):
I've not happy with passing the pre_check argument through `prepare_*()` and `r_mount_bundle()`, but I don't have a better idea at the moment.

If we made `pre-check` a global option in the config, we could avoid passing it through all the calls and just run the check early in `r_mount_bundle()`. I'm skeptical if that's a good idea, though.

Also we need tests cases for the error paths as well. We could move the `flip_bits_*()` helpers to `test/common.c` and use the to prepare a corrupted bundle copy.

src/bundle.c
chhitz: Uncrustify did not automatically fix this formatting. I have extended the uncrustify settings to catch theses.

src/bundle.c
chhitz: I have taken the read code from `readable_sections()`.

`O_DIRECT` lead to a `Invalid argument` error while reading in the "plain" bundle case.

src/main.c
chhitz: done

chhitz:
> Also we need tests cases for the error paths as well. We could move the `flip_bits_*()` helpers to `test/common.c` and use the to prepare a corrupted bundle copy.

I have moved the `flip_bits_*()` helpers to `test/common.c` and created a test for the error path.

src/bundle.c
jluebbe: Hmm, for `O_DIRECT`, you need to have 4kiB-alignment for the memory and the block to read. To you have the failing code somewhere we can take a look?

Otherwise, `--pre-check` should probably be skipped for plain bundles, as the signature verification verifies the full payload already (see `bundle->payload_verified`).

jluebbe: Thinking about this again, perhaps we should skip O_DIRECT for now and see whether this is actually needed in practice. Nevertheless, you should use a buffer on the heap instead of 64kiB on the stack.

chhitz: Moved buffer to heap

src/bundle.c
```diff
@@ -2502,6 +2502,50 @@ gboolean load_manifest_from_bundle(RaucBundle *bundle, RaucManifest **manifest,
...
+		int err = errno;
+		g_set_error(error,
+				G_FILE_ERROR,
+				g_file_error_from_errno(err),
+				"Failed to open %s: %s", dev, g_strerror(err));
+		return FALSE;
+	}
+
+	lseek(fd, 0, SEEK_SET);
+
```
jluebbe, line 2526: This `lseek` is redundant, as newly opened files are always at offset 0.

src/bundle.c
```diff
@@ -2502,6 +2502,50 @@ gboolean load_manifest_from_bundle(RaucBundle *bundle, RaucManifest **manifest,
...
+	buf = g_malloc0(65536);
+
+	for (guint sector = 0;; sector++) {
```
jluebbe, line 2529: This should use `goffset` instead of `guint`, as it's operating on file offsets. Also it's not iterating over sectors (which are usually either 512 or 4Ki bytes on Linux), but 64kiB. Perhaps just call it `i` or `chunk`.

src/bundle.c
```diff
@@ -2502,6 +2502,50 @@ gboolean load_manifest_from_bundle(RaucBundle *bundle, RaucManifest **manifest,
...
+		r = pread(fd, buf, sizeof(buf), sector*sizeof(buf));
+		if (r == 0)
+			break;
+		else if (r < 0)	{
+			int err = errno;
+			g_set_error(error,
+					G_FILE_ERROR,
+					g_file_error_from_errno(err),
+					"Check whole %s device failed: %s", dev, g_strerror(err));
```
jluebbe, line 2538: This could print the failing offset, to help with debugging. See `G_GOFFSET_FORMAT` for formatting these values.

jluebbe (changes requested):

jluebbe:
@chhitz Would a config option in system.conf to enable pre-checking for install and mount (instead of a per-operation option) work for you?

chhitz:
> @chhitz Would a config option in system.conf to enable pre-checking for install and mount (instead of a per-operation option) work for you?

Yes, this would work for us as well. We always want this behavior.

src/bundle.c
chhitz: removed

src/bundle.c
chhitz: Now prints range within the corruption occured

chhitz:
> > @chhitz Would a config option in system.conf to enable pre-checking for install and mount (instead of a per-operation option) work for you?
> 
> Yes, this would work for us as well. We always want this behavior.

I reworked the PR to use a config option.

docs/reference.rst
```diff
@@ -192,6 +192,13 @@ Example configuration:
   If set, neither ``variant-name`` nor ``variant-file`` must be set.
   Refer chapter :ref:`sec-variants` for more information.
 
+.. _perform-pre-check:
+
+``perform-pre-check``
+  This boolean value controls if a RAUC bundle is read completely before it is
+  mounted. This prevents RAUC from trying to install a corrupt bundle. Its 
+  default value is ``false`` which means that this check is not performed.
```
jluebbe, line 200:
```suggestion
``perform-pre-check``
  For ``verity`` and ``crypt`` bundles, this boolean value controls whether the complete
  bundle is checked for data corruption before it is mounted.
  Normally, this option is not needed as every access to the bundle payload during
  installation is already protected by ``dm-verity``.
  The default value is ``false`` which means that this pre-check is not performed.
  
  This option is useful when the installation should be aborted early even if the corrupt
  part of the bundle is not used during installation (perhaps due to adaptive updates or
  image variants).

  It has no effect for ``plain`` bundles, as the signature verification already checks the
  whole bundle.  
```

jluebbe (changes requested):
Just two minor comments. Looks good otherwise.

src/bundle.c
```diff
@@ -2738,6 +2766,15 @@ gboolean mount_bundle(RaucBundle *bundle, GError **error)
...
 
+		if (r_context()->config->perform_pre_check) {
+			res = read_complete_dm_device(loopname, &ierror);
+			if (!res) {
+				g_propagate_error(error, ierror);
+				ierror = NULL;
+				goto umount;
+			}
+		}
+
```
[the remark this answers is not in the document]
chhitz: Is removed

jluebbe (approved):

jluebbe:
Thanks for your contribution!

PR #1107, opened 2023-04-05
Rework Parts of Slot State Determination

ejoerns:
This splits up slot state determination and external mount point detection and ensures slot state determination is called quite early in the service.

This removes unnecessary re-determination of static data and allows us to access slot state information earlier in the future.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -85,7 +85,7 @@ static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 	return g_strchomp(content);
 }
 
-static gboolean update_external_mount_points(GError **error)
+gboolean update_external_mount_points(GError **error)
 {
 	GList *mountlist = NULL;
 	GHashTableIter iter;
--- a/src/install.c
+++ b/src/install.c
@@ -85,25 +85,12 @@ static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 	return g_strchomp(content);
 }
 
-gboolean determine_slot_states(GError **error)
+static gboolean update_external_mount_points(GError **error)
 {
-	g_autoptr(GList) slotlist = NULL;
 	GList *mountlist = NULL;
-	RaucSlot *booted = NULL;
 	GHashTableIter iter;
 	RaucSlot *slot;
 	GError *ierror = NULL;
-
-	g_assert_nonnull(r_context()->config);
-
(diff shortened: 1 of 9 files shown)


src/main.c
```diff
@@ -299,6 +299,13 @@ static gboolean install_start(int argc, char **argv)
 			goto out_loop;
 		}
 	} else {
+		if (!determine_slot_states(&error)) {
```
jluebbe, line 302: The commit message doesn't mention that this is copied here for the non-service install case.

test/rauc.t
```diff
@@ -1297,6 +1297,7 @@ test_expect_success ROOT,!SERVICE "rauc install (no service)" "
   test ! -s ${SHARNESS_TEST_DIRECTORY}/images/rootfs-1 &&
   rauc \
     --conf=${SHARNESS_TEST_DIRECTORY}/minimal-test.conf \
+    --override-boot-slot=system0 \
```
jluebbe, line 1300: With this change, do we still have tests for parsing the cmdline?

jluebbe (changes requested):

test/rauc.t
ejoerns: We have a few explicit ones for `/dev/nfs` and `rauc.external` in `test/context.c`. Maybe these should be extended.

In sharness tests we don't have any as far as I see.

The ones that I have changed were more accidentally than intentionally testing /proc/cmdline.
The alternative would be to add explicit tests that depend on running in qemu or to expect all tests to run in qemu and remove all override-boot-slots.

src/main.c
ejoerns: Not sure if I understand what you mean. It says:

>    An explicit call to determine_slot_states() needs to be re-added for
>    install_start() in the non-service case.

test/rauc.t
ejoerns: I've understood this as a request to add more tests and provided them now.

src/install.c
```diff
@@ -140,13 +125,32 @@
 	}
 	g_list_free_full(mountlist, (GDestroyNotify)g_unix_mount_free);
 
+	return TRUE;
+}
+
+gboolean determine_slot_states(GError **error)
```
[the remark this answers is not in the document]
ejoerns: fixed..

PR #1109, opened 2023-04-05
utils: add helper to atomically update symlinks

jluebbe:
This will be used for artifact updates.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -562,3 +562,46 @@ gchar *r_prepare_env_key(const gchar *key, GError **error)
 
 	return g_steal_pointer(&result);
 }
+
+gboolean r_update_symlink(const gchar *target, const gchar *name, GError **error)
+{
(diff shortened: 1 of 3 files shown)


src/utils.c
```diff
@@ -562,3 +562,46 @@ gchar *r_prepare_env_key(const gchar *key, GError **error)
 
 	return g_steal_pointer(&result);
 }
+
+gboolean r_update_symlink(const gchar *name, const gchar *target, GError **error)
```
ejoerns, line 566: Wouldn't

```suggestion
gboolean r_update_symlink(const gchar *target, const gchar *name, GError **error)
```

be more intuitive for this since it matches the argument order of `symlink()`?

jluebbe: OK. Will change.

ejoerns (approved):

PR #1117, opened 2023-05-05
src/bundle: allow exFAT as local file system

lategoodbye:
At least since Linux 5.7 the Kernel supports the exFAT file system. So allow it to use modern flash drives as local file system in rauc.

The changes has been tested on a i.MX6ULL SoC embedded device called Tarragon (Linux 6.1) with a common flash drive (Sandisk Ultra USB 3.0 64 GB).

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -33,6 +33,9 @@
 #ifndef CRAMFS_MAGIC
 #define CRAMFS_MAGIC 0x28cd3d45
 #endif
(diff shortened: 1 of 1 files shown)


jluebbe (approved):

jluebbe:
Thanks for your contribution, @lategoodbye.

Pleasen note that the `verity` bundle format doesn't require this check and protects against TOCTOU attacks using an emulated USB storage device as well, so I'd suggest to use it instead of the old `plain` format.

PR #1118, opened 2023-05-08
Allow to skip bundle verification

ensc:
During development, bundles are often signed by self signed certificates which change e.g. with every run of the CI.

To test such bundles on existing systems, it is required to add the used certificate to /etc/rauc/ca.cert.pem.  This is cumbersome because /etc might be read-only or the used certificate might not be available.

Patch adds a '--no-verify' option to 'install' which allows to skip bundle verification.

Support for this option is disabled by default to reduce the impact of other configuration problems.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -58,6 +58,16 @@ AS_IF([test "x$enable_create" != "xno"], [
         AC_DEFINE([ENABLE_CREATE], [0])
 ])
 
+AC_ARG_ENABLE([no-verify],
+        AS_HELP_STRING([--enable-no-verify], [Allow to skip bundle verification during installation]),
+	[], [enable_no_verify=no]
+)
+AS_IF([test "x$enable_no_verify" != "xno"], [
+        AC_DEFINE([ALLOW_NO_VERIFY], [1], [Define to 1 to allow skipping bundle verification during installation])
+], [
+        AC_DEFINE([ALLOW_NO_VERIFY], [0])
+])
+
 # Checks for libraries.
 PKG_CHECK_MODULES([GLIB], [glib-2.0 >= 2.50 gio-2.0 gio-unix-2.0])
 # Sanity checks to not use new glib methods unintentionally.
@@ -285,4 +295,5 @@ AC_MSG_RESULT([
 
         network:                ${enable_network}
         streaming:              ${enable_streaming}
+        no-verify:              ${enable_no_verify}
 ])
--- a/include/install.h
+++ b/include/install.h
@@ -34,6 +34,7 @@ typedef struct {
 	gint status_result;
 	/* install options */
 	gboolean ignore_compatible;
+	gboolean no_verify;
 	RaucBundleAccessArgs access_args;
 } RaucInstallArgs;
 
--- a/meson.build
+++ b/meson.build
@@ -49,6 +49,7 @@ systemddep = dependency('systemd', required : false)
 conf.set10('ENABLE_SERVICE', get_option('service'))
 conf.set10('ENABLE_CREATE', get_option('create'))
 conf.set10('ENABLE_JSON', jsonglibdep.found())
+conf.set10('ALLOW_NO_VERIFY', get_option('allow-no-verify'))
 
 c_warn_flags = '''
   -Wundef
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -29,6 +29,11 @@ option(
   type : 'feature',
   value : 'auto',
   description : 'Enable/Disable GPT support')
+option(
+  'allow-no-verify',
+  type : 'boolean',
+  value : 'false',
+  description : 'Allow to skip bundle verification during installation')
 
 # other options
 option(
--- a/src/install.c
+++ b/src/install.c
@@ -1088,6 +1088,7 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
 	gboolean res = FALSE;
 	g_autoptr(RaucBundle) bundle = NULL;
 	GHashTable *target_group;
+	CheckBundleParams params = CHECK_BUNDLE_DEFAULT;
 
 	g_assert_nonnull(bundlefile);
 	g_assert_null(r_context()->install_info->mounted_bundle);
@@ -1103,10 +1104,13 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
 		goto out;
 	}
 
+	if (args->no_verify)
+		params |= CHECK_BUNDLE_NO_VERIFY;
+
 	// TODO: mount info in context ?
 	install_args_update(args, "Checking and mounting bundle...");
 
-	res = check_bundle(bundlefile, &bundle, CHECK_BUNDLE_DEFAULT, &args->access_args, &ierror);
+	res = check_bundle(bundlefile, &bundle, params, &args->access_args, &ierror);
 	if (!res) {
 		g_propagate_error(error, ierror);
 		goto out;
--- a/src/main.c
+++ b/src/main.c
@@ -261,7 +261,8 @@ static gboolean install_start(int argc, char **argv)
 		g_auto(GVariantDict) dict = G_VARIANT_DICT_INIT(NULL);
 
 		g_variant_dict_insert(&dict, "ignore-compatible", "b", args->ignore_compatible);
-		g_variant_dict_insert(&dict, "no_verify", "b", args->no_verify);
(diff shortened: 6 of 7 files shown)


jluebbe:
This is intentional. We don't want to have any code in the install path to disable verification.

Furthermore, for testing, it's much better to test the same code paths in CI as in the real product, just with different key material.

ensc:
> Furthermore, for testing, it's much better to test the same code paths in CI as in the real product, just with different key material.

Here, three different kind of bundles are existing:

- productive builds in CI (protected branches); stable key pair
- test builds in CI (unprotected branches; e.g. merge requests); random key pair (CI does not allow per developer keys here)
- local builds; stable key pair (but per developer)

Going to productive builds from local and test builds is possible by adding its public key.

But switching between local and CI test builds is extremely cumbersome and requires manual modification of `/etc/rauc/ca.cert.pem`.  This kind of installation (local <-> test or test <-> test) is the most common one during development and it would be really nice to simplify it.

ejoerns:
@ensc How do you build your bundle? Is this part of the BSP build? Or done separately?

My recommendation would otherwise be to just have the development keys available for everyone as part of the BSP (or from any other shared source) so that they do not have to be (re-)generated individually.

These are then on a security level of 'no keys', but (as @jluebbe said) you always develop and test with the signing and verification flow / code path that is quite essential for RAUC.

ensc:
> How do you build your bundle? Is this part of the BSP build

yes; OE with meta-rauc and `bundle.bbclass`.

> just have the development keys available for everyone as part of the BSP

I think, publishing any key material publicly is a bad idea because at some point, somebody will start to use it for convenience for productive purposes.

> you always develop and test with the signing and verification flow / code path that is quite essential for RAUC.

I have to trust that verification in RAUC works.  Testing for false positives (accepting bundles signed by another authority) is usually not part of normal workflow.


The develop flow is now:

- look for `rauc.crt` in the build artifacts and download them
- `cat rauc.crt` + mark everything with mouse
- `mount -o remount,rw /` on the target
- `cat >> /etc/rauc/ca.cert.pem` + double click
- run `rauc install`

With patch and when functionality is explicitly enabled, it will be

- run `rauc install --no-verify`

In both cases, there is a difference to the workflow in productive systems.

ejoerns:
> > How do you build your bundle? Is this part of the BSP build
> 
> yes; OE with meta-rauc and `bundle.bbclass`.

Ok, thanks for clarifying!

> > just have the development keys available for everyone as part of the BSP
> 
> I think, publishing any key material publicly is a bad idea because at some point, somebody will start to use it for convenience for productive purposes.

I understand your concerns but would also argue that it is a matter of release process that one should have under control. If security/verification plays a role, one must care for key handling anyway. If not, one is probably fine with shipping BSP global keys).

I guess the same argumentation of 'someone will start to use it for production' applies to introducing a `--no-verify` option, too.
And, in addition, you need to maintain different *build-time* configurations between what you develop with and what you ship. I fear this is more error-prone.

> > you always develop and test with the signing and verification flow / code path that is quite essential for RAUC.
> 
> I have to trust that verification in RAUC works. Testing for false positives (accepting bundles signed by another authority) is usually not part of normal workflow.

It was less meant like testing RAUC itself with this. It is more about consistently using and testing the whole workflow and handling in your setup.

I understand that in your use case the CI serves as the line of defense against using developer features in the field, but I am unsure if this applies to others, too.

> The develop flow is now:
> 
> * look for `rauc.crt` in the build artifacts and download them
> * `cat rauc.crt` + mark everything with mouse
> * `mount -o remount,rw /` on the target
> * `cat >> /etc/rauc/ca.cert.pem` + middle click
> * run `rauc install`

Yes, I go with you that with the given restrictions this is as necessary as it is inconvenient.

> With patch and when functionality is explicitly enabled, it will be
> 
> * run `rauc install --no-verify`
> 
> In both cases, there is a difference to the workflow in productive systems.

True. But I would still argue that the manual handling is less prone to land in the final product.

Personally, I am unsure how to proceed here so that it fits both your's and RAUC's needs. For the background: We have a longer history of discussing and rejecting similar requests that could (accidentially) lead to introducing code paths to bypass signature verification for installation. I hope that you understand our general concerns here.

RAUC was also not intended to be a developer tool but of course we also find it very convenient to use it for deploying newly in-development software states to devices :wink:

jluebbe:
> The develop flow is now:
> 
> * look for `rauc.crt` in the build artifacts and download them
> * `cat rauc.crt` + mark everything with mouse
> * `mount -o remount,rw /` on the target
> * `cat >> /etc/rauc/ca.cert.pem` + middle click
> * run `rauc install`

When using a separate set of development key pairs, you'd use a normal `rauc install http://build-server.example/../foo.raucb`. That will use the exact same RAUC binaries and code paths as installation in the field. This is critical, as any issue in the update installation process that is missed during development can lead to systems in the field which boot successfully after an update (and run `rauc status mark-good`) but are unable to install further updates. Using the same code paths as in the field during development exposes this kind of issues much earlier.

The approach of using an explicit development key pair for each production key pair is especially relevant when using secure/verified boot. Without integrating a well-known set of development keys, it's impossible to have development devices which always exercise the full verified boot flow, but are still usable for all developers, even for code built locally.

When testing/developing on (unlocked) devices which boot any software, a significant part of the system's complexity is unused (and untested) until it reaches devices fused with production keys, which means that errors are discovered late and expensive to fix. See [signing.bbclass](http://cgit.openembedded.org/meta-openembedded/tree/meta-oe/classes/signing.bbclass) in meta-oe and my [meta-code-signing](https://github.com/jluebbe/meta-code-signing/) for an example how this can be used.


With the current approach, RAUC's behavior is simple: Only bundles signed with keys in the keyring can be installed, no exceptions. This simplicity is a very useful property to have for security critical software. RAUC's design is intentionally opinionated and we think this is a benefit compared to alternatives. :)

So while we won't change the behavior of `rauc install`, I think a `rauc info <bundle> --no-verify --extract-certs=/tmp/extracted-certs/` would be acceptable. The user would still need to copy a cert from that directory to the keyring manually, making the security implications obvious. Each certificate from the chain should probably go into a separate file.

ensc:
> When using a separate set of development key pairs, you'd use a normal `rauc install http://build-server.example/../foo.raucb`
>
> Using the same code paths as in the field during development exposes this kind of issues much earlier.

Noble goal but this will never be the case.  Development code paths **are** different from these in the field.

E.g. in the field you will never type `rauc install` and run it.  There will always be some frontend or other mechanism which triggers the update process.  Typing `rauc install` in the console is a development task.

I agree that final code paths must be tested somehow.  But it must be also possible, to do development on other parts of the systems without reconfiguring it or being forced to do dozens of manual steps before.


> The approach of using an explicit development key pair for each production key pair is especially relevant when using secure/verified boot.

Testing secure boot is an important task.  But development which is unrelated to it should not be hindered by it.


> With the current approach, RAUC's behavior is simple

Well, the behaviour might be simple.  But its use becomes very difficult...

jluebbe:
> > When using a separate set of development key pairs, you'd use a normal `rauc install http://build-server.example/../foo.raucb`
> > Using the same code paths as in the field during development exposes this kind of issues much earlier.
> 
> Noble goal but this will never be the case. Development code paths **are** different from these in the field.

Depends on the project. There may be requirements which mean that 100% cannot be reached in a given project, but one can certainly avoid may differences and get very close.

> E.g. in the field you will never type `rauc install` and run it. There will always be some frontend or other mechanism which triggers the update process. Typing `rauc install` in the console is a development task.

When using the D-Bus service, both ways result in identical D-Bus calls and use the same code paths in the service.

> I agree that final code paths must be tested somehow. But it must be also possible, to do development on other parts of the systems without reconfiguring it or being forced to do dozens of manual steps before.
> 
> > The approach of using an explicit development key pair for each production key pair is especially relevant when using secure/verified boot.
> 
> Testing secure boot is an important task. But development which is unrelated to it should not be hindered by it.

I very much agree that development of unrelated parts of the system must not be hindered, as that usually leads to local workarounds invented by individual developers, defeating the purpose of a deterministic build and test process.

> > With the current approach, RAUC's behavior is simple
> 
> Well, the behaviour might be simple. But its use becomes very difficult...

We've suggested a way (separate development key pairs), which means that the update authentication is completely transparent to developers working on development devices. They can simply use RAUC as on the production systems, while not needing access to confidential key material.

So far, we've used this approach successfully in dozens of customer projects.

PR #1119, opened 2023-05-09
Fix memory leak of manifest hash

securitykernel:
Fix leaking memory of the manifest hash. Found using libFuzzer + address sanitizer.


ejoerns:
@securitykernel Good catch! Seems I introduced this with 3a58163767d370ca7e4605663a6dc5fcb0a1ced0. Could you add your [Signed-Off-by](https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin) to the commit message?

jluebbe:
@securitykernel Is the setup you use to run libFuzzer published somewhere? I think it would be useful to add this to the main repository, so it's easier to run regularly.

securitykernel:
> @securitykernel Good catch! Seems I introduced this with [3a58163](https://github.com/rauc/rauc/commit/3a58163767d370ca7e4605663a6dc5fcb0a1ced0). Could you add your [Signed-Off-by](https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin) to the commit message?

Sure. Force pushed with a signed-off-by now.

securitykernel:
> @securitykernel Is the setup you use to run libFuzzer published somewhere? I think it would be useful to add this to the main repository, so it's easier to run regularly.

It is a work-in-progress, but I definitely want to contribute it. The memory leak was detected when libFuzzer loaded a valid manifest also used by the test suite, so I think it would be useful to run the tests with sanitizers enabled in the first iteration, e.g., add a `.github/workflow/sanitizers.yml`, and then add jobs for ASan, UBSan, TSan and the like. I got lots of false positives in glib when I did this locally, probably because their allocation code is not instrumented, so the sanitizer builds likely would have to use a custom glib. I need to verify if this helps.

jluebbe (approved):
I've adjusted the commit title style and added a fixes tag via a force push.

ejoerns (approved):

jluebbe:
> I got lots of false positives in glib when I ran just ASan without libFuzzer locally, probably because their allocation code is not instrumented, so the sanitizer builds likely would have to use a custom glib.

You may need to set some options. `G_SLICE=always-malloc G_DEBUG=gc-friendly ASAN_OPTIONS=malloc_context_size=64:fast_unwind_on_malloc=0 build/test/manifest-test` seems to work for me.

jluebbe:
@securitykernel Now that we have fixed a large part of the leaks in the test suite, could you perhaps share your libFuzzer setup?

securitykernel:
@jluebbe Sure, I'll prepare a PR.

Issue #1028, opened 2022-12-27
rauc extract is not extracting rauc bundle.


samannazz:
Creating a Bundle (Host):

step 1: `mkdir content-dir/`

step 2: adding a custom rootfs.ext4  → {rauc/content-dir}, because `cp $SOURCE/rootfs.ext4 content-dir/` is not working.
 

step 3: creating manifest.raucm

```
cat >> content-dir/manifest.raucm << EOF
[update]
compatible=FooCorp Super BarBazzer
version=2019.01-1
[image.rootfs]
filename=rootfs.ext4
EOF
```

step 4: creating key and cert. {/rauc}

`openssl req -x509 -newkey rsa:4096 -nodes -keyout demo.key.pem -out demo.cert.pem -subj "/O=rauc Inc./CN=rauc-demo"
`
![Screenshot from 2022-12-27 15-08-46](https://user-images.githubusercontent.com/89451747/209658227-b98574d1-58b2-420d-8568-6087101c1545.png)

step 5: creating a bundle from this: {/rauc}

`rauc --cert demo.cert.pem --key demo.key.pem bundle content-dir/ update-2019.01-1.raucb
`
output:
![Screenshot from 2022-12-27 14-57-01](https://user-images.githubusercontent.com/89451747/209658349-7a528854-163e-4a32-b713-3db2e877df04.png)

The output of the command indicates that the RAUC bundle is being created in the "plain" format, which is a simple format that includes all of the files and directories in the bundle without any additional metadata or compression.

`rauc-Message: 13:59:54.136: No keyring given, skipping signature verification` indicates that the RAUC was unable to find a keyring to use for verifying the signature of the RAUC bundle.

How can i create keyring?

Step 6: now i am extracting ruac bundle so that i can launch virtual machine i.e. qemu.

![Screenshot from 2022-12-27 14-05-50](https://user-images.githubusercontent.com/89451747/209658553-066e718e-6e19-4290-b764-0ecd2e254379.png)

How can I solve this problem? Maybe it is because of the custom rootfs.ext4, or maybe I haven't created a keyring?

samannazz:
this error resloved by using this:
`sudo rauc extract -d update-2019.01.1.raucb --keyring deo.cert.pem /tmp/bundle`

ejoerns:
@samannazz The error message

> FATAL ERROR:dir_scan: failed to make directory output, because File exists

[...]

PR #1123, opened 2023-05-11
src/bundle: ensure that the output directory does not exist before extracting

jluebbe:
Otherwise, unsquashfs will just fail with a possibly confusing error message.

Fixes: #1028

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -2425,6 +2425,12 @@ gboolean extract_bundle(RaucBundle *bundle, const gchar *outputdir, GError **err
 
 	r_context_begin_step("extract_bundle", "Extracting bundle", 2);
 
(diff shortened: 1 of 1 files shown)


src/bundle.c
```diff
@@ -2425,6 +2425,11 @@ gboolean extract_bundle(RaucBundle *bundle, const gchar *outputdir, GError **err
 
 	r_context_begin_step("extract_bundle", "Extracting bundle", 2);
 
+	if (g_file_test(outputdir, G_FILE_TEST_EXISTS)) {
+		g_set_error(error, G_FILE_ERROR, G_FILE_ERROR_EXIST, "output directory %s exists already", outputdir);
```
ejoerns, line 2429: Maybe set `res = FALSE` here for clarity and to avoid potential future issues?

ejoerns (approved):

PR #1125, opened 2023-05-11
recommend verity bundles by default

jluebbe:
There is no reason to start new projects with plain format bundles, so
recommend against it in the docs and when creating new bundles.
Also point to the migration section for existing projects.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -208,6 +208,8 @@ static gboolean parse_manifest(GKeyFile *key_file, RaucManifest **manifest, GErr
 				"  To silence this warning, select the 'plain' format explicitly."));
 		g_ptr_array_add(raucm->warnings, g_strdup(
 				"  See https://rauc.readthedocs.io/en/latest/reference.html#sec-ref-formats for details.'"));
+	} else {
+		raucm->bundle_format_explicit = TRUE;
 	}
 	if (tmp == NULL || g_strcmp0(tmp, "plain") == 0) {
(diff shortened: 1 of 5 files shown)


jluebbe:
Rebased onto master.

src/manifest.c
```diff
@@ -198,6 +201,14 @@ static gboolean parse_manifest(GKeyFile *key_file, RaucManifest **manifest, GErr
 	tmp = key_file_consume_string(key_file, "bundle", "format", NULL);
 	if (tmp == NULL || g_strcmp0(tmp, "plain") == 0) {
 		raucm->bundle_format = R_MANIFEST_FORMAT_PLAIN;
+		g_ptr_array_add(raucm->warnings, g_strdup(
```
ejoerns, line 204: Two issues here:
* The first is that when being within this if statement, the warning will still be printed even if 'plain' was selected explicitly. This is not what the warning message suggests.

  This should be solvable by handling `tmp == NULL` and `g_strcmp0(tmp, "plain") == 0)` in different if statements and emitting the warning only on `tmp == NULL`.
* The second is that the manifest parsing is not idempotent on a `format=plain` since when writing the manifest back, a `format=plain` will just be ignored and the `[bundle]` section will thus be deleted. This means even if we had set `format=plain` explicitly, a second call would re-trigger the warning.

  If the parsing would always write `format=plain` then we would have the same issue but the other way around.

ejoerns (approved):
Looks like a good solution to both problems mentioned! :+1:

PR #1126, opened 2023-05-12
fix some leaks and add a address-sanitizer workflow

jluebbe:
For now, there are still several tests for which address sanitizer reports leaks, but we can still check for regressions in the others.

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,51 @@
+name: sanitizers
+
+on: [push, pull_request]
+
+jobs:
+  address:
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Inspect environment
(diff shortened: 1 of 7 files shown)


.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,49 @@
...
+    - uses: actions/checkout@v3
+
+    - name: Run meson
+      run: |
+        meson setup build -Db_sanitize=address
+        meson compile -C build
+
+    - name: Run selected tests
+      run: |
+        export G_SLICE=always-malloc G_DEBUG=gc-friendly
```
ejoerns, line 28: Maybe we should add some hints in the commit message about why these are required or where this is inspired from?

ejoerns (approved):

PR #1143, opened 2023-05-23
fix memory leaks found by address sanitizer and enable even more tests

jluebbe:


test/common.h
```diff
@@ -4,7 +4,7 @@
 
 #include "manifest.h"
 
-#define memdup(x) (g_memdup(x, sizeof(*x)))
```
ejoerns, line 7: This seems to break the test suite (test/nbd-test).

jluebbe: Should be fixed by the new approach.

test/common.c
```diff
@@ -408,3 +408,26 @@ void replace_strdup(gchar **dst, const gchar *src)
 	g_free(*dst);
 	*dst = g_strdup(src);
 }
+
+void* dup_test_mem(GPtrArray *ptrs, const void *mem, gsize len)
```
ejoerns, line 412: Maybe it should somehow reflect that something is added to the list of 'to free' candidates? But I also don't really have a good name that covers both the copying and the freeing aspect...

ejoerns (commented):
Looks really good overall! Just some minor comments.

test/bundle.c
```diff
@@ -67,17 +67,17 @@ static void prepare_bundle(BundleFixture *fixture, gconstpointer user_data)
 static void bundle_fixture_set_up_bundle(BundleFixture *fixture,
 		gconstpointer user_data)
 {
-	r_context_conf()->certpath = g_strdup("test/openssl-ca/dev/autobuilder-1.cert.pem");
-	r_context_conf()->keypath = g_strdup("test/openssl-ca/dev/private/autobuilder-1.pem");
+	replace_strdup(&r_context_conf()->certpath, "test/openssl-ca/dev/autobuilder-1.cert.pem");
+	replace_strdup(&r_context_conf()->keypath, "test/openssl-ca/dev/private/autobuilder-1.pem");
```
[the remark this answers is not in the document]
jluebbe: I'd like to do that refactoring separately, as it needs a much more thorough review (compared to these simple changes).

test/common.c
jluebbe: Yes, I've thought about that for some time as well and this is the best I came up with. Can we keep it for now?

jluebbe:
I've force pushed this PR. Only the commit order, separation and messages have changed, so the diff is empty.

jluebbe:
I've rebased on master to ensure no new leaks have been introduced in the meantime. There are no changes to the commits in this PR.

test/bundle.c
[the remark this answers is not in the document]
ejoerns: Ok. In order to unblock follow-up work (and as it already allows us to enable more sanity tests), I guess we can merge it like it is first of all.

ejoerns (approved):

PR #1144, opened 2023-05-31
src/main: add missing --key argument for 'rauc info'

ejoerns:
For the decryption use case, this specifies the decryption key.
The global `--key` option has been replaced by command-specific ones but the purpose of `--key` for 'rauc info' has not been considered.

Since the former global options are still available but hidden, using the options always worked but it has not been shown in the help output or man page.

Fixes #875.

### DIFF
--- a/rauc.1
+++ b/rauc.1
@@ -164,6 +164,10 @@ Extract the bundle content to a directory.
 .RS 4
 
 .TP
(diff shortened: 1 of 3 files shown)


jluebbe:
Don't we need this for `extract` and `extract-signature` as well? ( see https://github.com/rauc/rauc/blob/46ceb4a1a03e36c425c9b6fc2597dc3231aff26d/src/main.c#L2463-L2466)

Issue #1070, opened 2023-02-21
Concept: Configurable Event Logging


ejoerns:
So far, RAUC provides a status file where information about the latest installation is stored.
This is useful to know what is currently on the system but does not (or only partly) answer common questions such as:

* What has been installed on this device before?
* How many times did the installed system boot (successfully)?
* When has which redundancy group been activated?
* Did we run into a fallback?

A proper solution to gain more history-related information would be to establish a dedicated logging mechanism in RAUC.

Since the actual events to log might differ between different use cases as well as the logging policy may differ, we propose a configurable event logging framework in RAUC here.

The key idea is to have different loggers configurable in the system configuration file. Each logger can specify the events to log and where to write the log. A good default location for logging would be a file in RAUC's data-dir.
Additional options may define log rotation handling, compression, or fsync() handling.
Default output should be a human-readable format, but machine-readable formats may follow.

An initial draft of system.conf sections might look as follows:

``` ini
[log.boot]
filename=boot.log # relative in data-dir
events=booted-version
max-size=1M
max-files=5
compress="xz -9"

[log.install]
filename=/srv/rauc/install.log
events=install
sync=true # after each write: flush und fsync
max-size=0 # no size limit
max-files=0 # no rotation
```

Internally an event log message will contain meta information that can be used for printing, filtering, etc.
As these event log messages will replace parts of the current log messages, we can use the additional meta information to also extend the information sent to the systemd-journal, etc. E.g. with using a message [catalog](https://www.freedesktop.org/wiki/Software/systemd/catalog/) for providing more details.

The high-level structure of a log item might look as follows:

  * Log item:
      * timestamp
      * event type
      * event ID  (MESSAGE_ID=$(systemd-id128 new))
      * payload

A failure in logging should not be fatal for the installation process but might disable the corresponding log handler.

Two pieces of information that will be relevant for answering parts of the above-noted questions from looking at the logs are still missing in RAUC and need to be added:

**Transaction ID:** A unique ID that allows unambiguously identifying a specific installation process. It will change for each install invocation and might either be provided by the user or will be auto-generated if not.

**Manifest Hash:** Allows unambiguously identifying a bundle with all its content (for verity bundles that contain the verity hash in the manifest).

ejoerns:
FTR: After some pre-work, partly referenced here, the base implementation for the event logging noted in this concept was submitted as PR #1230.

[...]

PR #1150, opened 2023-06-05
src: add installation transaction UUID

ejoerns:
It can either be provided manually (via D-Bus or command line arg) or will be auto-generated.

The transaction UUID allows us to identify and track an installation run.

It is an important building block for adding future features like history logging, life cycle tracing, and more.

First of all, add a simple printout of the UUID when the installation starts so we can follow it in the logs.

Also see #1070

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -944,6 +944,7 @@ termed with the slot name (e.g. [slot.rootfs.1]) for the central status file:
   status=ok
   sha256=b14c1457dc10469418b4154fef29a90e1ffb4dddd308bf0f2456d436963ef5b3
   size=419430400
+  installed.transaction=dad3289a-7de1-4ad2-931e-fb827edc6496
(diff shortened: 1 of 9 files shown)


jluebbe (approved):

jluebbe:
@ejoerns Do you have plans on how to expose this via D-Bus?

ejoerns:
@jluebbe Not sure if I understand you correctly. It is exposed via `GetSlotStatus` method and can be set over D-Bus via `InstallBundle` method. Where else would you expect it?
Do you mean for later integration into a life-cycle flow with a deployment server? The first guess would be to add it to the `Completed` signal, but this cannot be extended. Thus something else might be required, but I guess there is more to be added to the D-Bus API anyway such as methods for manual confirmation, etc.

jluebbe:
I meant in the case where RAUC generates a random transaction UUID. Yes, it will probably need API changes for full integration later.

PR #1152, opened 2023-06-06
Add System Information HTTP Headers

ejoerns:
This implements passing system information upon the first HTTP request when accessing the URL of a streaming bundle.

This allows the server to collect more information about the device in order to create more verbose logs or to actually build deployment logic on this.

Besides some standard information that is configurable, custom information can be provided by the `system-info` handler.

This implements parts of concept #1114

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -478,6 +478,22 @@ round-trip-time (RTT) lead to longer installation times.
 This can be compensated somewhat by using a HTTP/2 server, as this supports
 multiplexing and better connection reuse.
 
+Additional HTTP Header Information
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Upon first HTTP request, RAUC can expose some additional information about the
+system in HTTP headers.
+This allows the receiving server to log this information or to build some
+simple logic and rollout handling on it.
+
+The actual information exposed to the server is configurable by
+``send-headers`` option in the :ref:`[streaming] section
+<streaming-config-section>` of ``system.conf``.
+
+Beside some standard information, like the *boot ID*, the system's *uptime* or
+the *installation transaction ID*, one can also expose custom information
+provided by the ``system-info`` :ref:`handler <sec_ref_handlers>`.
+
 .. _sec-encryption:
 
 Bundle Encryption
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -298,6 +298,22 @@ For more information about using the streaming support of RAUC, refer to
   This option can be used to set the path of the CA certificate which should be
   used instead of the system wide store of trusted TLS/HTTPS certificates.
 
+``send-headers``
+  This option takes a ``;``-separated list of information to send as HTTP
+  header fields to the server with the first request.
+
+  Supported values are:
+
+  * ``boot-id``: Enables sending the *boot_id* as ``RAUC-Boot-ID`` header field.
+  * ``machine-id``: Enables sending the *machine-id* as ``RAUC-Machine-ID`` header field.
+
+    .. note:: The machine ID should be considered "confidential" and thus not
+       be used over unauthenticated connections or with untrusted servers!
+  * ``serial``: Enables sending the *system serial* as ``RAUC-Serial`` header field.
+  * ``variant``: Enables sending the *variant* as ``RAUC-Variant`` header field.
+  * ``transaction-id``: Enables sending the *transaction UUID* as ``RAUC-Transaction-ID`` header field.
+  * ``uptime``: Enables sending the system's current uptime as ``RAUC-Uptime`` header field.
+
 **[encryption]**
 
 The ``encryption`` section contains information required to decrypt a 'crypt'
@@ -359,6 +375,8 @@ performed from a dedicated (recovery) slot.
   The full path of the bundle file to check for.
   If file at ``path`` exists, auto-install will be triggered.
 
+.. _sec_ref_handlers:
+
 **[handlers] section**
 
 Handlers allow to customize RAUC by placing scripts in the system that RAUC can
@@ -388,6 +406,16 @@ See details about using handlers in `Custom Handlers (Interface)`_.
 
   System information is made available to other handlers via environment
   variables that have the exact same name and value.
+
+  The ``system-info`` handler also allows to define custom information that is
+  forwarded to the server upon RAUC's first streaming request.
+  In order to define forwarded info, this must be returned as a key prefixed.
+  with ``RAUC_HTTP_``.
+  The generated header field will be name of the key (with out the prefix)
+  where an ``RAUC-`` is prepended and all underscores are converted to
+  hyphens.
+  E.g. ``RAUC_HTTP_MY_CUSTOM_INFO=dummyvalue`` will emit a header
+  ``RAUC-MY-CUSTOM-INFO: dummyvalue``.
 
 ``pre-install``
   This handler will be called right before RAUC starts with the installation.
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -73,6 +73,8 @@ typedef struct {
 
 	gchar *systeminfo_handler;
 
+	gchar **enabled_headers; /* standard HTTP headers to send */
+
 	/* streaming */
 	gchar *streaming_sandbox_user;
 	gchar *streaming_tls_cert;
--- a/src/install.c
+++ b/src/install.c
@@ -1175,9 +1175,32 @@ static gchar* get_uptime(void)
 	return g_strdup(uptime[0]);
 }
 
+/* If the input key starts with RAUC_HTTP_, it returns a valid HTTP header
+ * string with 'RAUC_HTTP_' replaced by 'RAUC-'.
+ * If the input string does not start with RAUC_HTTP_, NULL is returned.
+ */
+static gchar *system_info_to_header(const gchar *key, const gchar *value)
+{
+	g_autofree gchar *header_key = NULL;
+
+	g_return_val_if_fail(key, NULL);
+	g_return_val_if_fail(value, NULL);
+
+	if (!g_str_has_prefix(key, "RAUC_HTTP_"))
+		return NULL;
+
+	header_key = g_strdup(key + strlen("RAUC_HTTP_"));
+	for (size_t i = 0; i < strlen(header_key); i++) {
+		if (header_key[i] == '_')
+			header_key[i] = '-';
+	}
+
+	return g_strdup_printf("RAUC-%s: %s", header_key, value);
+}
+
 static gchar **assemble_info_headers(RaucInstallArgs *args)
 {
-	GPtrArray *headers = g_ptr_array_new_full(1, g_free);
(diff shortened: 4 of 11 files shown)


docs/reference.rst
```diff
@@ -298,6 +298,21 @@ For more information about using the streaming support of RAUC, refer to
...
+
+``send-headers``
+  This option takes a list of information to send as HTTP header fields.
+  Supported values are:
+
+  * ``boot-id``: Enables sending the *boot_id* as ``BOOT-ID`` header field.
+  * ``machine-id``: Enables sending the *machine-id* as ``MACHINE-ID`` header field.
+  * ``serial``: Enables sending the *system serial* as ``SERIAL`` header field.
+  * ``variant``: Enables sending the *variant* as ``VARIANT`` header field.
+  * ``transaction-id``: Enables sending the *variant* as ``TRANSACTION-ID`` header field.
```
jluebbe, line 313:
```suggestion
  * ``transaction-id``: Enables sending the *transaction UUID* as ``TRANSACTION-ID`` header field.
```

src/nbd.c
```diff
@@ -521,11 +522,16 @@ static void start_read(struct RaucNBDContext *ctx, struct RaucNBDTransfer *xfer)
 		g_error("unexpected error from curl_multi_add_handle in %s", G_STRFUNC);
 }
 
-static struct curl_slist *gstrv_to_slist(const GStrv strv)
+/* Appends Gstrv elements to curl_slist.
+ * If curs_slist does not exist yet (NULL passed), it will be created. */
```
jluebbe, line 526: This should be explicit about who will own `initial_list`.

docs/advanced.rst
```diff
@@ -478,6 +478,21 @@ round-trip-time (RTT) lead to longer installation times.
...
+Additional HTTP Header Information
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Upon first HTTP request, RAUC can expose some additional information about the
+system in HTTP headers.
+This allows the receiving server to log this information or to build some
+simple logic and rollout handling on it.
+
+The actual information exposed to the server is configurable in the
+``system.conf`` section ``[server]``.
```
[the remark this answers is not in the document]
ejoerns: Yes, this was my second thought on this. I was unsure whether we would need more options later for phased rollouts, etc. But maybe we can even move/duplicate these options later if that should be required.

docs/reference.rst
```diff
@@ -298,6 +298,21 @@ For more information about using the streaming support of RAUC, refer to
...
 
+**[server]**
+
+Settings related to interaction with a deployment server.
+
+``send-headers``
+  This option takes a list of information to send as HTTP header fields.
+  Supported values are:
+
+  * ``boot-id``: Enables sending the *boot_id* as ``BOOT-ID`` header field.
```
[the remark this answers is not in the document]
ejoerns: Sounds like a plan, yes.

include/config_file.h
```diff
@@ -32,6 +32,15 @@ typedef enum {
...
 
+typedef struct {
+	gboolean boot_id;
+	gboolean transaction_id;
+	gboolean machine_id;
+	gboolean serial;
+	gboolean variant;
+	gboolean uptime;
+} RConfigHTTPSendHeaders;
+
```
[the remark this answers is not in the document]
ejoerns: The idea was to be more performant with integer comparison rather than requiring string comparison. Maybe it's not that relevant here. I have changed it to use string lists..

ejoerns:
Updated to address parts of the review:

* Added `X-RAUC-` prefix to headers
* use string lists for storing headers types
* added minimal config tests
* updated documentation

docs/reference.rst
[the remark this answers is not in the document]
jluebbe: Now that I've looked into this a bit more, IETF has deprecated the `X-` prefix: https://www.rfc-editor.org/rfc/rfc6648

The recommendation seems to be to prefix them with a vendor name for disambiguation. So perhaps i.e. `RAUC-Boot-ID`?

src/install.c
```diff
@@ -1159,6 +1159,85 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 	return res;
 }
 
+static gchar* get_uptime(void)
+{
+	gchar *contents = NULL;
+	g_autoptr(GError) ierror = NULL;
+
+	if (!g_file_get_contents("/proc/uptime", &contents, NULL, &ierror)) {
```
jluebbe, line 1167: `/proc/uptime` contains two values, see `proc(5)`:
```
/proc/uptime
       This  file  contains  two numbers (values in seconds): the uptime of the
       system (including time spent in suspend) and the amount of time spent in
       the idle process.
```
Did you intentionally use both?

src/install.c
```diff
@@ -1159,6 +1159,85 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
...
+		if (header_key[i] == '_')
+			header_key[i] = '-';
+	}
+
+	return g_strdup_printf("X-RAUC-%s: %s", header_key, value);
+}
+
+static gchar **assemble_info_headers(RaucInstallArgs *args)
+{
+	GPtrArray *headers = g_ptr_array_new_full(1, g_free);
```
jluebbe, line 1201: `_new_full` with a fixed, low size isn't really useful. Perhaps just use `g_ptr_array_new_with_free_func`, as this function isn't called often.

src/install.c
```diff
@@ -1159,6 +1159,85 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
...
+	GHashTableIter iter;
+	gchar *key = NULL;
+	gchar *value = NULL;
```
jluebbe, line 1204: Move these into the `if (r_context()->system_info)` block where they are used.

docs/reference.rst
```diff
@@ -298,6 +298,17 @@ For more information about using the streaming support of RAUC, refer to
   This option can be used to set the path of the CA certificate which should be
   used instead of the system wide store of trusted TLS/HTTPS certificates.
 
+``send-headers``
+  This option takes a list of information to send as HTTP header fields.
```
jluebbe, line 302: This should mention `;` as the list element separator.

docs/reference.rst
```diff
@@ -298,6 +298,17 @@ For more information about using the streaming support of RAUC, refer to
...
+  Supported values are:
+
+  * ``boot-id``: Enables sending the *boot_id* as ``X-RAUC-BOOT-ID`` header field.
+  * ``machine-id``: Enables sending the *machine-id* as ``X-RAUC-MACHINE-ID`` header field.
```
a3f, line 306: jfyi, [Man page says: ](https://www.freedesktop.org/software/systemd/man/machine-id.html):
> This ID uniquely identifies the host. It should be considered "confidential", and must not be exposed in untrusted environments, in particular on the network.

I am not sure if that warrants using `sd_id128_get_machine_app_specific` or if we can just expect the user not to download over HTTP if they care about this.

jluebbe: We should point to the documentation and mention that this header should not be used over unauthenticated connections or if the server is not trusted. In most cases, I expect that the update server needs to know the real `machine-id` tough. If needed, we could add `*-app-specific` variants to `boot-id` and `machine-id` later.

docs/reference.rst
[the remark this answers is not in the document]
ejoerns: Will change to `RAUC-` instead of `X-RAUC-`.

src/nbd.c
ejoerns: Will extend doc

src/install.c
ejoerns: Not intentionally. Will limit to using first one only

src/install.c
ejoerns: Ack.

docs/reference.rst
ejoerns: Yes, will add

docs/reference.rst
ejoerns: Will add a note for this option.

ejoerns:
Fixed leaking `enabled_headers` revealed by sanitizer CI check.

src/install.c
```diff
@@ -1159,6 +1159,99 @@
...
+			gchar *header = system_info_to_header(key, value);
+			if (header)
+				g_ptr_array_add(headers, header);
+		}
+	}
+	g_ptr_array_add(headers, NULL);
+
+	return (gchar**) g_ptr_array_free(headers, FALSE);
+}
+
```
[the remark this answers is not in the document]
ejoerns: FTR: this references `do_install_bundle()` which existed before.

ejoerns:
Rebased onto master. No further changes.

docs/reference.rst
```diff
@@ -389,6 +407,16 @@ See details about using handlers in `Custom Handlers (Interface)`_.
...
 
+  The ``system-info`` handler also allows to define custom information that is
+  forwarded to the server upon RAUC's first streaming request.
+  In order to define forwarded info, this must be returned as a key prefixed.
+  with ``RAUC_HTTP_``.
+  The generated header field will be name of the key (with out the prefix)
+  where an ``X-RAUC-`` is prepended and all underscores are converted to
+  hyphens.
+  E.g. ``RAUC_HTTP_MY_CUSTOM_INFO=dummyvalue`` will emit a header
+  ``X-RAUC-MY-CUSTOM-INFO: dummyvalue``.
```
jluebbe, line 418:
```suggestion
  The generated header field will be name of the key (with out the prefix)
  where an ``RAUC-`` is prepended and all underscores are converted to
  hyphens.
  E.g. ``RAUC_HTTP_MY_CUSTOM_INFO=dummyvalue`` will emit a header
  ``RAUC-MY-CUSTOM-INFO: dummyvalue``.
```

src/install.c
```diff
@@ -1158,6 +1158,99 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
...
+		g_warning("Failed to get uptime: %s", ierror->message);
+		return NULL;
+	}
+
+	/* file contains two values and a newline, 'chomp' in-place and split then */
+	uptime = g_strsplit(g_strchomp(contents), " ", 2);
+
+	/* hand char referenced by uptime[0] over to total_time to prevent autofreeing */
+	total_time = uptime[0];
+	uptime[0] = NULL;
```
jluebbe, line 1178: Will uptime[1] be freed this way? `g_strfreev()` docs talk about a NULL-terminated array. Simply use `g_strdup()`.

jluebbe (changes requested):

src/install.c
ejoerns: Ack

jluebbe (approved):

jluebbe:
> Note: Preview for high-level feedback. Still lacks some details.

Is this still accurate or can we merge this?

ejoerns:
> > Note: Preview for high-level feedback. Still lacks some details.
> 
> Is this still accurate or can we merge this?

Not valid any more. I've removed the comment. Can be merged as soon as we have the next point release ready.

PR #1155, opened 2023-06-10
Initializing middle

b4yuan:
Variable `middle` coule be uninitialized at the access of the variable on line 481 in `hash_index.c`; adding a line that initializes middle to 0.

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -50,6 +50,7 @@ typedef struct {
 	gchar *mountprefix;
 	gchar *bootslot;
 	gchar *boot_id;
+	gchar *machine_id;
 
 	gchar *system_serial;
 	GHashTable *system_info; /* key/values of system information */
--- a/include/context.h
+++ b/include/context.h
@@ -49,6 +49,7 @@ typedef struct {
 	/* optional global mount prefix overwrite */
 	gchar *mountprefix;
 	gchar *bootslot;
+	gchar *boot_id;
 
 	gchar *system_serial;
 	GHashTable *system_info; /* key/values of system information */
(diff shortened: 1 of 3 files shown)


jluebbe:
> Variable `middle` coule be uninitialized at the access of the variable on line 481 in `hash_index.c`; adding a line that initializes middle to 0.

Line 481 is only executed if found is true, and in that case middle is always initialized.

For consistency, we could set it to 0, though. Could you add your Signed-off-by (see https://github.com/rauc/rauc/pull/1155/checks?check_run_id=14152982601)?

ejoerns (changes requested):
Branch modification by maintainers is disabled in the branch, thus I write my proposal for a proper commit message here:

>    src/hash_index: initialize 'middle' for consistency
>
 >   Note that it was not actually used uninitialized since using it depended on
 >   having 'found' set to TRUE which can only happen in a code path that also
 >   sets 'middle' to some value.

Must not be taken exactly but at least the path prefix (`src/hash_index`) should be included.

jluebbe:
This now includes both versions of the commit message and a merged. https://github.com/rauc/rauc/pull/1155/commits should only include a single patch with the new commit message.

b4yuan:
I'm not given the option of "Allow edits and access to secrets by maintainers" in the sidebar.

jluebbe:
> I'm not given the option of "Allow edits and access to secrets by maintainers" in the sidebar.

Hmm, perhaps this is related to the use of a organisation repository as the source. I'll open a new PR with your commit to get this in.

jluebbe:
If you'd like, you can add a review on https://github.com/rauc/rauc/pull/1171.

ejoerns:
> I'm not given the option of "Allow edits and access to secrets by maintainers" in the sidebar.

You should be able to force-push your own branches, anyway. But I guess #1171 does it, too.

PR #1157, opened 2023-06-13
contrib/uboot.sh: reset BOOT_ORDER to initial value

thomsc:
When setting the BOOT_ORDER environment variable to an invalid value, the boot sequence will be broken. This will result in an endless reset loop.
To make sure that the next boot will start using a valid BOOT_ORDER, reset it to its initial value BOOT_ORDER='A B'.

### DIFF
--- a/contrib/uboot.sh
+++ b/contrib/uboot.sh
@@ -29,6 +29,7 @@ else
   echo "No valid slot found, resetting tries to 3"
   setenv BOOT_A_LEFT 3
   setenv BOOT_B_LEFT 3
+  setenv BOOT_ORDER "A B"
   saveenv
   reset
 fi


ejoerns:
@thomsc How did you end up having the `BOOT_ORDER` set to an invalid value?
RAUC changes it only if a slot is marked 'active' or if it is marked 'bad'. In the latter case, the slot is fully deactivated which might happen on purpose. E.g. when you want to prevent an old system from being booted due to security concerns. Resetting the `BOOT_ORDER` to some 'random' value would then forcefully reactivate the inactive slots.

I see that there might be valid use cases for this, mainly if having something booting is the primary goal.
Thus I am unsure what the best proposal for this example file is. Maybe others that use the U-Boot script more often than I do have an opinion on this?

thomsc:
Hi Enrico,

Basically we want to secure the RAUC boot mechanism against unauthorized attacks to bootloader environment.

We have a special case on our system as we don't use the script directly. Instead, the commands from the script are stored within the bootloader environment.

However, the basic problem is the same when using the script. As the BOOT_ORDER variable is stored in bootloader environment, somebody could gain unauthorized access to the environment and set BOOT_ORDER to an invalid value. As a result, rauc will never be able to boot the system after that.

In u-boot we are able to protect boot commands and startup scripts by reading compiled-in defaults from bootloader binary, but this is not possible with RAUC parameters (BOOT_ORDER, BOOT_A_LEFT, BOOT_B_LEFT) as these are intended to be changed. The proposed fix resets the BOOT_ORDER to its default in case of no valid image is found, just the same way it already does with BOOT_A_LEFT and BOOT_B_LEFT.

Best regards,
Thomas

Von: Enrico Jorns ***@***.***>
Gesendet: Donnerstag, 15. Juni 2023 11:21
An: rauc/rauc ***@***.***>
Cc: Thomas Schaefer ***@***.***>; Mention ***@***.***>
Betreff: Re: [rauc/rauc] contrib/uboot.sh: reset BOOT_ORDER to initial value (PR #1157)

You don't often get email from ***@***.*** Learn why this is important<https://aka.ms/LearnAboutSenderIdentification>

@thomsc<https://github.com/thomsc> How did you end up having the BOOT_ORDER set to an invalid value?
RAUC changes it only if a slot is marked 'active' or if it is marked 'bad'. In the latter case, the slot is fully deactivated which might happen on purpose. E.g. when you want to prevent an old system from being booted due to security concerns. Resetting the BOOT_ORDER to some 'random' value would then forcefully reactivate the inactive slots.

I see that there might be valid use cases for this, mainly if having something booting is the primary goal.
Thus I am unsure what the best proposal for this example file is. Maybe others that use the U-Boot script more often than I do have an opinion on this?

-

jluebbe (dismissed):

jluebbe:
> I understand that resetting the BOOT_ORDER will break the intended behavior you described. Actually I don't have an idea how to solve this.
> 
> We are using a signed bootloader which is configured to boot only a properly signed linux image. We are not using RPMB and OP-TEE so far.
> 
> From linux, bootloader environment can be read and written using a user space tool (fw_setenv). Besides the RAUC control variables we have to store a few others in bootloader environment using the tool. So there is a way to write a bad BOOT_ORDER value accidentially or intentionally even if the system is protected against block device access otherwise.

If the block device containing your root and kernel slots (perhaps an eMMC) is not write-protected by the storage device itself, an attacker able to use fw_setenv could also simply corrupt all signed linux images by direct writes. In that case, the bootloader would keep trying all slots, but will always fail to boot. This means that additional protection of the boot configuration won't guarantee a system that can always boot.

If you want to protect only against accidental modification, you could add additional checks to the top (i.e. BOOT_ORDER must either be "A", "B", "A B", "B A". Otherwise reset to the default. I'm not sure if that's a realistic scenario, though. :)

This unnecessary complexity caused by U-Boot storing everything as string key-value pairs was one of the reasons why we built https://barebox.org/doc/latest/user/bootchooser.html and https://barebox.org/doc/latest/user/state.html with fixed data types and layouts (and HMAC authentication).

> I assume that we cannot protect against this even when using RPMB and OP-TEE in our system.

Take a look at https://optee.readthedocs.io/en/latest/architecture/secure_storage.html#rpmb-secure-storage. U-Boot seems to have support to read/write variables via OP-TEE: https://patchwork.ozlabs.org/project/uboot/patch/20200422171826.139753-1-ruchika.gupta@nxp.com/

I've not used that myself, but it should be relatively easy add support for that into RAUC (either natively or via a custom [bootloader backend](https://rauc.readthedocs.io/en/latest/integration.html#custom)).

jluebbe:
@thomsc Are you fine with me closing this PR?

thomsc:
Yes, thats okay. We will check your suggestions and decide how to proceed.
Thanks for your help.

PR #1158, opened 2023-06-14
Ensure Consistent Slot Status

ejoerns:
So far, the slot status has only been written after a successful (not-skipped) slot update.

This makes it hard to identify what the actual content and state of the slot is.

Also, the existing `status=` field had only one possible value which was `"ok"`.

Besides some minor adaptions, the main aspect of this PR is to ensure the slot status gets updated even when the slot update failed or was skipped due to the 'slot skipping' mechanism.

In the first case, it also sets the status to `"failed"` to let the user know this slot is in an unusable state. When having skipped, it sets the state to `"skipped"` indicating it matches the content of the noted bundle information but was not explicitly written.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -982,13 +982,29 @@ static gboolean handle_slot_install_plan(const RaucManifest *manifest, const RIm
 	load_slot_status(plan->target_slot);
 	slot_state = plan->target_slot->status;
 
-	/* In case we failed unmounting while reading status
+	/* In case we failed unmounting while reading per-slot status
 	 * file, abort here */
 	if (plan->target_slot->mount_point) {
 		g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_MOUNTED,
 				"Slot '%s' still mounted", plan->target_slot->device);
 		r_context_end_step("check_slot", FALSE);
 		return FALSE;
+	}
+
+	/* For global slot status: Clear checksum info and make status
+	 * 'pending' to prevent the slot status from looking valid later in
+	 * case we crash while installing. */
+	if (g_strcmp0(r_context()->config->statusfile_path, "per-slot") != 0) {
+		g_clear_pointer(&slot_state->status, g_free);
+		slot_state->status = g_strdup("pending");
+		g_clear_pointer(&slot_state->checksum.digest, g_free);
+		slot_state->checksum.size = 0;
+
+		if (!save_slot_status(plan->target_slot, &ierror)) {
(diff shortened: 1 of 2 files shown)


ejoerns:
Changes:

* For slot skipping, also just mark the status as "good" since skipping should be an internal optimization of RAUC
* Ensure invalidating a slot's central status before starting writing to prevent considering a broken slot a working after a RAUC crash.

src/install.c
```diff
@@ -997,6 +997,15 @@ static gboolean handle_slot_install_plan(const RaucManifest *manifest, const RIm
 
 		/* Dummy step to indicate slot was skipped */
 		r_context_begin_step("skip_image", "Copying image skipped", 0);
+
+		/* Update the status also for skipped slots */
+		g_message("Updating slot %s status", plan->target_slot->name);
+		update_slot_status(slot_state, "ok", manifest, plan, args);
+		if (!save_slot_status(plan->target_slot, &ierror)) {
+			g_propagate_prefixed_error(error, ierror, "Error while writing status file: ");
+			return FALSE;
```
jluebbe, line 1021: Why don't we need the `r_context_end_step` here?

src/install.c
```diff
@@ -946,7 +946,7 @@ static gboolean pre_install_checks(gchar* bundledir, GPtrArray *install_plans, G
 	return TRUE;
 }
 
-static void update_slot_status(RaucSlotStatus *slot_state, const RaucManifest *manifest, const RImageInstallPlan *plan)
+static void update_slot_status(RaucSlotStatus *slot_state, const gchar* status, const RaucManifest *manifest, const RImageInstallPlan *plan, const RaucInstallArgs *args)
```
jluebbe, line 949: The `args` argument seems unused.

jluebbe (changes requested):
Looks fine otherwise.

src/install.c
ejoerns: Seems, but isn't :wink:. It is used in

``` c
slot_state->installed_txn = g_strdup(args->transaction);
```

src/install.c
ejoerns: I've checked this and would say it was just missing. The same is true for the error case of the `skip_image` step which I also fixed now.

ejoerns:
Fixed (1st force push) and rebased onto current master (2nd force push)

jluebbe (changes requested):

jluebbe (approved):

PR #1160, opened 2023-06-14
Introduce System Status and Use For Boot Detection

ejoerns:
This introduces support for writing a global rauc state file in key-value format, located in the data-directory.
For now, the file will be used to store the system's boot ID:

    [system]
    boot-id=cfb75bf4-e678-42e7-8dc9-14fed56baf1a

With the boot-id information from the state file, the service can determine if the system was (re)booted.

This works by comparing the current boot ID with the ID stored in the status file.
If they differ, we know a reboot occurred and then store the current boot ID in the status file.
A subsequent restart of the service would the see the current and the stored boot ID being equal and just report a service restart.

We emit a normal log message for now but the plan is to use this later for dedicated logging of reboots.


src/main.c
```diff
@@ -1390,14 +1390,12 @@ static gchar* r_status_formatter_readable(RaucStatusPrint *status)
 
 	g_return_val_if_fail(status, NULL);
 
-	bootedfrom = r_slot_find_by_device(status->slots, status->bootslot);
-	if (!bootedfrom)
-		bootedfrom = r_slot_find_by_bootname(status->slots, status->bootslot);
+	bootedfrom = r_slot_get_booted(status->slots);
```
jluebbe, line 1393: If `r_slot_get_booted` returned NULL, that could also be mapped to `"unknown"`.

src/context.c
```diff
@@ -315,6 +316,15 @@ static gboolean r_context_configure_target(GError **error)
 		}
 	}
 
+	/* load global state */
+	g_clear_pointer(&context->statepath, g_free);
+	g_clear_pointer(&context->global_state, r_global_state_free);
+	context->statepath = g_build_filename(context->config->data_directory, "state", NULL);
```
jluebbe, line 322: `data_directory` can be unset. Perhaps move it inside the `if` above?

Also, `state` as a filename seems too generic. Perhaps something mentioning persistence?

src/main.c
```diff
@@ -2011,6 +2011,23 @@ static gboolean service_start(int argc, char **argv)
...
+		g_message("Booted into %s (%s)", booted_slot->name, booted_slot->bootname);
+
+		/* update boot ID */
+		g_free(r_context()->global_state->boot_id);
+		r_context()->global_state->boot_id = g_strdup(r_context()->boot_id);
+
+		if (!r_global_state_save(r_context()->statepath, r_context()->global_state, &ierror)) {
+			g_warning("Failed saving global state");
+			g_clear_error(&ierror);
+		}
```
jluebbe, line 2028: I find passing data from `r_context()` into a different module strange. Perhaps this should a helper in `context.h` or `r_global_state_save` should access the context directly?

jluebbe (changes requested):
I'm not convinced that a separate file is the best approach, as it means that we can't have atomic updates to the global state and the slot status.

Also, we should probably be explicit about the requirement for the data directory.

src/global_state.c
```diff
@@ -145,6 +146,17 @@ gboolean determine_slot_states(GError **error)
 				s->state = ST_INACTIVE;
 			}
 
+			/* Create and add dummy external slot */
+			ext_dummy = g_new0(RaucSlot, 1);
+			ext_dummy->name = g_strdup("external");
+			ext_dummy->sclass = g_strdup("external");
+			ext_dummy->type = g_strdup("virtual");
```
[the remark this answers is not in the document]
ejoerns: I still like the idea but I will omit it for the next PR update first of all

src/context.c
ejoerns: We discussed that due to atomicity using the existing status file should be preferred.
Then we still need to check for per-slot not being enabled.

I'll implement it this way but rethinking your proposal I am not sure if atomicity plays any role since the update of the slot status and the update of the system status would never happen at the same time I guess.

src/main.c
ejoerns: We discussed that (unlike for the load function which gets called before having context set up) passing path and state object explicitly is not required here. I will check if we can also use a wrapper method in `context.c` to hide the direct call to slot saving method.

src/config_file.c
```diff
@@ -631,7 +631,9 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 		if (c->data_directory) {
 			c->statusfile_path = g_build_filename(c->data_directory, "central.raucs", NULL);
 		} else {
-			g_message("Config option 'statusfile=<path>/per-slot' unset, falling back to per-slot status");
+			g_message("No data directory or status file set, falling back to per-slot status.\n"
+					"This will limit supported features.\n"
+					"Consider setting 'data-directory=<path>' or 'statusfile=<path>/per-slot' explicitly.");
```
jluebbe, line 636: `per-slot` will still limit supported features. Perhaps the limited features message should be triggered by `per-slot` only?

src/context.c
```diff
@@ -315,6 +316,15 @@ static gboolean r_context_configure_target(GError **error)
 		}
 	}
 
+	/* load system status if central status file is available */
+	if (g_strcmp0(context->config->statusfile_path, "per-slot") != 0) {
+		g_clear_pointer(&context->system_status, r_system_status_free);
+		context->system_status = g_new0(RSystemStatus, 1);
+		if (!r_system_status_load(context->config->statusfile_path, context->system_status, &ierror)) {
```
jluebbe, line 323: `ierror` needs to be cleared on in the error case, otherwise the next function can't use it.

src/main.c
jluebbe: You decided against the wrapper, but with out the explicit parameters, I'm fine with this.

src/context.c
```diff
@@ -315,6 +316,15 @@ static gboolean r_context_configure_target(GError **error)
...
+			g_message("Failed to load global state: %s\n", ierror->message);
```
jluebbe, line 324:
```suggestion
			g_message("Failed to load system status: %s\n", ierror->message);
```

include/status_file.h
```diff
@@ -0,0 +1,96 @@
...
+ * desinged to be called during context setup where we cannot access context,
+ * yet.
+ *
+ * @param filename File name to load state from
+ * @param state RSystemStatus to update from file
+ * @param[out] error Return location for a GError, or NULL
+ *
+ * @return
+ */
+gboolean r_system_status_load(const gchar *filename, RSystemStatus *state, GError **error)
```
jluebbe, line 80:
```suggestion
 * Note that filename and status are passed explicitly here since the method is
 * designed to be called during context setup where we cannot access context,
 * yet.
 *
 * @param filename File name to load status from
 * @param status RSystemStatus to update from file
 * @param[out] error Return location for a GError, or NULL
 *
 * @return
 */
gboolean r_system_status_load(const gchar *filename, RSystemStatus *status, GError **error)
```

include/status_file.h
```diff
@@ -0,0 +1,96 @@
...
+G_GNUC_WARN_UNUSED_RESULT;
+
+/**
+ * Save sytem status to central status file.
```
jluebbe, line 84:
```suggestion
 * Save system status to central status file.
```

include/status_file.h
```diff
@@ -0,0 +1,96 @@
...
+ */
+gboolean r_system_status_save(GError **error)
+G_GNUC_WARN_UNUSED_RESULT;
+
+/**
+ * Free system status.
+ *
+ * @param state RSystemStatus to free
+ */
+void r_system_status_free(RSystemStatus *state);
```
jluebbe, line 96:
```suggestion
 * @param status RSystemStatus to free
 */
void r_system_status_free(RSystemStatus *status);
```

src/main.c
```diff
@@ -2013,6 +2014,29 @@ static gboolean service_start(int argc, char **argv)
 		return TRUE;
 	}
 
+	if (r_context()->system_status) {
+		/* Boot ID-based boot vs service restart detection */
```
jluebbe, line 2018:
```suggestion
		/* Boot-ID-based system reboot vs service restart detection */
```

jluebbe (changes requested):

src/config_file.c
ejoerns: So we should probably just omit the limited feature line. For `per-slot` case in general I already I introduced the follow-up message that says

> Using per-slot statusfile. System status information not supported!

I'll also adapt the commit message accordingly

src/status_file.c
```diff
@@ -320,6 +323,28 @@ static gboolean save_slot_status_globally(GError **error)
 
 	g_debug("Saving global slot status");
 
+	if (!g_key_file_load_from_file(key_file, r_context()->config->statusfile_path, G_KEY_FILE_NONE, &ierror)) {
+		if (g_error_matches(ierror, G_FILE_ERROR, G_FILE_ERROR_NOENT)) {
+			g_info("Status file does not exist yet, ignore loading");
+			g_clear_error(&ierror);
+		} else {
+			g_propagate_error(error, ierror);
+			return FALSE;
```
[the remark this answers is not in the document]
ejoerns: Unsure. Having it broken in the save method would mean that it broke after loading it. Do we assume this might happen?

src/context.c
ejoerns: Will fix

ejoerns:
Also moved `config_file_statusfile_missing` check back to `test/config_file.c` since it is more of a config-related test.

ejoerns:
Rebased onto master to resolve conflict in `src/mark.c`.

src/status_file.c
```diff
@@ -0,0 +1,471 @@
...
+	return g_steal_pointer(&key_file);
+}
+
+/* Updates slot status information in status file while leaving system status
+ * information untouched */
+static gboolean save_slot_status_globally(GError **error)
+{
+	g_autoptr(GKeyFile) key_file = NULL;
+	GError *ierror = NULL;
+	g_auto(GStrv) groups = NULL;
```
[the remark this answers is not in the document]
ejoerns: Fixed

include/status_file.h
```diff
@@ -0,0 +1,96 @@
...
+ */
+gboolean r_slot_status_save(RaucSlot *dest_slot, GError **error)
+G_GNUC_WARN_UNUSED_RESULT;
+
+typedef struct {
+	gchar *boot_id;
+} RSystemStatus;
+
+/**
+ * Save sytem status from central status file.
```
jluebbe, line 68:
```suggestion
 * Load system status from central status file.
```

src/status_file.c
```diff
@@ -308,9 +308,63 @@ static gboolean save_slot_status_locally(RaucSlot *dest_slot, GError **error)
...
+ *
+ * With keep_prefix, one can define groups to keep in the returned GKeyFile.
+ * All other groups will be removed and can be newly populated.
+ *
+ * @param path Path to key file
+ * @param keep_prefix prefix for groups to keep when loading the file
+ *
+ * @return newly-allocated GKeyFile or NULL on error
+ */
+static GKeyFile* key_file_init_from_shared_status_file(const gchar *path, const gchar *keep_prefix, GError **error)
```
jluebbe, line 322: The path argument seems unused. Was this intended to be used in unit tests?

src/context.c
```diff
@@ -315,6 +316,16 @@ static gboolean r_context_configure_target(GError **error)
 		}
 	}
 
+	/* load system status if central status file is available */
+	if (g_strcmp0(context->config->statusfile_path, "per-slot") != 0) {
+		g_clear_pointer(&context->system_status, r_system_status_free);
+		context->system_status = g_new0(RSystemStatus, 1);
+		if (!r_system_status_load(context->config->statusfile_path, context->system_status, &ierror)) {
+			g_message("Failed to load sytem status: %s", ierror->message);
```
jluebbe, line 324:
```suggestion
			g_message("Failed to load system status: %s", ierror->message);
```

src/status_file.c
```diff
@@ -0,0 +1,470 @@
...
+
+	if (!g_key_file_load_from_file(key_file, r_context()->config->statusfile_path, G_KEY_FILE_NONE, &ierror)) {
+		if (g_error_matches(ierror, G_FILE_ERROR, G_FILE_ERROR_NOENT)) {
+			g_info("Status file does not exist yet, ignore loading");
+			g_clear_error(&ierror);
+		} else {
+			g_autofree gchar *broken_file = g_strconcat(r_context()->config->statusfile_path, ".broken", NULL);
+			/* Allow recovering from broken status file */
+			g_warning("Failed loading status file: %s\n", ierror->message);
+			g_warning("Moving status file to %s and re-creating it.", broken_file);
```
jluebbe, line 338:
```suggestion
			g_warning("Failed to load status file: %s\n", ierror->message);
			g_warning("Will move status file to %s and re-create it.", broken_file);
```

jluebbe (changes requested):
We also need test cases for the newly added functions including the broken file handling.

For the shared handling, perhaps add add save_/load_ of the other kind to see if something is damaged?

src/status_file.c
ejoerns: I guess this is a remnant from refactoring. Cannot remember having something in mind for it. Removed.

ejoerns:
I've addressed the review feedback and added a bunch of test cases.

test/status_file.c
```diff
@@ -0,0 +1,366 @@
...
+\n\
+[broken\n\
+";
+	pathname = write_tmp_file(fixture->tmpdir, "broken_config.conf", cfg_file, NULL);
+	g_assert_nonnull(pathname);
+
+	system_status = g_new0(RSystemStatus, 1);
+	g_assert_nonnull(system_status);
+
+	/* assert error-free loading */
```
jluebbe, line 176: This comment doesn't match the code.

test/status_file.c
```diff
@@ -0,0 +1,366 @@
...
+	GError *ierror = NULL;
+	gboolean res;
+	g_autoptr(RSystemStatus) system_status = NULL;
+
+	pathname = g_build_filename(fixture->tmpdir, "system_only.conf", NULL);
+	g_assert_nonnull(pathname);
+
+	r_context_conf()->configpath = g_strdup("test/test-global.conf");
+	r_context()->config->statusfile_path = pathname;
+	r_context()->system_status->boot_id = g_strdup("e02a2afe-cf45-4d50-a3f3-c223ca0f480a");
```
jluebbe, line 196: Perhaps use `replace_strdup`?

test/status_file.c
```diff
@@ -0,0 +1,366 @@
...
+[slot]\n\
+status=ok\n\
+sha256=e437ab217356ee47cd338be0ffe33a3cb6dc1ce679475ea59ff8a8f7f6242b27\n\
+";
+	pathname = write_tmp_file(fixture->tmpdir, "existing_config.conf", cfg_file, NULL);
+	g_assert_nonnull(pathname);
+
+	r_context_conf()->configpath = g_strdup("test/test-global.conf");
+	r_context()->config->statusfile_path = pathname;
+	r_context()->system_status->boot_id = g_strdup("e02a2afe-cf45-4d50-a3f3-c223ca0f480a");
```
jluebbe, line 261: Perhaps use `replace_strdup`?

test/status_file.c
```diff
@@ -0,0 +1,366 @@
...
+[broken\n\
+";
+	pathname = write_tmp_file(fixture->tmpdir, "existing_config.conf", cfg_file, NULL);
+	g_assert_nonnull(pathname);
+
+	r_context_conf()->configpath = g_strdup("test/test-global.conf");
+	r_context()->config->statusfile_path = pathname;
+	r_context()->system_status->boot_id = g_strdup("e02a2afe-cf45-4d50-a3f3-c223ca0f480a");
+
+	g_message("EXPECT FORM NOW!");
```
jluebbe, line 305: Leftover from debugging?

test/status_file.c
```diff
@@ -0,0 +1,366 @@
...
+			status_file_fixture_tear_down);
+	/* Combined tests */
+	g_test_add("/status-file/save-system-status-existing-slot-status", ConfigFileFixture, NULL,
+			status_file_fixture_set_up_global,
+			status_file_test_save_system_status_existing_slot_status,
+			status_file_fixture_tear_down);
+	g_test_add("/status-file/save-system-status-broken-existing-slot-status", ConfigFileFixture, NULL,
+			status_file_fixture_set_up_global,
+			status_file_test_save_system_status_broken_existing_slot_status,
+			status_file_fixture_tear_down);
```
jluebbe, line 363: Do we need the other side (save slot status in a file with existing system status) as well?

test/meson.build
```diff
@@ -16,6 +16,7 @@ tests = [
   'utils',
   'install',
   'service',
+  'status_file',
```
jluebbe, line 19: Add this new file to https://github.com/rauc/rauc/blob/master/.github/workflows/sanitizers.yml, otherwise we loose address sanitizer coverage for these tests.

jluebbe (changes requested):

test/status_file.c
ejoerns: Thought about this, too. Since it uses the same helper it should not fail the other way round, but, maybe better safe than sorry. Will add a test for this, too.

test/meson.build
ejoerns: Ack. Forgot that it is a separate list. Will add.

test/status_file.c
ejoerns: May contain traces of copy-paste :smirk:

ejoerns:
Collected a couple of further fixes for tests while working on the review feedback:

* renamed 'cfg/.conf' to 'status/.raucs' for test status file creation
* Fixed typos
* Renamed `ConfigFileFixture` to `StatusFileFixture`
* Use `configpath` initialization from fixture (was replicated in each test)
* removed unused `pathname` in `status_file_test_save_system_status_per_slot()`
* Added missing cleanup of `system_status` in `r_context_clean()`
* Fixed various memory leaks by using `replace_strdup()`
* Fixed further memory leaks in strings, GError, GKeyFile, etc.

test/status_file.c
```diff
@@ -0,0 +1,406 @@
...
+
+	/* assert loaded key file contains only newly added [system] group */
+	groups = g_key_file_get_groups(keyfile, &num_groups);
+	g_assert_cmpint(num_groups, ==, 1);
+	g_assert_true(g_strv_contains((const gchar * const *)groups, "system"));
+}
+
+/* Creates and saves system status into file with existing slot status */
+static void status_file_test_save_slot_status_existing_system_status(StatusFileFixture *fixture,
+		gconstpointer user_data)
```
jluebbe, line 319:
```suggestion
/* Creates and saves slot status into file with existing system status */
static void status_file_test_save_slot_status_existing_system_status(StatusFileFixture *fixture,
		gconstpointer user_data)
```

jluebbe (changes requested):
Just a minor issue with the comment. Please rebase and this can go in.

ejoerns:
Fixed comment. No rebase, yet.

ejoerns:
Rebased onto master now and fixed a faulty-fixuped hunk so that it is also now fully bisectable again.

jluebbe (approved):

PR #1161, opened 2023-06-15
fix memory leaks found by address sanitizer

jluebbe:


test/install.c
```diff
@@ -1044,7 +1044,10 @@ static gboolean install_cleanup(gpointer data)
 	g_assert_cmpint(args->status_result, ==, 0);
 	g_assert_false(g_queue_is_empty(&args->status_messages));
 
-	g_queue_clear(&args->status_messages);
+	while (!g_queue_is_empty(&args->status_messages)) {
```
ejoerns, line 1047: When moving to glib 2.60 someday, `g_queue_clear_full()` would be an alternative.

ejoerns (approved):

test/install.c
jluebbe: Yes, that was my first idea. :)

Issue #1165, opened 2023-06-16
meson build system installs /usr/libexec/rauc-service.sh without execute permissions


BubuOT:
**Describe the bug**
When building rauc with dbus support with meson, running it on a non-systemd system every call to rauc fails with the following error:

```
Error retrieving slot status via D-Bus: error creating proxy: Error calling StartServiceByName for de.pengutronix.rauc: Cannot launch daemon, file not found or permissions invalid
```

It seems that `/usr/libexec/rauc-service.sh` has no execute permissions. Comparing against a system with rauc 1.8 built with autotools, this use-case works correctly and this file has execute permissions.

When starting the service manually (with `rauc service &`) everything works afterwards.

**Background information**
rauc version v1.9

**To Reproduce**

S. above, then run i.e. `rauc status` on the system.

**Expected behavior**
`rauc status` works without having to manually start the service.

**Logs**
-

**Additional context**
-

ejoerns:
Confirmed. Using `install_data()` does not seem to be the right option for executables (as its name already suggests).

ejoerns:
> Confirmed. Using `install_data()` does not seem to be the right option for executables (as its name already suggests).

Ok, seems `install_data()` can be used, just lacks a proper `install_mode`

BubuOT:
Thanks! :)

PR #1167, opened 2023-06-16
Fix dbus_sources dependency

chris2511:
Create a source dependency on dbus_sources to "tests". These sources are required at compile time and might be generated too late otherwise.

The race condition may be provoked by adding a "sleep 10" to "gdbus-codegen"

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -39,6 +39,8 @@ extra_test_sources = files([
   'install-fixtures.c',
 ])
 
+test_dep = declare_dependency(sources : dbus_sources)
+
 foreach test_name : tests
   exe = executable(
     test_name + '-test',
@@ -47,7 +49,7 @@ foreach test_name : tests
     link_with : librauc,
     c_args : '-DTEST_SERVICES="' + meson.build_root() + '"',
     include_directories : incdir,
-    dependencies : rauc_deps)
+    dependencies : [ rauc_deps, test_dep ])
 
   test(
     test_name,


jluebbe:
@chris2511 I've force-pushed to adjust the commit title to our convention.

ejoerns:
We already have `dbus_sources` in the executable's source list. My assumption so far is that this is sufficient.

I also could not reproduce the described behavior. Which meson version did you test with?

chris2511:
We were working with "v1.9", and missed  97b0ecb981e5c8e6a70620cd0ff86d61b346137d, which is already in master.
Sorry for the noise...

ejoerns:
@chris2511 Thanks for clarifying. Indeed, this was fixed *after* the latest release v1.9. I did not have this in mind either.

I'd say that this might be sufficient to consider the meson support in v1.9 not sufficiently working to already drop automake with v1.10.

There were also a few more issues, like #1165 or #1088 that prevent the meson support from being usable in some scenarios.

@jluebbe Any thoughts on this? Might a bugfix release be appropriate here or should we just not fully recommend the meson support for v1.9 yet?

jluebbe:
> @jluebbe Any thoughts on this? Might a bugfix release be appropriate here or should we just not fully recommend the meson support for v1.9 yet?

If it was unusable, we'd have heard more about it, so we should probably just release v1.10 soon, but not drop autotools yet. Carrying it a bit longer isn't *that* painful.

ejoerns:
@jluebbe Unsure how many are actually using the meson support already. In Yocto it's only in master and mickledore and I guess all other build systems still use automaker.

A soon v1.10 release is a good idea anyway.

BubuOT:
buildroot uses it in the latest (2023.05) release now as well.

ejoerns:
> buildroot uses it in the latest (2023.05) release now as well.

@BubuOT Ah, cool. Thanks for the hint! It builds with `-Dtests=false` and thus should not hit the issue. I guess not building the tests at all is then the best advice anyway.

PR #1171, opened 2023-06-22
src/hash_index: initialize 'middle' for consistency

jluebbe:
Note that it was not actually used uninitialized since using it depended on having 'found' set to TRUE which can only happen in a code path that also sets 'middle' to some value.

Closes: #1155


ejoerns (approved):
I like the commit message!

PR #1182, opened 2023-07-10
improve docs

rohieb:
Some things I found while reading through the docs.

While "tar archive" could be consideres as a redundant acronym, it reads better in texts than a simple "tar". With some missing words I was not sure if filled in the correct ones, and had to extrapolate from the surrounding text; please review those carefully. Also use the "official" capitalisation of names like RAUC and PyPI.


jluebbe:
@rohieb I've made some changes to your commit.

jluebbe (approved):

rohieb:
@jluebbe Thanks, 👍🏼

PR #1183, opened 2023-07-12
get_system_info_from_handler: do not step on existing values

hcl-dr:
context->system_serial and context->config->system_variant was deleted even when no substitute is found in the script.

This breaks systems using DTB compatible variant setups when the script did not supply this information.
Keep off values not supplied by the handler script!

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -274,16 +274,15 @@ static GHashTable *get_system_info_from_handler(GError **error)
 		return NULL;
 	}
 
-	g_clear_pointer(&context->system_serial, g_free);
-	g_clear_pointer(&context->config->system_variant, g_free);
 	g_hash_table_iter_init(&iter, vars);
 	while (g_hash_table_iter_next(&iter, (gpointer*) &key, (gpointer*) &value)) {
 		/* legacy handling */
 		if (g_strcmp0(key, "RAUC_SYSTEM_SERIAL") == 0) {
+			g_clear_pointer(&context->system_serial, g_free);
 			context->system_serial = g_strdup(value);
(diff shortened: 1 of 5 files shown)


jluebbe:
You're right, this was broken by PR https://github.com/rauc/rauc/pull/1143 and released with v1.10.

We should probably also add a test case to ensure that we don't break this again in the future. Do you have time to add a test case as part of this PR?

src/context.c
```diff
@@ -274,12 +274,11 @@ static GHashTable *get_system_info_from_handler(GError **error)
 		return NULL;
 	}
 
-	g_clear_pointer(&context->system_serial, g_free);
-	g_clear_pointer(&context->config->system_variant, g_free);
 	g_hash_table_iter_init(&iter, vars);
 	while (g_hash_table_iter_next(&iter, (gpointer*) &key, (gpointer*) &value)) {
 		/* legacy handling */
 		if (g_strcmp0(key, "RAUC_SYSTEM_SERIAL") == 0) {
+			g_clear_pointer(&context->system_serial, g_free);
```
jluebbe, line 281: I'd prefer to use `g_clear_pointer` in both branches to keep them consistent. That could be done in the same commit.

jluebbe (changes requested):

src/context.c
hcl-dr: Sounds good. I'm on vacation right now and unable to build a test case at the moment, but will make the modification you suggested.

jluebbe (approved):

jluebbe:
I've added a regression test case.

ejoerns (approved):
Oh, good catch! Actually did not see this during the review despite it looks quite obvious now.

Added test case looks fine for me, too.

:+1:

Issue #1187, opened 2023-07-19
RAUC will only accidentally build against OpenSSL


fhars:
The `Makefile.am` references the variables `OPENSSL_LDFLAGS`and `OPENSSL_INCLUDES` in the definitions of  `AM_LDFLAGS` and `AM_CPPFLAGS`, and nothing ssl-related in `AM_CFLAGS`, while `configure` adds definitions for `OPENSSL_LIBS` and `OPENSSL_CFLAGS`. As a result, rauc will only build if OpenSSL happens to be installed in the same directory as some other library it depends on.
A similar thing is probably also true for the other libraries where the definition of `AM_LDFLAGS` uses different names that the one for `librauc_la_LIBADD`.

```patch
From d22046902a9e14ee0cbcaf9bab95c9441912f22f Wed Jul 19 00:00:00 2023
From: Florian Hars <florian@hars.de>
Date: Wed, 19 Jul 2023 15:01:39 +0200
Subject: [PATCH] actually link against the found libssl

---
 Makefile.am | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/Makefile.am b/Makefile.am
index 45d5994..24d880a 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -15,9 +15,9 @@ CODE_COVERAGE_IGNORE_PATTERN = "*-generated.c"

 GC_CFLAGS = -fdata-sections -ffunction-sections
 GC_LDFLAGS = -Wl,--gc-sections -Wl,-Map,$@.map
-AM_CFLAGS = -DG_LOG_DOMAIN=\"rauc\" $(GC_CFLAGS) $(WARN_CFLAGS) $(GLIB_CFLAGS) $(CURL_CFLAGS) $(NL3_CFLAGS)
-AM_LDFLAGS = $(GC_LDFLAGS) $(WARN_LDFLAGS) $(GLIB_LDFLAGS) $(CURL_LDFLAGS) $(OPENSSL_LDFLAGS)
-AM_CPPFLAGS = -I${top_srcdir}/include -include ${top_builddir}/config.h $(OPENSSL_INCLUDES)
+AM_CFLAGS = -DG_LOG_DOMAIN=\"rauc\" $(GC_CFLAGS) $(WARN_CFLAGS) $(GLIB_CFLAGS) $(CURL_CFLAGS) $(NL3_CFLAGS) $(OPENSSL_CFLAGS)
[...]

PR #1189, opened 2023-07-20
OpenSSL autotools flags cleanup

jluebbe:
Fixes: #1187


Makefile.am
```diff
@@ -16,7 +16,7 @@ CODE_COVERAGE_IGNORE_PATTERN = "*-generated.c"
 GC_CFLAGS = -fdata-sections -ffunction-sections
 GC_LDFLAGS = -Wl,--gc-sections -Wl,-Map,$@.map
 AM_CFLAGS = -DG_LOG_DOMAIN=\"rauc\" $(GC_CFLAGS) $(WARN_CFLAGS) $(GLIB_CFLAGS) $(CURL_CFLAGS) $(NL3_CFLAGS) $(OPENSSL_CFLAGS)
-AM_LDFLAGS = $(GC_LDFLAGS) $(WARN_LDFLAGS) $(GLIB_LDFLAGS) $(CURL_LDFLAGS) $(OPENSSL_LDFLAGS)
+AM_LDFLAGS = $(GC_LDFLAGS) $(WARN_LDFLAGS)
```
michaelolbrich, line 19: Implementation looks good. You might want to mention that PKG_CHECK_MODULES adds everything to _LIBS. Without seeing the code, the comment about librauc_la_LIBADD is a bit confusing.

michaelolbrich (approved):
One nitpick about a commit message, looks good either way.

Makefile.am
jluebbe: Thanks.

ejoerns:
From the 1st commit message:

> Makefile.am: use OPENSSL_CFLAGS in AM_CFLAGS and AM_CPPFLAGS
This didn't seem to cause problems so far, as it's empty on most systems.
Still, it's more correct to use it.

-> So this means not using the (mostly) empty `OPENSSL_CFLAGS` did not cause any problems, right? What about the `OPENSSL_INCLUDES` which was also removed in this patch?

From the 2nd commit message: 

> PKG_CHECK_MODULES doesn't define a _LDFLAGS variable, so this expands to nothing.

-> It is a bit unclear to me if this is always the case or only for openssl or for all packages (openssl, glib, curl) removed with this commit.

jluebbe:
> From the 1st commit message:
> 
> > Makefile.am: use OPENSSL_CFLAGS in AM_CFLAGS and AM_CPPFLAGS
> > This didn't seem to cause problems so far, as it's empty on most systems.
> > Still, it's more correct to use it.
> 
> -> So this means not using the (mostly) empty `OPENSSL_CFLAGS` did not cause any problems, right?

Yes.

> What about the `OPENSSL_INCLUDES` which was also removed in this patch?

This doesn't exist, so it's always empty.

> From the 2nd commit message:
> 
> > PKG_CHECK_MODULES doesn't define a _LDFLAGS variable, so this expands to nothing.
> 
> -> It is a bit unclear to me if this is always the case or only for openssl or for all packages (openssl, glib, curl) removed with this commit.

`PKG_CHECK_MODULES` only defines `__CFLAGS` and `_LIBS` (not `_LDFLAGS`) independent of the package.

ejoerns:
> > From the 1st commit message:
> > > Makefile.am: use OPENSSL_CFLAGS in AM_CFLAGS and AM_CPPFLAGS
> > > This didn't seem to cause problems so far, as it's empty on most systems.
> > > Still, it's more correct to use it.
> > 
> > 
> > -> So this means not using the (mostly) empty `OPENSSL_CFLAGS` did not cause any problems, right?
> 
> Yes.
> 
> > What about the `OPENSSL_INCLUDES` which was also removed in this patch?
> 
> This doesn't exist, so it's always empty.

Oh, ok. Maybe mentions this in the commit message, too?

> > From the 2nd commit message:
> > > PKG_CHECK_MODULES doesn't define a _LDFLAGS variable, so this expands to nothing.
> > 
> > 
> > -> It is a bit unclear to me if this is always the case or only for openssl or for all packages (openssl, glib, curl) removed with this commit.
> 
> `PKG_CHECK_MODULES` only defines `__CFLAGS` and `_LIBS` (not `_LDFLAGS`) independent of the package.

Ah, was confused. So only `_FLAGS` macro sets `_LDFLAGS`. Got it now from re-reading https://www.gnu.org/software/guile/manual/html_node/Autoconf-Macros.html. Good that we switch to meson :smirk: 

Rest looks fine for me :+1:

ejoerns (approved):

PR #1197, opened 2023-07-28
Fix bootchooser for the output of efibootmgr 18

dvzrv:
Since efibootmgr 18, the default output of `efibootmgr` is now more verbose [1], which breaks the assumptions made by RAUC in regards to parsing the output. This issue is also affecting others [2].

Fix the parsing logic by unconditionally splitting on a tab in the detected name part of the boot entry, so that the unused data part can be discarded.
Emit a debug message for each found boot entry, as this helps in detecting issues in the future.

[1] https://github.com/rhboot/efibootmgr/commit/8ec3e9dedb3cb62f19847794012420b90f475398
[2] https://github.com/rhboot/efibootmgr/issues/169

This fix is not related to an open ticket, as I noticed this quite recently and got in touch on matrix to discuss a possible fix.
I verified this to work in an Arch Linux VM using efibootmgr 18 and a RAUC 1.10 with this patch applied.

As writing C is not really my forte, I'd appreciate feedback (especially whether this works as intended with efibootmgr < 18).

### DIFF
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -1159,11 +1159,20 @@ static gboolean efi_bootorder_get(GList **bootorder_entries, GList **all_entries
 	}
 
 	while (g_match_info_matches(match)) {
+		gchar *tab_point = NULL;
 		efi_bootentry *entry = g_new0(efi_bootentry, 1);
 		entry->num = g_match_info_fetch(match, 1);
 		entry->name = g_match_info_fetch(match, 2);
+
+		/* Remove anything after a tab, as it is most likely path
+		 * information which we don't need. */
+		tab_point = strchr(entry->name, '\t');
+		if (tab_point)
+			*tab_point = '\0';
+
 		entries = g_list_append(entries, entry);
 		g_match_info_next(match, NULL);
+		g_debug("Detected EFI boot entry %s: %s", entry->num, entry->name);
 	}
 
 	g_clear_pointer(&regex, g_regex_unref);
--- a/include/utils.h
+++ b/include/utils.h
@@ -198,6 +198,16 @@ gchar * key_file_consume_string(
 G_GNUC_WARN_UNUSED_RESULT;
 
 /**
+ * Ensure that the input string contains neither whitespace nor tab.
+ *
+ * @param str string to check.
+ *
+ * @return TRUE if str contains neither whitespace nor tab, FALSE otherwise
+ */
+gboolean value_check_tab_whitespace(const gchar *str, GError **error)
+G_GNUC_WARN_UNUSED_RESULT;
+
+/**
  * Get integer argument from key and remove key from key_file.
  */
 gint key_file_consume_integer(
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -233,8 +233,16 @@ static GHashTable *parse_slots(const char *filename, const char *data_directory,
 			}
 
 			value = key_file_consume_string(key_file, groups[i], "bootname", NULL);
+
 			slot->bootname = value;
 			if (slot->bootname) {
+				/* Ensure that the bootname does not contain whitespace or tab */
+				if (!value_check_tab_whitespace(value, &ierror)) {
+					g_propagate_prefixed_error(error, ierror,
+							"Invalid bootname for slot %s: ", slot->name);
+					return NULL;
+				}
+
 				/* check if we have seen this bootname on another slot */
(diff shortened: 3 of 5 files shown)


src/bootchooser.c
```diff
@@ -1159,11 +1159,26 @@ static gboolean efi_bootorder_get(GList **bootorder_entries, GList **all_entries
...
+					R_BOOTCHOOSER_ERROR_FAILED,
+					"Extracting a boot entry name failed!");
+			res = FALSE;
+			goto out;
+		}
+
 		efi_bootentry *entry = g_new0(efi_bootentry, 1);
 		entry->num = g_match_info_fetch(match, 1);
-		entry->name = g_match_info_fetch(match, 2);
+		entry->name = g_strdup(name_data_splits[0]);
```
jluebbe, line 1177: Instead of using `g_strsplit`, you can also modify entry->name in place. Use strchr to find the first tab, and if there is one, replace it by `'\0'`. This also avoids any additional memory allocations.

jluebbe (changes requested):
To make sure this can work reliably, we need to check that the boot name doesn't contain whitespace. See https://github.com/rauc/rauc/blob/master/src/config_file.c#L235 for the place where it is read. That area also has examples for reporting an error. The bootname check should be in a helper function.

dvzrv:
> To make sure this can work reliably, we need to check that the boot name doesn't contain whitespace. See https://github.com/rauc/rauc/blob/master/src/config_file.c#L235 for the place where it is read. That area also has examples for reporting an error. The bootname check should be in a helper function.

Maybe I lack context here, but why can it not contain a whitespace? This seems certainly possible in the context of e.g. `efibootmgr`.

src/bootchooser.c
dvzrv: Ah good point. I'll try that... might take me some time though.
Better than to leak memory.. phew :sweat_smile:

jluebbe:
> > To make sure this can work reliably, we need to check that the boot name doesn't contain whitespace. See https://github.com/rauc/rauc/blob/master/src/config_file.c#L235 for the place where it is read. That area also has examples for reporting an error. The bootname check should be in a helper function.
> 
> Maybe I lack context here, but why can it not contain a whitespace? This seems certainly possible in the context of e.g. `efibootmgr`.

We currently don't have checks for the bootname, so it could contain a tab character. In that case, applying this change would break at runtime. Adding a check during config parsing would report that issue early. And while adding that check, we should also reject any unusual bootnames containing whitespace (to be better prepared for future efibootmgr changes).

dvzrv:
> We currently don't have checks for the bootname, so it could contain a tab character. In that case, applying this change would break at runtime. Adding a check during config parsing would report that issue early. And while adding that check, we should also reject any unusual bootnames containing whitespace (to be better prepared for future efibootmgr changes).

I see. But validating the config input and more robustly dealing with the output of efibootmgr are two separate issues (in my mind).
Are there any straightforward facilities to search for the right-most tab character? Maybe that would make more sense.

src/bootchooser.c
dvzrv: I'll try with cleaning up my current code further, as the glib provided functions are easier for me to use right now.
I'd be happy to see whether with the recent change I am circumventing the memory leak already.

jluebbe:
I've adjusted the commit titles to our usual format, made some changes to the code (see the fixup commits) and added a test-case for the config value check. @dvzrv does this look OK for you? Could you test it with efibootmgr 18?

dvzrv:
> I've adjusted the commit titles to our usual format, made some changes to the code (see the fixup commits) and added a test-case for the config value check. @dvzrv does this look OK for you? Could you test it with efibootmgr 18?

Thanks! I've rebuild the package for Arch Linux locally with the updated fixes. Tests run fine and after installing the updated package in a VM (with A/B setup) the output of `rauc status --detailed` works without issue.

jluebbe (approved):

jluebbe:
Spaces/tabs are also problematic for bootnames in other places, like the kernel cmdline, so it's better to reject them early.

The strchr approach seems simpler, so I prefer it for the point release.

ejoerns:
> Spaces/tabs are also problematic for bootnames in other places, like the kernel cmdline, so it's better to reject them early.

Since a broken config will fail early with this approach, go for it.

> The strchr approach seems simpler, so I prefer it for the point release.

Agreed for the point release.

ejoerns (approved):

Issue #1009, opened 2022-11-15
Image type of "boot-emmc" limited to "*.img" files


mschwan-phytec:
Currently, updating the bootloader in an eMMC boot partition is only allowed for files of type `.img`:
```c
RaucUpdatePair updatepairs[] = {
...
#if ENABLE_EMMC_BOOT_SUPPORT == 1
	{"*.img", "boot-emmc", img_to_boot_emmc_handler},
#endif
	{"*", "boot-emmc", NULL},
...
```
Is there a specific reason to limit it to that file type? Usual extensions for bootloaders also include `.bin` or even nothing at all. In general, I would assume the bootloader is almost always a binary file which needs to written as is to the eMMC boot partition.

ejoerns:
The idea was to be just as explicit as we are for the other slot types, too.

So the question is if we can be more relaxed for slot types that already clearly dictate the handling, but I understand your point of course.

This would drop some sanity checking of input types, but, to be honest, when e.g. using meta-rauc, the bundle class would also just take the given image and unconditionally append `*.img` to the filename if the slot type is set to `bootloader`.

@jluebbe Do you have any opinion on this?

jluebbe:
The selection of the update handler based on the file type is intended as a way to allow different handlers in the future, so just allowing any extension here would mean that the `boot-emmc` slot type can never gain a different handler.

I wouldn't like to have that limitation just to avoid the small inconvenience of renaming files while copying them into the bundle directory during creation.

mb-karo:
I also ran into this issue, and I also was unable to put any *.bin* file into a raw partition.

So I made this patch, and re-compiled RAUC and now it works as I need it. Any comments to this?
Maybe a bit "too" generic?

```
diff --git a/src/update_handler.c b/src/update_handler.c
index f800cf2..c783b0f 100644
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -2778,7 +2778,7 @@ RaucUpdatePair updatepairs[] = {
 #if ENABLE_EMMC_BOOT_SUPPORT == 1
        {"*.img", "boot-emmc", img_to_boot_emmc_handler},
 #endif
-       {"*", "boot-emmc", NULL},
+       {"*", "boot-emmc", img_to_boot_emmc_handler},
        {"*.vfat", "boot-mbr-switch", img_to_boot_mbr_switch_handler},
        {"*.img", "boot-mbr-switch", img_to_boot_mbr_switch_handler},
[...]

Issue #1177, opened 2023-06-28
Image type detection to include ".verity" rootfs extension


Gerry546:
**Is your feature request related to a problem? Please describe.**
Currently I'm working on integrating dm-verity support together with RAUC in a system. 
Using the bbclass provided in meta-security (https://gitlab.com/akuster/meta-security/-/blob/master-next/classes/dm-verity-img.bbclass) to add verity support to the image this class will ensure that the rootfs images are in the `DEPLOYDIR` and all are appended with `.verity`. These can then be included in `.wic` and bundles via raw copies.
As mentioned in the RAUC docs my `system.conf` is adjusted to do `raw` copies only. This also has as a consequence (AFAIK) that the rootfs bundled into the image needs to have the `.img` extension for the RAUC installer to properly handle the `raw` type.

**Describe the solution you'd like**
Ensure the RAUC understands any rootfs slot with the extension .verity (such as `.ext4.verity`, `.btrfs.verity`) when handling a raw copy.

**Describe alternatives you've considered**
Fow now I have patched rauc v1.9 to enable these.
Alternatively I was considering to simply rename the files during a build such that RAUC will correctly pick them up.

ejoerns:
@Gerry546  It's always a compromise between what we say are supported standard image types that we should have in the list and which are just some special cases following a custom naming convention. In the past we have adapted some of the built-in supported file name extensions to match at least what oe-core provides you. I am not sure if we should also cover meta-security or if this in the end makes the mapping scheme too complex and confusing.

So for now, what should work and should be supported by meta-rauc, in case you use this for integration is to use

```
RAUC_SLOT_verity[type] = file
```

This will let meta-rauc add a generic `.img` to the original file name and this in turn will then be correctly recognized and handled by the img_to_raw_handler in RAUC. Would that be sufficient for your use case?

Gerry546:
I understand your comment and it is indeed always a compromise. 
I will try with the `type = file` and the `.img` extension to see if this works.

Thank you for the comment so far

jluebbe:
@Gerry546 Is this issue solved for you?

Gerry546:
@jluebbe The issue can be considered resolved. I understand the necessary tradeoff between 'commonly' used data and extension types (like mentioned in #1201 ) and the need to keep this system to a sort of standarized set of items. 
[...]

PR #1201, opened 2023-08-01
src/update_handler: support *.bin for u-boot emmc-boot image

hthiery:
Since a common name for U-Boot binaries is flash.bin or u-boot.bin add support for the "*.bin" suffix for the emmc-boot type.

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -2777,6 +2777,7 @@ RaucUpdatePair updatepairs[] = {
 	{"*.squashfs-zst", "ubivol", img_to_ubivol_handler},
 #if ENABLE_EMMC_BOOT_SUPPORT == 1
 	{"*.img", "boot-emmc", img_to_boot_emmc_handler},
+	{"*.bin", "boot-emmc", img_to_boot_emmc_handler},
 #endif
 	{"*", "boot-emmc", NULL},
 	{"*.vfat", "boot-mbr-switch", img_to_boot_mbr_switch_handler},


ejoerns (approved):
Sounds reasonable to me.

a3f:
This would resolve #1009, but @jluebbe had [concerns about this ](https://github.com/rauc/rauc/issues/1009#issuecomment-1643503104):
> As mentioned in [#1009 (comment)](https://github.com/rauc/rauc/issues/1009#issuecomment-1327260133), we don't want to add more broad matches here. The easy fix is to name rename the images to `.img` in your build system. Even if we allow `.bin` now, that would only set a precedent for things like `.raw` or any other file extension that people like to use for images.

Paiusco:
@a3f I think would also add precedent towards #1177 as well...

ejoerns:
I would differentiate between the renaming applied by integration tools and very common original names set by the generating tools. We allow `.ext4` for ext4 images since this is the common extension.
I guess arguing that U-Boot is some kind of widespread embedded bootloader is not fully wrong... thus one could argue that `.bin` is a very common extension for bootloaders.
But on the other hand, I do not have a very strong opinion on this.

Also, if I understand @jluebbe in #1009 correctly, he was referring to 'broad matches' like `"*"` for `"boot-emmc"` or `"*.bin"` for all slot types. Explicitly handling `"*.bin"` for `"boot-emmc"` does not appear to be a 'broach match' to me.

jluebbe:
Instead of adding more and more extension matches, perhaps an explicit option in the manifest to set the file type would be better.

Louson:
> Instead of adding more and more extension matches, perhaps an explicit option in the manifest to set the file type would be better.

@jluebbe @ejoerns Would you have any clue how to do that ? Would you use an existing option in the manifest or create a new one ?

ejoerns:
@Louson This would be a new option. I guess it makes sense to call it `type`.

I have created https://github.com/rauc/rauc/issues/1530 to track this.

Thus I would close this PR for now.

Issue #1200, opened 2023-08-01
Casync update on NAND mtd partition


claudels:
Hello,

**Describe the bug**
I use RAUC as update system on custom board. All works fine but I am trying to set up casync updates to save Mo on remote updates.  I can't get a casync update to work on a nand partition. Here's the command I use and its output : 

```
# G_MESSAGES_DEBUG="all" rauc install /run/media/sda1/casync-guinault-dev-bundle-gmc160-hdmi-20230731144653.raucb -d
rauc-Message: 09:34:30.734: Debug log domains: 'all'
(rauc:4015): rauc-DEBUG: 09:34:30.741: install started
(rauc:4015): rauc-DEBUG: 09:34:30.741: input bundle: /run/media/sda1/casync-guinault-dev-bundle-gmc160-hdmi-20230731144653.raucb
(rauc:4015): rauc-DEBUG: 09:34:30.774: Trying to contact rauc service
installing
  0% Installing
  0% Determining slot states
 20% Determining slot states done.
 20% Checking bundle
 20% Verifying signature
 40% Verifying signature done.
 40% Checking bundle done.
 40% Checking manifest contents
 60% Checking manifest contents done.
 60% Determining target install group
 80% Determining target install group done.
 80% Updating slots
 80% Checking slot dtb.1
 83% Checking slot dtb.1 done.
 83% Copying image to dtb.1
 86% Copying image to dtb.1 failed.
 99% Updating slots failed.
100% Installing failed.
LastError: Installation error: Failed updating slot dtb.1: failed to run casync extract: Child process exited with code 1
Installing `/run/media/sda1/casync-guinault-dev-bundle-gmc160-hdmi-20230731144653.raucb` failed
```

I also noticed these messages in the logs : 
```
[ 1908.584153] loop0: detected capacity change from 0 to 285
[ 1908.627459] loop_set_status: loop0 () has still dirty pages (nrpages=4)
[ 1908.667460] loop0: detected capacity change from 285 to 280
[ 1908.712448] device-mapper: verity: sha256 using implementation "sha256-caam"
[ 1909.212847] nand: nand_do_write_ops: attempt to write non page aligned data
```

I can't get more log about the subprocess error. We use sysvinit instead of systemd in our distribution because we have boot time constraints.

**Background information**
[...]

PR #1210, opened 2023-08-08
src: implement --ignore-image option for rauc convert (casync)

ejoerns:
This option allows providing the slotclass of an image to ignore during conversion. Multiple images can be ignored by using the argument multiple times.

This can be useful to generate 'mixed' casync bundles that contain both .caidx/.caibx images as well as conventional ones.

Usage:

    rauc convert [..] --ignore-image=rootfs input-bundle.raucb mixed-bundle.raucb

Fixes: #1200

Note: This is a proof-of-concept and a base for discussion only for now.

### DIFF
--- a/include/bundle.h
+++ b/include/bundle.h
@@ -221,9 +221,10 @@ G_GNUC_WARN_UNUSED_RESULT;
  *
  * @param bundle RaucBundle struct as returned by check_bundle()
  * @param outbundle output location for converted casync bundle
+ * @param ignore_images string list of slot classes of images to ignore during conversion
  * @param error Return location for a GError
  */
-gboolean create_casync_bundle(RaucBundle *bundle, const gchar *outbundle, GError **error)
+gboolean create_casync_bundle(RaucBundle *bundle, const gchar *outbundle, const gchar **ignore_images, GError **error)
 G_GNUC_WARN_UNUSED_RESULT;
 
 /**
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -1055,7 +1055,7 @@ out:
 	return res;
 }
 
-static gboolean convert_to_casync_bundle(RaucBundle *bundle, const gchar *outbundle, GError **error)
+static gboolean convert_to_casync_bundle(RaucBundle *bundle, const gchar *outbundle, const gchar **ignore_images, GError **error)
 {
 	GError *ierror = NULL;
 	gboolean res = FALSE;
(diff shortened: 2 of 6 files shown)


ejoerns:
> One complication is that old RAUC versions would reject manifests with these new options, so we'd need to have a way to suppress the for compatibility... Perhaps that's still better than the option above.

I did not see any argument against a command line option, so why is config better?
The only issue I see is that when extending the use-cases of 'convert', the option might cause confusion, but there are casync-specific options existing, anyway. So maybe it's just a matter of naming?

I thought this is the least-invasive approach and is compatible with all existing setups.

include/bundle.h
```diff
@@ -223,7 +223,7 @@ G_GNUC_WARN_UNUSED_RESULT;
  * @param outbundle output location for converted casync bundle
  * @param error Return location for a GError
  */
-gboolean create_casync_bundle(RaucBundle *bundle, const gchar *outbundle, GError **error)
+gboolean create_casync_bundle(RaucBundle *bundle, const gchar *outbundle, const gchar **ignore_list, GError **error)
```
jluebbe, line 226:
```suggestion
gboolean create_casync_bundle(RaucBundle *bundle, const gchar *outbundle, const gchar **ignore_images, GError **error)
```

src/bundle.c
```diff
@@ -1055,7 +1055,7 @@ gboolean resign_bundle(RaucBundle *bundle, const gchar *outpath, GError **error)
 	return res;
 }
 
-static gboolean convert_to_casync_bundle(RaucBundle *bundle, const gchar *outbundle, GError **error)
+static gboolean convert_to_casync_bundle(RaucBundle *bundle, const gchar *outbundle, const gchar **ignore_list, GError **error)
```
jluebbe, line 1058:
```suggestion
static gboolean convert_to_casync_bundle(RaucBundle *bundle, const gchar *outbundle, const gchar **ignore_images, GError **error)
```

src/main.c
```diff
@@ -41,6 +41,7 @@ gchar **intermediate = NULL;
 gchar *signing_keyring = NULL;
 gchar *mksquashfs_args = NULL;
 gchar *casync_args = NULL;
+gchar **convert_ignore = NULL;
```
jluebbe, line 44:
```suggestion
gchar **convert_ignore_images = NULL;
```

src/main.c
```diff
@@ -2144,6 +2145,7 @@ static GOptionEntry entries_convert[] = {
 	{"signing-keyring", '\0', 0, G_OPTION_ARG_FILENAME, &signing_keyring, "verification keyring file", "PEMFILE"},
 	{"mksquashfs-args", '\0', 0, G_OPTION_ARG_STRING, &mksquashfs_args, "mksquashfs extra args", "ARGS"},
 	{"casync-args", '\0', 0, G_OPTION_ARG_STRING, &casync_args, "casync extra args", "ARGS"},
+	{"ignore-image", '\0', 0, G_OPTION_ARG_STRING_ARRAY, &convert_ignore, "ignore image during conversion", "SLOTCLASS"},
```
jluebbe, line 2148: It should be explicit that the slot class name is needed here.

jluebbe (changes requested):
OK. This looks like the least invasive change to allow ignoring images.

jluebbe (approved):

src/main.c
ejoerns: FTR: I haven't changed anything here, yet since I wasn't sure what's the best way to make it more explicit

PR #1214, opened 2023-08-10
src/install: fix error handling on status write failure during slot update failure

jluebbe:
Previously, if `r_slot_status_save()` failed after an slot handler failure, we reported the handler error with a mismatching prefix. Instead, capture the slot status error in a separate `GError` and report it explicitly.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -1052,15 +1052,16 @@ static gboolean handle_slot_install_plan(const RaucManifest *manifest, const RIm
 	r_context_begin_step_weighted_formatted("copy_image", 0, 9, "Copying image to %s", plan->target_slot->name);
 
 	if (!plan->slot_handler(plan->image, plan->target_slot, hook_name, &ierror)) {
+		g_autoptr(GError) ierror_status = NULL;
+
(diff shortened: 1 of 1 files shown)


src/install.c
```diff
@@ -1052,14 +1052,16 @@ static gboolean handle_slot_install_plan(const RaucManifest *manifest, const RIm
...
 				"Failed updating slot %s: ", plan->target_slot->name);
 		r_context_end_step("copy_image", FALSE);
 
 		g_message("Updating slot %s status", plan->target_slot->name);
 		update_slot_status(slot_state, "failed", manifest, plan, args);
-		if (!r_slot_status_save(plan->target_slot, NULL)) {
-			g_propagate_prefixed_error(error, ierror, "Error while writing status file: ");
+		if (!r_slot_status_save(plan->target_slot, &ierror_status)) {
+			g_warning("Error while writing status file after slot update failure: %s", ierror_status->message);
 			return FALSE;
```
ejoerns, line 1065: Maybe we should then also remove the return statement here to prevent confusion?

ejoerns (approved):

PR #1215, opened 2023-08-11
src/context: get current bootslot on custom bootloader

angeloc:
The Nvidia Tegra UEFI offers its own method to enable A/B redundancy, that it's handled internally by the L4T UEFI.

The nvbootctrl command can be used to do maintenance on A/B mechanism, inspect the status and change the next reboot booting slot.

It's not possible also, to modify the kernel cmdline from inside Linux because it is hardcoded in the bootloader at compile time.
To handle the A/B booting process, the system doesn't add the `root=` option to the cmdline, there is indeed a script which inspects the UEFI variables statuses and mount the correct rootfs at the right moment.

Therefore, using the cmdline to understand on which slot the system is running is not possible on this family platforms.

Using the custom backend is the only alternative.

I tested this on a jetson agx xavier board.


src/context.c
```diff
@@ -140,6 +140,39 @@ static gchar* get_cmdline_bootname(void)
...
+	g_autoptr(GSubprocessLauncher) launcher = NULL;
+	g_autoptr(GSubprocess) handle = NULL;
+	g_autoptr(GDataInputStream) datainstream = NULL;
+	g_autoptr(GPtrArray) args_array = NULL;
+	g_autofree gchar *outline = NULL;
+	GInputStream *instream;
+
+	args_array = g_ptr_array_new();
+	g_ptr_array_add(args_array, context->config->custom_bootloader_backend);
+	g_ptr_array_add(args_array, "get-primary");
```
ejoerns, line 154: 'primary' in our API is the slot (bootname) to be booted next.

Once you have installed an update or manually marked a slot active, this information will change and subsequent calls of this function without a reboot will give incorrect results.

If you have an explicit method that provides the information about the currently booted slot, it would be better to add an explicit API call in RAUC's bootchooser implementation (like `get-current`). For example, UEFI should already provide the current booted entry in the `BootCurrent` variable.
Maybe using this is an option, too?.

angeloc: > If you have an explicit method that provides the information about the currently booted slot, it would be better to add an explicit API call in RAUC's bootchooser implementation (like get-current).

I like this idea, but the `custom_backend_get` uses `r_context` internally, which is not available in context.c. Should I add a minimal method like the one I proposed?

> For example, UEFI should already provide the current booted entry in the BootCurrent variable.

Not this UEFI, this provides `781e084c-a330-417c-b678-38e696380cb9-BootChainOsCurrent`

ejoerns: > I like this idea, but the custom_backend_get uses r_context internally, which is not available in context.c. 

I see.

> Should I add a minimal method like the one I proposed?

Having it in bootchooser would be even better, but I guess we could also accept it in context only first of all.

However, the commit should then also touch the [Custom bootloader backend interface](https://rauc.readthedocs.io/en/latest/integration.html#custom-bootloader-backend-interface) documentation and mentions the purpose of `get-current`.

> Not this UEFI, this provides 781e084c-a330-417c-b678-38e696380cb9-BootChainOsCurrent

Oh, interesting. I've assumed this would be a standard var so far.

angeloc:
@ejoerns I think the PR should pass all the tests now

angeloc:
@ejoerns Sorry, I fixed the remaining issue, now it should really pass all the tests!

angeloc:
@ejoerns For the codecov, I'm not sure on how to proceed. Can you share some hints?

src/context.c
```diff
@@ -140,6 +140,43 @@ static gchar* get_cmdline_bootname(void)
...
+static gchar* get_custom_bootname(void)
+{
+	g_autoptr(GSubprocessLauncher) launcher = NULL;
+	g_autoptr(GSubprocess) handle = NULL;
+	g_autoptr(GDataInputStream) datainstream = NULL;
+	g_autoptr(GPtrArray) args_array = NULL;
+	g_autofree gchar *outline = NULL;
+	GInputStream *instream;
+
+	if (context->mock.custom_bootloader_output) {
```
ejoerns, line 152: This prevents all the added code from being tested. Please replace this by actually testing the code.

Your config file already references the custom bootloader test script `bin/custom-bootloader-script`. This should be extended to actually return something on a `get-current` argument invocation.

All mock code should be removed.

test/context.c
```diff
@@ -66,6 +66,19 @@ static void test_bootslot_no_bootslot(void)
 	r_context_clean();
 }
 
+static void test_bootslot_custom_bootloader(void)
+{
+	r_context_conf()->configpath = g_strdup("test/test-custom.conf");
+	r_context_conf()->configmode = R_CONTEXT_CONFIG_MODE_REQUIRED;
+	r_context_conf()->mock.proc_cmdline = "quiet";
+	r_context_conf()->mock.custom_bootloader_output = "A";
```
ejoerns, line 74: You might want to write something like

```
g_assert_true(g_setenv("CUSTOM_STATE_PATH", "test/custom-state-current", TRUE));
```

here and provide the file with some pre-set information. As an alternative, you could let your custom script just return a fixed bootname first of all.

test/test-custom.conf
```diff
@@ -0,0 +1,47 @@
+# testsuite system configuration
```
ejoerns, line 1: This file could be shorter since its purpose is just to test the `custom` bootloader backend. Maybe just have two rootfs slots here.

src/context.c
```diff
@@ -140,6 +140,43 @@ static gchar* get_cmdline_bootname(void)
 	return bootname;
 }
 
+static gchar* get_custom_bootname(void)
```
ejoerns, line 143: This could be a GError method. Alternatively, at least emit `g_warning()` messages for errors inside this function.

src/context.c
```diff
@@ -140,6 +140,43 @@ static gchar* get_cmdline_bootname(void)
...
+	g_autofree gchar *outline = NULL;
+	GInputStream *instream;
+
+	if (context->mock.custom_bootloader_output) {
+		return g_strdup(context->mock.custom_bootloader_output);
+	}
+
+	args_array = g_ptr_array_new();
+	g_ptr_array_add(args_array, context->config->custom_bootloader_backend);
+	g_ptr_array_add(args_array, (gchar *)("get-current"));
```
ejoerns, line 158: Sentinel is missing here. So this should never work.

Add
``` c
g_ptr_array_add(args_array, NULL)
```

src/context.c
```diff
@@ -140,6 +140,43 @@ static gchar* get_cmdline_bootname(void)
...
+	if (handle == NULL) {
+		return NULL;
+	}
+
+	instream = g_subprocess_get_stdout_pipe(handle);
+	datainstream = g_data_input_stream_new(instream);
+
+	outline = g_data_input_stream_read_line(datainstream, NULL, NULL, NULL);
+
+	g_subprocess_wait_check(handle, NULL, NULL);
```
ejoerns, line 172: Also add error handling here

src/context.c
```diff
@@ -140,6 +140,43 @@ static gchar* get_cmdline_bootname(void)
...
+	handle = r_subprocess_launcher_spawnv(launcher, args_array, NULL);
+
+	if (handle == NULL) {
+		return NULL;
+	}
+
+	instream = g_subprocess_get_stdout_pipe(handle);
+	datainstream = g_data_input_stream_new(instream);
+
+	outline = g_data_input_stream_read_line(datainstream, NULL, NULL, NULL);
```
ejoerns, line 170: Should gain an output check like in https://github.com/rauc/rauc/blob/master/src/bootchooser.c#L109

src/context.c
```diff
@@ -140,6 +140,43 @@ static gchar* get_cmdline_bootname(void)
...
+
+	g_subprocess_wait_check(handle, NULL, NULL);
+
+	if (outline)
+		return g_strdup(outline);
```
ejoerns, line 175: When having checked `outline` before, better use `g_steal_pointer` here in case of success then.

ejoerns (changes requested):
I have made a number of comments, mainly referring to testing and error handling.

ejoerns:
> @ejoerns For the codecov, I'm not sure on how to proceed. Can you share some hints?

The loss in coverage should be resolved by actually testing the added code. See my comments.

src/context.c
angeloc: I decide for keeping the method simple as the `get_cmdline_bootname`, so it simply outputs the errors and returns NULL when bootname is not found.

angeloc:
@ejoerns fixed the distcheck now, sorry!

angeloc:
@ejoerns `./uncrustify.sh` is not failing on my machine and not sure why it is failing here

angeloc:
> @ejoerns `./uncrustify.sh` is not failing on my machine and not sure why it is failing here

ok, found and fixed

angeloc:
I sincerely hope all the issues are solved now!

src/context.c
```diff
@@ -140,6 +140,43 @@ static gchar* get_cmdline_bootname(void)
...
+	if (context->mock.custom_bootloader_output) {
+		return g_strdup(context->mock.custom_bootloader_output);
+	}
+
+	args_array = g_ptr_array_new();
+	g_ptr_array_add(args_array, context->config->custom_bootloader_backend);
+	g_ptr_array_add(args_array, (gchar *)("get-current"));
+
+	launcher = g_subprocess_launcher_new(G_SUBPROCESS_FLAGS_STDOUT_PIPE);
+	handle = r_subprocess_launcher_spawnv(launcher, args_array, NULL);
```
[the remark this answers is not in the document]
ejoerns: This has been marked as resolved despite no `GError` being handled. Is there a good reason not to handle the GError?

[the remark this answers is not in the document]
angeloc: I commented on the first comment for context.c

> I decide for keeping the method simple as the get_cmdline_bootname, so it simply outputs the errors and returns NULL when bootname is not found.

Basically, the get_cmdline_bootname is not handling or rising errors, simply returns NULL if the bootname is not there. I decided to keep the same approach, but adding checks where needed and log messages.

src/context.c
ejoerns: This is not a check like the one I mentioned. We could abort earlier here. Most functions in RAUC have the error case in if statements and the 'good' case in the normal function flow. If you abort here, you could steal the pointer below unconditionally.

src/context.c
```diff
@@ -140,6 +140,51 @@ static gchar* get_cmdline_bootname(void)
...
+	}
+
+	instream = g_subprocess_get_stdout_pipe(handle);
+	datainstream = g_data_input_stream_new(instream);
+
+	outline = g_data_input_stream_read_line(datainstream, NULL, NULL, NULL);
+
+	res = g_subprocess_wait_check(handle, NULL, NULL);
+	if (!res) {
+		g_message("Failed to get custom backend output");
```
ejoerns, line 174: Since something went wrong, I would have put a `g_warning()` here. Did you have a reason to prefer `g_message()`?

ejoerns (changes requested):
Already looks quite good now, especially testing changes.

Just some minor comments you might want to address.

src/context.c
angeloc: I simply looked at the surrounding code, `get_cmdline_bootname` is doing

`g_message("Failed to resolve realpath for '%s'", bootname);`

But I'm happy to change to g_warning

docs/integration.rst
```diff
@@ -1110,11 +1120,16 @@ Custom bootloader backend interface
 According to :ref:`sec-boot-slot` the custom bootloader handler is called by 
 RAUC to trigger the following actions:
 
+* get the current booted slot
```
ejoerns, line 1123: I guess I suggested putting this at the end of the list as it is (unlike the others) more or less an optional method. So you prefer having it as the first element?

(If moving, also move the verbose description below)

ejoerns (approved):
Looks fine from my site now, thanks for your effort!

src/context.c
angeloc: Honestly, on the g_warning, I think it modifies the behaviour too much: when running tests, if there is a g_warning, the test bail out and abort instead of continuing and complaining about an assertion failed.
I think we must stay with g_message here.

ejoerns: > Honestly, on the g_warning, I think it modifies the behaviour too much

The tests are configured to handle unexpected warnings as fatal errors.
We normally emit a g_message() if the error is expected and just for information purposes.

It may be a question of whether we expect the custom backend script to return the 'get-current' information or not. If it is actually seen as optional you are probably right that a `g_message()` would be sufficient.

angeloc: Well, if we want to keep the same approach of `get_cmdline_bootname`, that function can fail and return null without any warning.
I would maintain the same approach, the custom handler could fail without any warning and return NULL.

ejoerns (approved):

jluebbe (approved):

PR #1218, opened 2023-08-11
github: workflows: coverity: don't complain about missing secrets

rohieb:
On forked repositories, the coverity token is usually not configured in the secrets, so running the coverity workflow will fail. Since it is configured to run on a nightly cron schedule, this will send one useless e-mail per night to the owner of the fork.

Run the workflow only if the access token is configured.

See e.g. https://github.com/rohieb/rauc/actions/runs/5809280301/job/15747760694 for the failed workflow.


.github/workflows/coverity.yml
```diff
@@ -24,6 +24,11 @@ jobs:
         export COVERITY_SCAN_BUILD_COMMAND="meson compile -C build"
         export COVERITY_SCAN_TOKEN="${{ secrets.COVERITY_SCAN_TOKEN }}"
 
+        if [ "$COVERITY_SCAN_TOKEN" != "rauc/rauc" ]; then
+            echo "Note: empty COVERITY_SCAN_TOKEN; not running coverity"
+            exit
+        fi
```
jluebbe, line 30: This doesn't check if the token is empty, but if it differs from "rauc/rauc".

jluebbe (changes requested):

.github/workflows/coverity.yml
rohieb: That is correct… my first try was testing for COVERITY_SCAN_PROJECT_NAME instead but then I decided that the token makes more sense… 🙈 Fixed it, and improved the code flow a bit too.

jluebbe: Hmm, I'd definitely prefer an early exit to indenting the normal case further.

rohieb: OK.

jluebbe (approved):

PR #1220, opened 2023-08-15
minor fixes

jluebbe:


src/main.c
```diff
@@ -2294,7 +2294,7 @@ static void create_option_groups(void)
 static void cmdline_handler(int argc, char **argv)
 {
 	gboolean help = FALSE, debug = FALSE, version = FALSE;
-	gchar *confpath = NULL, *keyring = NULL, *mount = NULL;
+	g_autofree gchar *confpath = NULL, *keyring = NULL, *mount = NULL;
```
ejoerns, line 2292: The values of these variables will be fed directly into context (like `r_context_conf()->configpath = confpath`).

I would assume that freeing them after leaving the function (without stealing them) will break something.

jluebbe: `cmdline_handler()` is only called from `main()`, just before `return r_exit_status`, so the context is not used after returning from `cmdline_handler()`.

ejoerns: Agreed.
But nor fully sure if the handling should rely on in which context the function is used since this is error-prone for refactoring.
A `g_steal_pointer()` should make it more consistent and would have the same effect, doesn't it?

Anyway, I'll accept it first of all and leave it up to you to change it or not.

ejoerns (approved):
Fixing the subprocess debug log flooding is great!

src/main.c
jluebbe: Without this fix, we leak these strings on errors, so I'd like to get the fix in first.

PR #1221, opened 2023-08-15
test: fix pre-check tests and generate slot artifacts only when needed

jluebbe:
This increases the flexibility for testing artifacts later.

### DIFF
--- a/test/install-fixtures.c
+++ b/test/install-fixtures.c
@@ -126,6 +126,9 @@ void fixture_helper_set_up_bundle(gchar *tmpdir,
 	g_autoptr(GError) error = NULL;
 	gboolean res = FALSE;
 
+	g_assert_nonnull(tmpdir);
+	g_assert_nonnull(options);
+
 	/* needs to run as root */
 	if (!test_running_as_root())
 		return;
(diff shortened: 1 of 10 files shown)


test/install-fixtures.c
```diff
@@ -139,34 +140,30 @@ void fixture_helper_set_up_bundle(gchar *tmpdir,
...
-			SLOT_SIZE, "/dev/zero") == 0);
-	g_assert_true(test_make_filesystem(tmpdir, "content/rootfs.ext4"));
-	g_assert_true(test_make_filesystem(tmpdir, "content/appfs.ext4"));
-	g_assert_true(test_make_filesystem(tmpdir, "content/bootloader.ext4"));
-	if (manifest_content) {
-		g_autofree gchar *tmpfile = write_tmp_file(tmpdir, "content/manifest.raucm", manifest_content, NULL);
-		g_assert_nonnull(tmpfile);
-	} else {
-		g_assert(test_prepare_manifest_file(tmpdir, "content/manifest.raucm", options) == 0);
+	if (manifest_content || (options && options->slots)) {
```
ejoerns, line 143: Why does this need to be guarded with `if (manifest_content)` ?

jluebbe: It's not needed. `manifest_content` is optionally used instead of a generated manifest.

ejoerns (approved):

PR #1222, opened 2023-08-15
update_utils: make copy_with_progress reusable

jluebbe:
This renames copy_with_progress to r_copy_stream_with_progress and moves it to
a new update_utils module. Also change the return value to a `gboolean`, as a
partial copy is an error case for this function.

### DIFF
--- a/src/update_utils.c
+++ b/src/update_utils.c
@@ -1,0 +1,49 @@
+#include <errno.h>
+#include <gio/gio.h>
+#include <glib.h>
+#include <glib/gstdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "update_utils.h"
+#include "context.h"
+
+gboolean r_copy_stream_with_progress(GInputStream *in_stream, GOutputStream *out_stream,
+		goffset size, GError **error)
+{
+	GError *ierror = NULL;
+	gsize out_size = 0;
+	goffset sum_size = 0;
+	gint last_percent = -1, percent;
+	gchar buffer[8192];
+	gssize in_size;
+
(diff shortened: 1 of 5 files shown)


src/update_utils.c
```diff
@@ -0,0 +1,49 @@
...
+#include <glib.h>
+#include <glib/gstdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "update_utils.h"
+#include "context.h"
+
+gboolean r_copy_stream_with_progress(GInputStream *in_stream, GOutputStream *out_stream,
```
ejoerns, line 12: I would have preferred not to mix the rename/move with changing the function's return value, but since it is not too invasive, it might still be ok.

ejoerns (changes requested):
Please add the file for meson, too

jluebbe:
> Please add the file for meson, too

Isn't this enough? https://github.com/rauc/rauc/pull/1222/files#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4

ejoerns:
> > Please add the file for meson, too
> 
> Isn't this enough? https://github.com/rauc/rauc/pull/1222/files#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4

Hmm, had a look for it but seems to have been somehow hidden by the diff view. Fine like it is.

ejoerns (approved):

src/update_utils.c
jluebbe: OK, I'll split it up.

PR #1225, opened 2023-08-16
create links to active slot devices in /run/rauc/slots/active

jluebbe:
This will be extended by artifact repository links later.


ejoerns (approved):
:bee:

PR #1230, opened 2023-08-17
Configurable Event Logging

ejoerns:
This is a first draft implementation of the event logging feature noted in concept issue #1070

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -1594,3 +1594,62 @@ https://github.com/caruhome/upparat
 It is also available via PyPI:
 
 https://pypi.org/project/upparat/
+
+.. _sec-advanced-event-log:
+
+RAUC Installation History and Event logging
+-------------------------------------------
+
+Even if RAUC mainly focuses on logging information to stdout or into the
+journal (when using systemd), this might be insufficient for some purposes and
+especially for keeping long-term history of what RAUC changed on the system.
+
+A common problem for example can be journal rotation. Since storage is limited
+and the journal contains a lot of other information, it needs to be rotated at
+some point.
+However, one might want to preserve the history of what RAUC installed on the
+system or when the system rebooted, went into fallback, etc. for very long or
+even the full life time of the device.
+
+Another motivation can be to have a clearly separated distinct log location
+where other system components or a service technician (that should not have
+access to the whole system) should have a look into.
+
+The RAUC 'event logging' handling targets these and other cases.
+It defines a distinct set of events that might be of interest for later
+introspection, debugging or informative output.
+
+Via RAUC's ``system.conf`` one or several loggers can be configured with
+selectable output format, event filters, and also basic log rotation is
+supported.
+
+A new logger can be registered with adding a ``log.<loggername>`` section to
+the ``system.conf``.
+
+To have e.g. an unlimited human-readable short log of the installations
+happened on the system, use::
+
+  [log.install-log]
+  filename=install.log
+  events=install
+  output-format=short
+
+Or, if you want a json-based log of all events, limited to 1M per log file and
+5 rotation files to keep, use::
+
+  [log.all-json-log]
+  filename=all-json.log
+  output-format=json
+  max-size=1M
+  max-files=5
+
+If an error occurs during logging (such as disk full or write errors), that
+logger is marked as broken and no longer used.
+An ongoing installation is **not** aborted.
+
+For a full reference of supported configuration options, see
+:ref:`logger sections reference <ref-logger-sections>`.
+
+.. note:: All events logged using the internal event logging framework will
+   also be forwarded to the default logger and thus be visible e.g. in the
+   journal (when using systemd).
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -697,6 +697,55 @@ No built-in slot update will run and no hook will be executed.
   ``<key>``.
   ``-`` is converted to ``_`` for use as an environment variable name.
 
+.. _ref-logger-sections:
+
+**[log.<logger>] sections**
+
+With a logger section, a *RAUC event logger* can be configured.
+The ``<logger>`` suffix determines the internal name of the logger and must be
+unique per ``system.conf``.
+
+For an overview over the event logging framework in RAUC and its purpose, have
+a look at :ref:`sec-advanced-event-log`.
+
+``filename``
+  The log file name used for logging.
+  If no absolute path is given, the location is assumed to be relative to the
+  ``data-directory``.
+  Using a relative file name without ``data-directory`` set will cause a
+  configuration error.
+
+``events``
+  Semicolon-separated list of events to log. Currently supported event types are:
+
+  * ``install`` - Logs start and end of installation
+  * ``boot`` - Logs boot information
+  * ``mark`` - Logs slot marking information
+  * ``all`` - Log all events (cannot be combined with other events)
+
+``format``
+  The output format used for the logger. Supported values are
+
+  * ``readable``: readable mutli-line output
+  * ``short``: Single-line readable output
+  * ``json``: single-line JSON output
+  * ``json-pretty``: formatted JSON output
+
+``max-size``
+  Allows to configure a basic log rotation.
+  When given, the logger's log file will be rotated before reaching
+  the size configured with ``max-size`` and renamed to ``<filename>.1``.
+  Existing rotation file names will be incremented by one.
+  The oldest file is removed.
+  To configure a maximum number of files to keep, see ``max-files``.
+  Values support common suffixes like ``K``, ``M``, ``G``, to ``T``.
+
+``max-files``
+  Configures the maximum number of files to keep per logger.
+  E.g. if set to ``3``, only ``<filename>``, ``<filename>.1`` and
+  ``<filename>.2`` will be kept during rotation.
+  Defaults to 10 if unset.
+
 .. _sec_ref_formats:
 
 Bundle Formats
--- a/qemu-test-rauc-config
+++ b/qemu-test-rauc-config
@@ -5,6 +5,11 @@ grubenv=/tmp/boot/grub/grubenv
 # HACK: must be on a separate partition for real systems
 data-directory=/tmp/rauc
 
+[log.testlogger]
+# create an event log file under /tmp/rauc/event.log
+filename=event.log
+events=all
+
 [keyring]
 path=ca.cert.pem
 
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -4,6 +4,7 @@
 #include "bootchooser.h"
 #include "config_file.h"
 #include "context.h"
+#include "event_log.h"
 #include "install.h"
 #include "manifest.h"
 #include "mount.h"
@@ -155,6 +156,174 @@ out:
 	if (res)
 		*mask = imask;
 	return res;
+}
+
+#define RAUC_LOG_EVENT_CONF_PREFIX "log"
+
+static gboolean r_event_log_parse_config_sections(GKeyFile *key_file, RaucConfig *config, GError **error)
+{
+	gsize group_count;
+	g_auto(GStrv) groups = NULL;
+	gint tmp_maxfiles;
+
+	g_return_val_if_fail(key_file, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	g_assert_null(config->loggers);
+
+	/* parse [log.*] sections */
+	groups = g_key_file_get_groups(key_file, &group_count);
+	for (gchar **group = groups; *group != NULL; group++) {
+		GError *ierror = NULL;
+		g_autoptr(REventLogger) logger = NULL;
+		const gchar *logger_name;
+		g_autofree gchar *log_format = NULL;
+		gsize entries;
+
+		if (!g_str_has_prefix(*group, RAUC_LOG_EVENT_CONF_PREFIX "."))
+			continue;
+
+		logger_name = *group + strlen(RAUC_LOG_EVENT_CONF_PREFIX ".");
+
+		logger = g_new0(REventLogger, 1);
+		logger->name = g_strdup(logger_name);
+
+		/* 'filename' option is currently mandatory for a logger group */
+		logger->filename = key_file_consume_string(key_file, *group, "filename", &ierror);
+		if (!logger->filename) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+		/* relative paths are resolved relative to the data-directory */
+		if (!g_path_is_absolute(logger->filename)) {
+			gchar *abspath;
+			if (!config->data_directory) {
+				g_set_error_literal(
+						error,
+						G_KEY_FILE_ERROR,
+						G_KEY_FILE_ERROR_INVALID_VALUE,
+						"Relative filename requires data-directory to be set");
+				return FALSE;
+			}
+
+			abspath = g_build_filename(config->data_directory, logger->filename, NULL);
+			g_free(logger->filename);
+			logger->filename = abspath;
+		}
+
+		log_format = key_file_consume_string(key_file, *group, "format", &ierror);
+		if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
+			log_format = g_strdup("readable");
+			g_clear_error(&ierror);
+		} else if (ierror) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+
+		if (g_strcmp0(log_format, "readable") == 0) {
+			logger->format = R_EVENT_LOGFMT_READABLE;
+		} else if (g_strcmp0(log_format, "short") == 0) {
+			logger->format = R_EVENT_LOGFMT_READABLE_SHORT;
+		} else if (g_strcmp0(log_format, "json") == 0) {
+#if ENABLE_JSON
+			logger->format = R_EVENT_LOGFMT_JSON;
+#else
+			g_set_error(
+					error,
+					G_KEY_FILE_ERROR,
+					G_KEY_FILE_ERROR_INVALID_VALUE,
+					"Invalid log format %s. RAUC is compiled without JSON support", log_format);
+			return FALSE;
+#endif
+		} else if (g_strcmp0(log_format, "json-pretty") == 0) {
+#if ENABLE_JSON
+			logger->format = R_EVENT_LOGFMT_JSON_PRETTY;
+#else
+			g_set_error(
+					error,
+					G_KEY_FILE_ERROR,
+					G_KEY_FILE_ERROR_INVALID_VALUE,
+					"Invalid log format %s. RAUC is compiled without JSON support", log_format);
+			return FALSE;
+#endif
+		} else {
+			g_set_error(
+					error,
+					G_KEY_FILE_ERROR,
+					G_KEY_FILE_ERROR_INVALID_VALUE,
+					"Unknown log format '%s'", log_format);
+			return FALSE;
+		}
+
+		logger->events = g_key_file_get_string_list(key_file, *group, "events", &entries, &ierror);
+		if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
+			logger->events = g_malloc(2 * sizeof(gchar *));
+			logger->events[0] = g_strdup("all");
+			logger->events[1] = NULL;
+			g_clear_error(&ierror);
+		} else if (ierror) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+
+		if (g_strv_length(logger->events) > 1 && g_strv_contains((const gchar * const *) logger->events, "all")) {
+			g_set_error(
+					error,
+					G_KEY_FILE_ERROR,
+					G_KEY_FILE_ERROR_INVALID_VALUE,
+					"Event type 'all' cannot be combined");
+			return FALSE;
+		}
+		for (gsize j = 0; j < entries; j++) {
+			/* FIXME: handle invalid values */
+		}
+		g_key_file_remove_key(key_file, *group, "events", NULL);
+
+		logger->maxsize = key_file_consume_binary_suffixed_string(key_file, *group, "max-size", &ierror);
+		if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
+			g_clear_error(&ierror);
+		} else if (ierror) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+
+		tmp_maxfiles = key_file_consume_integer(key_file, *group, "max-files", &ierror);
+		if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
+			tmp_maxfiles = 10;
+			g_clear_error(&ierror);
+		} else if (ierror) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+		if (tmp_maxfiles < 1) {
+			g_set_error_literal(
+					error,
+					G_KEY_FILE_ERROR,
+					G_KEY_FILE_ERROR_INVALID_VALUE,
+					"Value for 'max-files' must be >= 1");
+			return FALSE;
+		} else if (tmp_maxfiles > 1000) {
+			g_set_error_literal(
+					error,
+					G_KEY_FILE_ERROR,
+					G_KEY_FILE_ERROR_INVALID_VALUE,
+					"Value of %d for 'max-files' looks implausible");
+			return FALSE;
+		}
+		logger->maxfiles = (guint) tmp_maxfiles;
+
+		if (!check_remaining_keys(key_file, *group, &ierror)) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+
+		g_key_file_remove_group(key_file, *group, NULL);
+
+		/* insert new logger in list */
+		config->loggers = g_list_append(config->loggers, g_steal_pointer(&logger));
+	}
+
+	return TRUE;
 }
 
 static GHashTable *parse_slots(const char *filename, const char *data_directory, GKeyFile *key_file, GError **error)
@@ -873,6 +1042,11 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 	}
 	g_key_file_remove_group(key_file, "handlers", NULL);
 
+	if (!r_event_log_parse_config_sections(key_file, c, &ierror)) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
 	/* parse [slot.*.#] sections */
 	c->slots = parse_slots(filename, c->data_directory, key_file, &ierror);
 	if (!c->slots) {
@@ -938,6 +1112,7 @@ void free_config(RaucConfig *config)
 	g_strfreev(config->enabled_headers);
 	g_free(config->encryption_key);
 	g_free(config->encryption_cert);
+	g_list_free_full(config->loggers, (GDestroyNotify)r_event_log_free_logger);
 	g_clear_pointer(&config->slots, g_hash_table_destroy);
 	g_free(config->custom_bootloader_backend);
 	g_free(config->file_checksum);
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -275,7 +275,14 @@ static gboolean r_event_log_parse_config_sections(GKeyFile *key_file, RaucConfig
 			return FALSE;
 		}
 		for (gsize j = 0; j < entries; j++) {
-			/* FIXME: handle invalid values */
+			if (!r_event_log_is_supported_type(logger->events[j])) {
+				g_set_error(
+						error,
+						G_KEY_FILE_ERROR,
+						G_KEY_FILE_ERROR_INVALID_VALUE,
+						"Unsupported event log type '%s'", logger->events[j]);
+				return FALSE;
+			}
 		}
 		g_key_file_remove_key(key_file, *group, "events", NULL);
 
--- a/src/context.c
+++ b/src/context.c
@@ -4,6 +4,7 @@
 
 #include "config_file.h"
 #include "context.h"
+#include "event_log.h"
 #include "status_file.h"
 #include "network.h"
 #include "install.h"
@@ -374,6 +375,13 @@ static gboolean r_context_configure_target(GError **error)
 			g_message("Failed to load system status: %s", ierror->message);
 			g_clear_error(&ierror);
 		}
+	}
+
+	/* set up logging */
+	for (GList *l = context->config->loggers; l != NULL; l = l->next) {
+		REventLogger* logger = l->data;
+
+		r_event_log_setup_logger(logger);
 	}
 
 	if (context->config->system_variant_type == R_CONFIG_SYS_VARIANT_DTB) {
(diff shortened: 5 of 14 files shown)


src/event-log.c
```diff
@@ -0,0 +1,402 @@
...
+	/* determine number of existing rotation files */
+	for (guint num = 0; num < G_MAXUINT; num++) {
+		g_autofree gchar *tryfile = g_strdup_printf("%s.%d", logger->filename, num);
+
+		if (g_file_test(tryfile, G_FILE_TEST_EXISTS)) {
+			existing_files++;
+		} else {
+			break;
+		}
+	}
```
jluebbe, line 239: Can this be dropped by always trying to rotate `maxfiles` files and ignoring `ENOENT` on `rename`?

src/event-log.c
```diff
@@ -0,0 +1,402 @@
...
+void r_event_log_free_logger(REventLogger *logger)
+{
+	if (!logger)
+		return;
+
+	g_clear_pointer(&logger->name, g_free);
+	g_clear_pointer(&logger->filename, g_free);
+	g_clear_pointer(&logger->logstream, g_object_unref);
+
+
```
jluebbe, line 50: Redundant empty line.

src/event-log.c
```diff
@@ -0,0 +1,402 @@
...
+	gsize written = 0;
+
+	g_assert_nonnull(logger->filename);
+
+	if (!logger->logstream) {
+		g_autoptr(GFile) logfile = NULL;
+		logfile = g_file_new_for_path(logger->filename);
+		logger->logstream = g_file_append_to(logfile, G_FILE_CREATE_NONE, NULL, &ierror);
+		if (!logger->logstream) {
+			g_warning("Failed opening log file for appending: %s", ierror->message);
```
jluebbe, line 295: "Failed to open..."

src/event-log.c
```diff
@@ -0,0 +1,402 @@
...
+		default:
+			g_error("Unknown log format");
+	}
+	output = g_strdup_printf("%s\n", formatted);
+
+	/* Once we know how much to write, we can use this information for trimming */
+	rotate_logfiles(logger, strlen(output));
+
+	if (!g_output_stream_write_all(G_OUTPUT_STREAM(logger->logstream), output, strlen(output), &written, NULL, &ierror)) {
+		g_warning("Failed writing log file: %s", ierror->message);
```
jluebbe, line 330: "Failed to write..."

qemu-test-rauc-config
```diff
@@ -5,6 +5,10 @@ grubenv=/tmp/boot/grub/grubenv
 # HACK: must be on a separate partition for real systems
 data-directory=/tmp/rauc
 
+[log.testlogger]
+filename=event.log
```
jluebbe, line 9: Should this be somewhere in `/tmp`? If not, where does this end up?

ejoerns: default for relative paths is currently relative to data-directory. Thus `/tmp/rauc/event.log` here.

src/event-log.c
ejoerns: I've thought about this, too. But have not found a proper solution for not having maxfiles set. Or there must be a default maxfiles of like 100 files or so.

src/event-log.c
```diff
@@ -0,0 +1,405 @@
...
+
+/**
+ * Log formatter for json output.
+ *
+ * Simply converts all log fields into a json object.
+ *
+ * | {
+ * |   "date": "2023-06-14T21:15:41Z"
+ * |   "MESSAGE" : "Booted into rootfs.0 (A)",
+ * |   "MESSAGE_ID" : "e60e0add-d345-4cb8-b796-eae0d497af96",
```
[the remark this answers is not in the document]
ejoerns: Different in which way? It's just an example with two entries. Could be shortened to one.

src/event-log.c
jluebbe: I'd say a default of 10 (so 1…9) would be fine. That would keep the size fluctuation low enough.

qemu-test-rauc-config
jluebbe: Perhaps a comment here would be nice.

src/event-log.c
ejoerns: 10 would be fine for me. Currently maxfiles is defined as the number of rotation files to keep (thus would be 1..10). Do you think this should be changed?

src/event-log.c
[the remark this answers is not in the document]
ejoerns: Ok, misunderstood you. So the reason for having "date" as it is is just that i wasn't sure if its the right name and I first of all wanted to differentiate it from the other entries since it is the only one generated in the log function itself.

Thus could be changed to `"DATE"` or `"TIMESTAMP"`.

ejoerns:
Changes made:

* log files are now counted `log.1` to `log.<max-files>` instead of starting with `log.0`
* `max-files` now defaults to `10`
* non-existing log files will just be skipped during rotation (simplified code)
* `logger` was renamed to `loggers`
* added comment to location of testlogger in qemu test config

src/event-log.c
jluebbe: I would have included the currently active file in the count. So maxfiles would not be exceeded in total.

src/event-log.c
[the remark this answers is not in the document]
jluebbe: Is it useful to save some space in the JSON output? If so, perhaps `"TS"` and `"MSG(_ID)"`?

src/event-log.c
ejoerns: As an alternative: In `logrotate` the option is called `rotate` and a `rotate` of `0` means "old versions are removed rather than rotated". But `-1` there means "old logs are not removed at all" (the ability we just removed) so we probably shouldn't adapt to its naming anyway.

For `maxfiles` it does not feel that clear how it can be interpreted. But maybe counting this as the total number of files (as you suggested) is fine.

src/config_file.c
```diff
@@ -157,6 +158,141 @@ gboolean parse_bundle_formats(guint *mask, const gchar *config, GError **error)
...
+		}
+
+		logger->events = g_key_file_get_string_list(key_file, *group, "events", &entries, NULL);
+		for (gsize j = 0; j < entries; j++) {
+			if (!r_event_log_is_supported_type(logger->events[j])) {
+				g_set_error(
+						error,
+						G_KEY_FILE_ERROR,
+						G_KEY_FILE_ERROR_INVALID_VALUE,
+						"Unsupported event type '%s'", logger->events[j]);
```
jluebbe, line 259:
```suggestion
						"Unsupported event log type '%s'", logger->events[j]);
```

src/config_file.c
```diff
@@ -157,6 +158,141 @@ gboolean parse_bundle_formats(guint *mask, const gchar *config, GError **error)
...
+
+		logger->maxsize = key_file_consume_binary_suffixed_string(key_file, *group, "max-size", &ierror);
+		if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
+			g_clear_error(&ierror);
+		} else if (ierror) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+
+		logger->maxfiles = key_file_consume_integer(key_file, *group, "max-files", &ierror);
```
jluebbe, line 273: We should probably check that the provided value is in a sensible range (i.e. 1..1000).

src/context.c
```diff
@@ -325,6 +326,15 @@ static gboolean r_context_configure_target(GError **error)
 		}
 	}
 
+	/* set up logging */
+	if (context->config->loggers) {
+		for (GList *l = context->config->loggers; l != NULL; l = l->next) {
```
jluebbe, line 331: The if should not be needed, as the loop condition checks for NULL as well.

docs/reference.rst
```diff
@@ -697,6 +697,51 @@ No built-in slot update will run and no hook will be executed.
...
+  Using a relative file name without ``data-directory`` set will cause a
+  configuration error.
+
+``events``
+  Colon-separated list of events to log. Currently supported event types are:
+
+  * ``install`` - Logs start and end of installation
+  * ``boot`` - Logs boot information
+  * ``mark`` - Logs slot marking information
+  * ...
```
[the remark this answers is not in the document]
jluebbe: This should also mention `all`.

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+#include "context.h"
+#include "event-log.h"
+
+static const gchar *supported_event_types[] = {
+	"all",
+	R_EVENT_LOG_TYPE_BOOT,
+	R_EVENT_LOG_TYPE_INSTALL,
+	R_EVENT_LOG_TYPE_SERVICE,
+	R_EVENT_LOG_TYPE_WRITE_SLOT,
+	R_EVENT_LOG_TYPE_BOOT_SELECTION,
```
jluebbe, line 18: Perhaps some comments to clarify the respective expected content?

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+	g_autofree gchar *now_formatted = NULL;
+	g_autoptr(JsonGenerator) gen = NULL;
+	g_autoptr(JsonNode) root = NULL;
+	g_autoptr(JsonBuilder) builder = json_builder_new();
+
+	now = g_date_time_new_now_utc();
+	now_formatted = g_date_time_format(now, "%Y-%m-%dT%H:%M:%SZ");
+
+	json_builder_begin_object(builder);
+	json_builder_set_member_name(builder, "date");
```
jluebbe, line 103:
```suggestion
	json_builder_set_member_name(builder, "TS");
```

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+		const GLogField *fields, gsize n_fields, gboolean pretty)
+{
+	g_autoptr(GDateTime) now = NULL;
+	g_autofree gchar *now_formatted = NULL;
+	g_autoptr(JsonGenerator) gen = NULL;
+	g_autoptr(JsonNode) root = NULL;
+	g_autoptr(JsonBuilder) builder = json_builder_new();
+
+	now = g_date_time_new_now_utc();
+	now_formatted = g_date_time_format(now, "%Y-%m-%dT%H:%M:%SZ");
```
jluebbe, line 100: Perhaps use `g_date_time_format_iso8601()`, as this also provides sub-second precision? That could be useful to correlate events if as lot is happening.

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+		g_autofree gchar *to_file = g_strdup_printf("%s.%d", logger->filename, file_num);
+
+		/* skip non-existing files */
+		if (!g_file_test(from_file, G_FILE_TEST_EXISTS))
+			continue;
+
+		if (g_rename(from_file, to_file) == -1) {
+			int err = errno;
+			g_warning("Failed to rotate log file: %s", g_strerror(err));
+		}
```
jluebbe, line 241: Avoid the initial check and just handle `errno==ENOENT`.

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+	/* method is no-op if no maxsize is configured */
+	if (!logger->maxsize)
+		return;
+
+	/* return if size limit not reached yet */
+	if (logger->filesize + next_len <= logger->maxsize)
+		return;
+
+	/* Close open stream */
+	g_clear_pointer(&logger->logstream, g_object_unref);
```
jluebbe, line 226: Should we flush before closing? Or even `fsync()`?

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+		}
+		continue;
+	}
+
+	/* rotate current log file as next .1 file */
+	rotfile = g_strdup_printf("%s.1", logger->filename);
+	if (g_rename(logger->filename, rotfile) == -1) {
+		int err = errno;
+		g_warning("Failed to rotate log file: %s", g_strerror(err));
+		return;
```
jluebbe, line 250: In this case, we no longer have a log file open, right? So writing the next log event would fail?

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+		default:
+			g_error("Unknown log format");
+	}
+	output = g_strdup_printf("%s\n", formatted);
+
+	/* Once we know how much to write, we can use this information for trimming */
+	rotate_logfiles(logger, strlen(output));
+
+	if (!g_output_stream_write_all(G_OUTPUT_STREAM(logger->logstream), output, strlen(output), &written, NULL, &ierror)) {
+		g_warning("Failed to write log file: %s", ierror->message);
```
jluebbe, line 312: Hmm, this will keep generating `g_warnings` for any further message. Perhaps we should have a `broken` flag per logger and drop any messages for broken loggers?

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+			break;
+		}
+	}
+
+	/* iterate over registered event loggers */
+	for (GList *l = r_context()->config->loggers; l != NULL; l = l->next) {
+		REventLogger* logger = l->data;
+
+		/* Filter out by event type */
+		if ((g_strcmp0(logger->events[0], "all") != 0) &&
```
jluebbe, line 353: Do we ensure that `all` is always the first type in this array?

src/event-log.c
```diff
@@ -0,0 +1,398 @@
...
+
+	/* query file information or create new ones */
+	if (!g_file_test(logger->filename, G_FILE_TEST_EXISTS)) {
+		logger->filesize = 0;
+	} else {
+		g_autoptr(GFile) logfile = NULL;
+		g_autoptr(GFileInfo) logfile_info = NULL;
+
+		logfile = g_file_new_for_path(logger->filename);
+		logfile_info = g_file_query_info(logfile, G_FILE_ATTRIBUTE_STANDARD_SIZE, G_FILE_QUERY_INFO_NONE, NULL, &ierror);
```
jluebbe, line 387: Couldn't this be done using `g_seekable_tell()` after opening the file?

jluebbe (changes requested):
Still missing is a test case for the `system.conf` parsing.

docs/reference.rst
```diff
@@ -697,6 +697,51 @@ No built-in slot update will run and no hook will be executed.
...
+  the size configured with ``max-size`` and renamed to ``<filename>.0``.
+  Existing rotation file names will be incremented by one.
+  To configure a maximum number of files to keep, see ``max-files``.
+  Values support common suffixes like ``K``, ``M``, ``G``, to ``T``.
+
+``max-files``
+  Configures the maximum number of rotation files to keep.
+  E.g. if set to ``3``, only ``<filename>.0``, ``<filename>.1`` and
+  ``<filename>.2`` will be kept during rotation.
+  Defaults to 10 if unset.
```
jluebbe, line 743: This is no longer correct.

docs/reference.rst
```diff
@@ -697,6 +697,51 @@ No built-in slot update will run and no hook will be executed.
   ``<key>``.
   ``-`` is converted to ``_`` for use as an environment variable name.
 
+.. _log.logger-sections:
+
+**[log.<logger>] sections**
+
+With a logger section, a *RAUC event logger* can be configured.
```
jluebbe, line 704: We should point to the section in `advanced.rst` here for the general behavior.

docs/advanced.rst
```diff
@@ -1594,3 +1594,56 @@ https://github.com/caruhome/upparat
...
+
+Or, if you want a json-based log of all events, limited to 1M per log file and
+5 rotation files to keep, use::
+
+  [log.all-json-log]
+  filename=all-json.log
+  output-format=json
+  max-size=1M
+  max-files=5
+
```
jluebbe, line 1643: We should specify what happens if rotation or logging fails.

docs/reference.rst
ejoerns: The mentioning of `.0` is invalid, yes. The default of 10 however was just added and should be valid.

docs/reference.rst
ejoerns: Indeed, this did not exist yet when I started here, will fix.

docs/advanced.rst
ejoerns: Yes. I was not quite sure about this, yet. But maybe marking the logger broken as suggested above could be a way to handle this.

src/event-log.c
ejoerns: Two possible solutions I see here:

a) ignore that we failed to rename and continue with re-opening (the probably still existing) log file

b) abort and disable logger

the disadvantage of a) is that it will probably fail on each further iteration and exceeds `max-size`. The disadvantage of b) is that we do not log anymore despite we probably could.

docs/advanced.rst
```diff
@@ -1594,3 +1594,56 @@ https://github.com/caruhome/upparat
...
+For a full reference of supported configuration options, see
+:ref:`log.logger-sections`.
+
+.. note:: All events logged using the internal event logging framework will
+   also be available on the journal.
+   Some of them also come with journal catalog entries to provide more details.
```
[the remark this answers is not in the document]
ejoerns: Yes, but I am not sure if this note should make it into this PR since adding the catalog is a follow-up for now.

src/event-log.c
ejoerns: We now ensure that `"all"` cannot be combined with other values (and will be set as default if no value is provided).

src/event-log.c
ejoerns: Added `broken` flag now (+test).

ejoerns:
Changes made:

* Add some internal module documentation for event-log
* Fix the above-noted documentation issues
* mention ``all`` as valid event type in doc
* replaced `"bootsel"` event type by `"mark"`
* added a couple of config parsing tests
* made `filename` option mandatory and fixed error handling
* let `logger->events` default to `{"all", NULL}` if no events given and error if `"all"` is combined with others in config
* fix freeing of `logger->events`
* added GError handling to `rotate_logfiles()`
* first of all do not abort on failing `g_rename()`
* added `logger->broken` `gboolean` flag and support deactivating logger on log file opening failure

ejoerns:
Changes made:

* fix freeing of `config->loggers`
* put chunk introducing `config->loggers` into appropriate commit

src/event-log.c
ejoerns: From the Glib `GOutputStream` documentation:

> Closing the stream will implicitly cause a flush.

Since a `GFileOutputStream` is also a `GOutputStream`, this will apply here, too.

However, calling `fsync()` should be possible when assuming having a `GUnixOutputStream` as we do for the update handlers already.

src/event-log.c
ejoerns: Already opening the logfile in the setup probably makes sense anyway. Since `GFileOutputStream` implements `GSeekable` this should work, yes.

src/event-log.c
ejoerns: Changed the code now to interpret `max-files` as the maximum of all files (including the logfile itself).

include/event-log.h
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * @param config Logger to free
+ */
+void r_event_log_free_logger(REventLogger *logger);
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(REventLogger, r_event_log_free_logger);
+
+/**
+ * Simple Logging convenience method.
```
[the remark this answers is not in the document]
ejoerns: More or less for reference. I am still unsure if it should be used at all.

src/event-log.c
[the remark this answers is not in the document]
ejoerns: I used the `journalctl -o json` as a reference which also just plainly prints out the internal names of structured logging.

ejoerns:
Changes made:

* added cross-link from `reference.rst` to `advanced.rst` chapter
* removed note of *catalog* from docs first of all (and made description more precise)
* Changed behavior of `max-files` to include the logfile itself in the count. Adapted tests.
* Added error handling, sanity checks and tests for ensuring `max-files > 0`
* changed `"date"` to `"TIMESTAMP"` in json output
* fixed broken logger handling
* open logger already during `r_event_log_setup_logger()` and use `g_seekable_tell()` for determining log file size

src/event-log.c
```diff
@@ -0,0 +1,406 @@
...
+	g_assert(logger->maxfiles >= 1);
+	for (guint file_num = logger->maxfiles - 1; file_num > 0; file_num--) {
+		g_autofree gchar *from_file = g_strdup_printf("%s.%d", logger->filename, file_num - 1);
+		g_autofree gchar *to_file = g_strdup_printf("%s.%d", logger->filename, file_num);
+
+		/* skip non-existing files */
+		if (!g_file_test(from_file, G_FILE_TEST_EXISTS))
+			continue;
+
+		if (g_rename(from_file, to_file) == -1) {
```
jluebbe, line 244: Add a comment that this will replace the .<maxfiles-1> file.

test/event-log.c
```diff
@@ -0,0 +1,299 @@
...
+		{"RAUC_EVENT_TYPE", "mark", -1},
+		{"RAUC_SLOT", "rootfs.0", -1},
+		{"BUNDLE_HASH", "b970468f-89e4-4793-9904-06c922902b25", -1},
+		{"SLOT_BOOTNAME", "A", -1},
+	};
+	g_autofree gchar *contents = NULL;
+
+	logger = g_new0(REventLogger, 1);
+	logger->name = g_strdup("testlogger");
+	logger->filename = g_strdup("/this/path/does/not/exist/testfile.log");
```
jluebbe, line 98: Another case would be a full device. You can use `/dev/full` for that:
```
# echo foo > /dev/full 
-bash: echo: write error: No space left on device
```

docs/advanced.rst
jluebbe:
```suggestion
If an error occurs during logging (such as disk full or write errors), that logger is marked as broken and no longer used.
An ongoing installation is **not** aborted.
```

docs/reference.rst
```diff
@@ -697,6 +697,54 @@ No built-in slot update will run and no hook will be executed.
...
+The ``<logger>`` suffix determines the internal name of the logger and must be
+unique per ``system.conf``.
+
+For an overview over the event logging framework in RAUC and its purpose, have
+a look at :ref:`sec-advanced-event-log`.
+
+``filename``
+  The log file name used for logging.
+  If not absolute path is given, the location is assumed to be relative to the
```
jluebbe, line 713:
```suggestion
  If no absolute path is given, the location is assumed to be relative to the
```

docs/reference.rst
```diff
@@ -697,6 +697,54 @@ No built-in slot update will run and no hook will be executed.
...
+  ``data-directory``.
+  Using a relative file name without ``data-directory`` set will cause a
+  configuration error.
+
+``events``
+  Colon-separated list of events to log. Currently supported event types are:
```
jluebbe, line 719:
```suggestion
  Semicolon-separated list of events to log. Currently supported event types are:
```

docs/reference.rst
```diff
@@ -697,6 +697,54 @@ No built-in slot update will run and no hook will be executed.
...
+
+  * ``readable``: readable mutli-line output
+  * ``short``: Single-line readable output
+  * ``json``: single-line JSON output
+  * ``json-pretty``: formatted JSON output
+
+``max-size``
+  Allows to configure a basic log rotation.
+  When given, the logger's log file will be rotated before reaching
+  the size configured with ``max-size`` and renamed to ``<filename>.0``.
```
jluebbe, line 737:
```suggestion
  the size configured with ``max-size`` and renamed to ``<filename>.1``.
```

docs/reference.rst
```diff
@@ -697,6 +697,54 @@ No built-in slot update will run and no hook will be executed.
...
+  Existing rotation file names will be incremented by one.
```
jluebbe, line 738:
```suggestion
  Existing rotation file names will be incremented by one.
  The oldest file is removed.
```

src/config_file.c
```diff
@@ -157,6 +158,174 @@ gboolean parse_bundle_formats(guint *mask, const gchar *config, GError **error)
...
+		} else if (ierror) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+		if (tmp_maxfiles < 1) {
+			g_set_error_literal(
+					error,
+					G_KEY_FILE_ERROR,
+					G_KEY_FILE_ERROR_INVALID_VALUE,
+					"Value for 'max-files' muste be >= 1");
```
jluebbe, line 310:
```suggestion
					"Value for 'max-files' must be >= 1");
```

jluebbe (changes requested):
Some comments from the previous review are still open.

ejoerns:
Changes made:

* Added and fixed documentation pieces as requested
* Fixed typos
* Added all so-far-known fields to formatter
* Made code comments clearer where requested
* Added test for setting up logger on full device
* Added tests for actually testing the code through structured logging (includes actually testing filtering)

test/event-log.c
ejoerns: Added test case "/event-log/log-writer/no-space-left" for this.

ejoerns:
Changes made:

* Added two missing doc suggestions
* Simply error-out on failed log rotation (as requested)

include/event-log.h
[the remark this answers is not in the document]
ejoerns: It's not used by the added structured logging tests.

src/event-log.c
ejoerns: Done so.

include/event-log.h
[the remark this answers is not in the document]
jluebbe: So, should it be removed or remain untested?

src/config_file.c
```diff
@@ -157,6 +158,141 @@ gboolean parse_bundle_formats(guint *mask, const gchar *config, GError **error)
...
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	g_assert_null(config->loggers);
+
+	/* parse [log.*] sections */
+	groups = g_key_file_get_groups(key_file, &group_count);
+	for (gchar **group = groups; *group != NULL; group++) {
+		GError *ierror = NULL;
+		g_autoptr(REventLogger) logger = NULL;
+		gchar *logger_name;
```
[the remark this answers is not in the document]
jluebbe: Still applies to current version.

src/config_file.c
jluebbe: This still applies to the current version.

src/config_file.c
jluebbe: This still applies to the current version.

src/event-log.c
jluebbe: These are `#define`d in `include/event-log.h`, so the comments should go there. Still applies to the current version.

src/event-log.c
jluebbe: This now uses `TIMESTAMP`, but not `TS`. My intention there was to *reduce* the verbosity and size of the log keys a bit.

src/event-log.c
jluebbe: This still applies to the current version.

include/event-log.h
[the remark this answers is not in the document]
ejoerns: My fault. The above command should have been "It's **now** used by the added structured logging tests."

src/event-log.c
ejoerns: I chose the format as this is consistent with the existing ones.

Another thing is that `g_date_time_format_iso8601()` requires glib 2.62 as a minimal version.

If we agree that it is worth the effort, we can either bump glib minimal version or conditionally import the method.

2.62 is available since Yocto/poky `dunfell`, Ubuntu 20.04, Debian oldstable (`bullseye`)

ejoerns:
* Added comments to log type `#define`s
* Added comments to `REventLogFormat` enum
* Made `logger_name` const pointer
* Added check for `max-files` being <= 1000 in config parsing
* Remove not needed `if` guard when iterating over loggers in `context.c`
* Renamed `"TIMESTAMP"` to shorter `"TS"` in json log output
* Combined file existence check with `g_rename` handling by ignoring `err == ENOENT`

src/event-log.c
ejoerns: Without actually using it, the scope is probably still a bit vague but tried to be a bit clearer.

Might need some adaptions and extensions when adding actual log message printouts later anyway.

src/config_file.c
ejoerns: We've already ensured "> 1" and I've now added a check for "< 1000"

src/event-log.c
ejoerns: I tried to stay close to what systemd journal json output would provide you, but since this uses far longer names anyway (like `_SOURCE_MONOTONIC_TIMESTAMP`, it might not be relevant here, anyway).

ejoerns:
Rebased onto current `master`.

src/event-log.c
jluebbe: As the current format is also ISO8601 (just without microsecond precision), we can use this for now and move to `g_date_time_format_iso8601()` later.

src/event-log.c
```diff
@@ -0,0 +1,405 @@
...
+
+	/* rotate current log file as next .1 file */
+	rotfile = g_strdup_printf("%s.1", logger->filename);
+	if (g_rename(logger->filename, rotfile) == -1) {
+		int err = errno;
+		g_set_error(error, G_FILE_ERROR, g_file_error_from_errno(err), "Failed to rotate log file: %s", g_strerror(err));
+		return FALSE;
+	} else {
+		logger->filesize = 0;
+	}
```
jluebbe, line 271: It doesn't really make sense to have an `else` here after the error return above.

The `filesize` reinit belongs in the re-open part below.

src/event-log.c
```diff
@@ -9,6 +9,20 @@
 #include "context.h"
 #include "event-log.h"
 
+static const gchar *supported_event_types[] = {
+	"all",
+	R_EVENT_LOG_TYPE_BOOT,
+	R_EVENT_LOG_TYPE_MARK,
+	R_EVENT_LOG_TYPE_INSTALL,
+	R_EVENT_LOG_TYPE_SERVICE,
+	R_EVENT_LOG_TYPE_WRITE_SLOT,
```
jluebbe, line 18: This needs to be `NULL` terminated for use with `g_strv_contains`.

jluebbe (changes requested):
We currently don't use dashes (but underscores) in filenames. Perhaps rename `event-log.[ch]`? Looks good otherwise.

include/event-log.h
```diff
@@ -0,0 +1,123 @@
...
+ * * "MESSAGE" : The log message
+ * * "MESSAGE_ID" : The ID of the message. Must be unique per message type.
+ *
+ * Optional fields for providing context information are:
+ *
+ * * "BOOT_ID" : The boot ID associated with the log message
+ * * "BUNDLE_HASH" : The bundle hash associated with the log message
+ * * "BUNDLE_VERSION" : The bundle version associated with the log message
+ * * "TRANSACTION_ID" : The transaction ID associated with the log message
+ * * "RAUC_SLOT" : The slot name associated with the log message
```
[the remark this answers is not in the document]
ejoerns: A bit historically because at first, I was unsure how the fields should be named.
Probably `SLOT_NAME` fits better here.

src/event-log.c
ejoerns: Yes, absolutely makes sense.

ejoerns:
* Renamed  `event-log.[ch]` to `event_log.[ch]`
* Renamed `RAUC_SLOT` to `SLOT_NAME`
* Moved resetting of `logger->filesize` to log file re-opening
* Added `NULL` sentinel to `supported_event_types`

jluebbe (approved):

jluebbe:
The cross build failure is unrelated. It seems the multiarch containers are outdated.

PR #1231, opened 2023-08-17
test/Dockerfile: pin desync to version 0.9.3

jluebbe:
Later versions fail to build with:
```
/go/pkg/mod/golang.org/x/crypto@v0.8.0/curve25519/curve25519_go120.go:9:8: package crypto/ecdh is not in GOROOT (/usr/lib/go-1.15/src/crypto/ecdh)
/go/pkg/mod/golang.org/x/net@v0.9.0/publicsuffix/table.go:5:8: package embed is not in GOROOT (/usr/lib/go-1.15/src/embed)
/go/pkg/mod/golang.org/x/net@v0.9.0/http2/transport.go:19:2: package io/fs is not in GOROOT (/usr/lib/go-1.15/src/io/fs)
```

### DIFF
--- a/test/Dockerfile
+++ b/test/Dockerfile
@@ -51,10 +51,11 @@ RUN apt-get update && apt-get install -y \
   curl -sLo /usr/bin/codecov https://codecov.io/bash && \
   chmod +x /usr/bin/codecov
 
-# Install the optional desync
+# Install the optional desync (pinned to version 0.9.3)
 ENV GOPATH=/go
 RUN git clone https://github.com/folbricht/desync.git /tmp/desync && \
     cd /tmp/desync/cmd/desync && \
+    git checkout c508eeb0865a5a7c2c9b1158a5f0414265d869df && \
     go install && \
     cp /go/bin/desync /usr/bin/desync && \
     rm -rf /tmp/desync


jluebbe:
@RyuzakiKK I had do pin the desync version to avoid the avoid build error. Do you have an idea how to fix this properly?

jluebbe:
@ejoerns A successful run can be seen at https://github.com/jluebbe/rauc/actions/runs/5891909531/job/15980034107

RyuzakiKK:
The minimum Go version declared by Desync is 1.19.

Debian Bullseye (oldstable) has Go 1.15, while the 1.19 is in both bullseye-backports and Bookworm (stable).
I just tested this in Docker and I confirm that if you install golang 1.19 from bullseye-backports this issue is fixed.

So I guess that the proper solution is to either bring golang from bullseye-backports or bump the VM to the latest Debian stable image instead of relying on Bullseye.

jluebbe:
> The minimum Go version declared by Desync is 1.19.
> 
> Debian Bullseye (oldstable) has Go 1.15, while the 1.19 is in both bullseye-backports and Bookworm (stable). I just tested this in Docker and I confirm that if you install golang 1.19 from bullseye-backports this issue is fixed.
> 
> So I guess that the proper solution is to either bring golang from bullseye-backports or bump the VM to the latest Debian stable image instead of relying on Bullseye.

Ah, thanks for the details. I'd prefer to update to bookworm, rather than keep bullseye for longer than needed. We'll need to check for other regressions, though (i.e. due to deprecation warnings and -Werror).

PR #1232, opened 2023-08-21
Expose transaction ID to pre/post install handlers

zyga:
I've implemented a RAUC + HawkBit stack that informs HawkBit of a "commit" or "rollback" of a particular update. I'm using a custom HawkBit client (not rauc-hawkbit-updater) that contains some other business logic that is not generic. The HawkBit updater and a set of RAUC handlers cooperate to understand when a system image is committed or rolled back.

I've since updated to v1.10.1 and noticed that the new RAUC transaction ID is the perfect way to identify a particular installation in HawkBit. I'm attaching a small set of patches that expose transaction ID to pre-install and post-install handlers as `RAUC_TRANSACTION_ID`.

The way this works for me is that the transaction ID is given to RAUC by the custom HawkBit client and also simultaneously sent back to HawkBit as an update message. Once commit or rollback decision is made I can cross check this with state maintained by custom handlers.

I think that the transaction ID being visible to handlers is a good idea in general, so I'm sharing my three patches.

Word of caution: I did not manage to run the test suite with qemu locally while I was traveling. I will update this PR if the test suite needs any adjustments having tried it on my main system.


ejoerns (approved):
Added just two minor doc changes.

Looks fine to me. :+1:

jluebbe (approved):

PR #1233, opened 2023-08-21
Add libFuzzer setup

securitykernel:
As promised in #1119, this PR adds a basic setup of libFuzzer in the build system and two initial fuzz targets.

### Setup

Basically adds a meson build option `fuzzing` that enables building the fuzz targets in `fuzz/`. Use like:

```
export CC=clang
meson setup build -Dfuzzing=true
ninja -C build
```

It is currently hardcoded to use in combination with address sanitizer (`-fsanitize=fuzzer,address`), but libFuzzer can be combined with other sanitizers as well. How to integrate this into the build system is left as an exercise.

### Fuzz targets

Adds two initial fuzz targets:

#### bundle_fuzzer

A fuzz target that tries to load a bundle using `check_bundle()`. Since this is function is parsing untrusted data, it makes a good first fuzz target.

Run the target using the following command line:

```bash
mkdir corpus-bundle
cp test/*.raucb corpus-bundle
build/fuzz/bundle_fuzzer corpus-bundle/
```

It uses existing bundles shipped with the test suite as good/bad examples for seeding the corpus.

#### manifest_fuzzer

A fuzz target that tries to load a manifest using `load_manifest_mem()`.

Run the target using the following command line:

```bash
mkdir corpus-manifest
cp test/manifest.raucm corpus-manifest
build/fuzz/manifest_fuzzer -use_value_profile=1 -only_ascii=1 -dict=fuzz/manifest.dict corpus-manifest
```

It uses a dictionary containing keywords contained in RAUC manifests created from the docs as well as an example manifest from the test suite for seeding the corpus.

The fuzz targets can probably be improved greatly with more knowledge of the parser code.


jluebbe:
When compiling this, I get:
```
[10/149] Compiling C object fuzz/manifest_fuzzer.p/manifest.c.o
../fuzz/manifest.c:10:5: warning: no previous prototype for function 'LLVMFuzzerTestOneInput' [-Wmissing-prototypes]
int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
    ^
../fuzz/manifest.c:10:1: note: declare 'static' if the function is not intended to be used outside of this translation unit
int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
^
static 
```

securitykernel:
> When compiling this, I get:
> 
> ```
> [10/149] Compiling C object fuzz/manifest_fuzzer.p/manifest.c.o
> ../fuzz/manifest.c:10:5: warning: no previous prototype for function 'LLVMFuzzerTestOneInput' [-Wmissing-prototypes]
> int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
>     ^
> ../fuzz/manifest.c:10:1: note: declare 'static' if the function is not intended to be used outside of this translation unit
> int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
> ^
> static 
> ```

Yeah I saw this. No way to fix this other than disabling `Wmissing-prototypes` for the fuzz targets. This function needs to be defined as-is, as it is called by libFuzzer.

fuzz/meson.build
```diff
@@ -7,12 +7,14 @@ fuzzers = [
   'bundle',
 ]
 
+extra_c_args = cc.get_supported_arguments('-Wno-missing-prototypes')
```
ejoerns, line 10: This name is quite generic and could be set by another meson.build file, too (like `extra_test_sources` already is) Maybe use something like `fuzz_c_args=`?

ejoerns (approved):
Looks fine to me. Should be another good first step towards further improving/maintaining the stability and security of RAUC. :+1:

ejoerns (approved):

jluebbe:
With the manifest fuzz target, I found an issue in glib: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3555

securitykernel:
> With the manifest fuzz target, I found an issue in glib: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3555

I remember seeing this warning too when setting this up. I looked through rauc's code, because the warning message from glib suggests it to be an application issue, but obviously found none. I did not think of an issue in glib itself. Nice finding!

PR #1240, opened 2023-08-30
add meson wrap files for glib, json-glib and openssl

jluebbe:
This allows us to build against a local copies of these dependencies (with --wrap-mode=forcefallback) and also supports static compilation for use in oss-fuzz.

These files were installed from meson's wrapdb.

### DIFF
--- a/subprojects/glib.wrap
+++ b/subprojects/glib.wrap
@@ -1,0 +1,11 @@
+[wrap-file]
+directory = glib-2.76.4
+source_url = https://download.gnome.org/sources/glib/2.76/glib-2.76.4.tar.xz
+source_fallback_url = https://github.com/mesonbuild/wrapdb/releases/download/glib_2.76.4-1/glib-2.76.4.tar.xz
+source_filename = glib-2.76.4.tar.xz
+source_hash = 5a5a191c96836e166a7771f7ea6ca2b0069c603c7da3cba1cd38d1694a395dda
+wrapdb_version = 2.76.4-1
+
+[provide]
+dependency_names = gthread-2.0, gobject-2.0, gmodule-no-export-2.0, gmodule-export-2.0, gmodule-2.0, glib-2.0, gio-2.0, gio-windows-2.0, gio-unix-2.0
+program_names = glib-genmarshal, glib-mkenums, glib-compile-schemas, glib-compile-resources, gio-querymodules, gdbus-codegen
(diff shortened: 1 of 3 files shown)


ejoerns:
When not having these libraries available on a system, meson will automatically use the libraries from the wrap file. Is this our intention?

I also got

> WARNING: Could not detect glib version, assuming 2.54. You may get build errors if your glib is older.

which I am not sure if this is problematic

jluebbe:
> When not having these libraries available on a system, meson will automatically use the libraries from the wrap file. Is this our intention?

Not in my tests, as fallback is not configured for the dependencies.

> I also got
> 
> > WARNING: Could not detect glib version, assuming 2.54. You may get build errors if your glib is older.
> 
> which I am not sure if this is problematic

As the glib wrap is newer, this shouldn't be an issue.

jluebbe:
> > When not having these libraries available on a system, meson will automatically use the libraries from the wrap file. Is this our intention?
> 
> Not in my tests, as fallback is not configured for the dependencies.

Ah, it seems I was mistaken, meson defaults to downloading.

For example glib will download pcre2 if it's not found, so we should probably just stick to the defaults. Build systems can override this (i.e. Yocto's `meson.bbclass` uses `--wrap-mode nodownload`).

ejoerns:
> > When not having these libraries available on a system, meson will automatically use the libraries from the wrap file. Is this our intention?
> 
> Not in my tests, as fallback is not configured for the dependencies.

Re-checked:

```
$ sudo aptitude remove libglib2.0-dev
$ rm -rf build
$ meson setup build

% meson setup build
The Meson build system
Version: 1.0.1
Source dir: /home/enrico/Code/rauc
Build dir: /home/enrico/Code/rauc/build
Build type: native build
Project name: rauc
Project version: 1.10.1
C compiler for the host machine: cc (gcc 12.2.0 "cc (Debian 12.2.0-14) 12.2.0")
C linker for the host machine: cc ld.bfd 2.40
Host machine cpu family: x86_64
Host machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (1.8.1)
Found CMake: /usr/bin/cmake (3.25.1)
Run-time dependency glib-2.0 found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency glib-2.0

Executing subproject glib 

glib| Project name: glib
glib| Project version: 2.76.4
```

ejoerns:
> > > When not having these libraries available on a system, meson will automatically use the libraries from the wrap file. Is this our intention?
> > 
> > 
> > Not in my tests, as fallback is not configured for the dependencies.
> 
> Ah, it seems I was mistaken, meson defaults to downloading.
> 
> For example glib will download pcre2 if it's not found, so we should probably just stick to the defaults. Build systems can override this (i.e. Yocto's `meson.bbclass` uses `--wrap-mode nodownload`).

Ok, should be fine first of all. Might also make it easier for some people to build the project.

ejoerns (approved):

PR #1241, opened 2023-08-30
src: detect config file change and warn in service log

ejoerns:
A common usage issue with RAUC is to change the system configuration but not to restart the rauc service to re-read the configuration.

To make this more obvious, at least in the service log output, this implements a simple checksum-based checking for changes in the system.conf.

On config file loading, the checksum of the original parsed configuration is cached and used as reference for later comparisons of re-calculated checksums.

The checker function r_config_file_sanity_check() that implements the comparison is used first of all in the service on calls of 'install' or 'status' actions and prints a warning.

### DIFF
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -88,6 +88,7 @@ typedef struct {
 	GHashTable *slots;
 	/* flag to ensure slot states were determined */
 	gboolean slot_states_determined;
+	gchar *file_checksum;
 } RaucConfig;
 
 typedef enum {
@@ -164,3 +165,11 @@ G_GNUC_WARN_UNUSED_RESULT;
 void free_config(RaucConfig *config);
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(RaucConfig, free_config);
+
+/**
+ * Checks if the current file checksum of the system config matches the one currently loaded.
+ *
+ * Prints a warning if the checksums of the config file does not match the one
+ * recorded during config parsing.
+ */
+void r_config_file_modified_check(void);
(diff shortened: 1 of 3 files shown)


include/config_file.h
```diff
@@ -164,3 +165,11 @@ G_GNUC_WARN_UNUSED_RESULT;
 void free_config(RaucConfig *config);
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(RaucConfig, free_config);
+
+/**
+ * Checks if the file checksum of 'filename' matches the one stored in config.
```
jluebbe, line 170:
```suggestion
 * Checks if the current file checksum of the system config matches the one currently loaded.
```

src/config_file.c
```diff
@@ -395,9 +395,36 @@ static GHashTable *parse_slots(const char *filename, const char *data_directory,
...
+	GError *ierror = NULL;
+	g_autofree gchar *data = NULL;
+	gsize length;
+	g_autofree gchar *new_checksum = NULL;
+
+	if (!r_context()->config->file_checksum)
+		return;
+
+	if (!g_file_get_contents(r_context()->configpath, &data, &length, &ierror)) {
+		g_warning("Failed to compare config");
```
jluebbe, line 408: Perhaps
```suggestion
		g_warning("Failed to compare config %s", ierror->message);
```

jluebbe (changes requested):
The commit message contains a different function name.

ejoerns:
Addressed changes requested and also fixed and simplified error handling.

ejoerns:
Rebased onto master to resolve conflict.

jluebbe (dismissed):

PR #1244, opened 2023-09-01
Boot nor fallback

### LINKED ISSUE #1235: boot fallback slot type for nor flash
As far as I can tell the `boot-raw-fallback` slot type does not work for `/dev/mtd*` partitions.

Implement `boot-nor-fallback` handler. The issue with that is that `flashcp` does not support write offsets as far as I can tell. The alternative would be to provide two mtd partitions set up in device tree as the primary and secondary targets. But for that the RaucSlot structure might need to be expanded to pass through those parameters.

I have not tried alternatives, as I need to be able to have an atomic update of a zynq boot.bin that has to be on the QSPI flash.

If there is a different way to implement atomic QSPI bootloader updates on zynq that would be great. Then this feature is not necessary.

esden:
This PR adds a special handler that allows bootloader fallback updates on `nor` (aka. QSPI) flash.

The first patch implements this feature using the current version of flashcp. It will work on systems that check the full integrity of the bootloader before trying to load it. It will not work right on systems that only check the bootloader header. This is where the second patch comes in. It uses a new option in flashcp that delays writing of a specified amount of bytes till the very end.

I have written and [submitted the necessary patch](http://lists.infradead.org/pipermail/linux-mtd/2023-August/100952.html) to mtd-utils project.

The new slot type is called `boot-nor-fallback` and we pass two mtd devices to the handler by separating them with a colon `:`. First device is the **primary** and the second device is the **secondary**. Using two mtd devices avoids us having to also add the ability to write at an offset to `flashcp`.

Resolves: #1235

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -905,8 +905,20 @@ static gboolean pre_install_check_slot_mount_status(RaucSlot *slot, RaucImage *m
 	return FALSE;
 }
 
+static gboolean check_device(gchar* device, GError **error)
+{
+	if (!g_file_test(device, G_FILE_TEST_EXISTS)) {
+		g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_NODST,
+				"Destination device '%s' not found", device);
+		return FALSE;
+	}
+	return TRUE;
+}
+
 static gboolean pre_install_checks(gchar* bundledir, GPtrArray *install_plans, GHashTable *target_group, GError **error)
 {
+	gchar** devices = NULL;
+
 	for (guint i = 0; i < install_plans->len; i++) {
 		const RImageInstallPlan *plan = g_ptr_array_index(install_plans, i);
 
@@ -934,10 +946,19 @@ static gboolean pre_install_checks(gchar* bundledir, GPtrArray *install_plans, G
 		}
 
 skip_filename_checks:
-		if (!g_file_test(plan->target_slot->device, G_FILE_TEST_EXISTS)) {
-			g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_NODST,
-					"Destination device '%s' not found", plan->target_slot->device);
-			return FALSE;
+		if (g_strcmp0(plan->target_slot->type, "boot-nor-fallback") == 0) {
+			devices = g_strsplit(plan->target_slot->device, ":", -1);
+			for (int i = 0; i < g_strv_length(devices); i++) {
+				if (!check_device(devices[i], error)) {
+					g_strfreev(devices);
+					return FALSE;
+				}
+			}
+			g_strfreev(devices);
+		} else {
+			if (!check_device(plan->target_slot->device, error)) {
+				return FALSE;
+			}
 		}
 
 		if (!pre_install_check_slot_mount_status(plan->target_slot, plan->image, error)) {
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -2527,6 +2527,102 @@ out:
 	return res;
 }
 
+static gboolean img_to_boot_nor_fallback_handler(RaucImage *image, RaucSlot *dest_slot, const gchar *hook_name, GError **error)
+{
+	GError *ierror = NULL;
+	gboolean res = FALSE;
+	gchar** devices = NULL;
+	gboolean primary_clear;
+	char *part_names[2] = {
+		"primary",
+		"fallback"
+	};
+	int first_part_desc_index = 1;
+
+	g_return_val_if_fail(image, FALSE);
+	g_return_val_if_fail(dest_slot, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	/* run slot pre install hook if enabled */
+	if (hook_name && image->hooks.pre_install) {
+		res = run_slot_hook(hook_name, R_SLOT_HOOK_PRE_INSTALL, image, dest_slot, &ierror);
+		if (!res) {
+			g_propagate_error(error, ierror);
+			goto out;
+		}
+	}
+
+	devices = g_strsplit(dest_slot->device, ":", -1);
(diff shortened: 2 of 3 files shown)


jluebbe:
Is this nor-specific or could it also work on other flashes (like NAND)?

I'm not sure yet if overloading the device parameter with ':' is a good idea. Everywhere else in RAUC it's just a single path. An alternative would be to allow an additional option to specify the secondary MTD.

esden:
I think it should be able to use all the flashes supported by the mtd subsystem. I do think that `flashcp` can write to all of them including NAND. I just have QSPI to test with though.

Yeah having an extra `device_fallback` config variable would be a good solution. I just wanted to keep the patch rather less than more invasive. And adding a config variable means adding things to the config struct and I did not want to touch it. :)

If you want me to change to the use of a separate variable I am happy to do it.

esden:
As a heads up. The patch for `flashcp` that this feature requires is now merged up stream: http://lists.infradead.org/pipermail/linux-mtd/2023-September/101077.html

src/install.c
```diff
@@ -934,10 +946,19 @@ static gboolean pre_install_checks(gchar* bundledir, GPtrArray *install_plans, G
 		}
 
 skip_filename_checks:
-		if (!g_file_test(plan->target_slot->device, G_FILE_TEST_EXISTS)) {
-			g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_NODST,
-					"Destination device '%s' not found", plan->target_slot->device);
-			return FALSE;
+		if (g_strcmp0(plan->target_slot->type, "boot-nor-fallback") == 0) {
+			devices = g_strsplit(plan->target_slot->device, ":", -1);
```
jluebbe, line 950: This `g_strsplit` should not be at multiple places. Instead, the handling for multiple devices should be explicit in the `RaucSlot` struct and should be split during parsing in `config_file.c`. For consistency, it should use `g_key_file_get_string_list`, which splits at `;`. If only one device is given, it should behave as before, storing it in the `gchar *device`. Otherwise, a new `gchar **devices` would be used and `device` would be `NULL`.

This way, it's explicit if one or multiple devices are configured. Then, we'd need a consistency check at the end of `load_config()` which ensures that only `boot-nor-fallback` passes two devices. All others would need only one.

src/update_handler.c
```diff
@@ -2527,6 +2532,102 @@ static gboolean check_if_area_is_clear(const gchar *device, guint64 start, gsize
...
+	GError *ierror = NULL;
+	gboolean res = FALSE;
+	gchar** devices = NULL;
+	gboolean primary_clear;
+	char *part_names[2] = {
+		"primary",
+		"fallback"
+	};
+	int first_part_desc_index = 1;
+	const gsize header_size = 512; /* TODO: this value should be settable through the system config file. */
```
jluebbe, line 2546: Yes, this needs to be configurable.

My suggestion would be a `header-size` field in the system config and a `guint64 header_size` in `RaucSlot`, similar to how region_size works.

src/update_handler.c
```diff
@@ -2527,6 +2532,102 @@ static gboolean check_if_area_is_clear(const gchar *device, guint64 start, gsize
...
+	g_message("Executing nor fallback handler on slot %s with primary device %s and fallback device %s",
+			dest_slot->name, devices[0], devices[1]);
+
+	/* If the primary partition is not fully programmed, it most likely means that the fallback
+	 * partition was used to boot and is therefore valid. To avoid ending up with two broken partitions,
+	 * upgrade the primary partition first.
+	 *
+	 * We are only checking the header area, as it is the last thing written.
+	 */
+	res = check_if_area_is_clear(devices[0], 0, header_size, &primary_clear, &ierror);
```
jluebbe, line 2577: Compared to block devices, NOR flash doesn't have atomic writes for 512 byte blocks. That results in some additional cases that need to be handled.

- Power loss while writing the primary header (after successfully writing the secondary image)
- Primary header is incomplete, but not clear
- Secondary image is used for booting
- Second update attempt erases secondary image, as primary header is not clear ⇒ temporarily no valid image

This means we'd need a better check to detect if the header is incomplete/corrupt. For simple cases, it might be enough if the last byte of the header region is != 0.

- Power loss while writing the primary header (after successfully writing the secondary image)
- Primary header is incomplete, but seems valid to the ROM code
- Boot fails

This depends on the SoC: For headers with a proper checksum, this shouldn't be a problem. If not it probably needs SoC specific research on how it actually behaves for partially written headers.


An alternative solution would be to keep some state in the RAUC data partition, to be able to detect the different cases.

If these corner-cases are acceptable in your use case, I'd also be OK with only the current approach of checking if the header is clear. In that case, there needs to be a big warning in the documentation, though.

jluebbe (changes requested):
This also needs documentation:

- A new section in https://rauc.readthedocs.io/en/latest/advanced.html#updating-the-bootloader
- Entries for the `header-size` key in `[slot.<slot-class>.<idx>] section` in https://rauc.readthedocs.io/en/latest/reference.html#system-configuration-file

jluebbe:
@esden What's your current status for this PR? We don't mind keeping it open for as long as you want to work on it.

daterdots:
@jluebbe - @esden and I are talking about this. We hope to get it into good mergable shape soon, but we need at least until February 2024 if that's OK with you. Other stuff is slamming us.

jluebbe:
> @jluebbe - @esden and I are talking about this. We hope to get it into good mergable shape soon, but we need at least until February 2024 if that's OK with you. Other stuff is slamming us.

What's the current state of this on your side?

esden:
hi @jluebbe, sorry for the delay. We are hoping to get some support to work on this in the next couple of months. We will then implement the changes that were suggested and improve the PR.

jluebbe:
Thanks for the update!

PR #1246, opened 2023-09-06
test: add tar archives for artifact testing

jluebbe:
These will become useful for testing conversion from `.tar` archives to directory trees in the bundle.


ejoerns (approved):
Looks fine to me :+1:

PR #1248, opened 2023-09-13
README: add OSS-Fuzz badge

jluebbe:


jluebbe:
OSS-Fuzz hasn't built RAUC yet, so we need to wait a bit.

Emantor (approved):

Emantor (approved):

PR #1258, opened 2023-10-10
Fix installation of rauc.1 manpage

ukleinek:
`install_data()` as currently used puts the manpage to `/usr/share/man/rauc.1`, however the right place is `/usr/share/man/man1/rauc.1`. Use the dedicated meson macro `install_man()` to get this right.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -227,7 +227,7 @@ dbuswrapperdir = get_option('libexecdir')
 
 install_data('src/de.pengutronix.rauc.Installer.xml', install_dir : dbusinterfacesdir)
 
-install_data('rauc.1', install_dir : get_option('mandir'))
+install_man('rauc.1')
 
 subdir('data')
 subdir('docs')


ejoerns (approved):

ukleinek:
the arm64 cross test failed with:

    Err:1 http://deb.debian.org/debian bullseye InRelease
      The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 0E98404D386FA1D9 NO_PUBKEY 6ED0E7B82643E131 NO_PUBKEY 605C66F00D6C9793
    W: GPG error: http://deb.debian.org/debian bullseye InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 0E98404D386FA1D9 NO_PUBKEY 6ED0E7B82643E131 NO_PUBKEY 605C66F00D6C9793
    E: The repository 'http://deb.debian.org/debian bullseye InRelease' is not signed.

That is a problem orthogonal to the change in this pull request.

ejoerns:
@ukleinek Yes, this is a known issue @jluebbe is just working on. It requires switching to the official Debian multiarch containers which have deprecated the ones we used before.

jluebbe (approved):

PR #1259, opened 2023-10-10
.github: workflows: coverity: disable git unsafe directory check

ejoerns:
Otherwise job will fail with

>  fatal: detected dubious ownership in repository at '/__w/rauc/rauc' | To add an exception for this directory, call:
> 
> 	git config --global --add safe.directory /__w/rauc/rauc

since the container user is not the repository owner.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -29,6 +29,9 @@ jobs:
           exit
         fi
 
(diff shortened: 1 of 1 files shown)


ejoerns:
Fixes currently broken coverity workflow

jluebbe (approved):

PR #1260, opened 2023-10-10
.github/workflows/tests: switch to official multiarch Debian containers

jluebbe:
riscv64 is not available for bookworm, but we'll be able it use it with the next release.


ejoerns:
This also includes an update from 'bullseye' to 'bookworm'. Should that be noted in the commit message?

ejoerns (approved):

ejoerns:
FTR: I have updated the branch protection rules to cover the new cross-build names.

PR #1261, opened 2023-10-10
Remove Autotools Support

ejoerns:
We have migrated to meson as build system with RAUC v1.9 and kept autotools support in parallel for two releases now to give packagers the chance to migrate.

To reduce the maintenance overhead caused by autotools, we will now finally remove it.

Since the cgi example still uses autotools, some of the files previously shared via symlinks are still required and thus moved into the `contrib/cgi/` directory.

All workflows that still targeted autotools were either adapted or removed.

A number of examples and messages in the code still referenced autotools-specific options. They were replaced by their meson pendant.


ejoerns:
Changes:

* fixed commit path prefixes
* removed more entries from `.gitignore`

jluebbe (changes requested):

jluebbe (approved):

Issue #1251, opened 2023-09-22
Document the system.conf tokens in the manual page


evelikov:
**Is your feature request related to a problem? Please describe.**

Currently `man rauc` mentions `/etc/rauc/system.conf` and that it has an ini style format. Although it does no cover the sections and options one can use within. Currently one has to (admittedly very useful) readthedocs website.

Having the information available offline would be amazing.

**Describe the solution you'd like**

Enhance the existing rauc manual entry to cover the sections/option/etc supported by system.conf

**Describe alternatives you've considered**

Depending on size or how annoying it's to auto-generate, a separate manual page might be easier? I don't have any opinion on the topic.

ejoerns:
I'd say a man page is primarily useful if the system you read it on is the system you use this information for.
So the man page should primarily be useful for 'host tool' usage, i.e. for creating or modifying bundles.

The `system.conf` is typically something that is generated offline (as in 'not on the live system') by a build system or similar.
For this use case, I would consider the reference documentation from the docs/ folder (or from readthedocs) sufficient.

How would you use the man page? Do you write the `system.conf` on your live system?

jluebbe:
It seems a full example `system.conf` (which package based distros could install to `/usr/share/doc/`) would be useful to have.

If we can figure out nice way to generate the manpage and the HTML docs from the same input, that would be nice. Otherwise, perhaps the HTML (or text?) docs should be built and installed to `/usr/share/doc/` on systems where end-users are expected to log in?

jluebbe:
OK, we two easier improvements, which could be done without duplicating the information from the `reference.rst`:

- [ ] add a meson to install the generated html documentation to `/usr/share/doc/rauc/html`
- [ ] create a commented example `system.conf`, explaining the common options

With this, the distros can decide if they want to install that by default.

PR #1269, opened 2023-10-24
meson: add htmldocs option for installing the docs

evelikov:
Currently there is a limited set of documentation installed by default. We already have the sphinx docs wired into the meson build, although there is no way for the end-user to install them.

Add an `htmldocs` option which guards both build and install of the docs. Thus people can opt out of, even if they have sphinx installed.

Issue: https://github.com/rauc/rauc/issues/1251
---
Thanks for wiring up sphinx to the meson build team. I've secretly waited for the autotools build to be merged, since wiring that wasn't fun :-P

Silliness aside, this PR addresses half the referenced issue hence why I didn't use "Fixes".

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,7 +52,10 @@ jobs:
 
     - name: Build documentation
       run: |
-        meson compile doc -C build
+        rm -rf build/
+        meson setup -Dtests=false -Dhtmldocs=true build
+        meson configure build
+        meson compile -C build
 
     - name: Build CGI example
       run: |
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -61,6 +61,11 @@ option(
 
 # build options
 option(
+  'htmldocs',
+  type : 'boolean',
+  value : 'false',
+  description : 'Enable/Disable HTML documentation')
+option(
   'tests',
   type : 'boolean',
   value : 'true',
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -1,4 +1,4 @@
-sphinx = find_program('sphinx-build', required: false)
+sphinx = find_program('sphinx-build', required: get_option('htmldocs'))
 
 if not sphinx.found()
   subdir_done()
@@ -27,5 +27,7 @@ custom_target(
   output: 'html',
   depend_files: sources_doc,
   command: [sphinx, '-b', 'html', meson.current_source_dir(), meson.current_build_dir() / 'html'],
-  build_by_default: false
+  build_by_default: get_option('htmldocs'),
+  install_dir: join_paths(datadir, 'doc', 'rauc'),
+  install: get_option('htmldocs'),
 )


jluebbe:
Perhaps it should default to false?

ejoerns:
This removes the ability to build the documentation without building the code.

Can't we have both at once when doing something like

```patch
index 31b99cef..2db1ba01 100644
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -1,4 +1,4 @@
-sphinx = find_program('sphinx-build', required: false)
+sphinx = find_program('sphinx-build', required: get_option('htmldocs'))
 
 if not sphinx.found()
   subdir_done()
@@ -27,5 +27,7 @@ custom_target(
   output: 'html',
   depend_files: sources_doc,
   command: [sphinx, '-b', 'html', meson.current_source_dir(), meson.current_build_dir() / 'html'],
-  build_by_default: false
+  build_by_default: get_option('htmldocs'),
+  install_dir: docsdir,
+  install: get_option('htmldocs')
 )
```
?

What I am not yet sure about is if we should call this option ``htmldocs`` or just ``docs``, i.e. if we should have a distinct option for installing possible later non-HTML docs or not.

ejoerns:
> The existing non-html docs (man, conf) seem miniscule at 16K. That said, I have no opinion on the toggle name - `htmldocs` `html-docs` `docs` other?

I thought more about potential additional documentation to generate like config examples, additional man pages, etc.

> Let me know your preferences and I'll update the PR.

Let's stick to the name you used (`htmldocs`).

.github/workflows/tests.yml
```diff
@@ -52,7 +52,10 @@ jobs:
 
     - name: Build documentation
       run: |
-        meson compile doc -C build
+        rm -rf build/
+        meson setup -Dservice=false -Dcreate=false -Dnetwork=false -Dstreaming=false -Djson=disabled -Dgpt=disabled -Dtests=false -Dhtmldocs=true build
```
ejoerns, line 56: I guess we should just leave the fields to their default and only set `-Dtests=false -htmldocs=true`.
We build the code anyway and build performance should not be an issue.

meson_options.txt
```diff
@@ -60,6 +60,11 @@ option(
   description : 'D-Bus interfaces directory')
 
 # build options
+option(
+  'htmldocs',
+  type : 'boolean',
+  value : 'false',
+  description : 'Enable/Disable html documentation')
```
ejoerns, line 67:
```suggestion
  description : 'Enable/Disable HTML documentation')
```

ejoerns (commented):
Just two minor nitpicks. The rest looks fine for me!

(I have no really strong opinion on having the patch history in the commit message. Just wanted to note that this is uncommon for other PRs so far).

ejoerns (approved):
Thank you for the changes!

PR #1273, opened 2023-10-25
utils: add fakeroot helpers

jluebbe:
These support a persistent fakeroot environment, so multiple subprocess can be executed within the same environment (for example extracting a tar and generating a squashfs).

### DIFF
--- a/test/utils.c
+++ b/test/utils.c
@@ -165,6 +165,97 @@ static void update_symlink_test(void)
 	g_clear_pointer(&target, g_free);
 }
 
+static void fakeroot_test(void)
+{
+	g_autofree GError *ierror = NULL;
+	g_autofree gchar *envpath = NULL;
+	g_autoptr(GPtrArray) chown_args = g_ptr_array_new_with_free_func(g_free);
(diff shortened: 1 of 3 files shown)


test/utils.c
```diff
@@ -165,6 +165,97 @@ static void update_symlink_test(void)
...
+	g_assert_cmpstr(g_strchomp(stdout), ==, "0");
+
+	res = g_subprocess_wait_check(sproc, NULL, &ierror);
+	g_assert_no_error(ierror);
+	g_assert_true(res);
+
+	g_clear_pointer(&stdout, g_free);
+	g_clear_pointer(&sproc, g_object_unref);
+
+	/* chown /dev/zero to user 1*/
```
ejoerns, line 217: I guess this was changed to `/etc/hostname` during development? (same below).

ejoerns (changes requested):
Looks fine! Just some minor notes (and maybe change s/subprocess/subprocesses/ in the commit message).

ejoerns (approved):

PR #1275, opened 2023-10-25
add grubalt bootchooser

mbessler:
The grubalt bootchooser is an alternative Grub-based bootchooser that works more like the uboot bootchoser.

The grubalt bootchooser requires Grub built with decrement support. Currently this feature requires patching Grub: https://src.fedoraproject.org/rpms/grub2/blob/db4a99687c8306760702abda9c01a2e74d19a3f3/f/0253-Reimplement-boot_counter.patch


jluebbe:
Rebased to master to retrigger CI.

jluebbe (commented):
The code looks looks good, although I'd prefer something like "grub-counting" as the name, as "alt" is too generic.

Before merging this, we'd prefer having the corresponding support upstream in grub. @mbessler, are you aware of attempts to get this merged? I've not found any in a quick search.

ejoerns:
@mbessler We could have both patch sets in meta-rauc, yes. This would at least ensure that we have a consistent environment.
However, at the moment I am not yet very convinced that this is a better option than finding someone who has an interest in pushing these things towards GRUB mainline, because it means additional maintenance effort for meta-rauc without knowing if it is actually used or a benefit for others.

PR #1276, opened 2023-10-26
Add 'crypt' Bundle Support to 'resign' and 'replace-signature' Subcommands

ejoerns:
So far, `rauc resign` and `rauc replace-signature` did not support crypt bundles.

While fixing some underlying inconsistencies, this PR adds support for both.

- [x] add tests

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -1286,10 +1286,6 @@ gboolean encrypt_bundle(RaucBundle *bundle, const gchar *outbundle, GError **err
 	GError *ierror = NULL;
 	GBytes *encdata = NULL;
 	gboolean res = FALSE;
-	g_autoptr(GFile) bundlefile = NULL;
-	g_autoptr(GFileIOStream) bundlestream = NULL;
-	GOutputStream *bundleoutstream = NULL; /* owned by the bundle stream */
-	guint64 offset;
 
 	g_return_val_if_fail(bundle != NULL, FALSE);
 	g_return_val_if_fail(outbundle != NULL, FALSE);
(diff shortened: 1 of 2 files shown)


hthiery:
@ejoerns Thank you for quickly provide this PR. I tested it and it looks good so far.

ejoerns:
This introduces a lot of restructuring and cleanup of the bundle manipulation code to reduce duplication, useless calls and fix behavior.

test/rauc.t
```diff
@@ -903,6 +903,30 @@ test_expect_success "rauc resign" "
...
+test_expect_success "rauc resign (crypt bundle)" "
+  cp -L ${SHARNESS_TEST_DIRECTORY}/good-crypt-bundle-unencrypted.raucb ${TEST_TMPDIR}/ &&
+  test_when_finished rm -f ${TEST_TMPDIR}/good-crypt-bundle-unencrypted.raucb &&
+  rm -f ${TEST_TMPDIR}/out.raucb &&
+  rauc \
+    --cert $SHARNESS_TEST_DIRECTORY/openssl-ca/dev/autobuilder-1.cert.pem \
+    --key $SHARNESS_TEST_DIRECTORY/openssl-ca/dev/private/autobuilder-1.pem \
+    --keyring $SHARNESS_TEST_DIRECTORY/openssl-ca/dev-ca.pem \
+    resign ${TEST_TMPDIR}/good-crypt-bundle-unencrypted.raucb ${TEST_TMPDIR}/out.raucb &&
+  test -f ${TEST_TMPDIR}/out.raucb
```
jluebbe, line 927: Should we also run `rauc info` or even `rauc extract` in both cases to check the resulting signature?

src/bundle.c
```diff
@@ -1095,16 +1093,7 @@ gboolean resign_bundle(RaucBundle *bundle, const gchar *outpath, GError **error)
...
-		g_print("Reading bundle in 'plain' format\n");
-	} else if (manifest->bundle_format == R_MANIFEST_FORMAT_VERITY || manifest->bundle_format == R_MANIFEST_FORMAT_CRYPT) {
-		g_print("Reading bundle in '%s' format\n", manifest->bundle_format == R_MANIFEST_FORMAT_VERITY ? "verity" : "crypt");
-		g_assert(bundle->size > (goffset)manifest->bundle_verity_size);
-	} else {
-		g_error("unsupported bundle format");
-		res = FALSE;
-		goto out;
-	}
+	g_print("Resigning '%s' format bundle\n", r_manifest_bundle_format_to_str(manifest->bundle_format));
```
jluebbe, line 1096: Why is the check for unsupported bundle formats not needed anymore? Perhaps `check_allowed_bundle_format` should also reject unknown formats, then we know after `check_bundle`, that the format is supported.

jluebbe (changes requested):
Looks good otherwise. Nice cleanup!

test/rauc.t
ejoerns: Probably, yes. At least I can remember having thought about this :smirk:

src/bundle.c
ejoerns: It was a sanity check only, leading to a hard abort (notice it uses `g_error()`).

No valid code path should allow ending up with an unknown bundle format.
The reason we had it in there was mainly to have an `else` branch documented.

I can also put it into the string conversion if you like.

jluebbe: Ah, the reason we can't error out here is that `parse_manifest()` already rejects any unsupported format.

ejoerns: "Can't" as in "shouldn't if the parsing code remains valid, yes". This is how I understand the general usage of `g_error()`.

ejoerns:
I've now extended the resign tests to verify stricter (with different pre- and post-verification keyrings) and with additional 'rauc info' calls.

PR #1278, opened 2023-11-13
qemu-test-init: add an ll alias for ls -l

jluebbe:


ejoerns (approved):
It's a common alias and might be helpful for people who often navigate through the qemu system while developing/testing. So fine for me.

PR #1280, opened 2023-11-14
simplify subprocess handling

jluebbe:
We often run external tools without the need to communicate with them while running.
So add a helper which calls `r_subprocess_newv()` and `g_subprocess_wait_check()` instead of duplicating that everywhere.


ejoerns (approved):
Looks fine to me! :+1:

PR #1281, opened 2023-11-14
qemu-test: allow specifiying test to run

jluebbe:
This can be used with `git bisect run`.

### DIFF
--- a/qemu-test-init
+++ b/qemu-test-init
@@ -26,6 +26,9 @@ for x in $(cat /proc/cmdline); do
     export ASAN_OPTIONS=malloc_context_size=64:fast_unwind_on_malloc=0
   fi
 done
+if [ -n "$test" ]; then
+  MESON_TEST="$test"
+fi
 
 hostname qemu-test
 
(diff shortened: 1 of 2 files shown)


qemu-test-init
```diff
@@ -148,7 +151,7 @@ if [ -n "$SERVICE" ]; then
   rauc status mark-good
 fi
 
-if ninja test -C $BUILD_DIR -v; then
+if meson test -C $BUILD_DIR -v "$MESON_TEST"; then
```
ejoerns, line 154: I am not sure if this is intentional, but this significantly changes the output since `-v` in `ninja` means "show all command lines while building" while in for `meson test` it means "Do not redirect stdout and stderr".

As a result, one will always get the full log dumped into the console which is possibly not what is desired or intended?

jluebbe: Do you have a suggestion on how to keep the behavior similar?

ejoerns: Most similar should be to just omit the `-v` flag.

ejoerns (approved):

PR #1283, opened 2023-11-15
test/service: simplify service fixture setup

jluebbe:
We need the same system setup in both cases, but the bundle in only one of them. Refactor the fixtures into a generic one (`service_fixture_set_up`) and one for install tests (`service_install_fixture_set_up`).

### DIFF
--- a/test/service.c
+++ b/test/service.c
@@ -24,7 +24,7 @@ typedef struct {
 GMainLoop *testloop = NULL;
 RInstaller *installer = NULL;
 
-static void service_install_fixture_set_up(ServiceFixture *fixture, gconstpointer user_data)
+static void service_fixture_set_up(ServiceFixture *fixture, gconstpointer user_data)
 {
 	g_autofree gchar *contents = NULL;
 	g_autofree gchar *filename = NULL;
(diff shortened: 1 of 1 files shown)


ejoerns (commented):
I am unsure if we need a full system with dummy target slots, etc. set up just for testing the invocation of `rauc info` on a bundle.

I would say this was the motivation for adding a more lightweight fixture for info that just loads a config file.

jluebbe:
> I am unsure if we need a full system with dummy target slots, etc. set up just for testing the invocation of `rauc info` on a bundle.
> 
> I would say this was the motivation for adding a more lightweight fixture for info that just loads a config file.

For the artifact repos, I want to check existence during service startup.  And I'd argue that we should do the same for slots as well, so we'd need dummy slots anyway.

ejoerns:
> > I am unsure if we need a full system with dummy target slots, etc. set up just for testing the invocation of `rauc info` on a bundle.
> > I would say this was the motivation for adding a more lightweight fixture for info that just loads a config file.
> 
> For the artifact repos, I want to check existence during service startup. And I'd argue that we should do the same for slots as well, so we'd need dummy slots anyway.

Sounds reasonable. I guess we had this on our wish list anyway. Thus I'm fine with the fixture rework (which also has little impact anyway).

PR #1284, opened 2023-11-22
docs/using: be more specific about how hooks work

ejoerns:
Emphasize more that the hook script

* should be part of the bundle
* is invoked based on hooks= settings
* is responsible for multiplexing hook calls

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -296,9 +296,10 @@ to stdout, like ``RAUC_SYSTEM_SERIAL`` or ``RAUC_SYSTEM_VARIANT``.
 Bundle-Based Customization: Hooks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Unlike handlers, hooks are part of the update bundle and must be
-specified in the bundle's :ref:`sec_ref_manifest` file and handled by a common
-executable.
+Unlike handlers, hooks are defined in the update bundle and must be
+specified in the bundle's :ref:`sec_ref_manifest` file.
+All hooks are handled by a common executable that must be included in the
+bundle.
 Hooks allow the author of a bundle to add or replace functionality for the
 installation of a specific bundle.
 This can be useful for performing additional migration steps, checking for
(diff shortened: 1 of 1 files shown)


docs/using.rst
```diff
@@ -298,7 +298,7 @@ Bundle-Based Customization: Hooks
 
 Unlike handlers, hooks are part of the update bundle and must be
 specified in the bundle's :ref:`sec_ref_manifest` file and handled by a common
-executable.
+executable that should be part of the bundle.
```
jluebbe, line 301: `part of the (update) bundle` is now at the beginning and end of the sentence, which sounds strange.

docs/using.rst
```diff
@@ -313,9 +313,16 @@ handled by a single executable that is registered in the bundle's manifest:
   [hooks]
   filename=hook
 
-Each hook must be activated explicitly and leads to a call of the hook executable
-with a specific argument that allows to distinguish between the different hook
-types. Multiple hook types must be separated with a ``;``.
+The ``filename`` must match the name of the script or binary executable placed
+inside the content folder the bundle is generated from.
+
+The actual hook invocation must be registered explicitly in the manifest via
```
jluebbe, line 319:
```suggestion
The actual hook invocation must be registered for each `[image.*]` in the manifest via
```

jluebbe (changes requested):

docs/using.rst
ejoerns: I had this first, but this would exclude "bundle hooks" (i.e. install-check) which are not inside an `image` section. So the other alternative would be to mention both. `[image.*]` and `[hooks]`.

docs/using.rst
```diff
@@ -296,9 +296,10 @@ to stdout, like ``RAUC_SYSTEM_SERIAL`` or ``RAUC_SYSTEM_VARIANT``.
 Bundle-Based Customization: Hooks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Unlike handlers, hooks are part of the update bundle and must be
-specified in the bundle's :ref:`sec_ref_manifest` file and handled by a common
-executable.
+Unlike handlers, hooks are defined in the update bundle and must be
+specified in the bundle's :ref:`sec_ref_manifest` file.
+All hooks are handled by a common executable that should also be part of the
+bundle.
```
jluebbe, line 302:
```suggestion
All hooks are handled by a common executable that must be included in the
bundle.
```

jluebbe (changes requested):

docs/using.rst
ejoerns: I actually assumed that it would be possible to have absolute paths here. But in fact it isn't really because of the indirect check from https://github.com/rauc/rauc/commit/c3b7c7d64e0891b9712e756fe092fd97f59a2fa0.

jluebbe (approved):

PR #1285, opened 2023-11-22
Create DBus method to dump the certificate

sanapci:
The changes will create a new DBus method, for other applications to get the certificate chain from a bundle.

There is a command which prints a readable format from CLI, but with these changes we could be able to retrieve certificate chain programmatically and work on the "consumer" side.

**What do you use the feature for?** Get the certificate from a bundle and validate it outside of the RAUC .
**How does RAUC benefit from the feature?** This could be useful also for other users if want to handle certificates outside of the RAUC, in their application. For example, let's say there are multiple signing certificates and we want to get from the bundle which one was used and interpret behavior based on this.
**How did you verify the feature works?** Used OpenSSL commands to generate the certificate chain and compared it with the one from the output of the changes. Also the output of the changes can be parsed and printed with OpenSSL commands and print it in readable format.
**If hardware is needed for the feature, which hardware is supported and which
  hardware did you test with?** No hardware required, only a created bundle required.


ejoerns:
@sanapci Thank you for sharing your work!

> **How does RAUC benefit from the feature?** This could be useful also for other users if want to handle certificates outside of the RAUC, in their application. For example, let's say there are multiple signing certificates and we want to get from the bundle which one was used and interpret behavior based on this. 

I guess with 'multiple signing certificates' you mean alternative ones, since RAUC at the moment does not support multiple signers (at once), right?
Could you give an insight on what kind of interpretation you would do based on the signer? 
Does it have to be done *before* invoking the installation?

From inside the bundle one could for example use hooks and the exported SPKI hashes to perform actions based on the kind of certificate used for signing: https://rauc.readthedocs.io/en/latest/advanced.html#switching-the-keyring-spki-hashes

jluebbe:
This approach would make it easy to have a [TOC/TOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) vulnerability, as there is no guarantee that an installation started after calling this method would see the same bundle.

As @ejoerns mentioned, we'll need more details on your use case and an explanation why you see this approach is the best solution for it.

sanapci:
@ejoerns @jluebbe Could you give an update (remarks, opinions, etc.) about this based on my previous comment?

jluebbe:
> @ejoerns @jluebbe Could you give an update (remarks, opinions, etc.) about this based on my previous comment?

Sorry, but the scenario doesn't make sense to me, yet.

How do you authenticate the new certificate?

> > Could you give an insight on what kind of interpretation you would do based on the signer?
>
> Let's say, the bundle validation fails with the installed trustchain. In this case, maybe an update required for the trustchain on the system because it is an outdated one and the installation fails.

To know that you keyring is outdated, you shouldn't need the certificates, but just a dedicated error result from `InstallBundle` or `InspectBundle`. If you see that error, you'd know that an update is needed and the correct CA certificate could be fetched over a separate authenticated channel.

At that point I'm wondering though: If you have a way to authenticate the new CA certificate, why not just use the same certificate to authenticate RAUC bundles and avoid the whole issue from the start?

sanapci:
The changes in the pull requests are containing some errors and needs to be updated if required.

The details maybe were a bit misleading, so to be clear:
There is a root CA, an intermediate CA and the certificate used for bundle creation. By using the --intermediate, the bundle can be used with the root CA, what is more the intermediate CA can be extracted also. If the CA is updated online, but the system is offline or used offline, the certificate extraction from the bundle can be a solution to install new CA via an update bundle.

The extracted CA and certificate is verified with the root CA.

PR #1288, opened 2023-11-27
test/common: add test_lstat helper

jluebbe:
This will be used for artifact installation tests, where we need to check if file ownership and created device nodes are correct.

### DIFF
--- a/test/common.c
+++ b/test/common.c
@@ -151,6 +151,19 @@ int test_remove(const gchar *dirname, const gchar *filename)
 	return res;
 }
 
+int test_lstat(const gchar *dirname, const gchar *filename, GStatBuf *buf)
+{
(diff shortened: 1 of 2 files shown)


test/common.c
```diff
@@ -151,6 +151,19 @@ int test_remove(const gchar *dirname, const gchar *filename)
...
 
+int test_lstat(const gchar *dirname, const gchar *filename, GStatBuf *buf)
+{
+	g_autofree gchar *path = NULL;
+	int res;
+
+	path = g_build_filename(dirname, filename, NULL);
+	g_assert_nonnull(path);
+
+	res = g_lstat(path, buf);
```
ejoerns, line 162: Would have said that we do not need `res` here, but since this is consistent with the other methods: fine for me.

PR #1289, opened 2023-11-27
Extend Documentation of Image Extension/Type Handling & More

ejoerns:
This documentation update mainly addresses the relevance of image file name extensions in the context of update handler selection. This has repeatedly led to irritation.

While the [Installation and Storage Handling](https://rauc.readthedocs.io/en/latest/basic.html#installation-and-storage-handling) chapter slightly touches the topic, it is not well discoverable and, lacks a list of supported file name extensions, etc.

The chapter is now rewritten, an explicit list of supported image types was added to the manifest reference and a couple of hints that the file name extension is important have been added.

Since reading documentation (especially those that have existed for a while) always leads to finding things that are not fully correct anymore or could be described better, this PR became a collection of various documentation updates.

I was also surprised that there was no real section on how to create a bundle (except for the very general [Creating Bundles](https://rauc.readthedocs.io/en/latest/using.html#creating-bundles) one in the 'Using RAUC' chapter).


jluebbe (approved):

ejoerns (commented):
I'm fine with your version of my documentation update, too. :smirk:

PR #1293, opened 2023-12-17
Adding CRYPTO_AES documentation

angeloc:
CRYPTO_AES option is mandatory to have crypto support working in RAUC.

Adding a bit of documentation and a better error when RAUC fails setup DM for cypto format.


jluebbe (approved):

jluebbe:
Thanks!

ejoerns:
Does it fail with "Failed to load dm table" when CRYPTO_AES is disabled?

Since we know if we are setting up crypt or verity, we should also be able to point to the actual type that fails or is missing (maybe as a further improvement)?

angelo-zehnder:
It fails with the very same error.

Yeah, as a further improvements, we shall distinguish the two and give two different error messages.

PR #1302, opened 2023-12-20
qemu-test-init: print backtrace for service threads if gdb is available

jluebbe:
This makes debugging easier when something fails in the service.


ejoerns:
Did a quick test. Looks quite helpful for testing/debugging :+1:

ejoerns (approved):

PR #1303, opened 2023-12-20
update_utils: refactor output stream

jluebbe:
Depending on the situation, we want to write into existing devices (/files for testing) or make sure to create new files. Create separate functions for that.


ejoerns (commented):
While at it, we should maybe let this better match the glib naming scheme which would mean to have

* `r_unix_output_stream_open_device`
* `r_unix_output_stream_create_file`

and should also be more intuitive to read and get what they do?

ejoerns (approved):

PR #1306, opened 2023-12-20
remove dashes from message IDs and add note in changelog

jluebbe:
The journald catalog uses UUIDs without dashes, as generated by `systemd-id128 new` (without `-u`), so remove the dashes while the change won't cause backwards-compatibility issues.

We don't consider the details of the new log events fixed yet, so please use them as a preview and for testing.
In a future release, they will be documented in a [journald message catalog](https://www.freedesktop.org/wiki/Software/systemd/catalog/).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,16 @@ Release 1.11 (development)
 * Remove autotools build system.
   Support for meson was added in 1.9 and supported in parallel to autotools
   until 1.10.1.
+* Add log events for slot update, start/end of an installation, good/bad/active
+  marking and boot/service restart.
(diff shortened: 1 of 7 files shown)


CHANGES
```diff
@@ -27,6 +27,16 @@ Release 1.11 (development)
 * Remove autotools build system.
   Support for meson was added in 1.9 and supported in parallel to autotools
   until 1.10.1.
+* Add log events for slot update, start/end of an installation and boot/service
```
ejoerns, line 30:
```suggestion
* Add log events for slot update, start/end of an installation, good/bad/active marking and boot/service
```

ejoerns (approved):

PR #1310, opened 2023-12-22
test/service: run service tests only as root

ejoerns:
With the rework of 8e2c45ed, these tests cannot be run as unpriviledged user anymore.

Fixes #1283.

### DIFF
--- a/test/service.c
+++ b/test/service.c
@@ -353,6 +353,10 @@ static void service_test_info(ServiceFixture *fixture, gconstpointer user_data, 
 		return;
 	}
 
+	/* needs to run as root */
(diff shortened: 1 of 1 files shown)


ejoerns:
Note that this was not detected by CI since we run the test only in qemu. Should we also run them unprivileged? Maybe in the stable tests?

jluebbe (approved):

ukleinek:
I was too slow with my feedback, but for the record: I agree and this patch is already part of the Debian downstream package. Thanks to @ejoerns, 🌟🌟🌟🌟🌟

PR #1314, opened 2024-01-05
src/main: fix and add support for boot detection with rauc.external

### LINKED ISSUE #1313: Rauc service core dump on 1.11
**Describe the bug**
Rauc service core dump running from external media (USB) (first install scenario)

**Background information**
1.11

**To Reproduce**
Boot

**Expected behavior**
Service should run (ok on 1.10.1)

**Logs**
```
Jan 05 07:09:21 dr-imx8mp systemd[1]: Starting RAUC Update Service...
Jan 05 07:09:21 dr-imx8mp rauc[698]: Using central status file /service/rauc/status
Jan 05 07:09:21 dr-imx8mp rauc[698]: Failed to load system status: No such file or directory
Jan 05 07:09:21 dr-imx8mp rauc[698]: Getting Systeminfo: /usr/lib/rauc/info-provider.sh
Jan 05 07:09:21 dr-imx8mp rauc[698]: Detected explicit external boot, ignoring missing active slot
Jan 05 07:09:21 dr-imx8mp rauc[698]: r_slot_status_load: assertion 'dest_slot' failed
Jan 05 07:09:21 dr-imx8mp systemd[1]: rauc.service: Main process exited, code=dumped, status=5/TRAP
Jan 05 07:09:21 dr-imx8mp systemd[1]: rauc.service: Failed with result 'core-dump'.
Jan 05 07:09:21 dr-imx8mp systemd[1]: Failed to start RAUC Update Service.
```

**Additional context**
Boots from USB stick targeting install on eMMC

ejoerns:
The current boot detection handling did not take into account that there is no booted slot when using `rauc.external` in cmdline and called `r_slot_status_load()` with `NULL` argument causing a

    (rauc:676): rauc-CRITICAL **: 16:16:16.336: r_slot_status_load: assertion 'dest_slot' failed

Fix this by explicitly checking for external boot and also emit a dedicated log message since this might be relevant for the log, too.

Fixes #1313
Fixes b9e4cdce0879862126aeb786bdd7f6672e392ca9 ("src/main: log boot and service restart")

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2004,6 +2004,7 @@ static gboolean status_start(int argc, char **argv)
 }
 
 #define MESSAGE_ID_BOOTED "e60e0addd3454cb8b796eae0d497af96"
+#define MESSAGE_ID_BOOTED_EXTERNAL "dd237efdad1945d9b1e471bc2b994532"
 
 static void r_event_log_booted(const RaucSlot *booted_slot)
 {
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -2078,12 +2096,17 @@ static gboolean service_start(int argc, char **argv)
...
+			} else {
+				RaucSlot *booted_slot = r_slot_find_by_device(r_context()->config->slots, r_context()->bootslot);
+				if (!booted_slot)
+					booted_slot = r_slot_find_by_bootname(r_context()->config->slots, r_context()->bootslot);
 
-			r_event_log_booted(booted_slot);
+				r_slot_status_load(booted_slot);
+
+				r_event_log_booted(booted_slot);
+			}
```
jluebbe, line 2109: Shouldn't we try to handle the case where we still can't find the booted slot?

We may not be able to install in that state, but `rauc info` should still work.

jluebbe (changes requested):

src/main.c
ejoerns: I've thought about this, too but did not find a proper handling.

If the slot is invalid, the service will abort anyway with

    Failed to determine slot states: Did not find booted slot (matching 'XYZ')

In how far do you mean that `rauc info` will not work? It does not use the service, thus this should be unrelated.

jluebbe (approved):

PR #1316, opened 2024-01-09
src/bundle: remove signature size sanity check

### LINKED ISSUE #1308: Cannot create a large amount of device keys for encrypted bundles
I am using Yocto ( Kirkstone ) to create an image for an embeded Linux device with Linux Kernel version 5.15.

I have implemented encrypted bundles with rauc using this guide:
https://www.pengutronix.de/en/blog/2022-03-31-tutorial-rauc-bundle-encryption-using-meta-rauc.html

Everything works great with the resulting bundle and on the end device, unless I try to create a number of device keys greater than 256.

I use this bit of code to create the device keys:

```
# Create X number of key/certs
for i in $(seq 1 256); do
    echo "*** Generating this key: $i ..."
    openssl ecparam -name prime256v1 -genkey -noout -out ${OUTPUT_DIR}/private/private-key-$i.pem
    openssl ec -in ${OUTPUT_DIR}/private/private-key-$i.pem -pubout -out ${OUTPUT_DIR}/public-key-$i.pem
    openssl req -new -x509 -key ${OUTPUT_DIR}/private/private-key-$i.pem -out ${OUTPUT_DIR}/recipient-cert-$i.pem -days 365250 -subj "/O=RAUC/CN=RAUC Test $i"
done
```

I then do a Yocto build, and encrypt the resulting *.raucb bundle with commands like this on the build machine:

```
# Encrypt bundle
echo "*** Encrypting this bundle: $LATEST_RAUCB_BUNDLE ..."
oe-run-native rauc-native rauc encrypt \
  --keyring=$RAUC_CA_KEYRING \
  --to $RECIPIENTS_FILE $LATEST_RAUCB_BUNDLE \
  $ENCRYPTED_RAUCB

# Define keys location
KEYS_DIR="/workspaces/keys/rauc/rauc-device-keys"

# Check bundle
echo "*** Checking this bundle: $ENCRYPTED_RAUCB ..."
oe-run-native rauc-native rauc info \
  --keyring=$RAUC_CA_KEYRING \
  --key=${KEYS_DIR}/private/private-key-1.pem --dump-recipients \
  $ENCRYPTED_RAUCB
```

If I generate only 256 keys initially, this whole process works and the encrypted bundle can be used for updates. If I generate more than 256 keys, I get the following error - both on the build machine during the 'rauc info' stage, and also on the device if I try to update with the bundle.

`Signature size [..] exceeds bundle size`

I've seen this issue which references the same error - however the file md5sums match what I expect and without changing anything else, if I create less keys, then it works:

https://github.com/rauc/meta-rauc/issues/235

I would greatly appreciate any help in resolving this  - thanks.

ejoerns:
Historically, checking that the signature size is smaller than the bundle size was a sensible test.
However, with the added support for encrypted bundles, we may run in cases where the bundle size is actually smaller than the signature size. This happens because the CMS structure holds the individual recipients for encrypted bundles and with having thousands of recipients, the CMS structure size can reach a notable size. If that's combined with a quite small update image, the sanity check can be triggered and prevents from installing such a bundle.

Since this is just a sanity check, since we have other sanity checks and since we intend to support large amounts of encryption recipients, simply remove the sanity check.

Fixes #1308.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -1697,13 +1697,6 @@ static gboolean open_local_bundle(RaucBundle *bundle, GError **error)
 		res = FALSE;
 		goto out;
 	}
-	/* sanity check: signature should be smaller than bundle size */
-	if (sigsize > (guint64)offset) {
-		g_set_error(error, R_BUNDLE_ERROR, R_BUNDLE_ERROR_SIGNATURE,
-				"Signature size (%"G_GUINT64_FORMAT ") exceeds bundle size", sigsize);
-		res = FALSE;
-		goto out;
-	}
 	/* sanity check: signature should be smaller than 64KiB */
 	if (sigsize > MAX_BUNDLE_SIGNATURE_SIZE) {
 		g_set_error(error, R_BUNDLE_ERROR, R_BUNDLE_ERROR_SIGNATURE,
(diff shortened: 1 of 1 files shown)


ejoerns:
The check compares the signature size against the *complete bundle size*, not only the *payload size* as I initially thought. And the *complete bundle size* (which includes the signature) should, of course, *always* be smaller than just the signature size.

So there is no reason to remove the check.

PR #1318, opened 2024-01-10
build multiarch CI containers, use podman and run user-level tests

jluebbe:
This PR refactors our CI container builds to use podman instead of docker.

It also adds pre-built containers for arm/v5, arm/v7, arm64/v8 and 386, which allows us to drop the dependency installation under QEMU.

For now, we need to install a newer buildah/podman from Ubuntu noble, but we'll be able to drop that step when GitHub Actions adds an image for the Ubuntu 24.04 LTS release.

To show how the containers work, I've added one commit to pull them from jluebbe/rauc instead of the main repo. When this PR is reviewed, I'd drop that commit.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -6,6 +6,7 @@ on:
       - 'master'
       - 'container'
     paths:
+      - '.github/workflows/container.yml'
       - 'test/Dockerfile'
   workflow_dispatch: {}
 
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -41,6 +41,12 @@ jobs:
         echo "Image: ${{ steps.build-image.outputs.image }}"
         echo "Tags: ${{ steps.build-image.outputs.tags }}"
         echo "Tagged Image: ${{ steps.build-image.outputs.image-with-tag }}"
+    - name: List Images
+      run: |
+        buildah images
+    - name: Inspect Manifest
+      run: |
+        buildah manifest inspect "${{ steps.build-image.outputs.image-with-tag }}"
     - name: Push Images
       id: push-to-github
       uses: redhat-actions/push-to-registry@v2
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -34,7 +34,7 @@ jobs:
         image: rauc-ci
         tags: latest
         platforms: linux/amd64, linux/386, linux/arm/v5, linux/arm/v7, linux/arm64/v8
-        dockerfiles: |
+        containerfiles: |
           ./test/Dockerfile
     - name: Echo Outputs for Build
       run: |
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -14,15 +14,35 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+    - name: Install QEMU
+      run: |
+        sudo apt-get update
+        DEBIAN_FRONTEND='noninteractive' sudo apt-get install -qy qemu-user-static
+    # remove this when Ubuntu 24.04 is available on GitHub Actions
+    - name: Enable Ubuntu noble and update buildah
+      run: |
+        echo 'deb http://archive.ubuntu.com/ubuntu/ noble main restricted universe' | sudo tee -a /etc/apt/sources.list.d/lunar.list
+        echo 'deb http://archive.ubuntu.com/ubuntu/ noble-updates main restricted universe' | sudo tee -a /etc/apt/sources.list.d/lunar.list
+        sudo apt-get update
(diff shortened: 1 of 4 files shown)


test/Dockerfile
```diff
@@ -62,10 +62,6 @@ RUN git clone https://github.com/folbricht/desync.git /tmp/desync && \
 RUN mkdir -p /lib/modules && \
     mkdir -p /var/run/dbus
 
-# We want to run as non-root user equaling uid of Travis' user 'travis' (2000)
-ENV user travis
+RUN adduser --disabled-password --gecos "" rauc-ci
```
ejoerns: `man adduser` doesn't list `--gecos` anymore, but for `--comment` it says:

```man
       --comment comment
              Set  the  comment  field for the new entry generated.  adduser will not ask for the information if this option is given.  This field is also known under the name GECOS field and contains information that is used by the finger(1) com-
              mand.  This used to be the --gecos option, which is deprecated and will be removed after Debian bookworm.  Valid Modes: adduser, adduser --system.
```

test/Dockerfile
```diff
@@ -1,7 +1,9 @@
 FROM debian:bullseye
 
 # Required for building
-RUN apt-get update && apt-get install -y \
+RUN apt-get update -q && apt-get install -q -y --no-install-recommends \
+  gcc \
```
ejoerns, line 5: These need to be added since it was installed as a recommended package before?
Maybe give a short hint in the commit message.

test/Dockerfile
```diff
@@ -12,21 +14,31 @@ RUN apt-get update && apt-get install -y \
   libfdisk-dev \
   libnl-genl-3-dev
 
+# Required for building the cgi example
```
ejoerns, line 17: Shouldn't this be a separate change?

test/Dockerfile
```diff
@@ -75,6 +76,7 @@ RUN mkdir -p /lib/modules /var/run/dbus
 # Remove apt lists
 RUN rm -rf /var/lib/apt/lists/*
 
-RUN adduser --disabled-password --gecos "" rauc-ci
+RUN adduser --disabled-password --gecos "" rauc-ci && \
+    echo "rauc-ci ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/ptx
```
ejoerns, line 83: Maybe use a different file name here?

.github/workflows/container.yml
```diff
@@ -6,13 +6,18 @@ on:
...
     paths:
+      - '.github/workflows/container.yml'
       - 'test/Dockerfile'
   workflow_dispatch: {}
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+    - name: Install QEMU
```
ejoerns, line 17: This seems to be unrelated to `build on workflow changes`, thus maybe put it in another commit.

.github/workflows/tests.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/rauc/rauc/rauc-ci:latest
+      image: ghcr.io/jluebbe/rauc/rauc-ci:latest
```
ejoerns, line 9: This doesn't match the commit title (which still includes `WIP` anyway).

ejoerns (changes requested):
Looks like a great enhancement for CI testing :+1:

Two minor notes on commit messages:

* `test/Dockerfile: only build desync on x86_64` should be `test/Dockerfile: build desync only on x86_64`

* `.github/workflows/container: add multiimage debugging` -> `multi-image` is better to read

test/Dockerfile
jluebbe: `--comment` is only supported from bookworm. At that point of the series, it's still bullseye.

test/Dockerfile
jluebbe: No, as this is only necessary due to moving to explicitly installing dependencies (`--no-install-recommends`).

.github/workflows/tests.yml
jluebbe: l'll fix the commit message.

.github/workflows/container.yml
jluebbe: Move to `.github/workflows/container: build multi-platform image`.

.github/workflows/tests.yml
jluebbe: Yes, see the last paragraph in the description. :)

ejoerns: Got it, thx!

ejoerns (dismissed):

jluebbe:
I've removed the WIP commit.

ejoerns (approved):

PR #1322, opened 2024-01-11
test/common: add helper to build a GPtrArray from a ;-separated string

jluebbe:
This makes it easy to create a `GPtrArray` containing some strings for use in test cases.
It will be used in artifact tests.

### DIFF
--- a/test/common.c
+++ b/test/common.c
@@ -422,6 +422,18 @@ void replace_strdup(gchar **dst, const gchar *src)
 	*dst = g_strdup(src);
 }
 
+GPtrArray *test_ptr_array_from_strsplit(const gchar *input)
+{
+	GPtrArray *result = g_ptr_array_new_with_free_func(g_free);
+	/* The strings will be owned by the returned GPtrArray, so we don't
+	 * want to free them via g_auto(GStrv). */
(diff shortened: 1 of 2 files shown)


test/common.c
```diff
@@ -422,6 +422,16 @@ void replace_strdup(gchar **dst, const gchar *src)
 	*dst = g_strdup(src);
 }
 
+GPtrArray *test_ptr_array_from_strsplit(const gchar *input)
+{
+	GPtrArray *result = g_ptr_array_new_with_free_func(g_free);
+	g_auto(GStrv) strv = g_strsplit(input, ";", 0);
+
+	r_ptr_array_addv(result, strv, FALSE);
```
ejoerns, line 430: If I'm not mistaken, the elements of `strv` will be freed when leaving the function context and then we return a `GPtrArray` with dangling pointers. Maybe use `TRUE` here?

ejoerns (commented):
I'd suggest adding the PR's reasoning to the commit message, too.

test/common.c
jluebbe: Instead I should only free the array, not the strings.

ejoerns (approved):

PR #1324, opened 2024-01-12
docs/using: be more explicit on current pre-install hook implementation

ejoerns:
Pre-install hooks currently offer a guarantee that the target slot will be mounted. For this, we explicitly error-out if the target slot is unmountable for any reason.

This can cause unexpected troubles when running this from a system that does not have the target slot formatted (yet) or where the target slot's file system is broken for any reason.

Be more explicit in the documentation about this fact.

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -414,8 +414,11 @@ executable, see the :ref:`sec-slot-hook-interface` section.
 .. rubric:: Pre-Install Hook
 
 The pre-install hook will be called right before the update procedure for the
-respective slot will be started. For slot types that represent a mountable file
-system, the hook will be executed with having the file system mounted.
+respective slot will be started.
+For target slot types that represent a mountable file system, the hook will be
+executed with the target slots' file system mounted.
+Note that a broken or unformatted target slot will currently cause the
+installation to be aborted with an error.
 
 .. code-block:: cfg
 


OnkelUlla (approved):

jluebbe:
The range-diff:
```
1:  e9e21e30 ! 1:  e0e09895 docs/using: be more explicit on current pre-install hook implementation
    @@ Metadata
      ## Commit message ##
         docs/using: be more explicit on current pre-install hook implementation
     
    -    Pre-install hooks currently offer a guarantee that the target slot will
    -    be mounted. For this, we explicitly error-out if the target slot is
    -    unmountable for any reason.
    +    Pre-install hooks currently guarantee that the target slot is mounted.
    +    For this, we explicitly error-out the update installtion if the target
    +    slot is unmountable for any reason.
     
    -    This can cause unexpected troubles when running this from a
    -    system that does not have the target slot formatted (yet) or where the
    -    target slot's file system is broken for any reason.
    +    This can cause unexpected troubles when running this from a system that
    +    does not have the target slot formatted (yet) or where the target slot's
    +    file system is broken for any reason.
     
    -    Be more explicit in the documentation about this fact.
    +    Be more explicit in the documentation about this fact. We'll likely
    +    need a way to handle mount errors more gracefully in the future.
     
    @@ docs/using.rst: executable, see the :ref:`sec-slot-hook-interface` section.
      The pre-install hook will be called right before the update procedure for the
     -respective slot will be started. For slot types that represent a mountable file
     -system, the hook will be executed with having the file system mounted.
    -+respective slot will be started. For target slot types that represent a mountable file
    -+system, the hook will be executed with having the target slots' file system mounted.
    -+Note that this is expected to fail for broken or unformatted target slots.
    ++respective slot will be started.
    ++For target slot types that represent a mountable file system, the hook will be
    ++executed with the target slots' file system mounted.
    ++Note that a broken or unformatted target slot will currently cause the
    ++installation to be aborted with an error.
      
      .. code-block:: cfg
      

```

jluebbe (approved):

PR #1328, opened 2024-01-17
improve structure of config and manifest reference

jluebbe:
Using titles makes the index more useful and allows us to deep-link to a specific section. 

Sometimes, we noted required/optional keys in the description, but it's better make it easy to see.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -751,12 +751,12 @@ a look at :ref:`sec-advanced-event-log`.
   * ``install`` - Logs start and end of installation
   * ``boot`` - Logs boot information
   * ``mark`` - Logs slot marking information
-  * ``all`` - Log all events (cannot be combined with other events)
+  * ``all`` - Log all events (default, cannot be combined with other events)
 
 ``format`` (optional)
   The output format used for the logger. Supported values are
 
-  * ``readable``: readable mutli-line output
+  * ``readable``: readable mutli-line output (default)
   * ``short``: Single-line readable output
   * ``json``: single-line JSON output
   * ``json-pretty``: formatted JSON output
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -51,7 +51,8 @@ Example configuration:
 
 .. _system-section:
 
-**[system] section**
+``[system]`` Section
+~~~~~~~~~~~~~~~~~~~~
 
 ``compatible``
   A user-defined compatible string that describes the target hardware as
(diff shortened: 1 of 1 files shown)


docs/reference.rst
```diff
@@ -152,7 +152,7 @@ Example configuration:
 
 .. _data-directory:
 
-``data-directory``
+``data-directory`` (optional, recommended)
```
ejoerns, line 155: For all others, we have either the one or the other.

docs/reference.rst
```diff
@@ -424,17 +423,17 @@ See details about using handlers in `Custom Handlers (Interface)`_.
...
   bundle content.
 
-``post-install``
+``post-install`` (optional)
   This handler will be called after a successful installation.
   The bundle is still mounted at this moment, thus you could access data in it
   if required.
 
-``bootloader-custom-backend``
+``bootloader-custom-backend`` (optional)
```
ejoerns, line 436: It depends. It's mandatory for `bootloader = custom`.

docs/reference.rst
```diff
@@ -739,30 +738,30 @@ unique per ``system.conf``.
...
-``filename``
+``filename`` (required)
   The log file name used for logging.
   If no absolute path is given, the location is assumed to be relative to the
   ``data-directory``.
   Using a relative file name without ``data-directory`` set will cause a
   configuration error.
 
-``events``
+``events`` (required)
```
ejoerns, line 748: Optional since it defaults to `"all"`.

docs/reference.rst
```diff
@@ -739,30 +738,30 @@ unique per ``system.conf``.
...
   Semicolon-separated list of events to log. Currently supported event types are:
 
   * ``install`` - Logs start and end of installation
   * ``boot`` - Logs boot information
   * ``mark`` - Logs slot marking information
   * ``all`` - Log all events (cannot be combined with other events)
 
-``format``
+``format`` (required)
```
ejoerns, line 756: Optional, since it defaults to `"readable"`.

ejoerns (changes requested):

docs/reference.rst
jluebbe: For the others we have either required or optional. "recommended" is independent of that.

ejoerns: You mean like

    ``type=<type>`` (recommended)

and others?

jluebbe: `type=` should be (optional, recommended) as well.

ejoerns: Fine for me. Same then for `bundle-formats` and `format`.

ejoerns (approved):

PR #1332, opened 2024-01-18
src/nbd: initialize info_headers to NULL

ejoerns:
When `g_variant_dict_lookup()` does not find the given key, it does not modify the values of the arguments provided.

This results in `info_headers` being undefined and can cause the rauc-nbd-server to crash, which will result in a broken request on client side that lets RAUC terminate hard with:

    (rauc:769): rauc-nbd-ERROR **: 15:40:30.553: failed to recv nbd config reply header

The behavior of `g_variant_dict_lookup()` has also recently been clarified in the glib documentation:
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3334

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -562,7 +562,7 @@ static void start_configure(struct RaucNBDContext *ctx, struct RaucNBDTransfer *
 
 	/* only read from the client on the first try */
 	if (!ctx->url) {
-		GStrv info_headers; /* array of strings such as 'Foo: bar' */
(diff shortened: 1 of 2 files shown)


hnez:
Hi, just wanted to say that adding the changes from this PR as a patch in meta-lxatac (linux-automation/meta-lxatac#98) fixes the issues I have seen with `rauc info` failing.

Thanks for investigating this!

jluebbe (approved):

PR #1333, opened 2024-01-19
src/nbd: manually trigger NBD disconnect

ejoerns:
When invoking 'rauc info' the connection to the NBD server is never handed to the kernel and thus just terminated by the code in r_nbd_stop_server() that just closes the socket.

This leads to messages like

> rauc-Message: 23:00:27.704: nbd server failed with: failed to read request from client: Unexpected end of file
> rauc-nbd-Message: 23:00:27.714: failed to stop ndb server: failed to stop streaming subprocess: Child process exited with code 1

This can be handled better by manually invoking a nbd disconnect before that triggers a shutdown on NBD server side that prevents the error message and also prints stats etc.

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -1244,6 +1244,16 @@ gboolean r_nbd_stop_server(RaucNBDServer *nbd_srv, GError **error)
 	g_message("nbd server stopping");
 
 	if (nbd_srv->sock >= 0) {
+		struct nbd_request request = {0};
+
+		/* If socket is still active, manually invoke NBD_CMD_DISC
+		 * first, to trigger a graceful shutdown of the NBD server. */
+		request.magic = GUINT32_TO_BE(NBD_REQUEST_MAGIC);
+		request.type = GUINT32_TO_BE(NBD_CMD_DISC);
+		request.len = 0;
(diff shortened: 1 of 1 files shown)


src/nbd.c
```diff
@@ -1244,6 +1244,16 @@ gboolean r_nbd_stop_server(RaucNBDServer *nbd_srv, GError **error)
...
 	if (nbd_srv->sock >= 0) {
+		struct nbd_request request = {0};
+
+		/* If socket is still active, manually invoke NBD_CMD_DISC
+		 * first, to trigger a graceful shutdown of the NBD server. */
+		request.magic = GUINT32_TO_BE(NBD_REQUEST_MAGIC);
+		request.type = GUINT32_TO_BE(NBD_CMD_DISC);
+		request.len = 0;
+		if (!r_write_exact(nbd_srv->sock, (guint8*)&request, sizeof(request), NULL))
+			g_error("failed to send nbd read request header");
```
jluebbe, line 1255:
```suggestion
			g_error("failed to send nbd disconnect request");
```

jluebbe (changes requested):
Looks good otherwise.

jluebbe (approved):

PR #1335, opened 2024-01-21
src/nbd: slightly clearer log messages

ejoerns:
The architecture of streaming that involves rauc-internal NBD client+server, handover to kernel NBD client, and various HTTP requests is not super trivial.
While testing and debugging the streaming and NBD client/server handling, I found some log information that was potentially misleading, imprecise, or missing.
This PR adjusts some of them to hopefully make it a little easier to debug.


src/nbd.c
```diff
@@ -619,6 +619,8 @@ static void start_configure(struct RaucNBDContext *ctx, struct RaucNBDTransfer *
 	xfer->buffer_size = 4;
 	xfer->buffer_pos = 0;
 
+	g_debug("nbd server sending test range request to HTTP server");
```
jluebbe, line 622: This is not only a test, but also used to follow redirects.

```suggestion
	g_debug("nbd server sending initial range request to HTTP server");
```

jluebbe (changes requested):
Looks good otherwise.

jluebbe (approved):

PR #1341, opened 2024-02-05
src/install: add bundle/manifest build value to hook environment

hthiery:
Add the value as RAUC_MF_BUILD to the execution environment of the install hook.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -1245,6 +1245,10 @@ The following environment variables will be passed to the hook executable:
   The value of the version field as provided by the current bundle,
   e.g. ``"V1.2.1-2020-02-28"``
 
+``RAUC_MF_BUILD``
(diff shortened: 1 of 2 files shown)


docs/reference.rst
```diff
@@ -1245,6 +1245,10 @@ The following environment variables will be passed to the hook executable:
   The value of the version field as provided by the current bundle,
   e.g. ``"V1.2.1-2020-02-28"``
 
+``RAUC_MF_BUILD``
+  The value of the build field as provided by the current bundle,
+  e.g. ``"V1.2.1-2020-02-28"``
```
ejoerns, line 1250: This should not have an example string that is equal to what we have for `RAUC_MF_BUILD`.

Maybe a date string (e.g. `20240205092747`) or a UNIX timestamp as an example.

ejoerns (approved):

PR #1344, opened 2024-02-05
Slightly Stricter Style Checks

ejoerns:
This makes three uncrustify options stricter that have a manageable impact on the current code but can help to keep the style even more readable and consistent:

* enforce spaces around assignments, comparisons, and boolean operators
* add spaces after commas


jluebbe (approved):

jluebbe:
Please add the missing SoB.

jluebbe (approved):

PR #1345, opened 2024-02-06
add systemd catalog support for event log messages

ejoerns:
The systemd journal message catalog [1] is a proper way to enrich journal log messages with additional information.

Since the log messages emitted by the RAUC event logging/history framework have distinct MESSAGE_IDs set, we can leverage this mechanism to ship a catalog file for RAUC containing explanations for the defined messages.

This is a first attempt to provide helpful explanation texts.

Due to the limited consistency in error handling, especially in the context of installation errors, providing more specific help texts is not easily feasible at the moment.

The catalog file will be installed under `/usr/lib/systemd/catalog/rauc.catalog` unless otherwise configured.

To update the binary catalog index and thus to make the log messages appear in the journal, run

    $ journalctl --update-catalog

To see the RAUC journal log, enriched with catalog information, type

    $ journalctl -u rauc -x

[1] https://www.freedesktop.org/wiki/Software/systemd/catalog/

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -29,3 +29,5 @@ rauc_dbus_policy = configure_file(
   copy : true
 )
 install_data(rauc_dbus_policy, install_dir : dbuspolicydir)
+
+install_data('rauc.catalog', install_dir : systemdcatalogdir)
--- a/data/rauc.catalog
+++ b/data/rauc.catalog
@@ -1,0 +1,99 @@
+-- e60e0addd3454cb8b796eae0d497af96
+Subject: RAUC detected boot into @SLOT_NAME@ (@SLOT_BOOTNAME@)
+Defined-By: rauc
+Support: https://rauc.io
+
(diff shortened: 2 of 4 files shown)


data/meson.build
```diff
@@ -29,3 +29,5 @@ rauc_dbus_policy = configure_file(
   copy : true
 )
 install_data(rauc_dbus_policy, install_dir : dbuspolicydir)
+
+install_data('rauc.catalog', install_dir : systemdcatalogdir)
```
jluebbe, line 33: What will happen if `systemdcatalogdir` is empty, due to meson not finding systemd?

jluebbe (changes requested):
Looks good otherwise.

data/meson.build
ejoerns: Changed to use `catalogdir` pkconfig variable and default to `libdir / 'systemd' / 'catalog'` so it cannot be empty anymore.

jluebbe (approved):

PR #1354, opened 2024-02-21
src/install: not mark status if boot from an external device

Qingwu-Li:
if boots from an external device, maybe need to re-partition the boot device from an unknown status, in such case, mark bad or mark active may fail, should allow to installing. add the external device boot check and allow it to install.

For instance, here is the error message during installation：
  “Installation error: Failed marking slot fit.0 non-bootable:
  Failed marking slot fit.0 as bad:
  barebox backend: Failed to run barebox-state:
  Child process exited with code 1"


Qingwu-Li:
Hello maintainer，why those verify workflow takes 1 day and still no result yet?

ejoerns:
@Qingwu-Li For first-time contributors, the default is that the workflow needs to be approved and run manually.
I have triggered it now since the changes don't look suspicious.

However, this PR changes existing behavior in a way that might fit a certain use case but breaks other use cases.

When running `rauc install` (with images for bootable slots), the expected behavior is that the bootloader will boot the updated slot(s) afterwards. A typical use case for `rauc.external` would be a recovery system. You can also use it during bootstrapping, but only to populate existing partitions.
Creating partitions (or dealing with early bootstrapping) is normally out-of-scope for RAUC. What one could do instead is to use `rauc write-slot` to write images to slots as defined in the system.conf, but without the full installation logic.

Qingwu-Li:
@ejoerns Thanks, I'm using `rauc.external` for recovery, after formatting the whole eMMC, the `rauc` install was blocked due to the `barebox-state` mark bad/active, I reworked it, and now it will not impact the whole workflow just the return value, I want to install with the full installation logic.

Qingwu-Li:
fixed outof RAUC, closed.

ejoerns:
@Qingwu-Li May I ask how you fixed it?

We have a longer history of discussions about how and where RAUC fits the bootstrapping use case.

I would like to understand better if people run into problems with the current behavior, or if better documenting how proper bootstrapping handling could look like is the way to go.

a3f:
> What one could do instead is to use rauc write-slot to write images to slots as defined in the system.conf, but without the full installation logic.

This becomes cumbersome with an increasing number of artifacts.

> I would like to understand better if people run into problems with the current behavior, or if better documenting how proper bootstrapping handling could look like is the way to go.

I too think it could be nice to be able to install a bundle, but skip bootloader communication. I wouldn't base the decision though on a kernel command line parameter, but rather on a rauc parameter, e.g. `rauc install --all update.raucb` would install all slots and just omit bootloader configuration. Would this be an acceptable addition to RAUC?

Qingwu-Li:
I closed the PR because I thought you wouldn't accept it,
I fix with a workaround not really fix,  still I think it's better if we can do some change in RAUC.

Let me explain in detail the issue I encountered, I use barebox and eMMC storage:
ralated software: Barebox, dt-utils barebox-state, RAUC

Tested scenarios the issue may arise:
case 1. boot from a fresh/clean eMMC
case 2. eMMC has no barebox-status partition
case 3. boot from u-boot and want to switch to barebox dynamically

**For Barebox**
 since there is no " barebox-status" partition on storage, barebox will not create `status` devicetree node for kernel.
see [barebox-code](https://github.com/barebox/barebox/blob/02f8945abcb0e12d02b2a3186c7c4d4cac64f4e3/common/state/state.c#L650)

**For barebox-state**
since there is no `status` node, it always exited with error code 1, no matter mark bad or mark active
`Neither /aliases/state nor /state found`

see [arebox-state code](https://git.pengutronix.de/cgit/tools/dt-utils/tree/src/barebox-state.c#n391) 

**For RAUC**
It will not install the bundle since it failed mark the slot bad with barebox-state before install, as well as mark active after install,
in fact, RAUC is able to install the full bundle if barebox-state return 0.

the error message from RAUC

```
LastError: Installation error: Failed marking slot fit.0 non-bootable: Failed marking slot fit.0 as bad:  barebox backend: Failed to run barebox-state: Child process exited with code 1
```
In my opinin, it better to allow install, since we are try to recovery the systemd, but RAUC prevent to recovery even it can.

Qingwu-Li:
My workaround  is manully add state devicetree node for kernel for recovery.

PR #1355, opened 2024-02-22
document manifest syntax for image variants

jluebbe:
Also fix the rst syntax for the `statusfile` definition list entry.


ejoerns (approved):
Fixed the section underlining with force-push. Rest looks fine to me!

PR #1357, opened 2024-02-29
CodeQL Workflow: Update and Filter Results

ejoerns:
* Updates to codeql-action v3
* Filters out meson-private directories to prevent false positives.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v2
+        uses: github/codeql-action/init@v3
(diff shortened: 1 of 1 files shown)


ejoerns:
FTR: Tested in my fork where it reduced the number of code-scanning alerts to 42.

ejoerns:
A general alternative could be to switch the profile from 'security-and-quality' to 'security-extended':

https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#using-queries-in-ql-packs

This would remove notes like 'Poorly documented large function', 'Commented-out code', etc. (but also a number of probably useful hints). In my tests, this reduced the number of alerts to "5" of level "high".

PR #1364, opened 2024-03-15
add bundle format to shell and json output of rauc info

jacmet:
The bundle format is displayed in the standard "readable" output format for rauc info but it is missing from the shell and JSON formatters. Add it now.

Notice that I only added the format and not the format specific details shown by the "readable" format. I currently don't have a need for those, but we could add it if needed.

Example of the shell formatter:
```
rauc --output-format=shell info foo.raucb
RAUC_MF_COMPATIBLE='...'
RAUC_MF_VERSION='...'
RAUC_MF_DESCRIPTION='..'
RAUC_MF_BUILD='...'
RAUC_MF_HASH='085860cd4e09a942c6ed1bf4682a3b7d8cfd0574b1df2371a260ca25bf7f7f05'
RAUC_MF_FORMAT='verity'
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1125,6 +1125,9 @@ static gchar* info_formatter_json_base(RaucManifest *manifest, gboolean pretty)
 	json_builder_set_member_name(builder, "build");
 	json_builder_add_string_value(builder, manifest->update_build);
 
+	json_builder_set_member_name(builder, "format");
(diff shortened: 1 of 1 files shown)


jacmet:
@ejoerns @jluebbe Any specific reason why the bundle format is missing in the shell and JSON formats?

jluebbe:
I've force-pushed to remove the `.c` suffixes in the commit titles to align them to our convention.

PR #1367, opened 2024-03-21
bootchooser: add systemd-boot bootloader support

obbardc:


src/bootchooser.c
```diff
@@ -1456,6 +1457,51 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+		g_set_error(
+				error,
+				R_BOOTCHOOSER_ERROR,
+				R_BOOTCHOOSER_ERROR_NOT_SUPPORTED,
+				"Setting state for systemd-boot to bad is not yet supported");
+		return FALSE;
+	}
+
+	/* Set the default boot choice to this slot name */
+	/* TODO: allow the slot name to be configurable */
```
obbardc, line 1481: Should add a config option for this.

jluebbe: Shouldn't this be the slot's bootname?

jluebbe:
How would this interact with https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/, which is also implemented by systemd-boot?

jluebbe:
This would need to be rebased, as https://github.com/rauc/rauc/pull/1636 has split the bootloader backends into different files.

obbardc:
@jluebbe I rebased onto master and updated the PR. It is currently untested (apart from unit tests in RAUC); so take with a punch of salt.

PR #1368, opened 2024-03-21
Pass system version to all hooks

### LINKED ISSUE #531: Pass running system version to handler/hooks
It would be great to pass the running system version to hooks, specifically the `handlers.pre-install` hook and the `install-check` bundle hook.

This would allow the bundle and system hook scripts to do basic checking against the installed version and version to install.

Can you see any issue with passing this information to the hook scripts?
With your blessing, I will do a pull request to implement this feature.

obbardc:
Pass installed system version to bundle install-check hook script.

Fixes: #531

(This has been collecting dust downstream for a while now; time to get it upstreamed)

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -685,6 +685,7 @@ static gchar **add_system_environment(gchar **envp)
 
 	envp = g_environ_setenv(envp, "RAUC_SYSTEM_CONFIG", r_context()->configpath, TRUE);
 	envp = g_environ_setenv(envp, "RAUC_SYSTEM_VARIANT", r_context()->config->system_variant ?: "", TRUE);
+	envp = g_environ_setenv(envp, "RAUC_SYSTEM_VERSION", r_context()->system_version ?: "", TRUE);
 	envp = g_environ_setenv(envp, "RAUC_CURRENT_BOOTNAME", r_context()->bootslot, TRUE);
 	envp = g_environ_setenv(envp, "RAUC_MOUNT_PREFIX", r_context()->config->mount_prefix, TRUE);
 
@@ -906,6 +907,7 @@ static gboolean run_bundle_hook(RaucManifest *manifest, gchar* bundledir, const 
 
 	g_subprocess_launcher_setenv(launcher, "RAUC_SYSTEM_COMPATIBLE", r_context()->config->system_compatible, TRUE);
 	g_subprocess_launcher_setenv(launcher, "RAUC_SYSTEM_VARIANT", r_context()->config->system_variant ?: "", TRUE);
+	g_subprocess_launcher_setenv(launcher, "RAUC_SYSTEM_VERSION", r_context()->system_version ?: "", TRUE);
 
 	g_subprocess_launcher_setenv(launcher, "RAUC_MF_COMPATIBLE", manifest->update_compatible, TRUE);
 	g_subprocess_launcher_setenv(launcher, "RAUC_MF_VERSION", manifest->update_version ?: "", TRUE);
--- a/src/service.c
+++ b/src/service.c
@@ -612,6 +612,7 @@ static void r_on_bus_acquired(GDBusConnection *connection,
 
 	r_installer_set_compatible(r_installer, r_context()->config->system_compatible);
 	r_installer_set_variant(r_installer, r_context()->config->system_variant);
+	r_installer_set_system_version(r_installer, r_context()->system_version);
 	r_installer_set_boot_slot(r_installer, r_context()->bootslot);
 
 	r_polling_on_bus_acquired(connection);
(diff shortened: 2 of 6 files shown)


src/install.c
```diff
@@ -763,17 +763,28 @@ static gboolean run_bundle_hook(RaucManifest *manifest, gchar* bundledir, const
...
 	g_autofree gchar *hookreturnmsg = NULL;
+	g_autofree gchar *system_version = NULL;
 
 	g_assert_nonnull(manifest->hook_name);
 
 	hook_name = g_build_filename(bundledir, manifest->hook_name, NULL);
 
 	g_message("Running bundle hook %s", hook_cmd);
 
+	/* Pass the version of the currently running slot to the hook script */
```
obbardc, line 774: This (4 year old!) logic is a bit awful, but it seems to work.

src/install.c
```diff
@@ -763,17 +763,29 @@ static gboolean run_bundle_hook(RaucManifest *manifest, gchar* bundledir, const
...
+	s = find_config_slot_by_device(r_context()->config, r_context()->bootslot);
+	if (s) {
+		load_slot_status(s);
```
obbardc, line 778: This is the wrong API, from an older version of RAUC. I will fix it.

src/install.c
```diff
@@ -763,17 +763,28 @@ static gboolean run_bundle_hook(RaucManifest *manifest, gchar* bundledir, const
...
+		if (s->status)
+			system_version = g_strdup(s->status->bundle_version);
+	}
+	g_message("bootslot=%s, system_version=%s", r_context()->bootslot, system_version ?: "");
+
 	launcher = g_subprocess_launcher_new(G_SUBPROCESS_FLAGS_STDERR_PIPE);
 
 	g_subprocess_launcher_setenv(launcher, "RAUC_SYSTEM_COMPATIBLE", r_context()->config->system_compatible, TRUE);
 	g_subprocess_launcher_setenv(launcher, "RAUC_SYSTEM_VARIANT", r_context()->config->system_variant ?: "", TRUE);
+	g_subprocess_launcher_setenv(launcher, "RAUC_SYSTEM_VERSION", system_version ?: "", TRUE);
```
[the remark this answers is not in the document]
jluebbe: By now, that part was factored out to `add_system_environment`, which you could use in `run_bundle_hook`.

jluebbe:
As mentioned in https://github.com/rauc/rauc/issues/531#issuecomment-1640254105, `RAUC_SYSTEM_VERSION` would be a misleading name. Each slot remembers the bundle version which was installed to it last, so there are potentially multiple versions (one per slot).

If I understood your request correctly, you need the version of the currently booted slot. With the info suggested in the linked comment, you could do something like this:
```
eval CURRENT_VERSION=\$RAUC_SLOT_BUNDLE_VERSION_${RAUC_BOOTED_SLOT_IDX}
```

Alternatively I'd also be fine with exporting a `RAUC_BOOT_SLOT_BUNDLE_VERSION` with that value.

obbardc:
@jluebbe I changed approach here. I think it makes sense to expose this variable to hooks rather than the installed slot version, but let me know if I am missing something.

src/service.c
```diff
@@ -612,6 +612,7 @@ static void r_on_bus_acquired(GDBusConnection *connection,
 
 	r_installer_set_compatible(r_installer, r_context()->config->system_compatible);
 	r_installer_set_variant(r_installer, r_context()->config->system_variant);
+	r_installer_set_system_version(r_installer, r_context()->system_version);
```
obbardc, line 615: Not sure if `system_version` will be set at this point. Need to test it.

obbardc: well, it does seems to work when calling `rauc status` !

obbardc:
@jluebbe can you take a look, please? This solves my usecase of getting the current system version in various hooks.

ejoerns:
In contrast to the original approach, the rework (using the 'system version' rather than the 'bundle version')  looks plausible to me.

A small difference to adjacent properties like the 'compatible' or the 'variant' still is that the system version refers to the *currently booted* system, not the device in general. So it might change when updating/rebooting.

But since the variable isn't named 'booted system version' in other context either, it might be fine to expose it as 'system version'.

obbardc:
> A small difference to adjacent properties like the 'compatible' or the 'variant' still is that the system version refers to the _currently booted_ system, not the device in general. So it might change when updating/rebooting.
> 
> But since the variable isn't named 'booted system version' in other context either, it might be fine to expose it as 'system version'.

Would calling this variable something like `current system version` or `booted system version` across the codebase be more descriptive ?

jluebbe (approved):

PR #1370, opened 2024-03-27
src/config_file: add a hint that max-bundle-download-size is not used for streaming

jluebbe:
We've seen this option set in some cases even when streaming was enabled, so warn that it has no effect.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -737,6 +737,8 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 	} else if (ierror) {
 		g_propagate_error(error, ierror);
 		return FALSE;
+	} else if (ENABLE_STREAMING) {
+		g_message("Using max-bundle-download-size with streaming has no effect.");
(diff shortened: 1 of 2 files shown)


jacmet (approved):
looks sensible

src/config_file.c
```diff
@@ -737,6 +737,8 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 	} else if (ierror) {
 		g_propagate_error(error, ierror);
 		return FALSE;
+	} else if (ENABLE_STREAMING) {
+		g_message("Ignoring max-bundle-download-size, as streaming is enabled.");
```
ejoerns, line 741: Since we do not really ignore it here, maybe saying
```suggestion
		g_message("Using max-bundle-download-size with streaming has no effect.");
```
is more consistent with the code? Any opinion? Otherwise, current version would be fine for me, too.

ejoerns (approved):

PR #1374, opened 2024-03-28
fix some memory leaks in the NBD server and enable asan for the tests

jluebbe:
Most of the nbd asan tests are not running in the CI, as we're not using qemu there, yet.

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -287,13 +287,12 @@ struct RaucNBDContext {
 	gchar *tls_key; /* local file or PKCS#11 URI */
 	gchar *tls_ca; /* local file */
 	gboolean tls_no_verify;
-	GStrv headers; /* array of strings such as 'Foo: bar' */
+	struct curl_slist *headers_slist;
+	struct curl_slist *initial_headers_slist;
 
 	/* runtime state */
 	CURLM *multi;
 	gboolean done;
-	struct curl_slist *headers_slist;
-	struct curl_slist *initial_headers_slist;
 
 	/* statistics */
 	RaucStats *dl_size, *dl_speed, *namelookup, *connect, *starttransfer, *total;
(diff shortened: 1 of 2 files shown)


src/nbd.c
```diff
@@ -556,12 +556,12 @@ static void start_configure(struct RaucNBDContext *ctx, struct RaucNBDTransfer *
...
 	CURLMcode mcode = 0;
-	g_autofree guint8 *data = g_malloc(xfer->request.len);
-	g_autoptr(GVariant) v = NULL;
-	GVariantDict dict;
 
 	/* only read from the client on the first try */
 	if (!ctx->url) {
+		g_autofree guint8 *data = g_malloc(xfer->request.len);
+		g_autoptr(GVariant) v = NULL;
+		g_auto(GVariantDict) dict = G_VARIANT_DICT_INIT(NULL);
```
ejoerns, line 564: Maybe note that this changes from stack-allocated to heap-allocated and that the former variant lacked the `g_variant_dict_clear()` call?

Also, the commit contains an unrelated newline change and unrelated code move. But since the result is still clear enough to read I'd be fine with that.

src/nbd.c
```diff
@@ -951,6 +951,10 @@ gboolean r_nbd_run_server(gint sock, GError **error)
 
 	res = TRUE;
 out:
+	g_clear_pointer(&ctx.url, g_free);
```
ejoerns, line 957: That's quite close to justifying a cleanup method for `RaucNBDContext` :wink: 
But fine first of all.

ejoerns (commented):
Great to have the nbd c module asan-tested \o/

src/nbd.c
jluebbe: This is still stack-allocated, which is what `g_auto` is used for.

ejoerns: Ouh, right. My fault.

ejoerns (approved):

Issue #457, opened 2019-06-24
set options for EXT4 formatting in the manifest


embeddedrocks:
Hi,

is it possible to define options for `mkfs` on EXT4 partitions? Reading through doc and code, it doesn't seem so.

Best Regards

ejoerns:
Hi @embeddedrocks,

you're right, it is not supported yet.
But I don't see anything that speaks against implementing this.
Could you point out what your concrete use case for this is?

Best regards, Enrico

Bastian-Krause:
Assuming your mkfs.ext4 is a mke2fs symlink, you can configure it via /etc/mke2fs.conf, see `man mke2fs.conf`.

embeddedrocks:
Many thanks for your comments!

@ejoerns Use Case is to disable journal for the file system. Implementing it would be another great feature for RAUC (at least for us :-))

@Bastian-Krause Brilliant! Thanks a lot. That should do the trick by writing the file via pre-install hook and than go from there. Cheers mate! 

Best Regards,
Enrico

jluebbe:
@embeddedrocks, as options such as journal support are likely to differ between slots, I'd prefer an explicit way to configure it (like `extra-mount-opts=<options>`) in the system conf.
But maybe using the per-device defaults in mke2fs.conf is enough. I would ship it as part of the rootfs, though.

embeddedrocks:
@jluebbe many thanks for your further notes and suggestions. I agree, having an explicit way of configuring is preferred. But I would recommend configure it within the update bundle itself. This allows for changes when the next update takes place. The system conf only helps if the new system is already in place. In our case we need a procedure that takes place before the new update is installed and is shipped within the update file itself.
For now we try using the pre-install slot hook before rootfs is written (as you suggested), but we are more than happy to try a new version once that is implemented :-)

Again, many thanks and Best Regards,
Enrico

ejoerns:
I agree with @jluebbe that such an option will be useful in some cases and that a global mke2fs.conf will not reflect cases where different slots have different configuration.

Even if @embeddedrocks needs it for migration-purpose I would see this setting more in the system.conf. One always has the option to do a two-step update approach if really required.

@embeddedrocks would you like to implement such an option? Or is your pre-install solution sufficient for now and you don't need it during normal operation?

embeddedrocks:
[...]

PR #1375, opened 2024-04-01
Add extra-mkfs-opts to allow customising filesystem format options

djr-spectrummfg:
This PR adds an `extra-mkfs-opts` parameter to the system.conf file which can be used to provide extra options to `mkfs.ext4`/`mkfs.vfat`/`mkfs.ubifs` for each configured slot. 

An example configuration:
```
[slot.rootfs.0]
device=/dev/mmcblk0p5
type=ext4
extra-mkfs-opts=-i 8192 -L ROOT_A
bootname=SLOT_A

[slot.bootfs.0]
device=/dev/mmcblk0p2
type=vfat
extra-mkfs-opts=-F 32 -S 512 -n BOOT_A
parent=rootfs.0

[slot.rootfs.1]
device=/dev/mmcblk0p6
type=ext4
extra-mkfs-opts=-i 8192 -L ROOT_B
bootname=SLOT_B

[slot.bootfs.1]
device=/dev/mmcblk0p3
type=vfat
extra-mkfs-opts=-F 32 -S 512 -n BOOT_B
parent=rootfs.1
```

The meaning of the options depends on the type, see of course the man pages for `mkfs.ext4`/`mkfs.vfat`/`mkfs.ubifs` as appropriate.

See also issue #457, although note that this PR only allows setting the above in the system.conf file and not in the update manifest.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -465,6 +465,10 @@ hierarchical separator.
   See table :ref:`sec-slot-type` for a more detailed list of these different types.
   Defaults to ``raw`` if none given.
 
+``extra-mkfs-opts=<options>`` (optional)
+  Allows to specify custom filesystem creation options that will be passed to the slot's
+  ``mkfs.<type>`` call (ext4, vfat, and ubifs only).
+
 ``bootname=<name>`` (optional)
   Registers the slot for being handled by the
   :ref:`bootselection interface <bootloader-interaction>` with the ``<name>``
@@ -511,7 +515,7 @@ hierarchical separator.
   file system to an ext4 slot, i.e. if the slot has``type=ext4`` set.
 
 ``extra-mount-opts=<options>`` (optional)
-  Allows to specify custom mount options that will be passed to the slots
+  Allows to specify custom mount options that will be passed to the slot's
   ``mount`` call as ``-o`` argument value.
 
 .. _ref-logger-sections:
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -952,6 +952,7 @@ static gboolean ubifs_format_slot(RaucSlot *dest_slot, GError **error)
 
 	g_ptr_array_add(args, g_strdup("mkfs.ubifs"));
 	g_ptr_array_add(args, g_strdup("-y"));
+	r_ptr_array_addv(args, dest_slot->extra_mkfs_opts, TRUE);
 	g_ptr_array_add(args, g_strdup(dest_slot->device));
 	g_ptr_array_add(args, NULL);
 
@@ -995,7 +996,7 @@ static gboolean ext4_format_slot(RaucSlot *dest_slot, GError **error)
 {
 	GError *ierror = NULL;
 	gboolean res = FALSE;
-	g_autoptr(GPtrArray) args = g_ptr_array_new_full(4, g_free);
+	g_autoptr(GPtrArray) args = g_ptr_array_new_full(6, g_free);
 
 	g_ptr_array_add(args, g_strdup("mkfs.ext4"));
 	g_ptr_array_add(args, g_strdup("-F"));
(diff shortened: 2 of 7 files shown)


src/update_handler.c
```diff
@@ -952,6 +952,17 @@ static gboolean ubifs_format_slot(RaucSlot *dest_slot, GError **error)
 
 	g_ptr_array_add(args, g_strdup("mkfs.ubifs"));
 	g_ptr_array_add(args, g_strdup("-y"));
+
+	if (dest_slot->extra_mkfs_opts != NULL) {
+		gchar **extraopts = g_strsplit(dest_slot->extra_mkfs_opts, " ", 0);
```
jluebbe, line 957: This code is duplicated for each filesystem. I'd prefer to handle this during the config file parsing. 
Also, use `g_shell_parse_argv()`, similar to `casync_install_args` in `casync_extract()`. That will also make error detection easier.

Each format function would then only use our `r_ptr_array_addv()` helper.

docs/reference.rst
```diff
@@ -463,6 +463,10 @@ hierarchical separator.
   See table :ref:`sec-slot-type` for a more detailed list of these different types.
   Defaults to ``raw`` if none given.
 
+``extra-mkfs-opts=<options>`` (optional)
+  Allows to specify custom filesystem options that will be passed to the slot's
+  ``mkfs.type`` call (ext4, vfat, and ubifs only).
+
```
jluebbe, line 469: Is there a reason to have it here instead of close to `extra-mount-opts`?

jluebbe (changes requested):
Looks good overall, just some minor issues.

Please also add your PR description to the commit description, as it's very useful for context.

ejoerns:
I just had a look, too. My remarks would have been a subset and similar to what @jluebbe mentioned.

Since I argued for a system.conf option in #457, I'd be fine with that PR otherwise (even if maybe not primarily for the GRUB/label use case).

docs/reference.rst
djr-spectrummfg: I put it directly below `type` because you need to know the type to interpret the extra flags.

ejoerns (changes requested):
@djr-spectrummfg Thanks for the rework! CI checks still fail due to a style error and a 'mixed decleration and code' compiler error.

Also please squash the commits in the end to a single one.

djr-spectrummfg:
I will squash when the CI is happy.

djr-spectrummfg:
Looks like I have two more issues: 
- segfault caused by dereferencing null pointer in `r_ptr_array_addv` in the case where there are no extra opts
- CI failing builds on ubuntu 20.04 + debian bullseye due to glib version not having `g_assert_cmpstrv`

test/config_file.c
```diff
@@ -943,6 +955,45 @@ extra-mount-opts=ro,noatime\n";
...
+	gboolean res;
+	g_autofree gchar* pathname = NULL;
+	RaucSlot *slot = NULL;
+
+	const gchar *cfg_file = "\
+[system]\n\
+compatible=FooCorp Super BarBazzer\n\
+bootloader=barebox\n\
+mountprefix=/mnt/myrauc/\n\
+activate-installed=false\n\
```
ejoerns, line 972: This is a quite specific configuration and unrelated to the test, thus better remove it to make test purpose clearer.
```suggestion
```

ejoerns (commented):
Looks good overall. Just a few comments left from my side.

test/config_file.c
```diff
@@ -78,6 +79,7 @@ force-install-same=false\n\
 description=Root filesystem partition 1\n\
 device=/dev/rootfs-1\n\
 type=ext4\n\
+extra-mkfs-opts= \n\
```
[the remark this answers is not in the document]
djr-spectrummfg: It was

test/config_file.c
```diff
@@ -165,9 +170,16 @@ install-same=false\n";
 	slot = g_hash_table_lookup(config->slots, "appfs.1");
 	g_assert_cmpstr(slot->name, ==, "appfs.1");
 	g_assert_cmpstr(slot->description, ==, "Application filesystem partition 1");
-	g_assert_cmpstr(slot->device, ==, "/dev/appfs-1");
```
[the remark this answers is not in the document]
djr-spectrummfg: Accident, will undo

ejoerns (approved):

jluebbe (approved):

PR #1376, opened 2024-04-04
test/rauc.t: test faketime before relying on it

ukleinek:
On some platforms faketime doesn't work as intended due to not supporting the libc functions that provide 64bit time_t on 32bit architectures.

This problem already made rauc fail to build in Debian unstable, see [Debian bug #1032177 ](https://bugs.debian.org/1032177). With the expectation that the problem will also affect rauc mainline at some time in the future, precautionarily check faketime being functional before using it.

### DIFF
--- a/test/rauc.t
+++ b/test/rauc.t
@@ -158,7 +158,10 @@ test -f ${SOFTHSM2_MOD} &&
   test_set_prereq PKCS11
 
 # Prerequisite: faketime available [FAKETIME]
+# On some platforms faketime is broken, see e.g. https://github.com/wolfcw/libfaketime/issues/418
+# Only use it if it works for date
 faketime "2018-01-01" date &&
+faketime "2018-01-01" date -R | grep "Jan 2018" &&
   test_set_prereq FAKETIME
 
 # Prerequisite: grub-editenv available [GRUB]


test/rauc.t
```diff
@@ -158,7 +158,10 @@ test -f ${SOFTHSM2_MOD} &&
   test_set_prereq PKCS11
 
 # Prerequisite: faketime available [FAKETIME]
+# On some platforms faketime is broken, see e.g. https://github.com/wolfcw/libfaketime/issues/418
+# Only use it if it works for date
 faketime "2018-01-01" date &&
+faketime "2018-01-15" date -R | grep "Jan 2018" &&
```
jacmet, line 164: is the change of day done on purpose? To counter negative timezones? That probably is worth a comment

ukleinek: Yes, that was the idea. I find that obvious enough that I'd not add a comment, but I won't oppose strongly. Adding a comment for a single command in a &&-list is a bit ugly.

jacmet: Maybe just add a note about it in the commit message?

ukleinek: Ah, actually that's wrong, 2018-01-01 *does* work even with negative timezones, it just emits
```    
Mon, 01 Jan 2018 00:00:00 -0x00
```
then. I'll use that and add a note about this being safe in the commit log then.

jluebbe (approved):

jacmet (approved):

ejoerns (approved):

PR #1379, opened 2024-04-05
test/Dockerfile: add acl to control access to coverage files in qemu

jluebbe:
With some further changes, this will allow us to collect coverage data from the NBD helper as well.


ejoerns:
> With some further changes, this will allow us to collect coverage data from the NBD helper as well.

Should that be part of the commit message? :smirk:

jluebbe:
As a preview you can look at: https://app.codecov.io/github/jluebbe/rauc/commit/d23116e3a6a0fe370a7fa4bc5cbaa53efedb3ea8/blob/src/nbd.c

ejoerns (approved):

PR #1380, opened 2024-04-05
fix leaks discovered via shell based tests

jluebbe:
So far, we've only run the C-based test suite with address sanitized and fixed the resulting issues.

This PR fixes issues detected by running `./qemu-test` with a asan enabled and prepares for enabling these checks in CI as well.


ejoerns (approved):
Great work! :+1:

PR #1381, opened 2024-04-05
qemu-test-init: fix coverage for nbd helper

jluebbe:
As the nbd helper is running as a different user, it cannot write to the gcda files (which belong to the host user id). Instead of trying to solve this via umask or separate GCOV_PREFIXs per user with merging as post-processing, we use a /tmp/cov with a default ACL.

We need to create and configure that directory and then save the coverage data to the host before shutting down.


ejoerns (approved):
Fixed typo in commit title. Look good otherwise.

Overall coverage back at above 80% :+1:

PR #1384, opened 2024-04-08
manifest: ensure that size always defaults to -1 in RaucImage

jluebbe:
A size of -1 indicates that no `size` key was given in the manifest. Instead of hardcoding that in everywhere, create a `r_new_image()` constructor function and use it instead.

### DIFF
--- a/include/checksum.h
+++ b/include/checksum.h
@@ -14,6 +14,7 @@ GQuark r_checksum_error_quark(void);
 typedef struct {
 	GChecksumType type;
 	gchar *digest;
+	/* image size (-1 indicates no size in manifest) */
 	goffset size;
 } RaucChecksum;
 
--- a/include/manifest.h
+++ b/include/manifest.h
@@ -188,6 +188,12 @@ GVariant *r_manifest_to_dict(const RaucManifest *manifest)
 G_GNUC_WARN_UNUSED_RESULT;
 
 /**
+ * Creates a rauc image
+ */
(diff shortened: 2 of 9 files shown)


ejoerns:
> A size of -1 indicates that no `size` key was given in the manifest. Instead of hardcoding that in everywhere, create a `r_new_image()` constructor function and use it instead.

This does not replace any hard-coded setting of `checksum.size` as far as I can see (what the commit message suggests it does). So does this fix potential issues?

jluebbe:
> > A size of -1 indicates that no `size` key was given in the manifest. Instead of hardcoding that in everywhere, create a `r_new_image()` constructor function and use it instead.
> 
> This does not replace any hard-coded setting of `checksum.size` as far as I can see (what the commit message suggests it does). So does this fix potential issues?

I have some patches in my stack that perform stricter checking of input manifests. They enforce that no checksum or size is set, so a default constructed `RaucImage` would have size 0 and fail (especially in the test cases).

ejoerns:
So with mentioning the inconsistency in the commit message I'd be happy ;)

jluebbe:
@ejoerns I've updated the commit message.

ejoerns (approved):
I just wanted to add that using `r_rauc_image_free()` or `r_image_free()` would better fit our naming conventions, but then noted that we have `r_free_image()` already. Thus we should better rename these together at any later point.

PR #1392, opened 2024-04-22
Fix slot skipping (install-same) handling twice

ejoerns:
This fixes slot skipping handling on two levels.

With v1.10, slot skipping got broken while with v1.11 the slot skipping *detection* got broken (and thus hides the previously introduced issue).

The error that will be raised by v1.10 is:

```
Not enough substeps: update_slots (12/20)
```

Fix both and add a minimal test case that triggers an installation twice.

### DIFF
--- a/test/install.c
+++ b/test/install.c
@@ -176,6 +176,23 @@ adaptive=adaptive-test-method;block-hash-index";
 	fixture_helper_set_up_bundle(fixture->tmpdir, manifest_file, &data->manifest_test_options);
 }
 
+static void install_fixture_set_up_slot_skipping(InstallFixture *fixture,
+		gconstpointer user_data)
+{
+	InstallData *data = (InstallData*) user_data;
+	const gchar *manifest_file = "\
+[update]\n\
+compatible=Test Config\n\
+\n\
+[image.rootfs]\n\
+filename=rootfs.ext4";
+
+	fixture->tmpdir = g_dir_make_tmp("rauc-XXXXXX", NULL);
+
+	fixture_helper_set_up_system(fixture->tmpdir, "test/test-install-same-false.conf");
+	fixture_helper_set_up_bundle(fixture->tmpdir, manifest_file, &data->manifest_test_options);
+}
+
 static void install_fixture_set_up_system_conf(InstallFixture *fixture,
 		gconstpointer user_data)
 {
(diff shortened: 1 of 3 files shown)


test/install.c
```diff
@@ -1124,6 +1141,20 @@ static void install_test_bundle(InstallFixture *fixture,
...
 
+static void install_test_bundle_twice(InstallFixture *fixture,
+		gconstpointer user_data)
+{
+	/* First run */
+	install_test_bundle(fixture, user_data);
+
+	/* Manual partial cleanup */
+	test_umount(fixture->tmpdir, "mnt");
+	test_rm_tree(fixture->tmpdir, "mnt");
```
jluebbe, line 1152: `install_test_bundle()` should already handle the umount. Why is this needed?

test/install.c
```diff
@@ -1124,6 +1141,20 @@ static void install_test_bundle(InstallFixture *fixture,
...
+
+	/* Second run */
+	install_test_bundle(fixture, user_data);
```
jluebbe, line 1155: Can we check that the install was skipped? Perhaps by looking at the slot status data?

Also we should test the inverse: that the bundle is installed again if install-same is true.

test/test-install-same.conf
```diff
@@ -0,0 +1,45 @@
+# testsuite system configuration
```
jluebbe, line 1: This should perhaps be named `test/test-install-same-false.conf`

test/test-install-same.conf
```diff
@@ -0,0 +1,45 @@
...
+[slot.appfs.0]
+device=images/appfs-0
+type=ext4
+parent=rootfs.0
+
+[slot.appfs.1]
+device=images/appfs-1
+type=ext4
+parent=rootfs.1
+
```
jluebbe, line 45: If these are not needed, please remove them.

jluebbe (changes requested):

test/install.c
ejoerns: In fact, only the removal of the folder is needed. But this could be done by `install_test_bundle()` itself.

test/install.c
ejoerns: I don't see any way to check this with manageable effort at the moment or in the context of this fix.

Since we use the same bundle for installing, the only way to detect if an installation happened is by having a look at the time stamps in the status file. And I am not sure if enough time is spent between calls.

Anyway, adding support for parsing the status file here is quite complex from the layer we operate on.

With the current check, at least the older issue would have been detected. The other would have not.

test/test-install-same.conf
ejoerns: Will rename

PR #1395, opened 2024-04-24
src/nbd: handle closed socket during shutdown better

jluebbe:
As the nbd server can exit by itself, we shouldn't rely on the socket still being open at this point. Just log the failure cause and continue by closing the socket on the client side.


src/nbd.c
```diff
@@ -1278,8 +1278,9 @@ gboolean r_nbd_stop_server(RaucNBDServer *nbd_srv, GError **error)
 		request.magic = GUINT32_TO_BE(NBD_REQUEST_MAGIC);
 		request.type = GUINT32_TO_BE(NBD_CMD_DISC);
 		request.len = 0;
-		if (!r_write_exact(nbd_srv->sock, (guint8*)&request, sizeof(request), NULL))
-			g_error("failed to send nbd disconnect request");
+		if (!r_write_exact(nbd_srv->sock, (guint8*)&request, sizeof(request), &ierror)) {
```
ejoerns, line 1281: `ierror` needs to be cleared after printing it. Otherwise, it would be overridden on next usage (`g_subprocess_wait_check()`)

ejoerns (dismissed):

PR #1398, opened 2024-04-30
Fix PKCS11 URI handling in /etc/rauc/system.conf

dhollister0:
- RAUC should not try to resolve a PKCS11 URI like a file path.  This would happen if a URI was given as the key in the [encryption] section of /etc/rauc/system.conf.
- If a PKCS11 URI is detected in the resolve function, just return it back to the caller.
- I think this only affected a URI given in system.conf.  A URI passed in on the command line was correctly unmodified - which is why the unit tests related to PKCS11 were passing.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -170,6 +170,9 @@ gchar *resolve_path(const gchar *basefile, const gchar *path)
 	if (path == NULL)
 		return NULL;
 
+	if (g_str_has_prefix(path, "pkcs11:"))
+		return g_strdup(path);
+
 	if (g_path_is_absolute(path))
 		return g_strdup(path);
 


jluebbe:
I've force pushed to change the commit title to our convention.

jluebbe (approved):

ejoerns:
> * I think this only affected a URI given in system.conf.  A URI passed in on the command line was correctly unmodified - which is why the unit tests related to PKCS11 were passing.

Indeed, since only for `system.conf` RAUC attempts to resolve the relative paths relative to the `system.conf` file.

@dhollister0 Thanks for fixing!

ejoerns (approved):

PR #1400, opened 2024-05-02
semantic version comparison

js731ca:
This PR adds `min-version` and `max-version` configuration options to the system.conf, that can be set to a version triplet Major.Minor.Patch. If now bundles are versioned in a way that follows the [semantic versioning scheme](https://semver.org/) - e.g. `1.2.3-g35f26fd` - the bundle version can be compared against these configured limits, and the installation will be (gracefully) refused should the bundle version not fall within set limits. 

----
- What do you use the feature for?

  updates can introduce breaking changes, which could render a device inoperable should a downgrade happen; with this feature a breaking change could advance the `min-version` and from then on refuse to install "older" bundles.

- How does RAUC benefit from the feature?

  new feature :-D
  similar feature to the compatible check, but orthogonal

- How did you verify the feature works?
  
  added tests were feasible; manually on project/target hardware

- If hardware is needed for the feature, which hardware is supported and which
  hardware did you test with?

  no special requirements, just a hardware platform that uses rauc


js731ca:
v1.1: address issues raised by`style` and `stable (debian:testing)`

js731ca:
v1.2: second round of PR check fixes :-)
addressing(?) `CodeCL` and `stable (debian:testing)`

src/utils.c
```diff
@@ -743,3 +743,104 @@
...
+		if (pre_release != NULL) {
+			*pre_release = g_strdup(pre_release_pos + 1);
+			if (*pre_release == NULL)
+				goto out;
+		}
+		*pre_release_pos = '\0';
+	}
+
+	**version_parts = g_strsplit(version_copy, ".", 3);
+	if (version_parts == NULL)
```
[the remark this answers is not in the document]
js731ca: reworked with v1.2

jluebbe:
Thanks for your contribution!

I understand your "protect against downgrades" use-case for a `min-version` in the `system.conf`. So far, we've pointed users other ways this could be done:

- the `pre-install` handler in `system.conf`
- the `install-check` hook in the manifest
- rotating signing keys

Doing the bundle version comparison in RAUC itself would make it more explicit (which is good).

Before reviewing the code, I'd like to discuss some high-level questions:

- What's the use-case for a `max-version` in the `system.conf`? I'd need to know which future version will be incompatible with the version I'm currently developing, which seems difficult.
- `min-version` seems a bit unspecific. Maybe `min-bundle-version` would be better? Not sure though...
- Are there valid reasons where a bundle would have to override the version check?

Even if having a `max-version` in the `system.conf` may not be useful, there is the related use-case of "require intermediate updates". In that case, a *bundle* would have a `min-version` or `min-system-version`. The issue here is that RAUC doesn't really have concept of a system version (only the version of a bundle that was used to install to a slot).

A possible way to solve that would be to use the [system-info handler](https://rauc.readthedocs.io/en/latest/reference.html#handlers-section) to provide the system version. Another way would be to define that the bundle version of the booted slot is the system version. But what about systems installed from a disk image?

js731ca:
> Thanks for your contribution!

:-D

> I understand your "protect against downgrades" use-case for a `min-version` in the `system.conf`. So far, we've pointed users at other ways this could be done:
> 
> * the `pre-install` handler in `system.conf`
> * the `install-check` hook in the manifest
> * rotating signing keys

our stop-gap solution so far is a `pre-install` handler, but we found that it is somewhat lacking in status feedback: the (manual) `rauc install bundle` just fails, and one has to look into the service logs/journal to piece together why - doing this in rauc is clearer: versions mismatch, aborting bundle install

rotating signing keys puts (a lot of) burden on the PKI infrastructure... which is not always feasible/easy 

> Doing the bundle version comparison in RAUC itself would make it more explicit (which is good).
> 
> Before reviewing the code, I'd like to discuss some high-level questions:
> 
> * What's the use-case for a `max-version` in the `system.conf`? To use it, I'd need to know which future version will be incompatible with the version I'm currently developing, which seems difficult.

none so far, added it mostly for symmetry/completeness sake. 
i could picture someone doing hardware development and having a scheme where the hardware revision matches the major revision? that way a min 4.0.x and max 4.999.999 could make sense to filter out bundles that have the same compatible over hardware revisions...

> * `min-version` seems a bit unspecific. Maybe `min-bundle-version` would be better? Not sure though...

the way we're using it is system-version == bundle version; so it would depend on the definition

> * Are there valid reasons where a bundle would have to override the version check?

are you referring to the `--ignore-version-limit` parameter? that is meant for bootstrapping/manual intervention

> Even if having a `max-version` in the `system.conf` may not be useful, there is the related use-case of "require intermediate updates". In that case, a _bundle_ would have a `min-version` or `min-system-version`. The issue here is that RAUC doesn't really have concept of a system version (only the version of a bundle that was used to install to a slot).
> 
> A possible way to solve that would be to use the [system-info handler](https://rauc.readthedocs.io/en/latest/reference.html#handlers-section) to provide the system version. Another way would be to define that the bundle version of the booted slot is the system version. But what about systems installed from a disk image?

oooh, that the slot status (= "installed bundle version") is empty after an initial bringup from a diskimage is sth i missed - thnx!
as mentioned "the way we use it" is bye defining bundle-version = system version - since a bundle installs a rootfs/system of some known version

js731ca:
> But what about systems installed from a disk image?

on rethinking: those should also be ok, since they bring the `system.conf` with it; isn't the usual flow that a system is initially brought up from a disk image, and then updated via bundles?

or what was your train-o-thought that led to the above question?

js731ca:
v1.3: third round: addressing `style`->`codespell check` findings

ejoerns:
> our stop-gap solution so far is a `pre-install` handler, but we found that it is somewhat lacking in status feedback: the (manual) `rauc install bundle` just fails, and one has to look into the service logs/journal to piece together why - doing this in rauc is clearer: versions mismatch, aborting bundle install

Note that for this purpose, the 'install-check' handler has an explicit 'interface' for user-visible feedback:

> If available, RAUC will use the last line printed to standard error by the hook executable as the rejection reason message and provide it to the user

It's not ideal, but better than digging in logs :smirk: 

https://rauc.readthedocs.io/en/latest/using.html#install-hooks

jluebbe:
> > 6 failing and 10 successful checks
> 
> looking through those logs... can't see how those should be fixed :-S are they false positives?

Let's focus on the design questions for now. Then we can look at the implementation details. :)

js731ca:
> A different question: SemVer requires the version core to consist of exactly three parts. We've seen projects use date-based versions (such as `2024.05`, `2024.1` or similar). These would be rejected by your parser. Perhaps it would be useful in the context of RAUC to relax this requirement?

yes it would reject those, since it compares a triplet `guint[3]`.

> You've only implemented comparison for the version core, any specific reason for that?

the version core is `<numeric identifiers>` only = straight forward to compare,
but "pre-build" and "build" are `<alphanumeric identifier>`;  how would one tell which is less-then another?

jluebbe:
> > Hmm. I don't think `min-version`/`max-version` have symmetric use-cases. To me, `min-version` is about prohibiting downgrades. In that case, the running image is newer and the developer has the necessary information to set the correct limit.
> > For `max-version`, I only see the use-case of prohibiting updates to versions which no longer contain the necessary migration code from old installed versions or for which dependencies required during installation (e.g. by hooks in the bundle) are not available in old versions. In that case, the developer would know while creating the new bundle what the oldest supported installed version would be and accordingly, it should be part of the bundle manifest. To update from an older version, an intermediate update would need to be installed in that case. (That's what I meant with "require intermediate updates" above.)
> 
> we're starting to speculate possible use-cases :-D my thought was "if there is a lower-bound check; $someone will want an upper-bound check" - and since both are optional, it was easy to implement and there if $someone needs it

Any optional features still has costs, in maintenance, interactions with other features and complexity that needs to be documented. RAUC is opinionated and intended to discourage "incorrect" usage, so any new feature needs to have a clear use and be consistent with the rest of the design. That's not always easy. :)

> > > > * Are there valid reasons where a bundle would have to override the version check?
> > > 
> > > 
> > > are you referring to the `--ignore-version-limit` parameter? that is meant for bootstrapping/manual intervention
> > 
> > 
> > No, the `--ignore-version-limit` parameter was clear.
> > I meant that with the `install-check` approach, you could later decide to produce a bundle which allows a downgrade, as the check is implemented in code shipped in the bundle itself. Think about this scenario: You roll out a version `3.0.0` to the field which also sets `min-version=3.0.0`. Then it turns out that it contains a regression from `2.9.0`. With your approach, you'd need to create a bundle with version `3.0.1` but which actually contains the `2.9.0` images. This might be undesirable, if the bundle version is also used in other contexts.
> 
> or use the `--ignore-version-limit` to downgrade again? come to thing of that... the switch is for the commandline, but what about an install triggered from an application via dbus? hm...

A CLI or D-Bus switch wouldn't help, as these are normally not under the control of the person installing the update (except in a development scenario).

For now, I'd tend to add a warning to the documentation mentioning this potential issue.

> > That only relevant for the "require intermediate updates" case and not for "protect against downgrades". We don't need to implement both cases at the same time, but we need to make sure to have clear semantics which work for both.
> > You're right, that's the usual flow. I was thinking about: The factory originally installed `1.0.0` and later `2.0.0` via disk images. Some time passes and you now want to release a `4.0.0` update bundle, but can't support a direct upgrade from `1.0.0`. You could have a hypothetical `min-system-version=2.0.0` in the bundle manifest, but RAUC on a device fresh from the factory would have no information in the slot status. So it couldn't reject installation on the `1.0.0` devices and allow it on `2.0.0` devices.
> > So the question becomes: How do we get an "installed bundle version" on devices which have never installed a bundle. If that's even a valid concept. :)
> 
> i was wrong there: the comparison happens between system.conf <> new bundle; the rauc.status (which is initially empty, and only after bundle installation has versions) does not factor in

I'm confident enough that we can handle these use-cases separately, so we don't need to discuss he "require intermediate updates" case here.

js731ca:
v2: changes included, based on the discussions so far:
- rename('min-version', 'min-bundle-version')
- remove('max-version')
- include pre_release in comparison (and add more tests for that)
- relax version parsing to allow: '2020.02' or just '321'

js731ca:
v2.1: fixing `style` findings once more
rechecking `build` errors on the semver_parse_test: false positive? seems ok locally

js731ca:
v2.2: another `style` fix, and hopefully a fix for the failing `build` test run

test/utils.c
```diff
@@ -256,6 +256,138 @@ static void fakeroot_test(void)
...
 }
 
+static void semver_parse_test(void)
+{
+	g_autofree gchar *version_string;
+	g_autofree guint version[3];
+	g_autofree gchar *pre_release;
+	g_autofree gchar *build;
+
+	version_string = g_strdup("1.2.3");
```
jluebbe, line 266: These leak memory and will cause address sanitizer errors. Take a look at our `replace_strdup` helper and/or use const string literals if modification is not needed.

src/utils.c
```diff
@@ -743,3 +743,169 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+		is_num_a = g_ascii_isdigit(pre_fields_a[i][0]);
+		is_num_b = g_ascii_isdigit(pre_fields_b[i][0]);
+
+		if (is_num_a && is_num_b) {
+			//	compare numerically
+			num_a = g_ascii_strtoull(pre_fields_a[i], NULL, 10);
+			num_b = g_ascii_strtoull(pre_fields_b[i], NULL, 10);
+			if (num_a < num_b) {
+				g_strfreev(pre_fields_a);
+				g_strfreev(pre_fields_b);
```
jluebbe, line 879: Avoid these `g_strfreev` lines by using `g_auto(GStrv)` the result of `g_strsplit`.

That way, you can also avoid the `result` variable.

src/utils.c
```diff
@@ -743,3 +743,169 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+
+gboolean semver_less_equal(const guint version_a[3], gchar *pre_release_a, const guint version_b[3], gchar *pre_release_b)
+{
+	gchar **pre_fields_a = NULL;
+	gchar **pre_fields_b = NULL;
+	gboolean is_num_a = FALSE;
+	gboolean is_num_b = FALSE;
+	guint num_a = 0;
+	guint num_b = 0;
+	gint cmp = 0;
```
jluebbe, line 842: Move these declarations into the loop scopes.

src/utils.c
```diff
@@ -743,3 +743,169 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+
+	i = 0;
+	while (pre_fields_a[i] != NULL && pre_fields_b[i] != NULL) {
+		is_num_a = g_ascii_isdigit(pre_fields_a[i][0]);
+		is_num_b = g_ascii_isdigit(pre_fields_b[i][0]);
+
+		if (is_num_a && is_num_b) {
+			//	compare numerically
+			num_a = g_ascii_strtoull(pre_fields_a[i], NULL, 10);
+			num_b = g_ascii_strtoull(pre_fields_b[i], NULL, 10);
```
jluebbe, line 876: This can overflow on 32-bit architectures. All the core versions should probably be `guint64` to match the return value of `g_ascii_strtoull`.

src/utils.c
```diff
@@ -743,3 +743,169 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+	if (version_parts[1] != NULL && version_parts[2] == NULL) {
+		version_parts[2] = version_parts[1];
+		version_parts[1] = version_parts[0];
+		version_parts[0] = NULL;
+	}
+
+	for (int i = 0; i < 3; i++) {
+		if (version_parts[i] != NULL) {
+			// Note that negative numbers are already implicitly excluded due to the search for the '-pre_release'
+			version_core[i] = g_ascii_strtoll(version_parts[i], NULL, 10);
```
jluebbe, line 813: Why parse it as a signed number?

I'd suggest using `g_ascii_string_to_unsigned()` for parsing there and below, perhaps with a small wrapper which always allows full 64-bit range decimal numbers. That make the error handling much clearer and produces better error messages for the user.

src/utils.c
```diff
@@ -743,3 +743,169 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+	}
+
+	// compare	dot-separated fields of pre-release identifiers
+	pre_fields_a = g_strsplit(pre_release_a, ".", 0);
+	pre_fields_b = g_strsplit(pre_release_b, ".", 0);
+
+	i = 0;
+	while (pre_fields_a[i] != NULL && pre_fields_b[i] != NULL) {
+		is_num_a = g_ascii_isdigit(pre_fields_a[i][0]);
+		is_num_b = g_ascii_isdigit(pre_fields_b[i][0]);
```
jluebbe, line 871: The SemVer spec requires that "Identifiers consisting of only digits are compared numerically", but this only checks the first character. Using `g_ascii_string_to_unsigned` here could solve that, as the different cases a reported via `GErrors` that can be handled explicitly.

src/utils.c
```diff
@@ -743,3 +743,169 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+		if (version_parts[i] != NULL) {
+			// Note that negative numbers are already implicitly excluded due to the search for the '-pre_release'
+			version_core[i] = g_ascii_strtoll(version_parts[i], NULL, 10);
+			// handle g_ascii_strtoll returning zero on a conversion error
+			if ((version_core[i] == 0) && (g_strcmp0(version_parts[i], "0") != 0)) {
+				g_strfreev(version_parts);
+				goto out;
+			}
+		} else
+			version_core[i] = 0;
```
jluebbe, line 820: Use braces here as well to symmetry, because the true-arm needs them.

src/utils.c
```diff
@@ -743,3 +743,169 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+	version_parts = g_strsplit(version_copy, ".", 3);
+	// Note that g_strsplit returns an empty array when splitting: ""
+	if (version_parts == NULL)
+		goto out;
+	// and NULL terminates arrays it creates
+	if (version_parts[0] == NULL)
+		goto out;
+	// only on component -> shift right
+	if (version_parts[1] == NULL) {
+		version_parts[2] = version_parts[0];
```
jluebbe, line 800: This would write beyond the end of `version_parts`.

Why is that needed at all? I would have expected '1.0.0' < '2'.

There should be test cases for shorter versions.

docs/reference.rst
```diff
@@ -60,6 +60,17 @@ Example configuration:
...
 
+``min-bundle-version`` (optional)
+  An optional user-defined version string that follows the
+  `semantic versioning scheme <https://semver.org>`_.
+  The 'version core' and '-pre_release' parts from the ``version`` string in the
+  update manifest is compared against. This allows to roll out updates with some
+  (breaking) change, which set the ``min-bundle-version`` accordingly, and
+  through that disallow the installation of prior version bundle - which would
+  render the device inoperable.
+
```
jluebbe, line 71: You should add a comment that shorted version cores are allowed as well.

jluebbe (commented):
I'm not convinced that the split into `semver_parse` and `semver_less_equal` is useful.

`semver_less_equal` could encounters overflows while parsing the pre-release parts, so it would need to be able to return `GErrors` as well.

What do you think about keeping the version as a string in both `RaucConfig` and `RaucManifest`, getting rid of `semver_parse` and handling the core version by iterating through a `GStrv` in `semver_less_equal`, similar to how you implemented the pre-release parsing?

During config parsing, one could use `semver_less_equal("0", config_version, &ierror) == TRUE` to check if the given `min-bundle-version` is sensible.

js731ca:
> I'm not convinced that the split into `semver_parse` and `semver_less_equal` is useful.
> 
> `semver_less_equal` could encounters overflows while parsing the pre-release parts, so it would need to be able to return `GErrors` as well.
> 
> What do you think about keeping the version as a string in both `RaucConfig` and `RaucManifest`, getting rid of `semver_parse` and handling the core version by iterating through a `GStrv` in `semver_less_equal`, similar to how you implemented the pre-release parsing?
> 
> During config parsing, one could use `semver_less_equal("0", config_version, &ierror) == TRUE` to check if the given `min-bundle-version` is sensible.

thats an idea! basically pushing the whole handling (=parsing+comparing) into one place, and back to the point in time bundles are installed (or created...) -- i'll rework the PR towards that; and also pickup all the other comments&findings = thnx!

src/config_file.c
```diff
@@ -632,6 +632,24 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 		g_propagate_error(error, ierror);
 		return FALSE;
 	}
+
+	data = key_file_consume_string(key_file, "system", "min-bundle-version", &ierror);
```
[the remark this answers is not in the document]
js731ca: fixed with v3

src/manifest.c
```diff
@@ -198,6 +198,19 @@ static gboolean parse_manifest(GKeyFile *key_file, RaucManifest **manifest, GErr
...
+	if (!semver_parse(raucm->update_version, raucm->update_version_core, &raucm->update_version_prerelease, NULL)) {
+		g_ptr_array_add(raucm->warnings, g_strdup(
+				"Warning: The manifest specifies a version not following the semantic versioning scheme,"));
+		g_ptr_array_add(raucm->warnings, g_strdup(
+				"  version comparison and checks won't be possible against:"));
+		g_ptr_array_add(raucm->warnings, g_strdup(raucm->update_version));
+		g_ptr_array_add(raucm->warnings, g_strdup(
+				"  See https://rauc.readthedocs.io/en/latest/reference.html#sec-system-settings for details.'"));
+	}
+
```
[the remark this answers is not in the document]
js731ca: removed with v3

test/utils.c
js731ca: changed to replace_strdup with v3

src/utils.c
js731ca: ACK, done with v3

src/utils.c
js731ca: ACK, done with v3

src/utils.c
js731ca: fixed in v3

src/utils.c
js731ca: not quite - this only checks if a number starts here = do we need to compare numerically or lexicographically; later the "whole" number is parsed an compared

src/utils.c
js731ca: fixed with v3

src/install.c
```diff
@@ -550,6 +550,37 @@ static gboolean verify_compatible(RaucInstallArgs *args, RaucManifest *manifest,
 	}
 }
 
+static gboolean check_version_limits(RaucInstallArgs *args, RaucManifest *manifest, GError **error)
+{
+	if (args->ignore_version_limit)
+		return TRUE;
+
+	// when no min-bundle-version is set in the system.conf the version is {0,0,0}
+	if ((r_context()->config->system_min_bundle_version_core[0] > 0) ||
```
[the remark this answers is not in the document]
js731ca: this shifted away with v3

js731ca:
v3: rebase unto master
- keep versions as strings, and only do a comparison in the installation step - if a minimum is set in system.conf
- add GErrors
- address review comments on variable handling
- extend tests to cover more cases

src/utils.c
```diff
@@ -746,3 +746,181 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+	g_autofree gchar *version_copy = g_strdup(version_string);
+	gchar *build_pos = NULL;
+	gchar *pre_release_pos = NULL;
+	g_auto(GStrv) version_parts = NULL;
+
+	if (pre_release != NULL)
+		*pre_release = NULL;
+
+	if (build != NULL)
+		*build = NULL;
```
jluebbe, line 761: This will leak memory, if this already points to a string, as in the tests.

Take a look at https://github.com/rauc/rauc/blob/master/src/mount.c#L115, there we use `g_return_val_if_fail` to check that the output parameter is either `NULL` or points to `NULL`. Later on success, `g_steal_pointer` is used to write the result back if needed: https://github.com/rauc/rauc/blob/master/src/mount.c#L221. On error, the string is freed via `g_autofree`.

src/utils.c
jluebbe: Hmm, if we have `1foo` in `pre_fields_a[i]` and a valid number in `pre_fields_b[i]`, `is_num_a` and `is_num_b` would be `TRUE`. Then we'd use the "compare numerically" branch, where `g_ascii_strtoull` would ignore the `foo` suffix.

I read the spec as in this case, they are not "consisting of only digits", so they need to be compared in ASCII order.

jluebbe (changes requested):

src/utils.c
js731ca: merde! you're absolutely right... regex to the rescue ;-)

jluebbe: Why not the approach with `g_ascii_string_to_unsigned` as I suggested above?

js731ca: am/was a bit reluctant since the rauc codebase does not use it (yet...) and that the min/max have to be set, only to end up at 0/MAXUINT; but the error return like you mentioned could be worth it - giving it a shot (-:

jluebbe: It's already included in the minimum glib version we required, so I don't see a problem with using it.

src/utils.c
js731ca: so this way a pointer *has* to be passed in - or its an error
but here my intention is to give the user the option to pass in NULL as in "don't care about that returned value"

jluebbe: Ah, right. The it should look like the `error` checks we have in many functions:
```g_return_val_if_fail(pre_release == NULL || *pre_release == NULL, FALSE);```

File: src/utils.c (hunk 7 above)
Comment (js731ca) [reply src/utils.c]: ACK, switched to string_to_unsigned with v4

File: src/utils.c (hunk 9 above)
Comment (js731ca) [reply src/utils.c]: > BTW, you could enable GitHub Actions on your fork of the repo and get CI results for any branch, even those which are not part of a PR, to get feedback quickly.

thanks for the hint! that just allowed some quick iterations to patch some leaks :-)

File: src/utils.c [hunk 15]
```diff
@@ -746,3 +746,181 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+		return FALSE;
+	}
+	if (!semver_parse(version_string_b, version_core_b, &pre_release_b, NULL, &ierror)) {
+		g_clear_error(&ierror);
+		g_set_error(error,
+				R_UTILS_ERROR,
+				R_UTILS_ERROR_SEMVER_PARSE,
+				"Failed to parse semantic version B for comparison.");
+		return FALSE;
+	}
```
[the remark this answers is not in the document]
js731ca: fixed with v4

src/utils.c
```diff
@@ -746,3 +746,181 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+
+out:
+	if (pre_release && *pre_release)
+		g_free(*pre_release);
+	if (build && *build)
+		g_free(*build);
+	g_set_error(error,
+			R_UTILS_ERROR,
+			R_UTILS_ERROR_SEMVER_PARSE,
+			"Failed to parse semantic version.");
```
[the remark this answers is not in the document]
js731ca: v4: added more&multiple errors that include the `version_string` and more

src/utils.c
js731ca: v4 works like i had initially intended - and does not seem to leak... so ok? (-:

jluebbe: I'd prefer to use a consistent pattern of checks in RAUC. :)

js731ca: Ok, changed it to the `g_return_val_if_fail` with v4.1

js731ca:
v4.1: minor change: use more `g_return_val_if_fail`

jluebbe (changes requested):
The core code looks fine now.

What's still missing is test coverage for the new code in `config_file.c` and `install.c`. For the config, you can add two cases (good and invalid) similar to `config_file_logger_full` and `config_file_logger_invalid_event` in `test/config_file.c`.

Testing the check during installation is more involved, so I'll probably add that myself after this is merged.

js731ca:
> The core code looks fine now.

(:tada:)
 
> What's still missing is test coverage for the new code in `config_file.c` and `install.c`. For the config, you can add two cases (good and invalid) similar to `config_file_logger_full` and `config_file_logger_invalid_event` in `test/config_file.c`.

added with v4.2 :-)
 
> Testing the check during installation is more involved, so I'll probably add that myself after this is merged.

wow true... managed to cobble sth together that seems to get the job done; but the boilerplate seems way thicker than it needs to be - feel free to take the blowtorch to it ;-D

while putting it together, was wondering if the actual version check should be moved into this  `r_install_make_plans` and then have a test similar to `test_install_image_variants` that just has a system.conf and compares multiple bundle-manifests; instead of the current incarnation that actually has to construct&install bundles... thoughts?

js731ca:
v4.2:
src/config:
- fixed minor memleak in the config (uncovered by the test :-)
- removed unneeded error pointer handling
test/install:
- fix grubenv errors
- and failing github/actions, since they run tests as non-root

Thought/suggestion: moving the `mountprefix=` into the tmp/rauc-XXXXX would allow test runs on systems that have a RO rootfs (the default mkdir  /mnt/rauc fails)

jluebbe:
> Thought/suggestion: moving the `mountprefix=` into the tmp/rauc-XXXXX would allow test runs on systems that have a RO rootfs (the default mkdir /mnt/rauc fails)

We're currently reworking the tests a bit (see #1401), hopefully simplifying this a bit.

In general, you should use `./qemu-test` / `./qemu-test shell` during development, as this allows running the largest set of tests without affecting your host system. Tests that have requirements that are not fulfilled by the host should be skipped.

No need to try things out in this PR right now, I'll take a closer look later.

js731ca:
2mon later -> gentle ping? :-)
anything i can do/contribute?

jluebbe:
I've rebased and force pushed to fix the merge conflict.

src/install.c
```diff
@@ -550,6 +550,36 @@ static gboolean verify_compatible(RaucInstallArgs *args, RaucManifest *manifest,
 	}
 }
 
+static gboolean check_version_limits(RaucInstallArgs *args, RaucManifest *manifest, GError **error)
```
ejoerns, line 553: should gain input guards

src/install.c
```diff
@@ -550,6 +550,36 @@ static gboolean verify_compatible(RaucInstallArgs *args, RaucManifest *manifest,
...
+{
+	GError *ierror = NULL;
+
+	if (args->ignore_version_limit)
+		return TRUE;
+
+	if (!r_context()->config->system_min_bundle_version) {
+		g_debug("No min_bundle_version configured in the system.conf. Version-limit check does nothing.");
```
ejoerns, line 561:
```suggestion
		g_debug("No min-bundle-version configured in the system.conf. Version-limit check does nothing.");
```

docs/reference.rst
```diff
@@ -60,6 +60,28 @@ Example configuration:
...
   update manifest.
 
+``min-bundle-version`` (optional)
+  An optional user-defined version string that follows the
+  `semantic versioning scheme <https://semver.org>`_.
+  The 'version core' and '-pre_release' parts from the ``version`` string in the
+  update manifest is compared against. This allows to roll out updates with some
+  (breaking) change, which set the ``min-bundle-version`` accordingly, and
+  through that disallow the installation of bundles with a prior version -
+  which would render the device inoperable.
```
ejoerns, line 70:
```suggestion
  The 'version core' and '-pre_release' parts from the ``version`` string in the
  update manifest are compared against the configured min-bundle-version.
  The update will be rejected if the bundle manifest version is smaller/older. 
  
  This allows to roll out updates with some (breaking) changes by shipping them with
  a system.conf that sets the ``min-bundle-version`` accordingly.
  Once the updated system runs, it will prevent the installation of future updates with smaller/older versions.
```

jluebbe:
I've noticed that the check failed when the bundle didn't declare a version, so I added handling and a test case for that.

jluebbe (approved):

docs/reference.rst
```diff
@@ -1440,6 +1462,10 @@ IN a{sv} *args*:
         installation of bundles on platforms that a compatible not matching the one
         of the bundle to be installed
 
+    :STRING 'ignore-version-limit', VARIANT 'b' <true/false>: Disables all
+        version checks, which normally would compare the bundle version against
+        the system.conf set ``min-version`` and/or ``max-version``.
+
```
ejoerns, line 1468:
```suggestion
    :STRING 'ignore-version-limit', VARIANT 'b' <true/false>: Disables the
    check for the minimal bundle version as configured by system.conf option
    ``min-bundle-version`` checks.
```

src/install.c
```diff
@@ -550,6 +550,43 @@ static gboolean verify_compatible(RaucInstallArgs *args, RaucManifest *manifest,
 	}
 }
 
+static gboolean check_version_limits(RaucInstallArgs *args, RaucManifest *manifest, GError **error)
+{
+	GError *ierror = NULL;
+
+	if (args->ignore_version_limit)
```
ejoerns, line 557: Should this get a `g_message()` printout about explicitly disabling version check?

Maybe ordered then after the test for `system_min_bundle_version`.

src/utils.c
```diff
@@ -826,3 +826,192 @@ gchar *r_bytes_unref_to_string(GBytes **bytes)
...
+	if (version_parts[0] == NULL) {
+		g_set_error(error,
+				R_UTILS_ERROR,
+				R_UTILS_ERROR_SEMVER_PARSE,
+				"Failed to parse semantic version '%s', 'version core' has no components.",
+				version_string);
+		goto out;
+	}
+
+	for (i = 0; i < 3; i++)
```
ejoerns, line 908: shadows outer i

src/utils.c
```diff
@@ -826,3 +826,192 @@ gchar *r_bytes_unref_to_string(GBytes **bytes)
...
+	 *
+	 * detect first the '+build', separate it out by null terminating the
+	 * version_copy at it's start position; repeat for the '-pre-release'; and
+	 * then parse the remaining 'version core'
+	 */
+
+	build_pos = g_strrstr(version_copy, "+");
+	if (build_pos != NULL) {
+		if (build != NULL) {
+			*build = g_strdup(build_pos + 1);
```
ejoerns, line 859: It would possibly simplify the flow here and later if we'd save the return values internally in an autopointer gchar and steal the pointer at the end with returning TRUE. But it works this way, too.

src/utils.c
```diff
@@ -826,3 +826,192 @@ gchar *r_bytes_unref_to_string(GBytes **bytes)
...
+		goto out;
+	}
+
+	for (i = 0; i < 3; i++)
+		version_core[i] = 0;
+	i = 0;
+	while (version_parts[i]) {
+		if (!g_ascii_string_to_unsigned(version_parts[i], 10, 0, G_MAXUINT64, &version_core[i], &ierror)) {
+			g_propagate_prefixed_error(error, ierror,
+					"Failed to parse core version component '%s' as uint.", version_parts[i]);
```
ejoerns, line 914:
```suggestion
					"Failed to parse core version component '%s' as unit: ", version_parts[i]);
```

ejoerns (changes requested):

docs/reference.rst
jluebbe: s/minimal/minimum/

src/install.c
jluebbe: While we don't have that for the compatible check, it's probably a good idea.

src/utils.c
jluebbe: I'd like it keep it this way for now.

src/utils.c
ejoerns: ah, i see.

ejoerns (approved):

ejoerns:
I've quickly addressed my open review comments.

jluebbe (approved):

Issue #1399, opened 2024-05-01
Support installing and running tests on a target


LeBlue:
**Is your feature request related to a problem?**
I have been working on running the rauc unit-tests on a target system through the yocto ptest framework. To get it to work, i have patched the rauc (test) build process a bit.

If there is general interest in these changes, I can submit the related PRs. 

**Describe the solution you'd like**
To support the on-target tests/ptest, the rauc build could support installing all unit-test programs and assets. 

Additional changes i had to make/want to make:

- Improve test selection based on selected compile time features (bundle create, service, etc.)
- Improve/fix drop-ins e.g. grub-editenv

**Describe alternatives you've considered**
Most (all?) of the changes could be done in the yocto recipe, but:
- rauc version specific test changes do not have to be maintained in the recipe.
- implementing the necessary bits here instead of the yocto recipe makes the feature (running tests on a target machine) available outside of the yocto build.

**Additional context**
For the yocto bits, i can only provide and test the changes on the 'kirkstone' branch. I would need some help with that.

In addition I have serveral features disabled, namely casync, networking, streaming. Therfore I cannot really verify the tests will run successfully on the target.

ejoerns:
@LeBlue Many of the tests require root access or specific kernel features to be enabled and are meant to be run in our qemu environment, but that should not prevent allowing those that do not have such strong restrictions.

ejoerns:
[Pressed Send accidentially] :smirk: 

> If there is general interest in these changes, I can submit the related PRs.

I'd be interested. I guess having ptest support is at least good to align with other oe-integrated packages.

> Improve test selection based on selected compile time features (bundle create, service, etc.)

If there are missing bits, I'd be happy to have them in the general test suite, yes.

I am also working on migrating the sharness-based tests (rauc.t) to pytest currently, thus I would be really interested if there are things I should pick up.

> For the yocto bits, i can only provide and test the changes on the 'kirkstone' branch. I would need some help with that.

I could support with this, probably best in the context of meta-rauc repo then.

LeBlue:
@ejoerns I have added a PR for the basic install changes. I am still working moving the recipe appends into `meta-rauc`.

> If there are missing bits, I'd be happy to have them in the general test suite, yes.
I am aware the the `rauc.t` changes are probably outdated, but the idea of filtering for the `create` build option is still valid.

> I am also working on migrating the sharness-based tests (rauc.t) to pytest currently, thus I would be really interested if there are things I should pick up.
[...]

PR #1408, opened 2024-05-07
Tests: add option to install tests and fixes

LeBlue:
Add a build option to install the tests and assets. The intended use case is to run the tests on a target.

Filter the `rauc.t` tests  with the CREATE build config option. A feature that may be disabled for embedded systems builds.

Fix the `grub-editenv` mock script, that tests can succeed when using the script instead of the real program. 

Relates-To: #1399

### DIFF
--- a/test/bin/grub-editenv
+++ b/test/bin/grub-editenv
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+if [ $# -eq 1 ] && [ "$1" = "-V" ]; then
+	echo "grub-editenv (RAUC test drop-in) 0.0.0"
+	exit 0
+fi
+
 # Variable argument count
 if [ $# -lt 2 ]; then
 	echo "At least path to grubenv file and command must be specified!"
@@ -15,7 +20,8 @@ function variables_load()
 {
 	if ! [ -f "$GRUBENV" ]; then
 		echo "GRUB environment file not found: $GRUBENV"
-		exit 1
+		echo -n "" > "$GRUBENV"
+		return 0
 	fi
 
 	while read -r line; do
@@ -39,16 +45,27 @@ function variables_store()
 	variable_store "B_TRY"
 	variable_store "A_OK"
 	variable_store "B_OK"
+
+	variable_store "system0_TRY"
+	variable_store "system1_TRY"
+	variable_store "system0_OK"
+	variable_store "system1_OK"
+
 	variable_store "ORDER"
 }
 
 # Validates and sets variable value
 function variable_set {
 	if [ "$1" = "ORDER" ]; then
-		if ! [ "$2" = "A B" -o "$2" = "B A" ]; then
-			echo "Invalid argument for ORDER: '$2'"
-			exit 1
-		fi
+		case "$2" in
+			"A B"|"B A"|"system1 system0"|"system0 system1"|"system1 system0 factory0"|"system0 system1 factory0")
+				:
+			;;
+			*)
+				echo "Invalid argument for ORDER: '$2'"
+				exit 1
+			;;
+		esac
 	elif [ "$1" = "A_OK" ]; then
 		if ! [ "$2" -ge 0 -a "$2" -le 1 ]; then
 			echo "Invalid argument for A_OK: '$2'"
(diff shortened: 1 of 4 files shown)


test/meson.build
```diff
@@ -49,6 +52,9 @@ foreach test_name : tests
     link_with : librauc,
     c_args : '-DTEST_SERVICES="' + meson.build_root() + '"',
     include_directories : incdir,
+    install_mode : 'rwxr-xr-x',
+    install : get_option('install-tests'),
+    install_dir : get_option('libdir') / 'rauc' / 'tests',
```
ejoerns, line 57: The test installation directory should be defined as a variable before using it. Either in meson.build or test/meson.build.

test/meson.build
```diff
@@ -77,3 +86,13 @@ test(
   env : ['SHARNESS_BUILD_DIRECTORY=' + meson.build_root()],
   workdir : meson.current_source_dir(),
 )
+
+install_data(
```
ejoerns, line 90: This will be installed unconditionally. Maybe guard the additional install data calls with an `if install_tests`:

``` meson
if install_tests
install_data(..)
install_subdir(..)
...
endif
```

test/meson.build
```diff
@@ -77,3 +86,13 @@ test(
...
+  meson.build_root() / 'config.h',
+  install_dir: get_option('libdir') / 'rauc' / 'tests',
+)
+
+install_subdir(
+  meson.source_root() / 'test',
```
ejoerns, line 96: This installs way too many files. Also test results, etc. Should be more explicit.

test/rauc.t
```diff
@@ -132,7 +137,7 @@ grep -q "ENABLE_SERVICE 1" $SHARNESS_BUILD_DIRECTORY/config.h &&
   select_system_or_session_bus
 
 # Prerequisite: openssl available [OPENSSL]
-openssl asn1parse -help &&
+openssl asn1parse -help 2>/dev/null &&
```
ejoerns, line 140: This and the other changes are unrelated to the commit they come with. Should be split into a separate commit and note why it's needed.

test/bin/grub-editenv
```diff
@@ -69,6 +86,26 @@ function variable_set {
 			echo "Invalid argument for B_TRY: '$2'"
 			exit 1
 		fi
+	elif [ "$1" = "system0_OK" ]; then
```
ejoerns, line 89: Are these names used in any of our tests?

test/bin/grub-editenv
```diff
@@ -121,6 +164,9 @@ case $CMD in
 	"list")
 		variables_list "$@"
 		;;
+	"create")
```
ejoerns, line 167: Do we use the `create` subcommand anywhere? We also don't define `unset`

ejoerns (changes requested):
@LeBlue many thanks for sharing your work!

I've had a look and added some comments about what is missing from my point of view.

As long as the grub-editenv changes are not required for running the tests on the target, I would leave them out here first of all.

At the moment we are not fully sure yet how much effort and additional maintenance adding on-target tests would mean. Thus if anyone has an opinion here, feel free to comment.

A minimum requirement for merging this would be however to add some CI tests that check that a sensible form of test execution is possible from the installed tests.

test/meson.build
jluebbe: So far, we've not split out the subset of files needed by the test (e.g. from the openssl-* directories), as they were available anyway.

LeBlue: I have also seen this and tried to move all assets into subdirectories. But this will require a lot of changes in the `rauc.t` script and the unit tests. The additional files installed are not too many (the unit-test c-files and some build support files).

This would be doable but i wanted a first feedback.

test/rauc.t
LeBlue: It is not needed and i can remove it again.

test/bin/grub-editenv
LeBlue: Definintely used/needed for the `rauc.t` `install *` tests. I think the `install.c` unit-test also uses it, but not sure.

These changes should also be usable with the qemu test, if one uses the script instead of the real `grub-editenv`

test/bin/grub-editenv
LeBlue: I have been using it in the ptest support script: Draft is here: https://github.com/KLSMartin/meta-rauc/blob/feature/integrate-ptest/recipes-core/rauc/files/run-ptest 

Note: There are still some bits missing.

I have also tried to run the qemu tests with the `grub-editenv` drop-in script, where the `create` is needed.

jluebbe:
@LeBlue As the pytest migration has been merged, this needs some refactoring.

PR #1412, opened 2024-05-08
Use meson TAP protocol parser

ejoerns:
An issue with the current test suite is that it is quite hard to see
what tests were actually executed and what tests were just skipped.

Since the glib test suite produces TAP-parsable output by default, we
could use this to gain more detailed information on skipped tests.

Also enable verbose subtest output for run-qemu.

A general known issue is that in glib versions shipped with recent distros, the test log output emitted is duplicated: One tap-conform output with `# ` prefixed and the non-tap-conform default logging output.
The latter will cause non-fatal parsing errors for meson >= 1.0 versions that slightly pollute the log still.

This is fixed by glib commit https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3140/diffs?commit_id=6c714c9f46344c37b10c6e42936a54977f888d2d

However, we still have some remaining direct writes to stdout (by using g_print() or by sub scripts) that are not fixable this way and might need to be addressed later, like:

> stdout: 1239: UNKNOWN: Creating 'verity' format bundle
> ERROR: Unknown TAP output lines for a supported TAP version.
> This is probably a bug in the test; if they are not TAP syntax, prefix them with a #

### DIFF
--- a/qemu-test-init
+++ b/qemu-test-init
@@ -185,7 +185,7 @@ if [ -n "$SERVICE" ]; then
   rauc status mark-good
 fi
 
-if meson test -C $BUILD_DIR "$MESON_TEST"; then
+if meson test -C $BUILD_DIR "$MESON_TEST" -v; then
   touch qemu-test-ok
   echo "RESULT: PASSED"
 else
--- a/test/bundle.c
+++ b/test/bundle.c
@@ -29,7 +29,7 @@ static void bundle_fixture_set_up(BundleFixture *fixture,
 {
 	fixture->tmpdir = g_dir_make_tmp("rauc-XXXXXX", NULL);
 	g_assert_nonnull(fixture->tmpdir);
-	g_print("bundle tmpdir: %s\n", fixture->tmpdir);
+	g_test_message("bundle tmpdir: %s\n", fixture->tmpdir);
 }
 
 static void prepare_bundle(BundleFixture *fixture, gconstpointer user_data)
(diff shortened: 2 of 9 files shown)


jluebbe:
Does this make it clearer which test case failed?

ejoerns:
> Does this make it clearer which test case failed?

Hope so

```
1/3 install        RUNNING       
>>> MALLOC_PERTURB_=95 /home/enrico/Code/rauc/build/test/install-test
▶ 1/3 /install/external_mounts SKIP          
▶ 1/3 /install/adaptive SKIP          
▶ 1/3 /install/slot-skipping SKIP          
▶ 1/3 /install/target-group/non-redundant OK            
▶ 1/3 /install/target-group/async OK            
▶ 1/3 rauc:ERROR:../test/install.c:522:test_install_target_group_sync: 'g_hash_table_contains(tgrp, "rootfsXXX")' should be TRUE ERROR         
1/3 install        ERROR            0.20s   killed by signal 6 SIGABRT
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
stderr:
mke2fs 1.47.0 (5-Feb-2023)
mke2fs 1.47.0 (5-Feb-2023)
mke2fs 1.47.0 (5-Feb-2023)
```

(just with colors)

jluebbe (approved):

PR #1413, opened 2024-05-10
qemu-test-init: configure gdb to save history and enable pretty printing

jluebbe:


qemu-test-init
```diff
@@ -55,6 +55,8 @@ ip link set eth0 up
 ip route add default via 10.0.2.2
 echo "nameserver 10.0.2.3" > /etc/resolv.conf
 
+export HOME="/root"
```
ejoerns, line 58: Is this required for gdb to read the config from the correct location?

ejoerns (approved):

PR #1419, opened 2024-05-14
Fix double close on ubivol descriptor

tvdstaaij:
The g_output_stream_close call was introduced at some point after the casync functionality for UBI was introduced (#501). This close call is fine for the copy_raw_image branch, but for the casync branch it causes the update to fail:

`rauc[250]: Installation error: Failed updating slot rootfs.1: Error closing file descriptor: Bad file descriptor`

This was also mentioned in #1200. However it was not investigated there but instead sidestepped by using a different update approach.

Investigation with strace shows that the desciptor is closed twice:

```
290   openat(AT_FDCWD, "/dev/ubi0_1", O_WRONLY|O_EXCL|O_LARGEFILE <unfinished ...>
290   <... openat resumed>)             = 8
[snip]
290   close(8 <unfinished ...>
290   <... close resumed>)              = 0
290   close(8 <unfinished ...>
290   <... close resumed>)              = -1 EBADF (Bad file descriptor)
```

This can be explained by the documentation of
g_subprocess_launcher_close():

> Closes all the file descriptors previously passed to the object with
> g_subprocess_launcher_take_fd(), g_subprocess_launcher_take_stderr_fd(), etc.
> (...)
> This function is called automatically when the #GSubprocessLauncher is disposed

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -1487,6 +1487,7 @@ static gboolean img_to_ubivol_handler(RaucImage *image, RaucSlot *dest_slot, con
 			g_propagate_error(error, ierror);
 			goto out;
 		}
+		/* at this point, out_fd has already been closed by glib */
 	} else {
 		/* copy */
 		res = copy_raw_image(image, outstream, 0, &ierror);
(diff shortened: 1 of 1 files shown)


jluebbe:
The intention with the explicit close is to be able to catch more errors (some of which are only reported on `close()`). So perhaps a better way is to pass a `dup()`ed FD to `g_subprocess_launcher_take_stdout_fd()`?

Sadly this code is not covered by the CI, as we didn't want to build a custom version of casync https://github.com/systemd/casync/pull/227.

tvdstaaij:
> The intention with the explicit close is to be able to catch more errors (some of which are only reported on `close()`). So perhaps a better way is to pass a `dup()`ed FD to `g_subprocess_launcher_take_stdout_fd()`?

Ah, I wasn't aware of the existence of `dup()`. I submitted a new PR.

jluebbe:
Note that you can force-push to your branch to update this PR, which keeps the discussion in one place. :)

PR #1421, opened 2024-05-15
src/update_handler: fix double close on ubivol descriptor

tvdstaaij:
The g_output_stream_close call was introduced at some point after the casync functionality for UBI was introduced (#501). This close call is fine for the copy_raw_image branch, but for the casync branch it causes the update to fail:

>rauc[250]: Installation error: Failed updating slot rootfs.1: Error closing file descriptor: Bad file descriptor

This was also mentioned in #1200. However it was not investigated there but instead sidestepped by using a different update approach.

Investigation with strace shows that the descriptor is closed twice:

```
290   openat(AT_FDCWD, "/dev/ubi0_1", O_WRONLY|O_EXCL|O_LARGEFILE <unfinished ...>
290   <... openat resumed>)             = 8
[snip]
290   close(8 <unfinished ...>
290   <... close resumed>)              = 0
290   close(8 <unfinished ...>
290   <... close resumed>)              = -1 EBADF (Bad file descriptor)
```

This can be explained by the documentation of
g_subprocess_launcher_close():

> Closes all the file descriptors previously passed to the object with
> g_subprocess_launcher_take_fd(), g_subprocess_launcher_take_stderr_fd(), etc.
> (...)
> This function is called automatically when the #GSubprocessLauncher is disposed

---

__Additional information__

Rework of #1419

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -476,8 +476,20 @@ static gboolean casync_extract(RaucImage *image, gchar *dest, int out_fd, const 
 	g_ptr_array_add(args, NULL);
 
 	launcher = g_subprocess_launcher_new(G_SUBPROCESS_FLAGS_NONE);
-	if (out_fd >= 0)
-		g_subprocess_launcher_take_stdout_fd(launcher, out_fd);
+
+	if (out_fd >= 0) {
+		/* Must be a copy: glib automatically closes stdout_fd, but we close out_fd as well */
+		int out_fd_copy = dup(out_fd);
+		if (out_fd_copy == -1) {
+			g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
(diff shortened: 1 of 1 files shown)


src/update_handler.c
```diff
@@ -439,6 +439,7 @@ static gboolean casync_extract(RaucImage *image, gchar *dest, int out_fd, const
 	GError *ierror = NULL;
 	gboolean res = FALSE;
 	g_autoptr(GPtrArray) args = g_ptr_array_new_full(5, g_free);
+	int out_fd_copy = -1;
```
jluebbe, line 442: Move the variable to the `if (out_fd >= 0)` scope.

jluebbe (changes requested):
Just make a force-push with an updated commit to change this PR.

jluebbe (approved):

PR #1425, opened 2024-05-15
use a temporary workdir during bundle creation

jluebbe:
Currently, we modify and add files in the input directory. While this
isn't much of a problem now, we will need to delete converted files when
converting tar archives to file trees in the context of artifacts. That
would require the user to always re-create the bundle input directory
after running 'rauc bundle', which would be annoying.

It seems the simplest way is to hard-link all files from the input dir
to a .rauc-workdir subdirectory. This way, we don't need actually copy
any data and keep the rest of the bundle creation code unaffected.

For simplicity, we abort on anything in the input dir which is not a
regular file. If someone relies on the old (undocumented) behaviour of
including directories and symlinks in the bundle, please contact us.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -201,3 +201,29 @@ When using Yocto, in case of the ``orphan_file`` file feature, you could use::
 in the image recipe or an appropriate conf file to disable the feature until
 all systems have been updated with versions of ``resize2fs`` which support this
 feature.
+
+Why does bundle creation fail with a "not supported as contents" error?
+-----------------------------------------------------------------------
+
+Previous versions of RAUC modified and added files in the input directory during
+bundle creation.
+While this wasn't much of a problem, we need to delete the original files when
+converting tar archives to file trees in the context of artifact updates.
+That would require the user to always re-create the bundle input directory after
+running ``rauc bundle``, which would be unexpected and annoying.
+
+Since version 1.12, RAUC hard-links all files from the input directory to a
+``.rauc-workdir`` subdirectory.
+This way, we don't actually need to copy any data and can perform any
+preparation of the contents without affecting the input directory.
+
+For simplicity, we abort on anything in the input directory which is not a
+regular file.
+In that case, one of the following errors will be shown:
+
+  * ``Failed to create bundle: symlinks are not supported as bundle contents (a_symlink)``
+  * ``Failed to create bundle: directories are not supported as bundle contents (a_directory)``
+  * ``Failed to create bundle: only regular files are supported as bundle contents (a_fifo)``
+
+If someone relies on the old undocumented behavior of including directories and
+symlinks in the bundle, please contact us.
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -10,6 +10,7 @@
 #include <sys/socket.h>
 
 #include <openssl/rand.h>
+#include <unistd.h>
 
 #include "bundle.h"
 #include "context.h"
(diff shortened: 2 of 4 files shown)


ejoerns (commented):
I really like the idea in general, since it also solves current confusion we have by modifying the manifes in-place.

My biggest concern ist to brake a number of use-cases we don't know about, yet.

Would it be an option to loudly warn and fall back to the 'contentdir' for a transition phase?

We should maybe also instruct users via error messages to contact us, since probably not all users of bundle creation will be familiar with looking into the code/commit message..

src/bundle.c
```diff
@@ -949,6 +1038,11 @@ gboolean create_bundle(const gchar *bundlename, const gchar *contentdir, GError
 		goto out;
 	}
 
+	if (workdir && !rm_tree(workdir, &ierror)) {
+		g_propagate_prefixed_error(error, ierror, "Failed to remove workdir: ");
```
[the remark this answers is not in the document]
jluebbe: If we can't remove the workdir here, something is seriously wrong and we shouldn't just continue.

src/bundle.c
```diff
@@ -881,11 +882,91 @@ static gboolean decrypt_bundle_payload(RaucBundle *bundle, GError **error)
...
 
+static gchar *prepare_workdir(const gchar *contentdir, GError **error)
+{
+	GError *ierror = NULL;
+	g_autofree gchar* workdir = NULL;
+	g_autoptr(GDir) dir = NULL;
+	const gchar *name;
+
+	/* This could be simplified with g_canonicalize_file() after bumping the
+	 * glib requirement to 2.58. */
```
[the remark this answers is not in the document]
jluebbe: Ubuntu bionic is still in "Extended Security Maintenance" and the complexity for us is low, so I wouldn't want to drop support for it just for this.

jluebbe:
I've added a FAQ entry as discussed internally.

docs/faq.rst
```diff
@@ -201,3 +201,29 @@ When using Yocto, in case of the ``orphan_file`` file feature, you could use::
...
+bundle creation.
+While this wasn't much of a problem, we need to delete the original files when
+converting tar archives to file trees in the context of artifact updates.
+That would require the user to always re-create the bundle input directory after
+running ``rauc bundle``, which would be unexpected and annoying.
+
+Since version 1.12, RAUC hard-links all files from the input dir to a
+``.rauc-workdir`` subdirectory.
+This way, we don't need actually copy any data and can perform any preparation
+of the contents without affecting the input directory.
```
ejoerns, line 218:
```suggestion
This way, we don't actually need to copy any data and can perform any preparation of the contents without affecting the input directory.
```

docs/faq.rst
```diff
@@ -201,3 +201,29 @@ When using Yocto, in case of the ``orphan_file`` file feature, you could use::
...
+-----------------------------------------------------------------------
+
+Previous versions of RAUC modified and added files in the input directory during
+bundle creation.
+While this wasn't much of a problem, we need to delete the original files when
+converting tar archives to file trees in the context of artifact updates.
+That would require the user to always re-create the bundle input directory after
+running ``rauc bundle``, which would be unexpected and annoying.
+
+Since version 1.12, RAUC hard-links all files from the input dir to a
```
ejoerns, line 215:
```suggestion
Since version 1.12, RAUC hard-links all files from the input directory to a
```

src/bundle.c
```diff
@@ -881,11 +882,92 @@ static gboolean decrypt_bundle_payload(RaucBundle *bundle, GError **error)
 	return res;
 }
 
+static gchar *prepare_workdir(const gchar *contentdir, GError **error)
```
ejoerns, line 885: Better add `g_return_val_if_fail()` input checks for the parameters

ejoerns (changes requested):
Some minor things left, looks fine otherwise.

ejoerns (approved):

PR #1433, opened 2024-05-23
test/conftest: avoid busy-waiting

jluebbe:
The service will need some time to be available on the bus, so wait a bit before each try. This also gives the service more CPU time, so it may be faster.


ejoerns (approved):
I thought I already fixed this during the review of my pytest PR. Apparently, I missed it :smirk:

PR #1435, opened 2024-05-24
update-handler: fallbacks for unsupported sources

### LINKED ISSUE #1177: Image type detection to include ".verity" rootfs extension

Louson:
Some sources are not supported and may require a fallback. The verity image generated by meta-security is an example.

This is a suggestion to fix #1177. Adding fallbacks per destination would let rauc ignore unknown sources such as verity images. All fallbacks are raw copy except for ubi types.

I haven't made any test so far. I can try on ext4 verity images if you think the idea is good.

tx

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -2630,6 +2630,11 @@ RaucUpdatePair updatepairs[] = {
 	{"*.ext4", "boot-gpt-switch", img_to_boot_gpt_switch_handler},
 	{"*.img", "boot-gpt-switch", img_to_boot_gpt_switch_handler},
 #endif
+	{"*", "ext4", img_to_raw_handler},
+	{"*", "raw", img_to_raw_handler},
+	{"*", "vfat", img_to_vfat_handler},
+	{"*", "ubivol", img_to_ubivol_handler},
+	{"*", "ubifs", img_to_ubifs_handler},
 	{"*", "boot-gpt-switch", NULL},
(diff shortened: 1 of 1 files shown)


ejoerns:
Allowing a fallback for most image types would make the image type more or less useless, I guess.
And it will also just pick a (wrong) handler if the image type is actually not supported.

An alternative for this, suggested in another PR discussing this topic, was https://github.com/rauc/rauc/pull/1201#issuecomment-1663668792: 

> Instead of adding more and more extension matches, perhaps an explicit option in the manifest to set the file type would be better.

jluebbe:
As @ejoerns said, this doesn't fit the design.

PR #1437, opened 2024-05-29
refactor shell and environment variable preparation

jluebbe:
We currently prepare the env variables for subprocesses and
shell-parsable output separately. To unify them, handle the shell
quoting at the end, during collection into a string.

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -96,6 +96,40 @@ static inline void r_ptr_array_addv(GPtrArray *ptrarray, gchar **argvp, gboolean
 }
 
 /**
+ * Adds a formatted string to the end of a GPtrArray
+ *
+ * This is a shorter alternative to:
+ * g_ptr_array_add(arr, g_strdup_printf("%s: %s", ...));
+ *
+ * @param ptrarray GPtrArray to add to
+ * @param format the printf-like format string
+ * @param ... the parameters for the format string
+ */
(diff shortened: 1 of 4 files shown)


ejoerns:
I suspect a commit using `r_environ_setenv_ptr_array()` is missing for completeness.

jluebbe:
> I suspect a commit using `r_environ_setenv_ptr_array()` is missing for completeness.

I've added a test case.

ejoerns:
> We currently prepare the env variables for subprocesses and shell-parsable output separately. To unify them, handle the shell quoting at the end, during collection into a string.

So the aim of this PR is not to unify but to prepare for later unifying? My expectation from the PR description was that it touches env variables, too.

jluebbe:
> > We currently prepare the env variables for subprocesses and shell-parsable output separately. To unify them, handle the shell quoting at the end, during collection into a string.
> 
> So the aim of this PR is not to unify but to prepare for later unifying? My expectation from the PR description was that it touches env variables, too.

Yes, this is split out to make the artifact PR smaller and easier to review.

jluebbe:
The subprocess part is not contained in this PR, but it builds some helpers that would allow us to use the same list of variables for CLI shell format output and handler/hook startup. That will be useful for output of artifact repo information.

ejoerns (approved):
With the test case added, the intended later usage should be quite clear, thx!

Looks fine to me and already simplifies the shell output code a bit.

PR #1442, opened 2024-06-10
Fix U-Boot contrib script checks

xaqbr:
When adopting the provided U-Boot script for my system's configuration, I noticed `if test 0x${BOOT_A_LEFT} -gt 0; then` and `if test 0x${BOOT_B_LEFT} -gt 0; then` were failing due to trying to compare string with integers. Not sure if there was another intended purpose for this, but getting rid of the `0x` fixed the code in my case, so I thought I'd share.

### DIFF
--- a/contrib/uboot.sh
+++ b/contrib/uboot.sh
@@ -11,14 +11,14 @@ for BOOT_SLOT in "${BOOT_ORDER}"; do
   if test "x${bootargs}" != "x"; then
     # skip remaining slots
   elif test "x${BOOT_SLOT}" = "xA"; then
-    if test 0x${BOOT_A_LEFT} -gt 0; then
+    if test ${BOOT_A_LEFT} -gt 0; then
       echo "Found valid slot A, ${BOOT_A_LEFT} attempts remaining"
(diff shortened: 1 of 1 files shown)


jluebbe:
This is the inverse of commit 3fc36bbc88c4f46af54377f87c5e4f6fc6725756 from PR #1078. Does U-Boot's comparison behavior depend on configuration options?

ejoerns:
There have been similar discussions (and suggestionts to use `itest` instead) in https://github.com/rauc/meta-rauc-community/pull/73.

The reason not to use `itest` for me was that it does not seem to be enabled in any defconfig shipped with u-boot (still isn't as I just re-evaluated).

I had a look at the hush tests found in u-boot and noticed that there are tests like 

    ('test 0xa -eq 10', True)

which strongly suggest that our script should work.

Shortly digging through the corresponding mails led me to
https://lists.denx.de/pipermail/u-boot/2019-September/383309.html.

Thus I would conclude that the version we use works well starting with U-Boot version v2020.01. But not with older ones!

@xaqbr Which U-Boot version did you test this with?

ejoerns:
@xaqbr any update on this? Otherwise, I'd like to close it.

PR #1450, opened 2024-06-18
bootchooser: add efibootguard bootloader support

### LINKED ISSUE #1443: bootchooser: efibootguard support
It'd be great to add efibootguard support to the bootchooser as it is a very suitable bootloader for A/B booting.

For reference there is [swupdate's efibootguard implementation](https://github.com/sbabic/swupdate/blob/master/bootloader/ebg.c)

This is something I am planning to work on very shortly.

obbardc:
Add bootchooser support for efibootguard.

Draft for now, for some early feedback.

Fixes #1443

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -532,6 +532,8 @@ tool to interact with it:
 :U-Boot: fw_setenv/fw_getenv (from `u-boot <http://git.denx.de/?p=u-boot.git;a=summary>`_)
 :GRUB: grub-editenv
 :EFI: efibootmgr
+:EFI Boot Guard: bg_printenv/bg_setenv
+                 (from `efibootguard <https://github.com/siemens/efibootguard>`_)
 
 Note that for running ``rauc info`` on the target (as well as on the host), you
 also need to have the ``unsquashfs`` tool installed.
@@ -1359,6 +1361,147 @@ argument in the ``efibootmgr`` call above:
   device=/dev/sdX4
   type=ext4
   parent=efi.1
+
+.. _sec-efibootguard:
+
+EFI Boot Guard
+~~~~~~~~~~~~~~
+
+`EFI Boot Guard <https://github.com/siemens/efibootguard>`_ is an EFI-based
+bootloader from Siemens designed for embedded systems with A/B redundancy.
+It stores a boot environment per EFI partition. This environment contains a
+``USTATE`` value which records the current state of the slot and a ``REVISION``
+counter which is incremented for each update and is used to prefer the newest
+valid slot when selecting the slot to boot.
+
+
+The ``USTATE`` values are encoded as follows:
+
+.. code-block:: text
+
+  OK        = 0
+  INSTALLED = 1
+  TESTING   = 2
+  FAILED    = 3
+
+EFI Boot Guard will boot the slot with the highest ``REVISION`` whose ``USTATE``
+is not ``FAILED``.
+
+RAUC uses the ``bg_printenv`` and ``bg_setenv`` tools to read and write these
+variables.
+
+The slot selection logic in RAUC is as follows:
+
+* **Primary slot** - the slot with the highest ``REVISION`` value whose
+  ``USTATE`` is not ``FAILED`` is selected as primary. When RAUC marks a slot
+  as primary, it sets the slot state to ``INSTALLED`` and assigns it a
+  ``REVISION`` to one higher than the current maximum across all slots.
+* **Slot state** - a slot is considered *good* when its ``USTATE`` is not
+  ``FAILED``. RAUC marks a slot as bad by setting ``USTATE`` to ``FAILED`` (and
+  resets its ``REVISION`` to ``0`` so it cannot be re-selected as the primary).
+  RAUC marks a slot as good by setting ``USTATE`` to ``OK``.
+
+The full update lifecycle, with the responsibilities split between RAUC and
+EFI Boot Guard, looks like this:
+
+.. code-block:: text
+
+   ┌─────────────────┐          ┌─────────────────┐         ┌─────────────────┐
+   │ initial:        │   RAUC   │ after install:  │  EBG    │ first boot of   │
+   │ USTATE    = OK  ├─────────>│ USTATE = INST.  ├────────>│ new slot:       │
+   │ (highest REV    │ install  │ REVISION = max+1│ reboot  │ USTATE = TESTING│
+   │  is current)    │          │                 │         │                 │
+   └─────────────────┘          └─────────────────┘         └───────┬─────────┘
+                                                                    │
+   ┌─────────────────┐          ┌─────────────────┐                 │
+   │ RAUC mark-good: │          │ application     │     boot ok     │
+   │ USTATE    = OK  │<─────────┤ confirms the    │<────────────────┤
+   │                 │          │ update          │                 │
+   └─────────────────┘          └─────────────────┘                 │
+                                                                    │
+   ┌─────────────────┐          ┌─────────────────┐                 │
+   │ RAUC mark-bad   │          │ EFI Boot Guard: │  boot failure   │
+   │ (optional):     │<─────────┤ USTATE = FAILED ├<────────────────┘
+   │ USTATE = FAILED │          │ (auto, via      │  (watchdog or   │
+   │ REVISION = 0    │          │  watchdog)      │   max retries)  │
+   └─────────────────┘          └─────────────────┘
+
+In words:
+
+#. **Install** - RAUC writes the new image to the inactive slot and calls
+   ``bg_setenv`` to set its ``USTATE`` to ``INSTALLED`` and its ``REVISION``
+   to one higher than the current maximum. The currently-running slot is
+   untouched.
+#. **Reboot** - on the next boot, EFI Boot Guard picks the slot with the
+   highest ``REVISION`` whose ``USTATE`` is not ``FAILED``. As it hands
+   control to that slot, EFI Boot Guard transitions ``USTATE`` from
+   ``INSTALLED`` to ``TESTING``.
+#. **Confirm** - once the application running on the new slot has verified
+   the system is healthy, it calls ``rauc status mark-good``, which sets
+   ``USTATE`` back to ``OK``.
+#. **Failure** - if the new slot fails to boot or hangs, EFI Boot Guard's
+   watchdog / retry logic sets ``USTATE`` to ``FAILED`` automatically and
+   the next boot falls back to the previous slot (which still has
+   ``USTATE = OK`` and the next-highest ``REVISION``). RAUC's
+   ``rauc status mark-bad`` performs the same transition explicitly and
+   additionally resets the slot's ``REVISION`` to ``0``.
+
+To enable EFI Boot Guard support in RAUC, write in your ``system.conf``:
+
+.. code-block:: cfg
+
+  [system]
+  ...
+  bootloader=efibootguard
+
+Assuming a simple A/B redundancy with two EFI Boot Guard partitions and two
+rootfs partitions, set up ``system.conf`` as follows.
+The ``bootname`` for each slot must match the partition index as accepted by
+the ``--part`` flag of ``bg_printenv`` and ``bg_setenv``:
+
+.. code-block:: cfg
+
+  [slot.efi.0]
+  device=/dev/sda1
+  type=vfat
+  bootname=0
+
+  [slot.efi.1]
+  device=/dev/sda2
+  type=vfat
+  bootname=1
+
+  [slot.rootfs.0]
+  device=/dev/sda3
+  type=ext4
+  parent=efi.0
+
+  [slot.rootfs.1]
+  device=/dev/sda4
+  type=ext4
+  parent=efi.1
+
+.. note::
+
+   EFI Boot Guard does not pass the active slot name via the kernel command
+   line automatically.
+   You must pass ``rauc.slot=<bootname>`` explicitly in the kernel command line
+   of each slot so that RAUC can identify the currently booted slot, e.g.:
+
+   .. code-block:: cfg
+
+     rauc.slot=0
+
+   See :ref:`sec-integration-boot-slot-detection` for the full list of slot
+   detection mechanisms.
+
+Before using RAUC with EFI Boot Guard, ensure each partition's environment has
+been initialised. For example, to prefer booting from slot ``0``:
+
+.. code-block:: console
+
+  # bg_setenv -p 0 -s OK -r 1
+  # bg_setenv -p 1 -s OK -r 0
 
 .. _sec-custom-bootloader-backend:
 
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -2,6 +2,7 @@
 #include "bootloaders/barebox.h"
 #include "bootloaders/custom.h"
 #include "bootloaders/efi.h"
+#include "bootloaders/efibootguard.h"
 #include "bootloaders/grub.h"
 #include "bootloaders/uboot.h"
 #include "config_file.h"
@@ -12,7 +13,7 @@ GQuark r_bootchooser_error_quark(void)
 	return g_quark_from_static_string("r_bootchooser_error_quark");
 }
 
-static const gchar *supported_bootloaders[] = {"barebox", "grub", "uboot", "efi", "custom", "noop", NULL};
+static const gchar *supported_bootloaders[] = {"barebox", "grub", "uboot", "efi", "efibootguard", "custom", "noop", NULL};
 
 gboolean r_boot_is_supported_bootloader(const gchar *bootloader)
 {
@@ -93,6 +94,8 @@ gboolean r_boot_get_state(RaucSlot *slot, gboolean *good, GError **error)
 		res = r_uboot_get_state(slot, good, &ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "efi") == 0) {
 		res = r_efi_get_state(slot, good, &ierror);
+	} else if (g_strcmp0(r_context()->config->system_bootloader, "efibootguard") == 0) {
+		res = r_efibootguard_get_state(slot, good, &ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "custom") == 0) {
 		res = r_custom_get_state(slot, good, &ierror);
 	} else {
@@ -131,6 +134,8 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
 		res = r_uboot_set_state(slot, good, &ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "efi") == 0) {
 		res = r_efi_set_state(slot, good, &ierror);
+	} else if (g_strcmp0(r_context()->config->system_bootloader, "efibootguard") == 0) {
+		res = r_efibootguard_set_state(slot, good, &ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "custom") == 0) {
 		res = r_custom_set_state(slot, good, &ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "noop") == 0) {
@@ -171,6 +176,8 @@ RaucSlot *r_boot_get_primary(GError **error)
 		slot = r_uboot_get_primary(&ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "efi") == 0) {
 		slot = r_efi_get_primary(&ierror);
+	} else if (g_strcmp0(r_context()->config->system_bootloader, "efibootguard") == 0) {
+		slot = r_efibootguard_get_primary(&ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "custom") == 0) {
 		slot = r_custom_get_primary(&ierror);
 	} else {
@@ -209,6 +216,8 @@ gboolean r_boot_set_primary(RaucSlot *slot, GError **error)
 		res = r_uboot_set_primary(slot, &ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "efi") == 0) {
 		res = r_efi_set_primary(slot, &ierror);
+	} else if (g_strcmp0(r_context()->config->system_bootloader, "efibootguard") == 0) {
+		res = r_efibootguard_set_primary(slot, &ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "custom") == 0) {
 		res = r_custom_set_primary(slot, &ierror);
 	} else if (g_strcmp0(r_context()->config->system_bootloader, "noop") == 0) {
--- a/src/bootloaders/efibootguard.c
+++ b/src/bootloaders/efibootguard.c
@@ -1,0 +1,314 @@
+#include "efibootguard.h"
+#include "bootchooser.h"
+#include "context.h"
+#include "utils.h"
+
+#define BG_PRINTENV_NAME "bg_printenv"
+#define BG_SETENV_NAME   "bg_setenv"
+
+/* Fetch a variable from efibootguard environment for a slot.
+ *
+ * @slot: bootname of the slot
+ * @key: variable to fetch
+ * @value: (out) string value (allocated, must be NULL on entry)
+ * @error: (out) error
+ */
+static gboolean efibootguard_env_get(const gchar *slot, const gchar *key, GString **value, GError **error)
+{
+	g_autoptr(GSubprocess) sub = NULL;
+	GError *ierror = NULL;
+	g_autoptr(GBytes) sub_stdout_buf = NULL;
+	g_autoptr(GPtrArray) sub_args = g_ptr_array_new_full(5, g_free);
+
+	g_return_val_if_fail(slot, FALSE);
+	g_return_val_if_fail(key, FALSE);
+	g_return_val_if_fail(value && *value == NULL, FALSE);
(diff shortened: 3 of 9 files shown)


src/bootchooser.c
```diff
@@ -1456,6 +1458,326 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
 	return TRUE;
 }
 
+// TODO add efibootguard bootchooser tests
+
+// Getting slots:
+// TODO call "bg_printenv --raw --part=1" & return k=>v array
```
obbardc, line 1464: Remove these lines.

src/bootchooser.c
```diff
@@ -1456,6 +1458,326 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
 	return TRUE;
 }
 
+// TODO add efibootguard bootchooser tests
```
obbardc, line 1461: Need to add some tests.

src/bootchooser.c
```diff
@@ -1456,6 +1458,326 @@ static gboolean efi_get_state(RaucSlot* slot, gboolean *good, GError **error)
...
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	sub_args = g_ptr_array_new_full(5, g_free);
+	g_ptr_array_add(sub_args, g_strdup(BG_PRINTENV_NAME));
+	g_ptr_array_add(sub_args, g_strdup("--raw"));
+	g_ptr_array_add(sub_args, g_strdup("-p"));
+	g_ptr_array_add(sub_args, g_strdup(slot));
+	g_ptr_array_add(sub_args, NULL);
+
+	sub = r_subprocess_newv(sub_args, G_SUBPROCESS_FLAGS_STDOUT_PIPE | G_SUBPROCESS_FLAGS_STDERR_MERGE, &ierror);
```
obbardc, line 1498: makes sense to rework this function to get of the data all at once perhaps returning a struct

obbardc:
@jluebbe @ejoerns  I've been using this patch in production for a while now, with an old version of RAUC and it's working well. We now need to upgrade to latest RAUC so that means doing something about this patch.

Rather than rebasing this dirty patch, I was considering rewriting this in a shell script with the "custom" bootloader backend. Would that be something you'd accept into RAUC or would you rather I reworked this patch to be less dirty? Happy either way

obbardc:
> @jluebbe @ejoerns I've been using this patch in production for a while now, with an old version of RAUC and it's working well. We now need to upgrade to latest RAUC so that means doing something about this patch.
> 
> Rather than rebasing this dirty patch, I was considering rewriting this in a shell script with the "custom" bootloader backend. Would that be something you'd accept into RAUC or would you rather I reworked this patch to be less dirty? Happy either way

Actually, I had a quick look. Rewriting the logic in a shell script is probably a bit too complex with efibootguard tools. Happy to hear your thoughts.

obbardc:
still needs testing, pushed so i can get some early feedback.

src/bootloaders/efibootguard.c
```diff
@@ -0,0 +1,287 @@
...
+			g_propagate_error(error, ierror);
+			return NULL;
+		}
+		if (!efibootguard_env_get_int(slot->bootname, "USTATE", &slot_ustate, &ierror)) {
+			g_propagate_error(error, ierror);
+			return NULL;
+		}
+		g_debug("efibootguard: slot %s revision=%" G_GUINT64_FORMAT ", ustate=%" G_GUINT64_FORMAT,
+				slot->name, slot_revision, slot_ustate);
+		if (slot_ustate != 3 && (!primary_found || slot_revision > highest_revision)) {
```
obbardc, line 190: nitpick: replace with enum  (reminder; `USTATE_MAP = {"OK": "0", "INSTALLED": "1", "FAILED": "3"}`)

test/bootchooser.c
```diff
@@ -972,6 +972,147 @@ bootname=system1\n";
...
+	rootfs0 = find_config_slot_by_name(r_context()->config, "rootfs.0");
+	g_assert_nonnull(rootfs0);
+	rootfs1 = find_config_slot_by_name(r_context()->config, "rootfs.1");
+	g_assert_nonnull(rootfs1);
+
+	/* part0: REVISION=1 (higher), USTATE=0 (OK)
+	 * part1: REVISION=0 (lower),  USTATE=0 (OK) */
+	test_efibootguard_initialize_state(fixture, 0, 1, 0, 0);
+
+	/* check both slots are considered good (USTATE != 3) */
```
obbardc, line 1052: nitpick: replace with enum

test/bootchooser.c
```diff
@@ -972,6 +972,147 @@ bootname=system1\n";
...
+	g_clear_pointer(&r_context_conf()->configpath, g_free);
+	r_context_conf()->configpath = pathname;
+	r_context();
+
+	rootfs0 = find_config_slot_by_name(r_context()->config, "rootfs.0");
+	g_assert_nonnull(rootfs0);
+	rootfs1 = find_config_slot_by_name(r_context()->config, "rootfs.1");
+	g_assert_nonnull(rootfs1);
+
+	/* part0: REVISION=1 (higher), USTATE=0 (OK)
```
obbardc, line 1048: nitpick: replace USTATE with enum

test/bootchooser.c
```diff
@@ -972,6 +972,147 @@ bootname=system1\n";
...
+	/* check rootfs.0 is primary (highest revision, not failed) */
+	primary = r_boot_get_primary(&error);
+	g_assert_no_error(error);
+	g_assert_nonnull(primary);
+	g_assert_true(primary == rootfs0);
+
+	/* mark rootfs.0 bad: USTATE should become 3 (FAILED) */
+	g_assert_true(r_boot_set_state(rootfs0, FALSE, &error));
+	g_assert_no_error(error);
+	g_assert_cmpint(test_efibootguard_get_var("part0", "USTATE"), ==, 3);
```
obbardc, line 1069: nitpick: replace with enum

test/bootchooser.c
```diff
@@ -972,6 +972,147 @@ bootname=system1\n";
...
+	g_assert_no_error(error);
+	g_assert_true(good);
+
+	/* check rootfs.0 is primary (highest revision, not failed) */
+	primary = r_boot_get_primary(&error);
+	g_assert_no_error(error);
+	g_assert_nonnull(primary);
+	g_assert_true(primary == rootfs0);
+
+	/* mark rootfs.0 bad: USTATE should become 3 (FAILED) */
```
obbardc, line 1066: nitpick: replace with enum

test/bootchooser.c
```diff
@@ -972,6 +972,147 @@ bootname=system1\n";
...
+	g_assert_no_error(error);
+	g_assert_true(good);
+
+	/* rootfs.1 is now the only non-failed slot, so it becomes primary */
+	primary = r_boot_get_primary(&error);
+	g_assert_no_error(error);
+	g_assert_nonnull(primary);
+	g_assert_true(primary == rootfs1);
+
+	/* mark rootfs.0 good again: USTATE should become 0 (OK) */
```
obbardc, line 1085: nitpick: replace with enum

test/bootchooser.c
```diff
@@ -972,6 +972,147 @@ bootname=system1\n";
...
+	g_assert_true(r_boot_set_state(rootfs0, TRUE, &error));
+	g_assert_no_error(error);
+	g_assert_cmpint(test_efibootguard_get_var("part0", "USTATE"), ==, 0);
```
obbardc, line 1088: nitpick: replace with enum

test/bootchooser.c
```diff
@@ -972,6 +972,147 @@ bootname=system1\n";
...
+
+	/* rootfs.0 has higher revision again, so it is primary */
+	primary = r_boot_get_primary(&error);
+	g_assert_no_error(error);
+	g_assert_nonnull(primary);
+	g_assert_true(primary == rootfs0);
+
+	/* mark rootfs.1 as primary: should get REVISION=2 (max+1), USTATE=1 (INSTALLED) */
```
obbardc, line 1096: nitpick: replace USTATE with enum

test/bootchooser.c
```diff
@@ -972,6 +972,147 @@ bootname=system1\n";
...
+	g_assert_true(r_boot_set_primary(rootfs1, &error));
+	g_assert_no_error(error);
+	g_assert_cmpint(test_efibootguard_get_var("part1", "USTATE"), ==, 1);
```
obbardc, line 1099: nitpick: replace with enum

obbardc:
I tested it and it works well !

obbardc:
@jluebbe I rebased onto master and updated the PR. Appreciate it if you can have a look, this is quite an interesting feature for me as it's shipping on a number of devices and I would like to get it into RAUC upstream :-).

jluebbe:
Thanks for updating this PR! A few quick questions:

- How would the kernel/initramfs know which rootfs to use? You document that `rauc.slot=` should be used, but how do you use different kernel commandlines for the A/B slots? In the PR I don't see any code to handle that. See https://github.com/rauc/rauc/pull/1855 for how the EFI backend handles that.
- Do you use this with UEFI Secure Boot? If so, how do you protected the kernel arguments against modification, such as adding `init=/bin/sh`?
- Why do you set the revision explicitly? It seems that `bg_setenv` [supports selecting the partition automatically](https://github.com/siemens/efibootguard/blob/master/docs/TOOLS.md#updating-a-configuration), if I read it correctly.
- If you use `/dev/sda[12]` as the config partitions, where should efibootguard itself be located?

CC @jan-kiszka

jan-kiszka:
I suppose a look at https://archive.fosdem.org/2025/schedule/event/fosdem-2025-4695-generating-immutable-a-b-updatable-securely-booting-debian-images/ and https://gitlab.com/cip-project/cip-core/isar-cip-core could be useful in this context. This is not the only way to do things, but it is a rather broadly used one by now:
- establish a cookie between UKI (normally in as kernel command line arg) and rootfs (some value in some file on the rootfs or the dm-verity root hash)
- select the rootfs based on that cookie in the initramfs

Kernel command line is protected via the signature of the UKI, where both systemd-boot and EBG embed it

jluebbe:
The cookie approach is somewhat problematic when you have the same content in A and B, as efibootguard might select UKI B which then finds and mounts rootfs A. In that situation, it seems to me that userspace can't find out which UKI was actually selected by efibootguard. If userspace has to guess which is the inactive UKI, it could guess wrong and brick the system.

I'd think that selecting an explicit kernel commandline from the UKI pointing to the A or B rootfs via UKI Profiles (`@1`/`@2`) should work with efibootguard as well.

@obbardc Which approach did you use on your current devices?

jan-kiszka:
Yes, the cookie approach implies that rootfs-A == rootfs-B is invalid. We have measures in place to detect that prior to even installing such an update to a device.

Furthermore, EBG provides you the information about the active vs. next slot to use. How to map that on your device slots can vary. We have https://gitlab.com/cip-project/cip-sw-updates/swupdate-handler-roundrobin for that in the SWUpdate ecosystem.

In any case, you can believe me that anything that does not bind UKI and rootfs via some common ID is only making things only more complicated and, thus, more fragile. You have to account for currently used slots, deal with partial swappings of them (UKI storage / rootfs part / persistency snapshot etc.), maybe even need to account for storage topologies or sign multiple images to handle the possible differences in the slot configurations. Been there, seen that - at least I don't want to return.

src/bootloaders/efibootguard.c
```diff
@@ -0,0 +1,287 @@
...
+			g_propagate_error(error, ierror);
+			return NULL;
+		}
+		if (!efibootguard_env_get_int(slot->bootname, "USTATE", &slot_ustate, &ierror)) {
+			g_propagate_error(error, ierror);
+			return NULL;
+		}
+		g_debug("efibootguard: slot %s revision=%" G_GUINT64_FORMAT ", ustate=%" G_GUINT64_FORMAT,
+				slot->name, slot_revision, slot_ustate);
+		if (slot_ustate != R_EFIBOOTGUARD_USTATE_FAILED && (!primary_found || slot_revision > highest_revision)) {
```
obbardc: I wonder if this check for slot status is correct?

Maybe we should invert the logic && check if `slot_ustate` is one of `OK`, `INSTALLED` or `TESTING` to be good? Perhaps this needs to go into a helper function.

cc @jan-kiszka

src/bootloaders/efibootguard.c
```diff
@@ -0,0 +1,287 @@
...
+
+	/* Function must not be called for slots without a bootname! */
+	g_assert_nonnull(slot->bootname);
+
+	if (!efibootguard_env_get_int(slot->bootname, "USTATE", &slot_ustate, &ierror)) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+	g_debug("efibootguard: slot %s ustate=%" G_GUINT64_FORMAT, slot->name, slot_ustate);
+	*good = (slot_ustate != R_EFIBOOTGUARD_USTATE_FAILED);
```
obbardc, line 311: again, I wonder if this check for slot status is correct (see previous comment)

obbardc:
> The cookie approach is somewhat problematic when you have the same content in A and B, as efibootguard might select UKI B which then finds and mounts rootfs A. In that situation, it seems to me that userspace can't find out which UKI was actually selected by efibootguard. If userspace has to guess which is the inactive UKI, it could guess wrong and brick the system.
> 
> I'd think that selecting an explicit kernel commandline from the UKI pointing to the A or B rootfs via UKI Profiles (`@1`/`@2`) should work with efibootguard as well.
> 
> @obbardc Which approach did you use on your current devices?

For this I have been using a custom initrd script to mount the rootfs, e.g. the verity hash or the UUID of the generated image. Agree this isn't the nicest way to do it, open to more friendly suggestions.

In the most simple case, for the sample Buildroot recipe I've created for efibootguard/RAUC, I use partlabel: https://github.com/obbardc/buildroot/blob/wip/obbardc/rauc-efibootguard/board/pc-efibootguard-rauc/post-image.sh

src/bootloaders/efibootguard.c
jan-kiszka: EBG always boots the highest revision. If it detects a failed revision (USTATE = TESTING), it clears its revision to 0. So, you should not have to check USTATE yourself here.

That said, it's sad that we apparently have no library function in EBG to retrieve the active slot number.

src/bootloaders/efibootguard.c
```diff
@@ -0,0 +1,287 @@
...
+	g_ptr_array_add(sub_args, g_strdup(BG_PRINTENV_NAME));
+	g_ptr_array_add(sub_args, g_strdup("--raw"));
+	g_ptr_array_add(sub_args, g_strdup("-p"));
+	g_ptr_array_add(sub_args, g_strdup(slot));
+	g_ptr_array_add(sub_args, NULL);
+
+	g_autoptr(GSubprocess) sub = r_subprocess_newv(
+			sub_args,
+			G_SUBPROCESS_FLAGS_STDOUT_PIPE | G_SUBPROCESS_FLAGS_STDERR_MERGE,
+			&ierror);
```
jan-kiszka, line 37: Why not using the EBG library? I know its API is horrible, but this looks even worse.

jan-kiszka (commented):
@stormc may provide more insides as the author of the SWUpdate integration.

src/bootloaders/efibootguard.c
obbardc: I guess that's a question for @jluebbe if he would be willing to link to the EBG lib in RAUC.

Calling out to other helper binaries is the pattern of how the other RAUC bootloader drivers and I think following that is ugly too, but whatever :-). I guess it makes sense for embedded use-cases to not link to an optional library that only a handful of users may use.

jan-kiszka: dlopen()

obbardc: ah, of course! I will leave the decision to @jluebbe on the direction to take.

docs/integration.rst
```diff
@@ -1360,6 +1362,98 @@ argument in the ``efibootmgr`` call above:
...
+
+The *USTATE* values are encoded as follows:
+
+.. code-block:: text
+
+  OK        = 0
+  INSTALLED = 1
+  TESTING   = 2
+  FAILED    = 3
+  UNKNOWN   = 4
```
stormc: `UNKNOWN` is used internally (e.g. as absence / error marker) and is not a proper external `ustate`.

src/bootloaders/efibootguard.c
stormc: [Here's](https://github.com/sbabic/swupdate/blob/master/bootloader/ebg.c#L531-L559) the example of how it's done in SWUpdate.

src/bootloaders/efibootguard.c
stormc: We do, [`ebg_env_open_current()`](https://github.com/siemens/efibootguard/blob/next/include/ebgenv.h#L84-L88) + [`ebg_env_getglobalstate()`](https://github.com/siemens/efibootguard/blob/next/include/ebgenv.h#L138-L142) which is however currently not wired to the CLI. The mechanics to determine the global state (over the EBG environments) can be found in the [`ebg_env_getglobalstate()`](https://github.com/siemens/efibootguard/blob/next/env/env_api.c#L237-L253) implementation.

> If it detects a failed revision (USTATE = TESTING), it clears its revision to 0. So, you should not have to check USTATE yourself here.

Just to clarify, this is done in the EFI Domain [here](https://github.com/siemens/efibootguard/blob/next/env/fatvars.c#L182-L201) with then setting `revision` to `REVISION_FAILED` = 0 and `ustate` to `USTATE_FAILED` = 3.

Maybe this state diagram as implemented in SWUpdate is of help:

```
 ┌─────────────────┐          ┌─────────────────┐         ┌─────────────────┐
 │                 │          │ SWUpdate:       │         │ EFI Boot Guard: │
 │ ustate    = 0   ├─────────>│ ustate    = 1   ├────────>│ ustate    = 2   │
 │ sysselect = 0   │ Install  │ sysselect = 0   │ Reboot  │ sysselect = 1   │
 └─────────────────┘          └─────────────────┘         └───────┬─────────┘
                                                                  │
 ┌ ─ ─ ─ ─ ─ ─ ─ ─ ┐          ┌─────────────────┐                 │ Boot
      Report                  │ SWUpdate:       │     Success     │
 │    Success/     ├ ─ ─ ─ ─ ─┤ ustate    = 0   │<────────────────◊
      Failure                 │ sysselect = 1   │                 │
 └ ─ ─ ─ ┬ ─ ─ ─ ─ ┘          └─────────────────┘                 │
                                                                  │
 ┌───────┴─────────┐          ┌─────────────────┐                 │
 │ SWUpdate:       │          │ EFI Boot Guard: │     Failure     │
 │ ustate    = 0   │<─────────┤ ustate    = 3   │<────────────────┘
 │ sysselect = 0   │    Boot  │ sysselect = 0   │     Watchdog
 └─────────────────┘          └─────────────────┘      Reboot

 ustate :: OK         → 0
 ustate :: INSTALLED  → 1
 ustate :: TESTING    → 2
 ustate :: FAILED     → 3
 ```

Comment (jan-kiszka) [reply src/bootloaders/efibootguard.c]: > We do, [`ebg_env_open_current()`](https://github.com/siemens/efibootguard/blob/next/include/ebgenv.h#L84-L88) + [`ebg_env_getglobalstate()`](https://github.com/siemens/efibootguard/blob/next/include/ebgenv.h#L138-L142) which is however currently not wired to the CLI. The mechanics to determine the global state (over the EBG environments) can be found in the [`ebg_env_getglobalstate()`](https://github.com/siemens/efibootguard/blob/next/env/env_api.c#L237-L253) implementation.

This alone does not give you a "slot number" or even a path to the target partitions that would take the next update, though.

File: src/bootloaders/efibootguard.c (hunk 14 above)
Comment (obbardc) [reply src/bootloaders/efibootguard.c:311]: seems OK according to the state diagram in https://github.com/rauc/rauc/pull/1450#discussion_r3267747608

File: src/bootloaders/efibootguard.c (hunk 13 above)
Comment (obbardc) [reply src/bootloaders/efibootguard.c]: > If it detects a failed revision (USTATE = TESTING), it clears its revision to 0. So, you should not have to check USTATE yourself here.

In RAUC we can set the USTATE to FAILED in `r_efibootguard_set_state` by calling e.g. `rauc status mark-bad` and then re-read the slots before rebooting, so this check is still valid I think. I added a comment in the code about it.

File: docs/integration.rst (hunk 16 above)
Comment (obbardc) [reply docs/integration.rst]: thanks, removed from docs.

File: src/bootloaders/efibootguard.c (hunk 13 above)
Comment (obbardc) [reply src/bootloaders/efibootguard.c]: I added a diagram like the above into the RAUC documentation. Hope it makes sense.

File: src/bootloaders/efibootguard.c (hunk 15 above)
Comment (obbardc) [reply src/bootloaders/efibootguard.c:37]: I spoke to @jluebbe on Matrix, he said:

> In general, we prefer to use subprocesses as that makes it easier to have robust error handling. A crash in a subprocess doesn't affect RAUC directly, compared to a library.
> Also, it makes sure that the same actions are possible on manually in the shell, e.g. for debugging.

So I will stick to using the subprocess for now, but thanks for the pointers!

Comment (jan-kiszka) [reply src/bootloaders/efibootguard.c:37]: weird

File: src/bootloaders/efibootguard.c [hunk 17]
```diff
@@ -0,0 +1,314 @@
...
+ * @key: variable to fetch
+ * @value: (out) string value (allocated, must be NULL on entry)
+ * @error: (out) error
+ */
+static gboolean efibootguard_env_get(const gchar *slot, const gchar *key, GString **value, GError **error)
+{
+	g_autoptr(GSubprocess) sub = NULL;
+	GError *ierror = NULL;
+	g_autoptr(GBytes) sub_stdout_buf = NULL;
+	g_autoptr(GPtrArray) sub_args = g_ptr_array_new_full(5, g_free);
```
obbardc, line 21: Is it worth switching to `g_ptr_array_new_with_free_func(g_free)` like in the other call to `bg_setenv`? I think not, because here we have a fixed number of args.

src/bootloaders/efibootguard.c
```diff
@@ -0,0 +1,314 @@
...
+	guint64 bg_revision = highest_revision + 1;
+	g_autofree gchar *bg_revision_str = g_strdup_printf("%" G_GUINT64_FORMAT, bg_revision);
+	g_message("efibootguard: setting slot %s state to INSTALLED, revision to %" G_GUINT64_FORMAT,
+			slot->name, bg_revision);
+
+	g_autoptr(GSubprocess) sub = r_subprocess_new(
+			G_SUBPROCESS_FLAGS_NONE,
+			&ierror,
+			BG_SETENV_NAME,
+			"-p", slot->bootname,
```
obbardc, line 272: nitpick: use the long argument names here

docs/integration.rst
```diff
@@ -1360,6 +1362,147 @@ argument in the ``efibootmgr`` call above:
...
+
+   See :ref:`sec-integration-boot-slot-detection` for the full list of slot
+   detection mechanisms.
+
+Before using RAUC with EFI Boot Guard, ensure each partition's environment has
+been initialised. For example, to prefer booting from slot ``0``:
+
+.. code-block:: console
+
+  # bg_setenv -p 0 -s OK -r 1
```
obbardc, line 1503: nitpick: use full parameter names

src/bootloaders/efibootguard.c
```diff
@@ -0,0 +1,314 @@
...
+	g_autoptr(GSubprocess) sub = NULL;
+	GError *ierror = NULL;
+	g_autoptr(GPtrArray) args = g_ptr_array_new_with_free_func(g_free);
+
+	g_return_val_if_fail(slot, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	g_message("efibootguard: setting slot %s state to %s", slot->name, good ? "OK" : "FAILED");
+
+	g_ptr_array_add(args, g_strdup(BG_SETENV_NAME));
```
obbardc, line 140: One thing I noticed while testing something unrelated earlier; if you call `bg_setenv` it overwrites values of any unset arguments with empty/default values. This can be a problem and could cause systems to be unbootable.

@jan-kiszka Am I right in thinking we need to read all config from the slot, overwrite the values we want to change then write back all values? Seems a bit awkward to me ;-)

src/bootloaders/efibootguard.c
```diff
@@ -0,0 +1,314 @@
...
+
+	guint64 bg_revision = highest_revision + 1;
+	g_autofree gchar *bg_revision_str = g_strdup_printf("%" G_GUINT64_FORMAT, bg_revision);
+	g_message("efibootguard: setting slot %s state to INSTALLED, revision to %" G_GUINT64_FORMAT,
+			slot->name, bg_revision);
+
+	g_autoptr(GSubprocess) sub = r_subprocess_new(
+			G_SUBPROCESS_FLAGS_NONE,
+			&ierror,
+			BG_SETENV_NAME,
```
obbardc, line 271: see also previous comment about configuration values going missing: https://github.com/rauc/rauc/pull/1450/changes#r3386537804

src/bootloaders/efibootguard.c
jan-kiszka: We designed our interface very carefully to contain as many surprises for users as possible :laughing:: You need `--preserve` for the desired behaviour.

obbardc: Makes sense; thank you! :-)

PR #1456, opened 2024-07-02
src/main: fix wrong log field numbers in r_event_log_booted

ejoerns:
The bundle hash must be set for `GLogField [7]` (`"BUNDLE_HASH"`), not `[5]` (`"SLOT_BOOTNAME"`).

Note that this left the `"BUNDLE_HASH"` field with a NULL pointer which the default log writer seems to handle but which fails when running in a systemd context where `g_log_writer_journald()` is used, too. There, the call to `strlen()` with a `NULL` pointer causes a segfault in RAUC.


hnez (approved):
I've tested the change on a device where the `rauc` service would previously fail with a `SIGSEGV`.
This change fixes said segmentation fault.

jluebbe (approved):

Issue #1461, opened 2024-07-15
GLib assertion fails when converting `tar.gz` slots using Casync


matteocarnelos:
**Describe the bug**
When converting a conventional RAUC bundle to a Casync bundle using `rauc convert`, the conversion of `tar.gz` slots fail with the following error:

```
(rauc:782068): GLib-CRITICAL **: 19:29:58.598: g_atomic_ref_count_dec: assertion 'old_value > 0' failed
```

With other slot types, such as `.ext4`, this error is not encountered.

**Background information**
RAUC built from source at commit cf4d63a46c6b0a7a5cba3b039d146bbd6fd59665 using Meson.

**To Reproduce**
1. Create a conventional RAUC bundle containing a `tar.gz` slot
2. Convert it to a Casync bundle using `rauc convert`

**Expected behavior**
The conversion should succeed.

**Logs**
Full console log with debug enabled:

```
$ rauc -d convert --cert=certfile.cert.pem --key=keyfile.key.pem --keyring=keyring.crt.pem conventional-bundle.raucb casync-bundle.raucb
rauc-Message: 19:38:53.649: Debug log domains: 'rauc'
(rauc:782129): rauc-DEBUG: 19:38:53.649: v1.11.1-267-gcf4d63a4 create=1 emmc-boot=1 gpt=0 json=1 network=1 service=1 streaming=1
(rauc:782129): rauc-DEBUG: 19:38:53.652: convert start
(rauc:782129): rauc-DEBUG: 19:38:53.652: input bundle: conventional-bundle.raucb
(rauc:782129): rauc-DEBUG: 19:38:53.652: output bundle: casync-bundle.raucb
rauc-Message: 19:38:53.652: Reading bundle: conventional-bundle.raucb
(rauc:782129): rauc-DEBUG: 19:38:53.657: Found valid CMS data
rauc-Message: 19:38:53.659: Verifying bundle signature...
rauc-Message: 19:38:53.664: Verified inline signature by 'O = SW, CN = Development'
(rauc:782129): rauc-DEBUG: 19:38:53.666: Got 3 elements for trust chain
rauc-Message: 19:38:53.666: Verifying bundle payload...
(rauc:782129): rauc-DEBUG: 19:38:53.666: Hash verification SHA256, data blocks 25662.
(rauc:782129): rauc-DEBUG: 19:38:53.666: Using 3 hash levels.
(rauc:782129): rauc-DEBUG: 19:38:53.666: Data size: 105111552 bytes.
[...]

PR #1462, opened 2024-07-15
Memory handling fixes for casync archive conversion

ejoerns:
Fixes memory leakages, over-allocations, and a sort of double-free crash for 'rauc convert' calls with archives as input images.

Fixes: #1461

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -204,7 +204,7 @@ static gboolean casync_make_arch(const gchar *idxpath, const gchar *contentpath,
 	gboolean res = FALSE;
 	g_autoptr(GPtrArray) args = g_ptr_array_new_full(5, g_free);
 	g_autoptr(GPtrArray) iargs = g_ptr_array_new_full(15, g_free);
-	const gchar *tmpdir = NULL;
+	g_autofree gchar *tmpdir = NULL;
 
 	g_return_val_if_fail(idxpath != NULL, FALSE);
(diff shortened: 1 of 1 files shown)


src/bundle.c
```diff
@@ -254,12 +255,14 @@ static gboolean casync_make_arch(const gchar *idxpath, const gchar *contentpath,
...
 	g_ptr_array_add(iargs, NULL);
+	istr = (gchar**) g_ptr_array_free(iargs, FALSE);
+	iargs = NULL; /* prevent autoptr clenaup */
 
 	/* Outer process calll */
 	g_ptr_array_add(args, g_strdup("fakeroot"));
 	g_ptr_array_add(args, g_strdup("sh"));
 	g_ptr_array_add(args, g_strdup("-c"));
-	g_ptr_array_add(args, g_strjoinv(" ", (gchar**) g_ptr_array_free(iargs, FALSE)));
+	g_ptr_array_add(args, g_strjoinv(" ", istr));
```
jluebbe, line 265: Why not
```suggestion
	g_ptr_array_add(args, g_strjoinv(" ", iargs->pdata));
```
as we already have a NULL ptr at the end of `iargs`? This would avoid the complexity of the temporary `istr`.

ejoerns: Indeed, this is way simpler. Fixed.

jluebbe (approved):

PR #1463, opened 2024-07-16
src/main: fix leaking strings when printing hooks with 'rauc info'

ejoerns:
The GPtrArrays are reference-counted correctly, but since they do not have a free func set, the newly allocated gchar's will not be freed.

Fix this by just using string literals instead.


jluebbe:
All these strings are copies of literals. Why don't we drop the g_strdup instead?

ejoerns:
> All these strings are copies of literals. Why don't we drop the g_strdup instead?

Since these are used in the method only, this should be ok. Will change.

jluebbe (approved):

PR #1466, opened 2024-07-18
Some fixes for meson dist and running tests

ejoerns:
* Removes outdated `version-gen`
* Add proper way to get current meson build directory for tests
* do not preserve bundle file permission during copy to tmpdir


ejoerns:
* Re-added `version-gen` since `conf.py` turned out to still be a user of `.tarball-version`.
* Add more explicit handling of `MESON_BUILD_DIR` (raise if unset, make absolute path if relative)
* set `MESON_BUILD_DIR` for `./qemu-test shell`

PR #1468, opened 2024-07-18
fix leak in tests and run the full test suite with ASAN and UBSAN

jluebbe:


ejoerns (approved):
Fixed a small typo in the commit message. Looks good otherwise.

PR #1473, opened 2024-07-22
src/nbd: warn about slow installation when using HTTP/1

jluebbe:
It may not be obvious that using HTTP/2 is important to get good performance when using streaming, so add a warning if an earlier version is used.

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -697,6 +697,7 @@ static gboolean finish_configure(struct RaucNBDContext *ctx, struct RaucNBDTrans
 	gboolean res = FALSE;
 	CURLcode code;
 	long response_code = 0;
+	long http_version = 0;
 	const char *effective_url = NULL;
 	g_auto(GVariantDict) dict = G_VARIANT_DICT_INIT(NULL);
 	g_autoptr(GVariant) v = NULL;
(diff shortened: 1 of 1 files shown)


src/nbd.c
```diff
@@ -753,6 +754,13 @@ static gboolean finish_configure(struct RaucNBDContext *ctx, struct RaucNBDTrans
 		ctx->url = g_strdup(effective_url);
 	}
 
+	code = curl_easy_getinfo(xfer->easy, CURLINFO_HTTP_VERSION, &http_version);
+	if (code == CURLE_OK) {
+		if (http_version == CURL_HTTP_VERSION_1_0 ||
```
a3f, line 759: Does HTTP 1.0 even have range requests? Would this code work without?

src/nbd.c
```diff
@@ -753,6 +754,13 @@ static gboolean finish_configure(struct RaucNBDContext *ctx, struct RaucNBDTrans
...
+		    http_version == CURL_HTTP_VERSION_1_1)
+			g_warning("using HTTP/1 for streaming, expect slow installation; enable HTTP/2 if possible");
```
a3f, line 761: Is this because of the reuse of the socket or are there other factors?

src/nbd.c
jluebbe: Server may respond strangely, so I'd include 1.0 just in case.

src/nbd.c
jluebbe: HTTP/2 keeps the connection open and allows multiple pipelined requests. For HTTP/1.1, even with keep-alive, connections may need to be reopened multiple times, which takes time, especially with TLS.

Bastian-Krause (approved):

PR #1478, opened 2024-07-25
improve log messages related to block hash indices

jluebbe:
As building a hash index can take multiple minutes, we should tell the
user by default. Usually, this should only happen for the initial
installations, as the index is stored after writing the image.

Also reword the chunk count check, which can trigger for partitions as
well as for images, to use 'image/partition' instead of 'data file'.

### DIFF
--- a/src/hash_index.c
+++ b/src/hash_index.c
@@ -77,7 +77,7 @@ static GBytes *hash_file(int data_fd, guint32 count, GError **error)
 				g_set_error(error,
 						R_HASH_INDEX_ERROR,
 						R_HASH_INDEX_ERROR_SIZE,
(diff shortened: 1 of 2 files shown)


Bastian-Krause (approved):
Looks good. I am unsure whether these messages should also be improved, though:
https://github.com/rauc/rauc/blob/cb34e09d78b61b1d9f337377afb4a9fe2eb60026/src/hash_index.c#L77-L80
https://github.com/rauc/rauc/blob/cb34e09d78b61b1d9f337377afb4a9fe2eb60026/src/hash_index.c#L533-L536

Bastian-Krause (approved):

Issue #807, opened 2021-11-10
Multiple variants in DTB seems not handled properly


hongkongkiwi:
Hi, in my DTS/DTB I have the following:
`compatible = "company1,company1-soc","company2,ourproduct";`

This works fine in linux and shows as a NULL character between in linux, e.g:
```
# cat /proc/device-tree/compatible
company1,company1-soc<NULL>company2,ourproduct
```
So, I might use this in bash to display with a space
`cat /proc/device-tree/compatible | xargs -0`

However, when using `rauc status`, it ignores everything after the first null.
```
# rauc status
Variant:     company1,company1-soc
......
```

So, I would expect to be able to see both, perhaps replacing null with a special char or something else? Lets say _ then in my bundle variant I can use "[slot.myslot.company1,company1-soc_company2,ourproduct".

As a workaround, I can use variant-file for this, but I prefer if I could use the dtb directly. 

I am not sure how common this use case is, but if it's planned not to fix, I think it would be good to add this into the docs somewhere to say the expected behaviour if the dtb is setup like this.

jluebbe:
As the most specific compatible comes first, you can list multiple variants in the manifest (and refer to the same file from several variants). Documenting this better would be useful, though.

hongkongkiwi:
I see, that makes sense, thanks.

Issue #690, opened 2021-01-25
Are there any error codes?


krutztq:
Hi rauc Team,

I have  a question:
Are there any error codes or did I missed something? If not, are there any error codes planned?

I want read the latest error with my application and localize it for the GUI. My app has it's own localization workflow therefore a gettext solution would not fit my concept.

'rauc install' with an error always exits with error code 1, regardless of the error.
I found the dbus LastError property, but the comparison of strings is not robust enough for me.

I can handle the errors from the hook script well, but I have to handle them separately.
Here, too, it would be nice if you could get the exit code of the hook script from rauc itself.

Or is there another solution?

Thanks in advance

jluebbe:
Sorry, but there are currently no external error codes. My preferred approach would be to define error codes on the DBus level. @ejoerns took a quick look at that some time ago. I think that would allow us to make the error enums visible to DBus clients. Enrico, is that correct?

For the exit code of the hook script, I think we could add an optional details dictionary to a replacement of the LastError property.

krutztq:
Thank you for the quick reply. DBus level error codes would be great.
An additional LastErrorCode property (backwards compatible) with a small area for hook exit codes would be fine with me as well.

jluebbe:
As this feature is not on our roadmap so far, I can only offer to review any PRs which work towards this.

ejoerns:
@jluebbe To the best of my knowledge, there is no concept of error codes in dbus. What we have is a message type 'error' and an `ERROR_NAME` field [1].
What I had a look into is using explicit errors instead of automatically mapped ones. By default, a Glib error (GError) passed over D-Bus will result in the error MESSAGE being set to the error message of the GError, but the ERROR_NAME will be something like `org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code30` which is a simple serialized form of the original glib error, but hard to parse for non-glib applications. Thus better use explicit error names, matching the interface naming convention, like: `de.pengutronix.rauc.Error.InavlidSiganture`.

In general, I really appreciate the idea to use error-specific exit codes. But, this would pretty sure be independent of the codes we currently use for error reporting as these have the concept of a (domain, code) tuple that we would have to map to a single code being returned.

This is probably also independent from the Glib mappings as these are for errors passed over dbus only and allow reflecting the code / domain tuble by using `..Error.<Code>.<Name> strings.`


[1] https://dbus.freedesktop.org/doc/dbus-specification.html

[...]

PR #1501, opened 2024-08-24
system_variant as g_array

js731ca:
Convert the `system_variant` kept in the context from a single element to an array; so that a bundle can specify a 'broader' compatibility than just the most exact one.

This PR targets the `variant-dtb=true` feature, where rauc now reads in the full list of compatibles (and not just the very first one...) stored in `/sys/firmware/devicetree/base/compatible` [1] and compares the variant set on a slot in the bundle against *each* of the elements in the devices compatibility hierarchy.

A usage scenario/example: 
On a system with secure-boot, where the bootloader has to be signed against one key/cert out of a set of keys/certs, which can be (individually) revoked.
The running bootloader "would know" which keys/certs are still "open" and append them as of-fixup to the devicetree/compatible.
Bundles would have a variant set on *every* slot (e.g. not leaving an unset/default), where the bootloader.variant could reflect the signature as a compatible; which would then ensure that the bundle installation fails if there are no compatible elements in the bundle - ensuring that the device could not get bricked.
```
# example bundle:
[update]
compatible=ACME Coorp
version=4.3.2
description=acme corp embedded device update bundle
build=20240823202207
...
[image.bootloader.acme,dev001-K2]
sha256=f734326e25a2e1f1b516d7013dd1aabe145441fc61ee5e508e6e83f1542b2a8c
size=1597980
filename=barebox-acme-dev001-s.img

# the bootloader would know which keys/certs are revoked, and fixup the compatible accordingly:
$> cat /sys/firmware/devicetree/base/compatible
acme,dev001-revP\0acme,dev001-K1\0acme,dev001-K3\0acme,dev001\0fsl,imx8mm\0

# so in this scenario, the bundle would just refuse to install
# -> not bricking the device
$> rauc install /tmp/acme-bootloader-bundle-dev001-20240823202207.raucb 
installing                                                                                                       
  0% Installing                                                                                                  
  0% Determining slot states                                                                                     
 10% Determining slot states done.                                                                               
 10% Checking bundle                                                                                             
 10% Verifying signature
 20% Verifying signature done.                                                                                   
 20% Checking bundle done.                                                                                       
 20% Checking manifest contents                                                                                  
 30% Checking manifest contents done.
 30% Determining target install group
 40% Determining target install group done.
100% Installing failed.
LastError: Installation error: Failed to find matching variant of image for bootloader
Installing `/tmp/acme-bootloader-bundle-dev001-20240823202207.raucb` failed
```

Note that this implementation is currently only focused on the concept/feature, and would need to be covered by more tests...
But first -> RFC :-)

Also: this "one-dimensionality" could be carried over to the other two: `variant-file` and `variant-name` features - which would make them more useful/powerful

[1] https://www.kernel.org/doc/html/latest/devicetree/usage-model.html#platform-identification

### DIFF
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -37,7 +37,7 @@ typedef struct {
 	gchar *system_compatible;
 	gchar *system_min_bundle_version;
 	RConfigSysVariant system_variant_type;
-	gchar *system_variant;
+	GArray *system_variant;
 	gchar *system_bootloader;
 	gchar *system_bb_statename;
 	gchar *system_bb_dtbpath;
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -801,6 +801,7 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 	g_key_file_remove_key(key_file, "system", "activate-installed", NULL);
 
 	c->system_variant_type = R_CONFIG_SYS_VARIANT_NONE;
+	c->system_variant = g_array_new(FALSE, FALSE, sizeof(gchar *));
 
 	/* parse 'variant-dtb' key */
 	dtbvariant = g_key_file_get_boolean(key_file, "system", "variant-dtb", &ierror);
@@ -835,7 +836,7 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 		}
 
 		c->system_variant_type = R_CONFIG_SYS_VARIANT_FILE;
-		c->system_variant = g_steal_pointer(&variant_data);
+		g_array_append_val(c->system_variant, variant_data);
 	}
 
 	/* parse 'variant-name' key */
@@ -858,7 +859,7 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 		}
 
 		c->system_variant_type = R_CONFIG_SYS_VARIANT_NAME;
-		c->system_variant = g_steal_pointer(&variant_data);
+		g_array_append_val(c->system_variant, variant_data);
 	}
 
 	/* parse data/status location
@@ -1138,7 +1139,8 @@ void free_config(RaucConfig *config)
 
 	g_free(config->system_compatible);
 	g_free(config->system_min_bundle_version);
-	g_free(config->system_variant);
+	if (config->system_variant)
+		g_array_free(config->system_variant, TRUE);
 	g_free(config->system_bootloader);
 	g_free(config->system_bb_statename);
 	g_free(config->system_bb_dtbpath);
--- a/src/context.c
+++ b/src/context.c
@@ -262,21 +262,37 @@ static gboolean launch_and_wait_variables_handler(gchar *handler_name, GHashTabl
 	return TRUE;
 }
 
-static gchar* get_system_dtb_compatible(GError **error)
+static gboolean get_system_dtb_compatible(GError **error)
 {
 	gchar *contents = NULL;
 	GError *ierror = NULL;
+	gsize length;
 
-	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
-	if (!g_file_get_contents("/sys/firmware/devicetree/base/compatible", &contents, NULL, &ierror)) {
+	if (!g_file_get_contents("/sys/firmware/devicetree/base/compatible", &contents, &length, &ierror)) {
 		g_propagate_error(error, ierror);
-		return NULL;
+		return FALSE;
 	}
 
 	g_assert_nonnull(contents); /* fixes scan-build false positive */
+	const char *start = contents;
+	const char *end = contents;
 
-	return contents;
+	while (end < contents + length) {
+		if (*end == '\0') {
+			size_t token_length = end - start;
+			if (token_length > 0) {
+				gchar *substring = g_strndup(start, token_length);
+				g_array_append_val(context->config->system_variant, substring);
+				g_debug("added devicetree compatible: %s", substring);
+			}
+			start = end + 1; // Move start to the next character after '\0'
+		}
+		end++;
+	}
+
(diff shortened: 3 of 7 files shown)


jluebbe:
> Bundles would have a variant set on _every_ slot (e.g. not leaving an unset/default), where the bootloader.variant could reflect the signature as a compatible; which would then ensure that the bundle installation fails if there are no compatible elements in the bundle - ensuring that the device could not get bricked.

If I understand correctly, you want to use RAUC's variant mechanism to reject installation of the bundle is specific cases. While it could be used for that, it's actually intended to be used to *select* one one several alternative images in the bundle which perform the same function in the system (i.e. bootloaders for different SoCs used with a common kernel and root filesystem).

RAUC's mechanism to allow/reject installation of a bundle is the compatible: either with fixed string or by using the [`install-check` hook](https://rauc.readthedocs.io/en/latest/using.html#install-hooks). When using a hook, the script can introspect the running system (e.g. which fuses are burned) and then decide based on logic in the script.

Bundle [meta-data](https://rauc.readthedocs.io/en/latest/reference.html#meta-label-sections) could be exposed to this hook as well. You'd have something like
```ini
[meta.hab]
signature-key=3
```
and then check in the `install-check` hook whether that key is still valid.

It may be useful to define a new hook, which doesn't replace the built-in compatible check. Instead it would only run if the build-in one was successful to implement additional restrictions.

js731ca:
Thanks for the input!
I'll take a look at the hook implementation

the "but" here is that their error response isn't "as nice" as an integrated feature would be - looking back at the stop-gap solution for the min-version check, which is/was also a hook: rauc would start installation and then in-medias-res just fail, with no good error response (unless one knew to look in the rauc.service log, and for what...) 
here the response is direct and pretty clear: incompatible, stop!

for the usecase described: yes, that's a way to (ab)use this feature, IMO it could be generally usefull though: think about a set of boards that define a compatible hierarchy:
acme,device-rev01; acme,device; fsl,imx8mm
acme,device-rev42; acme,device; fsl,imx8mm
acme,fooo-rev44; acme,foo; fsl,imx8mm
here a bundle slot compatible check could say "ok, for acme,device, take this bootloader"; or another one that runs on all "fsl,imx8mm" (without being the no-variant-set default)

ejoerns:
> Thanks for the input! I'll take a look at the hook implementation
> 
> the "but" here is that their error response isn't "as nice" as an integrated feature would be - looking back at the stop-gap solution for the min-version check, which is/was also a hook: rauc would start installation and then in-medias-res just fail, with no good error response (unless one knew to look in the rauc.service log, and for what...) here the response is direct and pretty clear: incompatible, stop!

This is not fully correct I guess. From https://rauc.readthedocs.io/en/latest/using.html#install-hooks:

> If available, RAUC will use the last line printed to standard error by the hook executable as the rejection reason message and provide it to the user:

And the final error returned over D-Bus should include this error as well.

Thus the only thing that might be missing is having distinguishable error codes as discussed in https://github.com/rauc/rauc/issues/690. But that would be the same with your approach.

> for the usecase described: yes, that's a way to (ab)use this feature, IMO it could be generally usefull though: think about a set of boards that define a compatible hierarchy: acme,device-rev01; acme,device; fsl,imx8mm acme,device-rev42; acme,device; fsl,imx8mm acme,fooo-rev44; acme,foo; fsl,imx8mm here a bundle slot compatible check could say "ok, for acme,device, take this bootloader"; or another one that runs on all "fsl,imx8mm" (without being the no-variant-set default)

We had a similar expectation in https://github.com/rauc/rauc/issues/807 and I remember vaguely having discussed something similar already. The thing is that adding this would increase complexity in handling variants, etc. Thus maybe we should find out first if there are actual use cases that are hard to solve with the existing mechanism..?

jluebbe:
@js731ca You can use https://github.com/rauc/rauc/pull/1506 to have access to the manifest meta-data in the install-check hook.

js731ca:
> We had a similar expectation in #807 and I remember vaguely having discussed something similar already.

rauc only looks at the very first compatible = the most specific one

>  The thing is that adding this would increase complexity in handling variants, etc. Thus maybe we should find out first if there are actual use cases that are hard to solve with the existing mechanism..?

agreed, "complexity is the enemy" :-D
enabling rauc to look through each element in the  devicetree/base/compatible would have been usefull for #807... since that is what this user was asking for: "how to have a variant targeting not-the-first compatible"

js731ca:
@bith3ad has the bigger picture of "our" usecases, and we're going to handle the problem via the infrastructure and signing-providers = outside of rauc, but inside meta-rauc/yocto

-> so the immediate need for this PR dissolved, should we close it or continue on since it "might still be useful"? (-:

jluebbe:
> agreed, "complexity is the enemy" :-D
> enabling rauc to look through each element in the devicetree/base/compatible would have been useful for #807... since that is what this user was asking for: "how to have a variant targeting not-the-first compatible"

Yes, but "a variant targeting not-the-first compatible" may not be the best solution for their use-case when considering RAUC's design and long-term maintenance.

> @bith3ad has the bigger picture of "our" usecases, and we're going to handle the problem via the infrastructure and signing-providers = outside of rauc, but inside meta-rauc/yocto

Please check if you can handle it with an `install-check` hook and meta-data.

> -> so the immediate need for this PR dissolved, should we close it or continue on since it "might still be useful"? (-:

I'll close this for now. Some things that would need to be solved before merging something similar:

- [ ] analyze how multiple variants should interact with multiple compatibles (sooner or later someone would ask why multiple variants are supported but not multiple compatibles)
- [ ] write documentation, with guidance on how to use multiple variants, especially with regards to forwards/backwards compatibility
- [ ] decide how multiple variants interact with bundles with/without default images (i.e. is there an implicit "default" variant at the end of the list?)
- [ ] decide how to expose the variant list via D-Bus and via environment vars (which currently use only a single string)
- [ ] review the design for corner cases and possibly surprising behaviour

PR #1503, opened 2024-08-28
improve rauc info/status/help output by avoiding tabs

jluebbe:
Using "\t" in the strings makes them hard to align in the source code and
results in tab characters in the output, which tends to disrupt the
alignment when copying the output somewhere else. So, keep it simple by
aligning using spaces in the source as well.


ejoerns (approved):
:+1:

PR #1505, opened 2024-08-29
expose manifest meta-data in handler environment

jluebbe:
The manifest meta-data can be useful in the pre/post-install handlers for customization. By exposing them as environment variables it's no longer necessary to parse files from the bundle in the script.

### DIFF
--- a/include/shell.h
+++ b/include/shell.h
@@ -1,0 +1,16 @@
+#pragma once
+
+#include <glib.h>
+
+#include "manifest.h"
+
+/**
+ * Add the meta-data contents from a manifest as newly allocated strings to a
+ * GPtrArray, formatted for use as shell variables.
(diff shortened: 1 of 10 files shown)


jluebbe:
Testing features like this is so much easier with pytest now. :)

include/shell.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <glib.h>
+
+#include "manifest.h"
+
+/**
+ * Add the meta-data contents from a manifest to a GPtrArray, formatted for use
+ * as shell variables.
```
ejoerns, line 9: For clarity: Maybe note that newly allocated strings will be added to the GPtrArray?

ejoerns (approved):

PR #1506, opened 2024-08-29
expose manifest meta-data in install-check hook environment

jluebbe:
This uses the same variable names (and implementation) as `rauc info`.

Only the install-check hook is updated for now, as the slot hooks
currently don't have access to the manifest. We'd need to pass the
manifest or prepared variables down through all the update handles, but
that's a larger change.

Also add a helper class to build custom bundles in pytest test cases, based
on code from my artifact tests.

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -1300,6 +1300,7 @@ static gboolean unpack_archive(RaucImage *image, gchar *dest, GError **error)
  * @param hook_cmd first argument to the hook script
  * @param image image to be installed (optional)
  * @param slot target slot
+ * @param variables extra environment variables, or NULL
  * @param error return location for a GError, or NULL
  *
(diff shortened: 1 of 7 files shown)


src/update_handler.c
```diff
@@ -1300,6 +1300,7 @@ static gboolean unpack_archive(RaucImage *image, gchar *dest, GError **error)
  * @param hook_cmd first argument to the hook script
  * @param image image to be installed (optional)
  * @param slot target slot
+ * @param variables extra environment variables of NULL
```
ejoerns, line 1303:
```suggestion
 * @param variables extra environment variables, or NULL
```

ejoerns (changes requested):
Minor typo, looks good otherwise.

ejoerns (approved):

PR #1507, opened 2024-08-29
check if encryption recipient cert file containes at least one certificate

Fotte:
When rauc is used to encrypt a bundle, and a certificate file is provided which does not contain any certificate, then the encryption process was not stopped. Check the number of certificates when reaching end of certificate file. If this is zero, display an error and abort.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -316,7 +316,8 @@ static STACK_OF(X509) *load_certs_from_file(const gchar *certfile, GError **erro
 		if (cert_x509 == NULL) {
 			err = ERR_peek_last_error();
 
-			if (ERR_GET_REASON(err) == PEM_R_NO_START_LINE) {
+			if (ERR_GET_REASON(err) == PEM_R_NO_START_LINE &&
+			    sk_X509_num(certs) != 0) {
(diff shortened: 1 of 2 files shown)


ejoerns:
Tested locally and make RAUC successfully error-out with 

    Failed to parse cert file 'test/openssl-enc/keys/ecc/public-key-000.pem': Expecting: CERTIFICATE

@Fotte Could you add you [Signed-off-by](https://rauc.readthedocs.io/en/latest/contributing.html#sec-dco) and maybe adapt the commit message to the prefix style we use?

A test case would be great, too. (e.g. by encrypting with the pubkey file I used in my test).

Fotte:
@ejoerns: Will do.

Fotte:
@ejoerns:
Fixed the style issue. It took me some time to realize that i appended a newline at the end of the test_encrypt.py file.... it should be fixed now. Thanks for your support and patience.

ejoerns (approved):

jluebbe (approved):

jluebbe:
I've adjusted the commit message to our conventions.

PR #1509, opened 2024-08-30
src/bundle: fix hash_index fd close handling

### LINKED ISSUE #334: Calling rauc install from udev is not working
Hi!

I currently use `rauc` (0.1.1) to update some bundle on a custom distrib made with yocto. The bundle comes from a USB device. So, I add a udev rule to get my bundles as soon as the key is inserted on the device.

My hook contains (version A):

```bash
echo ${PATH} #/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin
echo `fw_printenv slot` #slot=A
rauc_output=`rauc install ${bundle}`
echo ${PATH} #/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin
```

But with this version, it fails with `starting fw_setenv failed: Failed to execute child process "fw_setenv" (No such file or directory)` if launched via udev (If I launch manually the script it works).

If I use this version:

```bash
echo ${PATH} #/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin
echo `fw_printenv slot` #slot=A
rauc_output=`PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin rauc install  ${bundle}`
echo ${PATH} #/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin
```

It works (both manually and with udev)

I don't really know what the heck is happening here. The PATH is ok before and after the install, but... not when calling rauc via udev. Moreover, `fw_setenv` is correctly found if I call it just before `rauc`.

The code to call `fw_setenv` in rauc is (4b6453e8678a48aa793c41871e3bfcfdbafce881):

```c
	sub = g_subprocess_new(G_SUBPROCESS_FLAGS_NONE, &error, UBOOT_FWSETENV_NAME,
			       key, value, NULL);
	if (!sub) {
		g_warning("starting fw_setenv failed: %s", error->message);
		g_clear_error(&error);
		goto out;
	}
```

`g_subprocess_new` is searching in `PATH`. So this should be ok.

Any clue?

ejoerns:
When calling `r_hash_index_open()`, the ownership of the file descriptor provided should be handed over to the hash index and will be later closed in `r_hash_index_free()`.

Thus, the file descriptor provided must not be closed by the calling code anymore.

In `generate_adaptive_data()` however, the provided pointer was closed manually in the later error case, which would lead to a double-call of `g_close()` on the same pointer happened.

This was wrong anyway but results in a critical error in RAUC since glib version 2.75.0 which comes with a pickier `g_close()` handling: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2964

Fix this and add a hint to the function documentation.

Fixes: https://github.com/rauc/meta-rauc/issues/334

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -396,7 +396,7 @@ static gboolean generate_adaptive_data(RaucManifest *manifest, const gchar *dir,
 				g_autofree gchar *indexname = g_strconcat(image->filename, ".block-hash-index", NULL);
 				g_autofree gchar *indexpath = g_build_filename(dir, indexname, NULL);
 				g_autoptr(RaucHashIndex) index = NULL;
-				int fd = -1;
+				g_auto(filedesc) fd = -1;
 
 				if (image_is_archive(image)) {
 					g_warning("Generating block hash index requires a block device image but %s looks like an archive", image->filename);
@@ -419,7 +419,6 @@ static gboolean generate_adaptive_data(RaucManifest *manifest, const gchar *dir,
 							error,
 							ierror,
 							"Failed to generate hash index for %s: ", image->filename);
-					g_close(fd, NULL);
 					return FALSE;
 				}
 
@@ -428,7 +427,6 @@ static gboolean generate_adaptive_data(RaucManifest *manifest, const gchar *dir,
 							error,
 							ierror,
 							"Failed to write hash index for %s: ", image->filename);
-					g_close(fd, NULL);
 					return FALSE;
 				}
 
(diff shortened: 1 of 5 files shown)


jluebbe (approved):

jluebbe (changes requested):
Hmm, I'm not convinced this is correct. The `g_close` in `r_hash_index_free` could be called via the `RaucHashIndex` autoptr in an error case in `r_hash_index_open_slot`, leadding to the double close.

jluebbe:
We discussed this internally. The `r_hash_index_open` should dup the passed file descriptor and only close the copy. Then caller is responsible for closing his own copy.

jluebbe (approved):

PR #1510, opened 2024-09-09
minor updates to the test infrastructure

jluebbe:
This helped with testing the artifact update code (#1301).

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -441,10 +441,13 @@ def rauc_dbus_service_with_system_adaptive(tmp_path, dbus_session_bus, create_sy
 
 
 class Bundle:
-    def __init__(self, tmp_path, name):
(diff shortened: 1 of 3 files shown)


test/conftest.py
```diff
@@ -421,10 +421,13 @@ def rauc_dbus_service_with_system_adaptive(tmp_path, dbus_session_bus, create_sy
 
 
 class Bundle:
-    def __init__(self, tmp_path, name):
+    def __init__(self, tmp_path, name=None):
```
ejoerns, line 424: Wouldn't `[..], name="test")` be more simple here?

jluebbe: With `=None`, the function itself can distinguish between the default and explicit cases, which is often useful to when refactoring.

ejoerns (approved):

PR #1523, opened 2024-09-14
src/manifest: throw parser error for invalid slot hook combination

ejoerns:
When using an `install` hook, the `pre-install` and `post-install` hooks will not be called. To avoid confusion, prevent generating bundles having both 'install' and 'pre-install' (or 'post-install') hook set for the same slot.

Fixes https://github.com/rauc/meta-rauc/issues/321

### DIFF
--- a/include/manifest.h
+++ b/include/manifest.h
@@ -132,6 +132,20 @@ gboolean check_manifest_external(const RaucManifest *manifest, GError **error)
 G_GNUC_WARN_UNUSED_RESULT;
 
 /**
+ * Check a new manifest for consistency during bundle creation.
+ *
+ * This is used for checks that could otherwise break compatibility of new RAUC
+ * versions with old bundles.
+ *
+ * @param manifest Pointer to the manifest to check
+ * @param error return location for a GError, or NULL
+ *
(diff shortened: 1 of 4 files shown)


jluebbe:
Wouldn't it be better to have the check in `check_manifest_common`? https://github.com/rauc/rauc/blob/master/src%2Fmanifest.c#L402

Perhaps we should enforce it only when creating a bundle, as otherwise it might reject previously accepted bundles when downgrading.

ejoerns:
> Wouldn't it be better to have the check in `check_manifest_common`? https://github.com/rauc/rauc/blob/master/src%2Fmanifest.c#L402
> 
> Perhaps we should enforce it only when creating a bundle, as otherwise it might reject previously accepted bundles when downgrading.

Yes, to ensure it is only enforced when creating a bundle I put it into the image parsing. If you prefer to have explicit support for creation vs. installation checking in `check_manifest_*()` methods, I could rework this.

jluebbe:
I've added a fixup! commit to implement the changes I requested. @ejoerns Please review and squash if you're OK with the change.

jluebbe (approved):

PR #1525, opened 2024-09-19
Support Installation Date-based Slot Selection (A/B/C Update)

ejoerns:
So far, RAUC theoretically supported having three slot groups (e.g. `rootfs.0`, `rootfs.1`, `rootfs.2`), but the slot selection algorithm prevented their use for an A/B/C update.

The algorithm simply selected the 'first' inactive slot group.
The 'first' group is implementation-defined by the data type used (i.e. by the GHashTable).

As a result, updating always alternated between two slot groups.

This PR now introduces a slightly more sophisticated slot selection algorithm which always selects the slot group with the oldest installation date.
Selecting the oldest installation date automatically ensures that all slots get their turn since once a slot is updated, it has the most recent timestamp and will be selected last.
A slot that has no installation date yet is considered outdated.

Not that the same behavior can also be used in an A/B setup together with `rauc.external` to ensure that two consecutive calls of `rauc install` populate *both* inactive slot groups A and B.


src/install.c
```diff
@@ -323,21 +325,63 @@ static gchar** get_all_manifest_slot_classes(const RaucManifest *manifest)
 static RaucSlot *select_inactive_slot_class_member(const gchar *rootclass)
 {
 	RaucSlot *iterslot;
+	RaucSlot *selectslot = NULL;
 	GHashTableIter iter;
 
 	g_return_val_if_fail(rootclass, NULL);
 
+	if (g_strcmp0(r_context()->config->statusfile_path, "per-slot") == 0) {
+		g_debug("Selecting inactive slot for class '%s'. Fallback strategy: 'first found'", rootclass);
```
jluebbe, line 334:
```suggestion
		g_debug("Selecting inactive slot for class '%s'. Strategy: 'first found'", rootclass);
```

src/install.c
```diff
@@ -323,21 +325,63 @@ static gchar** get_all_manifest_slot_classes(const RaucManifest *manifest)
...
+			continue;
+
+		/* For per-slot status, we don't try to access the timestamp information
+		 * and just take the first appropriate slot */
+		if (g_strcmp0(r_context()->config->statusfile_path, "per-slot") == 0) {
+			g_debug("Selected slot '%s'", iterslot->name);
+			return iterslot;
+		}
+
+		g_assert_nonnull(iterslot->status);
```
jluebbe, line 354: Is this guaranteed? What happens if the global slot status is empty?

src/install.c
```diff
@@ -323,21 +325,63 @@ static gchar** get_all_manifest_slot_classes(const RaucManifest *manifest)
...
+			g_debug("Consider slot '%s' as first candidate.", iterslot->name);
+			selectslot = iterslot;
+			continue;
+		}
+
+		/* if currently checked slot is older than candidate use as new candidate */
+		gint comp = g_date_time_compare(iterslot->status->installed_timestamp, selectslot->status->installed_timestamp);
+		if (comp < 0) {
+			gchar *found = g_date_time_format(iterslot->status->installed_timestamp, RAUC_FORMAT_ISO_8601);
+			gchar *current = g_date_time_format(selectslot->status->installed_timestamp, RAUC_FORMAT_ISO_8601);
```
jluebbe, line 373: Use `g_autofree`.

jluebbe (changes requested):
Do we still have enough test cases for the `per-slot` status cases?

ejoerns:
Rebased onto `master` to resolve conflict in `conftest.py` (no further changes).

jluebbe:
We should consider selecting bad slots (e.g. due to an interrupted installation) before looking at the dates. Otherwise, repeated installation attempts would lead to all inactive slots being marked as bad.

ejoerns:
Rebased on master and incorporated changes from #1567 and #1572 for now.

ejoerns:
Rebased onto master to drop the patch that got merged with #1584 already.

ejoerns:
> Do we still have enough test cases for the `per-slot` status cases?

`cat build/meson-logs/testlog.txt | grep "Using per-slot statusfile" | wc -l`
reports "485". Thus, I assume yes.

src/install.c
ejoerns: The status of all slots is initialized during context setup. I've added some comments to make this more obvious.

jluebbe (approved):

Issue #1477, opened 2024-07-25
I want to add a tutorials section to the documentation, starting with a tutorial on how to run and test RAUC locally


Erik-Sovereign:
**Is your feature request related to a problem? Please describe.**
At the moment I'm writing a Rust tool that is going to interact with the RAUC D-Bus API on an embedded device. I wanted to test the tool on my local Ubuntu PC with a working RAUC D-Bus API (I think, this is a common thing to do). I realized there was no throughout documentation on how to do this. So I wrote one, found here: https://erik-sovereign.github.io/Software-Engineering-Blog/2024/07/10/Test-the-RAUC-D-Bus-API-locally.html

**Describe the solution you'd like**
It would make even more sense to add a Tutorials section to the documentation, where maintainers or people like me can create and collaboratively update tutorials on how to achieve things with RAUC, that are commonly needed. Explainations that don't belong into other sections of the documentation, because they could confuse people or make the structure of the documentation more confusing.

**Describe alternatives you've considered**
The alternative is the blog post, which is considerably less findable than an official place in the documentation and RAUC website.

**Additional context**
I would be happy to create the tutorial section and the first tutorial. I would love to get some feedback on the tutorial, when it is ready.

Is this wanted by the RAUC team? When yes, then I will create the tutorial section and the tutorial.

jluebbe:
There is a built-in test environment:
-  https://rauc.readthedocs.io/en/latest/using.html#reproducing-issues-using-qemu-test-setup
-  https://rauc.readthedocs.io/en/latest/contributing.html#qemu-test-runner-qemu-test

Perhaps that's not easy to find or missing some details. A Tutorial section could describe how to use `qemu-test system` in more detail and the existing sections could refer to that. What do you think?

Erik-Sovereign:
Thank you for the answer :)

This is definitely not easy to find and even when found, its not obvious that this should be used to test RAUC locally.

Unfortunately `./qemu-test` also crashed early for me with 
```
qemu-test-init: 60: cannot create /etc/resolv.conf: Directory nonexistent
[    1.861537][    T1] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000200
```

jluebbe:
> This is definitely not easy to find and even when found, its not obvious that this should be used to test RAUC locally.

It's also mentioned in https://github.com/rauc/rauc?tab=readme-ov-file#running-the-test-suite, but without any background.
 
[...]

PR #1526, opened 2024-09-24
qemu-test-init: remove potential symlink at /etc/resolv.conf

jluebbe:
resolv.conf can be a symlink to /run (or somewhere else), so we need to remove the symlink to reliably replace it with our configuration. This works because we already mount an overlayfs at /etc.


KarlK90 (approved):
LGTM

JSydll:
Thanks a lot for addressing this. It solved the issues I mentioned in the community chat. It should also help to resolve #1477, if I'm not wrong.

PR #1528, opened 2024-09-25
src/bundle: gracefully abort remote bundle install if support is disabled

a3f:
Currently installing a remote bundle without having first enabled networking support at build-time results in a warning followed by a crash:
```
  Mounting remote bundle not supported, recompile with --enable-network
  installing http://192.168.8.15:8080/TODO-prod.raucb: Checking and mounting bundle...
  g_str_has_suffix: assertion 'str != NULL' failed
  rauc.service: Main process exited, code=dumped, status=5/TRAP
  rauc.service: Failed with result 'core-dump'.
```

Fix this by early exiting in this case and propagating the error up.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -2041,7 +2041,10 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, CheckBundleP
 		}
 		g_debug("Downloaded temp bundle to %s", ibundle->path);
 #else
-		g_warning("Mounting remote bundle not supported, recompile with -Dnetwork=true");
+		g_set_error(error, R_BUNDLE_ERROR, R_BUNDLE_ERROR_UNSUPPORTED,
+				"Remote bundle access not supported, recompile with -Dstreaming=true");
+		res = FALSE;
+		goto out;
 #endif
 	} else {
 		ibundle->path = g_strdup(bundlename);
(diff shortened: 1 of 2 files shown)


src/bundle.c
```diff
@@ -2041,7 +2041,10 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, CheckBundleP
 		}
 		g_debug("Downloaded temp bundle to %s", ibundle->path);
 #else
-		g_warning("Mounting remote bundle not supported, recompile with -Dnetwork=true");
+		g_set_error(error, R_BUNDLE_ERROR, R_BUNDLE_ERROR_FORMAT,
```
ejoerns, line 2044: Not sure how well `R_BUNDLE_ERROR_FORMAT` fits here which is used otherwise to show an issue with the bundle itself. Maybe define a new generic one like `R_BUNDLE_ERROR_UNSUPPORTED`.

src/bundle.c
```diff
@@ -2041,7 +2041,10 @@ gboolean check_bundle(const gchar *bundlename, RaucBundle **bundle, CheckBundleP
...
+				"Mounting remote bundle not supported, recompile with -Dnetwork=true");
```
ejoerns, line 2045: I'd leave it up to you but touching this would be a great option to fix the message. "Mounting" is valid only for streaming, thus maybe it's better to just omit the term. And `-Dnetwork=true` is not sufficient for streaming which is probably what we would like to suggest here.

Thus maybe:

```suggestion
				"Remote bundle not supported, recompile with -Dstreaming=true");
```

jluebbe: Perhaps "Remote bundle access not supported, …"?

ejoerns (approved):

a3f:
> I've implemented the requested changes.

Thanks!

PR #1538, opened 2024-10-16
allow local symlinks in bundle contents

jluebbe:
Local symlinks in the same directory (containing no slashes) seem
harmless and can be useful in special cases for backwards compatibility,
so allow them.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -997,14 +997,37 @@ out:
 
 static gboolean check_workdir_content(const gchar *path, GError **error)
 {
+	GError *ierror = NULL;
 	g_autofree gchar *name = g_path_get_basename(path);
 
-	/* It's not clear what symlinks in the bundle input should mean, so
-	 * reject them. If you have a use case for this, contact us. */
+	/* It's not clear what absolute or relative symlinks (with slashes)
+	 * in the bundle input should mean, so reject them. If you have a use
+	 * case for this, contact us. */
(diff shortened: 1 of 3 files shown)


src/bundle.c
```diff
@@ -879,6 +879,57 @@ static gboolean decrypt_bundle_payload(RaucBundle *bundle, GError **error)
 	return res;
 }
 
+static gboolean check_workdir_content(const gchar *path, GError **error)
+{
+	GError *ierror = NULL;
+	g_autofree gchar *name = g_path_get_basename(path);
+
+	/* It's not clear what realative symlinks in the bundle input should mean,
```
ejoerns, line 887:
```suggestion
	/* It's not clear what absolute symlinks (or relative symlinks that contain slashes) in the bundle input should mean,
```

ejoerns (changes requested):
We should also update the corresponding FAQ entry.

src/bundle.c
jluebbe: And we also need to actually copy the symlinks...

jluebbe: It turns out that symlinks can be hard-linked just fine, so there is no need to special case that. I've mentioned this in the commit message.

ejoerns:
Fixed a double 'symlinks' in the comment.

ejoerns (approved):

PR #1540, opened 2024-10-21
.github/workflows/tests: upload coverage separately for service=true/false

jluebbe:
This allows codecov to show the coverage for each case separately.


jluebbe:
You can see the resulting output at: https://app.codecov.io/gh/jluebbe/rauc/tree/codecov-flags

ejoerns (approved):
Not yet fully sure what to do with the results (since service-only paths can be followed by service tests, only) but looks interesting anyway.

Issue #953, opened 2022-08-19
add a way to pass config options via the command line


jluebbe:
This is the remaining open aspect of #729.

> > * [ ]   we should add a way to pass config options via the command line
> 
> Hi all, what is the status concerning the 2nd task "pass config options via the command line"? Will this be realized in near future? Thanks, Daniel

We don't have concrete plans for this, but would be happy to discuss concrete design and take a PR.

_Originally posted by @jluebbe in https://github.com/rauc/rauc/issues/729#issuecomment-1220659737_

ch-aurich:
Hi,
I'd like to contribute on this one... anything I need to be aware of before starting work?
Best Regards
Christian

jluebbe:
> I'd like to contribute on this one... anything I need to be aware of before starting work?

From the user's perspective, it could probably look like `rauc -c base-system.conf -C keyring:check-purpose=codesign -C keyring:check-crl=false`. I'm not sure if `:` is a good separator between section and key, but at least it's unused so far.

You'll need to collect all such options from the command line and add them to the `RaucContext`. The `load_config()` function can access them from there. Then, after parsing the config into a `GKeyFile`, you'd use [`g_key_file_set_value()`](https://docs.gtk.org/glib/method.KeyFile.set_value.html) to extend/override the loaded config file. This way, the existing loading of the individual options can stay as-is.

Issue #729, opened 2021-04-20
Unsupported certificate purpose when creating bundle


incaseoftrouble:
EDIT: Problem solved, two subsequent issues 1) `rauc bundle` should complain if config file given via `-c <file>` does not exist 2) it should be possible to specify `check-purpose=codesign` on the CLI (e.g. `--check-purpose=<purpose>`)

I'm creating a bundle using `rauc bundle --key=<key> --cert=<cert> --keyring=<ca> --intermediate=<intermediate> dir`. This fails with `Failed to create bundle: failed to sign manifest: signature verification failed: Verify error:unsupported certificate purpose`. 
Without `--keyring`, i.e. without verification on the host, bundle creation works. I suspect that an error occurs when parsing the keyring file.

The certificate chain is valid:
```
$ openssl verify -CAfile <ca> -untrusted <intermediate> <cert>
<cert>: OK
$  openssl verify -CAfile <(cat <intermediate> <ca>) <cert>
<cert>: OK
```
I also verified the  X509v3 Subject/Authority Key Identifier fields.

I have a "normal" root -> intermediate -> signing setup, where
```
[ v3_root ]
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid:always,issuer
basicConstraints = critical, CA:true
keyUsage = critical, digitalSignature, cRLSign, keyCertSign

[ v3_intermediate ]
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid:always,issuer
basicConstraints = critical, CA:true, pathlen:0
keyUsage = critical, digitalSignature, cRLSign, keyCertSign

[ v3_codesign ]
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid:always,issuer:always
basicConstraints = CA:FALSE
extendedKeyUsage = critical, codeSigning
```
with the last according to the [docs](https://rauc.readthedocs.io/en/latest/advanced.html#certificate-key-usage-attributes)

This might be related to #713 . The problem also occurs when I run `rauc` with `-c system.conf` with a `[keyring] check-purpose=codesign`

My version:
```
$ rauc --version
rauc 1.5.1
```

incaseoftrouble:
Addition: Reading throught the docs again:

> Also, you can optionally use extendedKeyUsage attributes on intermediate CA certificates to limit which ones are allowed to issue code signing certificates.

So it might be the case that the root / intermediate certs also need the codeSigning usage?

jluebbe:
[...]

PR #1542, opened 2024-10-25
Add -C option to overwrite config file parameters

ch-aurich:
add -C option to set/overwrite config file parameters (currently only for keyring:check-crl=false and keyring:check-purpose=codesign)

Fixes: #953

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -23,6 +23,12 @@ typedef enum {
 } RContextConfigMode;
 
 typedef struct {
+	gchar *section;
+	gchar *name;
+	gchar *value;
+} ConfigFileOverride;
+
+typedef struct {
 	/* a busy context must not be reconfigured */
 	gboolean busy;
 	gboolean pending;
@@ -30,6 +36,7 @@ typedef struct {
 	/* system configuration data */
 	RContextConfigMode configmode;
 	gchar *configpath;
+	GList *configoverride;
 	RaucConfig *config;
 
 	/* system status (not available when using per-slot status file) */
--- a/src/main.c
+++ b/src/main.c
@@ -2646,6 +2646,34 @@ static void create_option_groups(void)
 	g_option_group_add_entries(service_group, entries_service);
 }
 
+// Callback function to handle the repeated -C option
+static gboolean collect_config_values(const gchar *option_name, const gchar *value, gpointer data, GError **error)
+{
+	if (value == NULL) {
+		g_set_error(error, G_OPTION_ERROR, G_OPTION_ERROR_BAD_VALUE, "Error in parsing override: you must specify it in format SECTION:KEY=VALUE");
+		return FALSE;
+	}
+
+	g_auto(GStrv) key_value_split = g_strsplit(value, "=", 2);
+	if (g_strv_length(key_value_split) != 2) {
+		g_set_error(error, G_OPTION_ERROR, G_OPTION_ERROR_BAD_VALUE, "Error in parsing override: Missing '='");
+		return FALSE;
+	}
+
+	g_auto(GStrv) section_key_split = g_strsplit(key_value_split[0], ":", 2);
+	if (g_strv_length(section_key_split) != 2) {
+		g_set_error(error, G_OPTION_ERROR, G_OPTION_ERROR_BAD_VALUE, "Error in parsing override: Missing ':'");
+		return FALSE;
+	}
+
+	ConfigFileOverride *override = g_new(ConfigFileOverride, 1);
+	override->section = g_steal_pointer(&section_key_split[0]);
+	override->name = g_steal_pointer(&section_key_split[1]);
+	override->value = g_steal_pointer(&key_value_split[1]);
+	r_context_conf()->configoverride = g_list_append(r_context_conf()->configoverride, override);
+	return TRUE;
+}
+
 static void cmdline_handler(int argc, char **argv)
 {
 	gboolean help = FALSE, debug = FALSE, version = FALSE;
@@ -2654,6 +2682,7 @@ static void cmdline_handler(int argc, char **argv)
 	g_autoptr(GOptionContext) context = NULL;
 	GOptionEntry entries[] = {
 		{"conf", 'c', 0, G_OPTION_ARG_FILENAME, &confpath, "config file", "FILENAME"},
+		{"confopt", 'C', G_OPTION_FLAG_OPTIONAL_ARG, G_OPTION_ARG_CALLBACK, collect_config_values, "config settings that override parameters from the config file", "SECTION:KEY=VALUE" },
 		/* NOTE: cert and key kept for backwards-compatibility, but made invisible */
 		{"cert", '\0', G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_FILENAME, &certpath, "cert file or PKCS#11 URL", "PEMFILE|PKCS11-URL"},
 		{"key", '\0', G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_FILENAME, &keypath, "key file or PKCS#11 URL", "PEMFILE|PKCS11-URL"},
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -1248,12 +1248,13 @@ Command Line Tool
     rauc [OPTION?] <COMMAND>
 
   Options:
-    -c, --conf=FILENAME     config file
-    --keyring=PEMFILE       keyring file
-    --mount=PATH            mount prefix
-    -d, --debug             enable debug output
-    --version               display version
-    -h, --help              display help and exit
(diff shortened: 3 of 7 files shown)


jluebbe:
Thanks for your contribution! Why did you limit it to those two keyring options?

ch-aurich:
> Thanks for your contribution! Why did you limit it to those two keyring options?

I wanted to be able to verify the input, which is not an easy task to do for all the possible config parameters - hence the limitation to only the two parameters discussed in #953 and #729. But thinking about it again I would tend to leave that step (semantic check) out on this stage making it equivalent to the config file parsing where the g_key_file_load_from_data call takes everything that is syntactically correct. The semantic checks are done later nevertheless - so I could change it to accept all inputs as long as it is in the format SECTION:KEY=VALUE.

include/context.h
```diff
@@ -22,6 +22,12 @@ typedef enum {
 	R_CONTEXT_CONFIG_MODE_REQUIRED, /* require config file */
 } RContextConfigMode;
 
+typedef struct {
+	gchar *overwrite_section;
+	gchar *overwrite_name;
+	gchar *overwrite_value;
+} ConfigFileOverwrite;
+
```
jluebbe, line 30: I'd use [override instead of overwrite](https://www.grammar.com/override_vs_overwrite). Also there's no need to repeat prefix all structure members with the same string.

```suggestion
typedef struct {
	gchar *section;
	gchar *name;
	gchar *value;
} ConfigFileOverride;
```

ee

jluebbe:
Yes, at this level we should just allow all inputs, as long as they fit the pattern.

Please also squash the commits to remove the intermediate changes and force-push this PR. A single commit is probably fine for this feature.

Could you also add a test case for the new code in `src/main.c`? Perhaps in [`test/test_info.py`](https://github.com/rauc/rauc/blob/master/test/test_info.py) by checking that `system:bundle-formats=verity` rejects plain bundles.

a3f:
Just some bikeshedding: Should we maybe call this `-o` (for option) instead? Being similar to SSH's `-o` may be easier to remember and I like the `option` term more than `--cmdconf`

ch-aurich:
> Just some bikeshedding: Should we maybe call this `-o` (for option) instead? Being similar to SSH's `-o` may be easier to remember and I like the `option` term more than `--cmdconf`

+1 from my side... reusing an existing pattern makes a ton of sense!

Just waiting for the confirmation from @jluebbe because you brought up the initial idea (https://github.com/rauc/rauc/issues/953#issuecomment-2346256501)

jluebbe:
I quite like the the correspondence between `-c` and `-C`. I don't think really need the long form, as that way the command line will grow really long. Otherwise perhaps `--confopt`?

The current version fails to compile.

Also please squash the commits as mentioned in https://github.com/rauc/rauc/pull/1542#issuecomment-2437685305.

ch-aurich:
> I quite like the the correspondence between `-c` and `-C`. I don't think really need the long form, as that way the command line will grow really long. Otherwise perhaps `--confopt`?

I changed it that way

> The current version fails to compile.

sorry, I now fixed it

> Also please squash the commits as mentioned in [#1542 (comment)](https://github.com/rauc/rauc/pull/1542#issuecomment-2437685305).

did that as well

ch-aurich:
something seems to be wrong with my new test in test_info.py - sorry, will fix that

src/main.c
```diff
@@ -2646,6 +2646,41 @@ static void create_option_groups(void)
...
 }
 
+static gboolean append_config_override(gchar *section, gchar *name, gchar *value)
+{
+	ConfigFileOverride *override = g_new(ConfigFileOverride, 1);
+	override->section = section;
+	override->name = name;
+	override->value = value;
+	r_context_conf()->configoverride = g_list_append(r_context_conf()->configoverride, override);
+	return TRUE;
```
ejoerns, line 2656: why do you return a boolean here?

src/main.c
```diff
@@ -2646,6 +2646,41 @@ static void create_option_groups(void)
...
+	*colon_delimiter = 0;
+	colon_delimiter++;
+	gchar* equal_delimiter = strstr(colon_delimiter, "=");
+	if (equal_delimiter == NULL) {
+		return FALSE;
+	}
+	*equal_delimiter = 0;
+	equal_delimiter++;
+
+	return append_config_override(copy, colon_delimiter, equal_delimiter);
```
ejoerns, line 2681: Not sure if this requires an extra method.

test/config_file.c
```diff
@@ -1092,6 +1100,34 @@ check-purpose=codesign\n";
 	g_assert_true(config->keyring_allow_partial_chain);
 	g_assert_true(config->keyring_check_crl);
 	g_assert_cmpstr(config->keyring_check_purpose, ==, "codesign-rauc");
+
+	g_free(pathname);
```
ejoerns, line 1104: This does not really test the added functions but manually re-implements parts of their internal behavior. Maybe simply assigning allocated strings to each member of ConfigFileOverride is easier in any way?

src/main.c
```diff
@@ -2646,6 +2646,41 @@ static void create_option_groups(void)
...
+	return TRUE;
+}
+
+// Callback function to handle the repeated -C option
+static gboolean collect_config_values(const gchar *option_name, const gchar *value, gpointer data, GError **error)
+{
+	if (value == NULL) {
+		return FALSE;
+	}
+	gchar* copy = g_strdup(value);
```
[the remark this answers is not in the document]
ch-aurich: Mh, any suggestions how else to do?

src/main.c
ch-aurich: I put it in a separate function because initially I had the code to only accept known parameters and there I used it to not have duplicate code for different parameters. Now since we accept all parameters at this stage I can inline it again.

test/config_file.c
jluebbe: I'd be fine with that.

src/main.c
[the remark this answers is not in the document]
ejoerns: It would maybe be okay when adding explicit comments in the code, the struct and the freeing code and if you manage to fix the memory leakage.

The alternative would be to use `g_strsplit()` as e.g. the slot parsing does. This gives an array of newly allocated strings, but two rounds are needed to split the items properly.

If you use autoptr cleanup, then you also don't need to care for the cleanup:

``` c
g_auto(GStrv) key_value_split = g_strsplit(value, "=", 2);

if (g_strv_length(key_value_split) != 2) {
	g_set_error(error, G_OPTION_ERROR, G_OPTION_ERROR_BAD_VALUE, "Missing '=' ... ");
	return FALSE;
}
```

You can collect the items in the end with autoptr stealing, like:

``` c
override->value = g_steal_pointer(&key_value_split[1]);
```

Autoptr cleanup could be used for your approach, too, maybe.

ch-aurich:
not sure why codecov does not notice the coverage at this point - from my point of view the test in pytest should cover the red lines as well:
image

The missing parts that are really not tested are the error paths in main.c - should I add tests for that as well?

ejoerns (commented):
Parsing and memory handling looks really good to me now, Thank you :+1: 

Just a few final notes from my side, but nothing critical.

ejoerns:
> not sure why codecov does not notice the coverage at this point - from my point of view the test in pytest should cover the red lines as well: image

Mh, looks more like an off-by-one by codecov.

> The missing parts that are really not tested are the error paths in main.c - should I add tests for that as well?

As this should be quite straightforward, maybe add at least a test with a missing section name or so. This ensures that error propagation basically works.

src/main.c
```diff
@@ -2654,6 +2682,7 @@ static void cmdline_handler(int argc, char **argv)
 	g_autoptr(GOptionContext) context = NULL;
 	GOptionEntry entries[] = {
 		{"conf", 'c', 0, G_OPTION_ARG_FILENAME, &confpath, "config file", "FILENAME"},
+		{"confopt", 'C', G_OPTION_FLAG_OPTIONAL_ARG, G_OPTION_ARG_CALLBACK, collect_config_values, "config settings that override parameters from the config file", "SECTION:KEY=VALUE" },
```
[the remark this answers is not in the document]
ejoerns: > It not only allows overriding what's in the config file but also allows adding new options that haven't been present before. But maybe that's a matter of perspective.
> 
> There should at least be a hint in the documentation, maybe a note box under docs/reference.rst?

Should have been more specific here. I meant a hint under the "System Configuration File" chapter that this option exists and how it can be overridden. But for me, your version is fine, too.

> Adding a parameter also requires adapting the manpage (`/rauc.1`).

Looks good, especially since it adds the details that `-h` output lacks.

ejoerns (approved):

jluebbe (approved):

PR #1548, opened 2024-11-05
Print bootname during install

rical:
This makes the output more user friendly on systems where the bootname is the common way to refer to a slot.

### Example output
```
...
40% Checking slot rootfs.1 (secondary)
46% Checking slot rootfs.1 (secondary) done.
...
```
If no bootname is specified

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -1092,7 +1092,11 @@ static gboolean handle_slot_install_plan(const RaucManifest *manifest, const RIm
 
 	install_args_update(args, "Checking slot %s", plan->target_slot->name);
 
-	r_context_begin_step_weighted_formatted("check_slot", 0, 1, "Checking slot %s", plan->target_slot->name);
+	r_context_begin_step_weighted_formatted("check_slot", 0, 1, "Checking slot %s%s%s%s",
+			plan->target_slot->name,
+			plan->target_slot->bootname ? " (" : "",
+			plan->target_slot->bootname ? plan->target_slot->bootname : "",
+			plan->target_slot->bootname ? ")" : "");
 
 	r_slot_status_load(plan->target_slot);
 	slot_state = plan->target_slot->status;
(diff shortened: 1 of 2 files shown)


jluebbe:
This breaks (at least) one test:
```
2024-11-12T21:44:12.1350422Z rauc:ERROR:../test/service.c:105:on_installer_changed: assertion failed (message == cmp_message): ("Checking slot rootfs.1 (system1)" == "Checking slot rootfs.1")
```

rical:
> I am not fully sure what "systems where the bootname is the common way to refer to a slot" are.

In this use-case it's an embedded device where the user can select which slot to boot from in the bootloader, `primary` or `secondary`. This terminology is present throughout the UI and documentation.

> Normally, the primary identification is the slot name which can also be named/choosen quite freely. Is this more for developers? Or for users?

Both. Perhaps I'm misunderstanding the slot naming. Can I use something like `[slot.rootfs.primary]` and `[slot.rootfs.secondary]` ?

rical:
Not sure what style check wants me to change? I thought rauc used kernel coding style so I ran it through [checkpatch.pl ](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/scripts/checkpatch.pl?h=v6.12) and the indentation appears correct.

ejoerns:
> Not sure what style check wants me to change? I thought rauc used kernel coding style so I ran it through [checkpatch.pl ](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/scripts/checkpatch.pl?h=v6.12) and the indentation appears correct.

No, we are a userspace tool :wink: We have our own coding style, bit it is heavily inspired by common best-practices (like kernel-coding style). 
The best way to ensure it matches the expectations is to run `./uncrustify.sh` locally, like mentioned here:

https://rauc.readthedocs.io/en/latest/contributing.html#code-style-uncrustify

jluebbe (approved):

PR #1549, opened 2024-11-08
src/signature: fix CMS debug output check

tamara-schmitz:
As specified in the documentation, setting `G_MESSAGES_DEBUG=all` is supposed to enable log output for all logging domains. However the check to print CMS info introduced by
2cd407600a35eb51995c53bf97fc9cdb0d0fade4
only checks if `G_MESSAGES_DEBUG` contains `-signature`.

The goal is to make this check consistent with the rest of the program and the documentation. This does not affect the `-d` command line option as this uses the `rauc` debug message domain. But when one requests `all`, there should be all messages.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -1133,9 +1133,10 @@ static void debug_cms_ci(CMS_ContentInfo *cms)
 	gchar *out_str = NULL;
 	long size;
 
-	if ((domains == NULL) || (strstr(domains, G_LOG_DOMAIN "-signature") == NULL)) {
+	if (domains == NULL)
 		return;
-	}
+	if (!g_str_equal(domains, "all") && strstr(domains, G_LOG_DOMAIN "-signature") == NULL)
+		return;
 
 	out = BIO_new(BIO_s_mem());
 	CMS_ContentInfo_print_ctx(out, cms, 2, NULL);


tamara-schmitz:
Revised my commit as my first revision was confusing and incorrect.

src/signature.c
```diff
@@ -1133,7 +1133,9 @@ static void debug_cms_ci(CMS_ContentInfo *cms)
 	gchar *out_str = NULL;
 	long size;
 
-	if ((domains == NULL) || (strstr(domains, G_LOG_DOMAIN "-signature") == NULL)) {
+	if (domains == NULL)
+		return;
+	if ((strstr(domains, "all") == NULL) && (strstr(domains, G_LOG_DOMAIN "-signature") == NULL)) {
```
ejoerns, line 1138: In the way glib and RAUC interpret the domains, `"all"` should only be used as a single element. Thus using `g_strcmp0()` or `g_str_equal()` should be preferred here. It would also ensure that no domain name that accidentally contains `"all"` as substring would match.

tamara-schmitz: You made me take a closer look at the GLib documentation. And there I found this note:

> It is recommended that custom log writer functions re-use the G_MESSAGES_DEBUG environment variable, rather than inventing a custom one, so that developers can re-use the same debugging techniques and tools across projects. Since GLib 2.68, this can be implemented by dropping messages for which g_log_writer_default_would_drop() returns TRUE.

I will revise my commit to use `g_log_writer_default_would_drop()`

jluebbe: We still support GLib 2.56, so we can use `g_log_writer_default_would_drop()` yet.

tamara-schmitz: Oh I see

ejoerns (approved):
Thanks for the rework. Looks good to me!

PR #1563, opened 2024-11-18
config_file: handle empty config file specially

a3f:
The rauc info and rauc extract commands will automatically consult a system.conf if available and otherwise fallback to a built-in default configuration.

When the system.conf is empty or symlinked to /dev/null, RAUC will report an error:
```
  Failed to initialize context: Failed to load system config
  (/run/rauc/system.conf): Key file does not have group ?system?
```
This is not in line with the UAPI Group's "[Configuration Files Specification](https://uapi-group.org/specifications/specs/configuration_files_specification/
)":

>   As a special override case, it must be possible to mask files across
>   different locations by creating a symlink to /dev/null or an empty file.

In other words: An empty RAUC config should ultimately behave as if no RAUC config at all was found. Let's support this by checking if the file is empty in load_config and making use of that information in r_context_configure.

Note that this will only affect `R_CONTEXT_CONFIG_MODE_AUTO`; `R_CONTEXT_CONFIG_MODE_NONE` already ignores config files and `R_CONTEXT_CONFIG_MODE_REQUIRED` requires a config file and will continue to trigger an error, albeit now with a different error message.

---

This is a follow-up to the discussion at https://github.com/rauc/rauc/pull/1557#issuecomment-2480608282


jluebbe:
> Note that this will only affect `R_CONTEXT_CONFIG_MODE_AUTO`; `R_CONTEXT_CONFIG_MODE_NONE` already ignores config files and `R_CONTEXT_CONFIG_MODE_REQUIRED` requires a config file and will continue to trigger an error, albeit now with a different error message.

The error message is now: `Failed to initialize context: Failed to load system config (<filename>): Input file is empty`

jluebbe:
Force pushed to fix the style issue.

jluebbe (approved):

ejoerns (approved):

PR #1567, opened 2024-11-25
tests: Make GRUB env handling more consistent and update qemu-test doc, etc.

ejoerns:
While dealing with tests for boot slot activation, some inconsistencies in grubenv handling, indirectly used for most tests, came up. This PR aims to clean up inconsistencies in this area.

While touching this, we also update the qemu-test documentation which lacks some information for developers that are not that familiar with RAUC.


qemu-test-init
```diff
@@ -109,10 +109,7 @@ $BUILD_DIR/test/fakerand
 # grub env
 mkdir -p /tmp/boot/grub
 mount --bind /tmp/boot /boot
-grub-editenv test/grubenv.test create
-touch /tmp/boot/grub/grubenv
-mount --bind test/grubenv.test /tmp/boot/grub/grubenv
-
+grub-editenv /tmp/boot/grub/grubenv create
```
jluebbe: In the next patch, `grub-editenv` seems to be enough. Do we need to use the `create` command here? If not, just drop it.

test/bin/postinstall.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/sh
 
-[ "x$RAUC_CURRENT_BOOTNAME" != "xsystem0" ] && exit 1
```
jluebbe, line 3: Perhaps still check that the variable is set and as nonzero length?

docs/contributing.rst
```diff
@@ -127,16 +127,46 @@ To run the entire test suite, type::
...
   ./qemu-test shell
 
+From which you run the tests manually (as shown in the previous sections).
+
+Additional parameters that you can add to a ``./qemu-test`` call are:
+
+:passthrough: for optimal performance (by passing through your host's CPU
+  features to the guest).
+
+:asan: sets up environment variable to support the address sanitizer.
```
jluebbe, line 150:
```suggestion
:asan: sets up environment variables to support the address sanitizer.
```

docs/contributing.rst
```diff
@@ -127,16 +127,46 @@ To run the entire test suite, type::
...
+Interactive Test System
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Beside providing a safe test environment, the ``./qemu-test`` script also
+supports running RAUC interactively in the QEMU environment by calling::
+
+  ./qemu-test system
+
+
+This setup initializes QEMU to run with the RAUC service started, alongside a
```
jluebbe, line 162:
```suggestion
  ./qemu-test system

This setup initializes QEMU to run with the RAUC service started, alongside a
```

docs/contributing.rst
```diff
@@ -127,16 +127,46 @@ To run the entire test suite, type::
...
+configured D-Bus and dummy target slots to simulate a real firmware update
+scenario. The configuration uses GRUB as a mock boot selection backend,
+allowing RAUC to interact with it as it would in a real system. Notably,
+rebooting the environment is not supported in this setup, meaning the testing
+is limited to a single boot cycle. This is sufficient for testing RAUC’s update
+mechanism but does not cover reboot-based validation.
```
jluebbe, line 168: Please use https://sembr.org/.

jluebbe (changes requested):

qemu-test-init
ejoerns: I don't think create is required. Will drop it.

jluebbe (approved):

PR #1569, opened 2024-11-27
Add missing qt type annotations

richardhozak:
Adds qt type annotations so classes can be generated with `qdbusxml2cpp`.

### DIFF
--- a/src/de.pengutronix.rauc.Installer.xml
+++ b/src/de.pengutronix.rauc.Installer.xml
@@ -36,7 +36,7 @@
       <arg name="compatible" type="s" direction="out" />
       <arg name="version" type="s" direction="out" />
     </method>
- 
+
     <!--
          InspectBundle: D-Bus variant of rauc info <bundle>
          @source: Path or URL to the queried bundle.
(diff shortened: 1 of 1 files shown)


ejoerns:
@richardhozak Thank you for contributing! Could you add your [Signged-off-by](https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin) line to make the PR ready for being merged? Also please align your commit title with the prefixes used for this file (`de.pengutronix.rauc.Installer.xml: `).

While at it, maybe also move the unrelated whitespace fix to a separate commit. Thanks!

ejoerns:
> I've signed off the commit, should I also remove the whitespace change? Seems unnecessary to create another commit for it.

Thank you for your adaption. I would have said adding another commit to the PR should not be a big deal, but I'd also be fine now with keeping it as-is.

jluebbe:
You're using a GitHub generated email address in the author field and SoB line: https://github.com/rauc/rauc/commit/87c40b6768189e84188584486b1cfad5b1fab62f.patch

Is that intentional?

richardhozak:
Yes, that's intentional, I use it for all my commits.

jluebbe (approved):

PR #1574, opened 2024-11-29
Fix "mark-{bad,good}  other" only works for A/B setups

N3xp7im3:
Marking multiple slots with "rauc status mark-{bad,good} other" only works for two slots.
As this behavior might not be intentional, this PR implements support for setups with more than two slots. So now when using the command above more than two slots can be marked.

In order to write test cases I rebased my patches onto the following PR [1].
This avoids reinventing the wheel by using the ABC config and fixtures used in their PR.

[1] https://github.com/rauc/rauc/pull/1525

### DIFF
--- a/docs/basic.rst
+++ b/docs/basic.rst
@@ -175,13 +175,20 @@ of a rootfs and application slot each that will be used together.
    :width: 500
    :align: center
 
-To detect the active slots, RAUC attempts to detect the currently booted slot.
+To detect the *active* slots, RAUC attempts to detect the currently booted slot.
 For this, it relies on explicit mapping information provided via the kernel command
 line, or attempts to find it out using mount information.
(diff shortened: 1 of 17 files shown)


ejoerns:
FTR: I have an updated PR for #1525 in progress which rebases onto current master and incorporates the changes made in #1567 and #1572. This should resove some of the conflicts.

ejoerns:
I've split up my test adaptions so that it should be sufficient to cherry-pick 38cb2e4b4aed8ad1a5c5f1af0884938f7a2f8eac here.

N3xp7im3:
Rebased my changes onto newest PR version of [#1525](https://github.com/rauc/rauc/pull/1525)
Furthermore, I deleted my abc_mark fixture and now use your system_abc fixture instead

N3xp7im3:
Opened new PR which already includes an improved version of this PR
See #1584

PR #1584, opened 2024-12-11
Add option to prevent fallback

N3xp7im3:
This PR aims to implement an option that automatically marks all _other_ slots as bad when the booted one is marked as good. This can be used to ensure not booting into the previous used slot after installing a new rootfs. The option is called "prevent-fallback" and can be set in the _system_ section.

For example:

``` ini
[system]
compatible=My BSP
bootloader=barebox
bundle-formats=-plain
data-directory=/data/rauc
prevent-late-fallback=true
....
```

The main logic for this patch is mostly based on the underlying "mark-bad other" command.
In order to make the prevent-fallback option to work for more than two slots, I have reworked the _get_slot_by_identifier_ function to be able to return more than one slot.

For the test case I decided to only test the "mark-bad other" command as the prevent-fallback option simply calls this command. I used the ABC pytest fixture from [38f7a2f8eac](https://github.com/rauc/rauc/commit/38cb2e4b4aed8ad1a5c5f1af0884938f7a2f8eac) for my test.

There is one minor issue with this PR, which is the message handling in the _mark_run_ function. When the fallback option is enabled, we call the _mark_run_ method recursively which results in overriding the variables _message_ and _slot_name_. This results in information loss as only the last call  sets the variables properly. One possible solution to fix this is by concatenating the messages and slot_names in some way?

### DIFF
--- a/src/mark.c
+++ b/src/mark.c
@@ -6,10 +6,11 @@
 #include "slot.h"
 #include "status_file.h"
 
-static RaucSlot* get_slot_by_identifier(const gchar *identifier, GError **error)
+static GList* get_slots_by_identifier(const gchar *identifier, GError **error)
 {
 	GHashTableIter iter;
-	RaucSlot *slot = NULL, *booted = NULL;
+	GList *slots = NULL;
+	RaucSlot *booted = NULL;
 
 	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
 
@@ -22,7 +23,7 @@ static RaucSlot* get_slot_by_identifier(const gchar *identifier, GError **error)
 
 	if (g_strcmp0(identifier, "booted") == 0) {
 		if (booted)
-			slot = booted;
+			slots = g_list_append(slots, booted);
 		else
 			g_set_error(
 					error,
@@ -32,12 +33,13 @@ static RaucSlot* get_slot_by_identifier(const gchar *identifier, GError **error)
 	} else if (g_strcmp0(identifier, "other") == 0) {
 		if (booted) {
 			g_hash_table_iter_init(&iter, r_context()->config->slots);
+			RaucSlot *slot = NULL;
 			while (g_hash_table_iter_next(&iter, NULL, (gpointer*) &slot)) {
-				if (slot->sclass == booted->sclass && !slot->parent && slot->bootname && slot != booted)
-					break;
-				slot = NULL;
+				if (slot->sclass == booted->sclass && !slot->parent && slot->bootname && slot != booted) {
+					slots = g_list_append(slots, slot);
+				}
 			}
-			if (!slot)
+			if (g_list_length(slots) == 0)
 				g_set_error(error,
 						R_SLOT_ERROR,
 						R_SLOT_ERROR_FAILED,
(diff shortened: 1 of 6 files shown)


src/mark.c
```diff
@@ -203,55 +206,87 @@ gboolean mark_run(const gchar *state,
...
+			}
+			g_string_append_printf(slot_names, "%s ", slot->name);
 		}
+
+		if (slot_names->len > 1)
+			g_string_truncate(slot_names, slot_names->len - 1);
+
+		*message = g_strdup_printf("marked slot[s] %s as bad", slot_names->str);
+		if (slot_name)
+			*slot_name = g_strdup("other");
```
jluebbe, line 269: Previously, this was set to the actual "resolved" name of the other slot. Accordingly, this should probably use the space separate slot names.

See below for a simpler solution to handle `slot_name`.

src/mark.c
```diff
@@ -203,55 +206,87 @@ gboolean mark_run(const gchar *state,
...
+		*message = g_strdup_printf("activated slot %s", slot->name);
 	} else {
-		res = FALSE;
 		*message = g_strdup_printf("unknown subcommand %s", state);
+		return FALSE;
 	}
-
-out:
-	if (res && slot_name)
-		*slot_name = g_strdup(slot->name);
```
jluebbe, line 252: With the new code, we only get here in the successful case. As `slot_name` should be set only then, you should be able to reduce complexity and code duplication by generating the space-separated string here only once at the end. That would move that part of the loop code in the "bad" case here.

src/mark.c
```diff
@@ -203,55 +206,87 @@ gboolean mark_run(const gchar *state,
...
 			*message = g_strdup(ierror->message);
-		} else {
-			res = TRUE;
-			*message = g_strdup_printf("marked slot %s as good", slot->name);
+			return FALSE;
+		}
+
+		if (slot_name)
+			*slot_name = g_strdup(slot->name);
+		*message = g_strdup_printf("marked slot %s as good", slot->name);
```
jluebbe, line 237: This string overwritten in the prevent-fallback case. Move it into an `else` branch below.

test/test_mark.py
```diff
@@ -61,6 +61,21 @@ def test_status_mark_good_non_bootslot(rauc_no_service):
     assert exitcode == 1
 
 
+@have_grub
+def test_status_mark_bad_other(rauc_dbus_service_with_system_abc):
```
jluebbe, line 65: Also add a test case for the `mark-good booted` with enabled `prevent-fallback`. You can use the `system` fixture to update the config and write it again from the test case.

jluebbe (changes requested):

jluebbe:
This needs to be rebased when #1601 is merged.

jluebbe:
I had factored out the switch to `g_auto*` in mark_run in https://github.com/rauc/rauc/pull/1600.

jluebbe (approved):
I've rebased this on my PRs #1600 and #1601 to make this easier to review and test. Compared to the last version, we now have a test which runs `mark-good` with enabled `prevent-late-fallback` (which is the new name we agreed on internally).

ejoerns (approved):

Issue #1554, opened 2024-11-14
Failing slot-hooks does not reinstall the slot if install-same=false


mschnelte:
**Describe the bug**

Give a slot with `install-same=false` and a slot-hook that fails a first run of rauc install will fail. A subsequent run with the same rauc bundle however will return success as the slot is simply skipped.

**Background information**

Version 1.12

**To Reproduce**

* Use a system.conf with install-same=false like so:

```
[slot.boot.1]
device=/dev/mmcblk3p2
type=ext4
parent=rootfs.1
install-same=false
```
* and this manifest:

```
[image.boot]
filename=boot.ext4
adaptive=block-hash-index
hooks=pre-install
```
* add a hook file that returns a non-zero exit code.
* install the bundle -> it fails (and the target partition of the slot is untouched)
* install the bundle again -> it succeeds but skips the slot installation (i.e. the target slot content is still the old one)

The behavior when using a failing post-install is similar:

On first run the slot is installed into the target, but the post-install fails. On second run the post-install is not executed.

**Expected behavior**

Slot is reinstalled when running again after a run that failed with a slot-hook.

ejoerns:
@mschnelte Thanks for reporting this!

I can reproduce the issue and it should not be limited to failing hooks.

The issue is that we write the status `"failed"` together with the correct checksum in the status file but check only for the checksum when skipping slots.

The options I see are:

a) not write the checksum in the 'failed' case
b) check that the status is "good" before considering the checksum valid.

PR #1585, opened 2024-12-18
Check the slot status in case of reinstalling of failed slot

samerkenawy:
This PR aims to fix slot selection by checking slot status with the checksum incase of reinstalling bad slot with the same image

Fixes: #1554

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -1111,7 +1111,7 @@ static gboolean handle_slot_install_plan(const RaucManifest *manifest, const RIm
 	}
 
 	/* if explicitly enabled, skip update of up-to-date slots */
-	if (!plan->target_slot->install_same && g_strcmp0(plan->image->checksum.digest, slot_state->checksum.digest) == 0) {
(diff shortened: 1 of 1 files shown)


ejoerns:
@samerkenawy Thank you for your fix! Could you add a short description of the problem and how this is fixed to the commit message?

E.g. it should be mentioned that "failed" slots currently have a valid checksum set.

In the future, the general handling of status vs. checksum etc. might need to be rethought, but for the concrete problem, this fix should be appropriate.

While at it, please also add the file you changed as prefix to the commit: `src/install: <message>`

ejoerns:
@samerkenawy Thank you for your adaption. Just two few details in the reasoning:

* The issue should not only be triggered by a failing hook, but by all sorts of slot installation failures
* The problematic handling is that the checksum is written for the "failed" case, too

Maybe:

```
When an installation of a slot with install-same=false fails, the initial rauc install attempt will fail.
However, a subsequent attempt with the same rauc bundle will succeed as the slot is skipped.

This happens because "failed" slots also have a valid checksum set and the slot status is not checked to be "ok" before comparing the checksum.
```

ejoerns (approved):

jluebbe (approved):

PR #1589, opened 2025-01-07
Add support for TFTP

troglobit:
Unlock TFTP support for upgrades.

Tested extensively in the [Infix NOS](https://github.com/kernelkit/infix/) project.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,7 +78,7 @@ Features
   * USB Stick
 * Controllable via **D-Bus** interface
 * Supports data migration
-* Network protocol support using libcurl (https, http, ftp, ssh, ...)
+* Network protocol support using libcurl (https, http, tftp/ftp, ssh, ...)
 * Several layers of update customization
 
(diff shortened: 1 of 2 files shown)


jluebbe (changes requested):
A commit message describing the use-case and that the tftp support is limited to bundle downloads (meaning no streaming) would be useful.

Looks good otherwise.

troglobit:
> A commit message describing the use-case and that the tftp support is limited to bundle downloads (meaning no streaming) would be useful.

Just to clarify the requirements, a force-push with an improved commit message about the limitations of the TFTP support is sufficient?

> Looks good otherwise.

Thanks mate!

jluebbe:
Yes, just force-push if you want to make any changes.

troglobit:
> Yes, just force-push if you want to make any changes.

There, fixed (I hope).

jluebbe (approved):

Issue #276, opened 2018-05-31
Slot update fails if partition could not be mounted


mhei:
I'm testing rauc on a custom board with a typical A/B rootfs + customerfs setup.

Just for understanding the following trace, the partitioning looks like this:

    label: dos
    label-id: 0x57b0f89d
    device: /dev/mmcblk0
    unit: sectors

    /dev/mmcblk0p1 : start=        2048, size=        6144, type=53
    /dev/mmcblk0p2 : start=        8192, size=     2097152, type=83
    /dev/mmcblk0p3 : start=     2105344, size=     2097152, type=83
    /dev/mmcblk0p4 : start=     4202496, size=     2627584, type=5
    /dev/mmcblk0p5 : start=     4204544, size=     2097152, type=83
    /dev/mmcblk0p6 : start=     6303744, size=      262144, type=83
    /dev/mmcblk0p7 : start=     6567936, size=      262144, type=83

During manufacturing process, p2 is filled with a valid ext4 rootfs, and p5 to p7
are simply formatted as ext4. p3 is left untouched, i.e. usually no valid filesystem
which could be mounted.

When system should be updated the first time, rauc comes into play and it should write the ext4
image contained in the bundle to p3. However, this fails:

    May 30 13:30:44 system rauc[212]: input bundle: /tmp/update-image.raucb
    May 30 13:30:44 system rauc[212]: Active slot bootname: A
    May 30 13:30:44 system rauc[665]: installing /tmp/update-image.raucb: installing
    May 30 13:30:44 system rauc[212]: installing /tmp/update-image.raucb: started
    May 30 13:30:44 system rauc[665]: installing /tmp/update-image.raucb:   0% Installing
    May 30 13:30:44 system rauc[665]: installing /tmp/update-image.raucb:   0% Determining slot states
    May 30 13:30:44 system rauc[665]: installing /tmp/update-image.raucb:  20% Determining slot states done.
    May 30 13:30:44 system rauc[212]: installing /tmp/update-image.raucb: Checking and mounting bundle...
    May 30 13:30:44 system rauc[212]: Reading bundle: /tmp/update-image.raucb
    May 30 13:30:44 system rauc[665]: installing /tmp/update-image.raucb:  20% Checking bundle
    May 30 13:30:44 system rauc[212]: Verifying bundle...
    May 30 13:30:44 system rauc[665]: installing /tmp/update-image.raucb:  20% Verifying signature
[...]

PR #1591, opened 2025-01-11
src/update_handler: ignore mount error before running slot hook

gportay:
The handler mounts the other-slot filesystem, and it returns in error if it fails to mount it.

The filesystem may be in bad health, or may be not yet made, and thus, it causes the handler to fail.

As a consequence, it results in a slot that is unupdatable until the filesystem is remade manually (i.e. externally).

This ignores the mount error before the slot hook is run, and leaves the hook to decide whether or not to fail.

---

Hello,

This PR intends to fix an issue if installing a bundle with a pre-install slot hook, on the device slot A system, with the slot B partitions created, but without the filesystems made yet.

The handler function [img_to_raw_handler()](https://github.com/rauc/rauc/blob/e3956840291874dce6a7734d19eef62c50c5fa66/src/update_handler.c#L2527) runs the [pre_install](https://github.com/rauc/rauc/blob/e3956840291874dce6a7734d19eef62c50c5fa66/src/update_handler.c#L2533-2539) hook and fails because the filesystem is not made yet. I will be ~~made~~ copy right after by the function [write_image_to_dev()](https://github.com/rauc/rauc/blob/e3956840291874dce6a7734d19eef62c50c5fa66/src/update_handler.c#L2542).

(Well, I supposed it comes from that handler, I have not checked it).

With this PR, I am suggesting to ignore mount failure in function [mount_and_run_slot_hook()](https://github.com/rauc/rauc/blob/e3956840291874dce6a7734d19eef62c50c5fa66/src/update_handler.c#L1429-L1433). This fixes the issue I am facing; however, I guess it may have some side-effects and I am not convinced yet.

What are you opinions?

The issue can be reproduced easily:
 - generate a bundle with a pre-hook,
 - corrupt the other filesystem slot (`dd if=/dev/zero of=/dev/other-slot`), and
 - install the bundle (`rauc install bundle.raucb`). 

Regards,
Gaël

---

Here are the failing logs on 1.12:

```
# journalctl -u rauc | cat
Oct 08 15:42:35 buildroot systemd[1]: Starting RAUC Update Service...
Oct 08 15:42:35 buildroot rauc[305]: Using central status file /boot/rauc/central.raucs
Oct 08 15:42:35 buildroot rauc[305]: Failed to load system status: No such file or directory
Oct 08 15:42:35 buildroot rauc[305]: Getting Systeminfo: /usr/lib/raspberrypi-firmware-rauc-bootloader-backend/system-info
Oct 08 15:42:35 buildroot rauc[305]: Booted into rootfs.0 (ROOTFS-A)
Oct 08 15:42:35 buildroot systemd[1]: Started RAUC Update Service.
Oct 08 15:42:36 buildroot rauc[305]: Marked slot rootfs.0 as good
Oct 08 15:42:36 buildroot rauc[305]: rauc mark: marked slot rootfs.0 as good
Jan 10 16:54:53 buildroot rauc[305]: input bundle: /tmp/buildroot.raucb
Jan 10 16:54:53 buildroot rauc[305]: Active slot bootname: /dev/mmcblk0p5
Jan 10 16:54:53 buildroot rauc[305]: installing /tmp/buildroot.raucb: started
Jan 10 16:54:53 buildroot rauc[305]: Installation 0f629fe9 started
Jan 10 16:54:53 buildroot rauc[305]: installing /tmp/buildroot.raucb: Checking and mounting bundle...
Jan 10 16:54:53 buildroot rauc[305]: Reading bundle: /tmp/buildroot.raucb
Jan 10 16:54:53 buildroot rauc[305]: Detected CRL but CRL checking is disabled!
Jan 10 16:54:53 buildroot rauc[305]: Verifying bundle signature... 
Jan 10 16:54:54 buildroot rauc[305]: Verified detached signature by 'O = Test Org, CN = Test Org Release-1'
Jan 10 16:54:54 buildroot rauc[305]: Mounting bundle '/tmp/buildroot.raucb' to '/run/rauc/bundle'
Jan 10 16:54:54 buildroot rauc[305]: Configured loop device '/dev/loop0' for 51494912 bytes
Jan 10 16:54:54 buildroot rauc[305]: Checking image type for slot type: ext4
Jan 10 16:54:54 buildroot rauc[305]: Image detected as type: *.ext4
Jan 10 16:54:54 buildroot rauc[305]: Checking image type for slot type: vfat
Jan 10 16:54:54 buildroot rauc[305]: Image detected as type: *.vfat
Jan 10 16:54:54 buildroot rauc[305]: Marking target slot rootfs.1 as non-bootable...
Jan 10 16:54:54 buildroot rauc[305]: Marked slot rootfs.1 as bad
Jan 10 16:54:54 buildroot rauc[305]: installing /tmp/buildroot.raucb: Updating slots...
Jan 10 16:54:54 buildroot rauc[305]: installing /tmp/buildroot.raucb: Checking slot rootfs.1
Jan 10 16:54:54 buildroot rauc[305]: Updating slot rootfs.1
Jan 10 16:54:54 buildroot rauc[305]: installing /tmp/buildroot.raucb: Updating slot rootfs.1
Jan 10 16:54:54 buildroot rauc[305]: Updating /dev/mmcblk0p6 with /run/rauc/bundle/rootfs.ext4
Jan 10 16:54:54 buildroot rauc[305]: Mounting slot /dev/mmcblk0p6
Jan 10 16:54:54 buildroot rauc[366]: mount: /run/rauc/rootfs.1: wrong fs type, bad option, bad superblock on /dev/mmcblk0p6, missing codepage or helper program, or other error.
Jan 10 16:54:54 buildroot rauc[366]:        dmesg(1) may have more information after failed mount system call.
Jan 10 16:54:54 buildroot rauc[305]: Updating slot rootfs.1 status
Jan 10 16:54:54 buildroot rauc[305]: Installation 0f629fe9 failed: Installation error: Failed updating slot rootfs.1: failed to mount slot: failed to run mount: Child process exited with code 32
Jan 10 16:54:54 buildroot rauc[305]: Installation error: Failed updating slot rootfs.1: failed to mount slot: failed to run mount: Child process exited with code 32
Jan 10 16:54:54 buildroot rauc[305]: installing /tmp/buildroot.raucb: Installation error: Failed updating slot rootfs.1: failed to mount slot: failed to run mount: Child process exited with code 32
Jan 10 16:54:54 buildroot rauc[305]: installing /tmp/buildroot.raucb: finished
Jan 10 16:54:54 buildroot rauc[305]: installing `/tmp/buildroot.raucb` failed: 1
```

Here are the logs on 1.12 + PR:
```
# journalctl -u rauc | cat
Oct 08 15:42:35 buildroot systemd[1]: Starting RAUC Update Service...
Oct 08 15:42:35 buildroot rauc[303]: Using central status file /boot/rauc/central.raucs
Oct 08 15:42:35 buildroot rauc[303]: Failed to load system status: No such file or directory
Oct 08 15:42:35 buildroot rauc[303]: Getting Systeminfo: /usr/lib/raspberrypi-firmware-rauc-bootloader-backend/system-info
Oct 08 15:42:35 buildroot rauc[303]: Booted into rootfs.0 (ROOTFS-A)
Oct 08 15:42:35 buildroot systemd[1]: Started RAUC Update Service.
Oct 08 15:42:36 buildroot rauc[303]: Marked slot rootfs.0 as good
Oct 08 15:42:36 buildroot rauc[303]: rauc mark: marked slot rootfs.0 as good
Jan 10 16:58:04 buildroot rauc[303]: input bundle: /tmp/buildroot.raucb
Jan 10 16:58:04 buildroot rauc[303]: Active slot bootname: /dev/mmcblk0p5
Jan 10 16:58:04 buildroot rauc[303]: installing /tmp/buildroot.raucb: started
Jan 10 16:58:04 buildroot rauc[303]: Installation 8c72cc7c started
Jan 10 16:58:04 buildroot rauc[303]: installing /tmp/buildroot.raucb: Checking and mounting bundle...
Jan 10 16:58:04 buildroot rauc[303]: Reading bundle: /tmp/buildroot.raucb
Jan 10 16:58:04 buildroot rauc[303]: Detected CRL but CRL checking is disabled!
Jan 10 16:58:04 buildroot rauc[303]: Verifying bundle signature... 
Jan 10 16:58:05 buildroot rauc[303]: Verified detached signature by 'O = Test Org, CN = Test Org Release-1'
Jan 10 16:58:05 buildroot rauc[303]: Mounting bundle '/tmp/buildroot.raucb' to '/run/rauc/bundle'
Jan 10 16:58:05 buildroot rauc[303]: Configured loop device '/dev/loop0' for 51494912 bytes
Jan 10 16:58:05 buildroot rauc[303]: Checking image type for slot type: ext4
Jan 10 16:58:05 buildroot rauc[303]: Image detected as type: *.ext4
Jan 10 16:58:05 buildroot rauc[303]: Checking image type for slot type: vfat
Jan 10 16:58:05 buildroot rauc[303]: Image detected as type: *.vfat
Jan 10 16:58:05 buildroot rauc[303]: Marking target slot rootfs.1 as non-bootable...
Jan 10 16:58:05 buildroot rauc[303]: Marked slot rootfs.1 as bad
Jan 10 16:58:05 buildroot rauc[303]: installing /tmp/buildroot.raucb: Updating slots...
Jan 10 16:58:05 buildroot rauc[303]: installing /tmp/buildroot.raucb: Checking slot rootfs.1
Jan 10 16:58:05 buildroot rauc[303]: Updating slot rootfs.1
Jan 10 16:58:05 buildroot rauc[303]: Updating /dev/mmcblk0p6 with /run/rauc/bundle/rootfs.ext4
Jan 10 16:58:05 buildroot rauc[303]: Mounting slot /dev/mmcblk0p6
Jan 10 16:58:05 buildroot rauc[303]: installing /tmp/buildroot.raucb: Updating slot rootfs.1
Jan 10 16:58:05 buildroot rauc[363]: mount: /run/rauc/rootfs.1: wrong fs type, bad option, bad superblock on /dev/mmcblk0p6, missing codepage or helper program, or other error.
Jan 10 16:58:05 buildroot rauc[363]:        dmesg(1) may have more information after failed mount system call.
Jan 10 16:58:05 buildroot rauc[303]: Ignoring mount error before slot hook error: failed to mount slot: failed to run mount: Child process exited with code 32
Jan 10 16:58:05 buildroot rauc[303]: Running slot 'slot-pre-install' hook for rootfs.1
Jan 10 16:58:05 buildroot rauc[303]: Running slot hook 'slot-pre-install' for rootfs.1
Jan 10 16:58:05 buildroot rauc[303]: opening slot device /dev/mmcblk0p6
Jan 10 16:58:05 buildroot rauc[303]: writing data to device /dev/mmcblk0p6
Jan 10 16:58:23 buildroot rauc[303]: Updating slot rootfs.1 status
Jan 10 16:58:23 buildroot rauc[303]: installing /tmp/buildroot.raucb: Updating slot rootfs.1 done
Jan 10 16:58:23 buildroot rauc[303]: installing /tmp/buildroot.raucb: Checking slot firmware.1
Jan 10 16:58:23 buildroot rauc[303]: Updating slot firmware.1
Jan 10 16:58:23 buildroot rauc[303]: Updating /dev/mmcblk0p3 with /run/rauc/bundle/firmwarefs.vfat
Jan 10 16:58:23 buildroot rauc[303]: opening slot device /dev/mmcblk0p3
Jan 10 16:58:23 buildroot rauc[303]: writing data to device /dev/mmcblk0p3
Jan 10 16:58:23 buildroot rauc[303]: installing /tmp/buildroot.raucb: Updating slot firmware.1
Jan 10 16:58:27 buildroot rauc[303]: Mounting slot /dev/mmcblk0p3
Jan 10 16:58:27 buildroot rauc[303]: Running slot 'slot-post-install' hook for firmware.1
Jan 10 16:58:27 buildroot rauc[303]: Running slot hook 'slot-post-install' for firmware.1
Jan 10 16:58:27 buildroot rauc[303]: Unmounting slot /dev/mmcblk0p3
Jan 10 16:58:27 buildroot rauc[303]: Updating slot firmware.1 status
Jan 10 16:58:27 buildroot rauc[303]: installing /tmp/buildroot.raucb: Updating slot firmware.1 done
Jan 10 16:58:27 buildroot rauc[303]: Marking target slot rootfs.1 as bootable...
Jan 10 16:58:27 buildroot rauc[303]: installing /tmp/buildroot.raucb: All slots updated
Jan 10 16:58:27 buildroot rauc[377]: 0x0000001c 0x80000000 0x00038064 0x00000004 0x80000004 0x00000000 0x00000000
Jan 10 16:58:27 buildroot rauc[303]: Marked slot rootfs.1 as active
Jan 10 16:58:27 buildroot rauc[303]: Installation 8c72cc7c succeeded
Jan 10 16:58:27 buildroot rauc[303]: installing /tmp/buildroot.raucb: finished
Jan 10 16:58:27 buildroot rauc[303]: installing `/tmp/buildroot.raucb` succeeded
```

Note: I have tested it with that [custom](https://github.com/Rtone/raspberrypi-firmware-rauc-bootloader-backend/) bootloader backend, and that [pre-install](https://github.com/Rtone/rtone-br2-external/blob/main/board/raspberrypi/raspberrypi-firmware-hook.sh) slot hook. The bundle is generated by that [genimage](https://github.com/Rtone/rtone-br2-external/blob/main/board/raspberrypi/genimage.cfg#L52-L78) config file.

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -1424,12 +1424,12 @@ static gboolean mount_and_run_slot_hook(const gchar *hook_name, const gchar *hoo
 	g_assert_nonnull(hook_name);
 	g_assert_nonnull(hook_cmd);
 
-	/* mount slot */
+	/* try mount slot */
 	g_message("Mounting slot %s", slot->device);
 	res = r_mount_slot(slot, &ierror);
 	if (!res) {
-		g_propagate_error(error, ierror);
-		goto out;
+		g_warning("Ignoring mount error before slot hook error: %s", ierror->message);
+		g_clear_error(&ierror);
 	}
 
 	/* run slot install hook */
@@ -1438,6 +1438,10 @@ static gboolean mount_and_run_slot_hook(const gchar *hook_name, const gchar *hoo
 	if (!res) {
 		g_propagate_error(error, ierror);
 	}
+
+	/* slot not mounted */
+	if (slot->mount_point == NULL)
+		return res;
 
 	/* finally umount slot */
 	g_message("Unmounting slot %s", slot->device);
@@ -1453,7 +1457,6 @@ static gboolean mount_and_run_slot_hook(const gchar *hook_name, const gchar *hoo
 		}
 	}
 
-out:
 	return res;
 }
 


ejoerns:
Fixes: #276

@gportay Thank you for addressing this! The discussion on the original issue halted back in 2018, but re-reading my [comment](https://github.com/rauc/rauc/issues/276#issuecomment-396188393) I would stick with what I said back then (and your solution is one of those proposed there).

I would also clarify in the docs that a pre-install hook (on a mountable fs slot type) must not assume the slot is mounted/readable but must explicitly check and handle this itself.

gportay:
> I would also clarify in the docs that a pre-install hook (on a mountable fs slot type) must not assume the slot is mounted/readable but must explicitly check and handle this itself.

I am thinking about adding a flag in parameter to the function [mount_and_run_slot_hook()](https://github.com/rauc/rauc/blob/e3956840291874dce6a7734d19eef62c50c5fa66/src/update_handler.c#L1418), so it returns in error or not if the filesystem is mounted or not. This way, we can behave differently if running a pre-install or post-install slot hook.

But, my opinion is to defer the error handling to the script for now and not adding any flags... this introduces a "behavior" change for bundles installing a blob that is not a filesystem (let's say dm-verity hashes) using a post-install slot hook (RAUC would no fail anymore). I wonder if it makes sense to control such things from the `system.conf` or the `manifest.raucm`.

ejoerns:
> > I would also clarify in the docs that a pre-install hook (on a mountable fs slot type) must not assume the slot is mounted/readable but must explicitly check and handle this itself.
> 
> I am thinking about adding a flag in parameter to the function [mount_and_run_slot_hook()](https://github.com/rauc/rauc/blob/e3956840291874dce6a7734d19eef62c50c5fa66/src/update_handler.c#L1418), so it returns in error or not if the filesystem is mounted or not. This way, we can behave differently if running a pre-install or post-install slot hook.

You could just check for the `hook_cmd` being `R_SLOT_HOOK_PRE_INSTALL` and ignore the warning for this case only.

> But, my opinion is to defer the error handling to the script for now and not adding any flags... this introduces a "behavior" change for bundles installing a blob that is not a filesystem (let's say dm-verity hashes) using a post-install slot hook (RAUC would no fail anymore). I wonder if it makes sense to control such things from the `system.conf` or the `manifest.raucm`.

When using a file system type, we must assume that the installed blob is a file system. Otherwise, a different slot type, like `raw` moust be used.

src/update_handler.c
```diff
@@ -1424,12 +1424,12 @@ static gboolean mount_and_run_slot_hook(const gchar *hook_name, const gchar *hoo
...
 	g_assert_nonnull(hook_cmd);
 
-	/* mount slot */
+	/* try mount slot */
 	g_message("Mounting slot %s", slot->device);
 	res = r_mount_slot(slot, &ierror);
 	if (!res) {
-		g_propagate_error(error, ierror);
-		goto out;
+		g_warning("Ignoring mount error before slot hook error: %s", ierror->message);
```
ejoerns, line 1431: This message doesn't make sense here. I guess it is copied and modified from below?

Please also ensure that this applies to pre-install handlers only as mentioned in the other comment.

ejoerns (changes requested):

jluebbe:
In general, we need to allow mount failures before the installation, as the slot can be un-mountable for valid reasons.

The way this PR implements it would lead to cases where `RAUC_SLOT_MOUNT_POINT` is unset in the hook script. That could cause it to do something else by mistake (e.g. `rm -rf $RAUC_SLOT_MOUNT_POINT/bin` would remove `/bin`).

I'm not sure how widespread this issue is or what the best way to handle it would be. We could:

- Pass `$RAUC_MOUNT_PREFIX/mount-error-$SLOT_NAME` in RAUC_SLOT_MOUNT_POINT, but create an immutable file instead. That would (probably?) lead to errors in the script if it tries to access the mount point as a directory.
- Use a different hook name for scripts that claim to handle mount errors correctly.
- Add a flag in the manifest to declare that mount errors are handled correctly. It could be required only if a pre-install hook was used. Older RAUC versions would reject those bundles, though, making it hard to deploy.
- Add a flag to the system conf, which allows executing hooks without mounted filesystems. It would be the responsibility of the user to set an appropriate `min-bundle-version` so that only bundles with correctly implemented hooks can be installed.

@gportay @ejoerns Do you see other options? Do you have preferences?

gportay:
> The way this PR implements it would lead to cases where `RAUC_SLOT_MOUNT_POINT` is unset in the hook script. That could cause it to do something else by mistake (e.g. `rm -rf $RAUC_SLOT_MOUNT_POINT/bin` would remove `/bin`).
> 

Oh that is a good catch!

> I'm not sure how widespread this issue is or what the best way to handle it would be. We could:
> 
>     * Pass `$RAUC_MOUNT_PREFIX/mount-error-$SLOT_NAME` in RAUC_SLOT_MOUNT_POINT, but create an immutable file instead. That would (probably?) lead to errors in the script if it tries to access the mount point as a directory.
> 
>     * Use a different hook name for scripts that claim to handle mount errors correctly.
> 
>     * Add a flag in the manifest to declare that mount errors are handled correctly. It could be required only if a pre-install hook was used. Older RAUC versions would reject those bundles, though, making it hard to deploy.
> 
>     * Add a flag to the system conf, which allows executing hooks without mounted filesystems. It would be the responsibility of the user to set an appropriate `min-bundle-version` so that only bundles with correctly implemented hooks can be installed.
> 
> @gportay @ejoerns Do you see other options? Do you have preferences?

Well, I have no preference for now.

jluebbe:
> I also would not know of good use cases where a _pre-install_ hook needs access to the _target_ partition. The way more common use case would be to use this in a _post-install_ hook, of course.

A _pre-install_ hook could be used to copy some data that needs to be restored later. I wouldn't count that as a good use case, but it's certainly currently possible, so there's probably someone doing that.

jluebbe:
We'd like to understand your use-case a bit better, because we're still unsure what the best approach would be.

What does your pre-install actually do? :) Would it behave differently depending on whether the target slot can be mounted?

jluebbe:
We just noticed that `archive_to_ext4_handler()` runs the `pre-install` hook **without** mounting the target slot, so there is already some inconsistency regarding behavior between raw images and archives.

gportay:
> We'd like to understand your use-case a bit better, because we're still unsure what the best approach would be.
> 
> What does your pre-install actually do? :) Would it behave differently depending on whether the target slot can be mounted?

The use case is simple, I write the A partitions set on the sd-card of the Raspberry Pi; the B partitions set are created at runtime using systemd-repart. The B filesystems are not made, and thus, cannot get mounted.

The bundle has a [pre-install](https://github.com/Rtone/raspberrypi-firmware-rauc-bootloader-backend/blob/main/support/hook.bash#L11-L13) script that generates the `cmdline.txt` with root= kernel parameter using to the variable `$RAUC_SLOT_DEVICE` of the rootfs slot. Then, the bundle's [post-install](https://github.com/Rtone/raspberrypi-firmware-rauc-bootloader-backend/blob/main/support/hook.bash#L15-L19) script installs the `cmdline.txt` to the firmware slot.

This is very hacky as the slot update order is important, but it makes the script agnostic from the partition indexes.

I have solutions for my use case; the command line can be hard-coded in the script; now, the raspberry pi firmware supports multiple `cmdline.txt` thanks to `boot_partition` filters in the file `config.txt`.

That use case is very very nit, and I do not think it helps in guiding to the best approach.

How does RAUC recover from the situation where the slot filesystem get corrupted (and cannot get mounted) and if the bundle has a pre-install script? because the RAUC daemon fails to mount the filesystem and it aborts the update.

PR #1593, opened 2025-01-12
src/config_file: assert NULL returns for key file errors

ejoerns:
The key files methods will return NULL when the key or group was not found. To make this clear and to fix some Coverity false-positives for potential memory leaks, add explicit `g_assert_null()` statements.

Fixes issues like:

    261     		logger->events = g_key_file_get_string_list(key_file, *group, "events", &entries, &ierror);
    262     		if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
    >>>     CID 1587498:  Resource leaks  (RESOURCE_LEAK)
    >>>     Overwriting "logger->events" in "logger->events = g_malloc(16UL)" leaks the storage that "logger->events" points to.
    263     			logger->events = g_malloc(2 * sizeof(gchar *));


jluebbe (approved):

jluebbe:
The style check errors have been fixed with https://github.com/rauc/rauc/pull/1592.

PR #1597, opened 2025-01-15
src/emmc: Error-Out for invalid boot partition configuration and use more explicit errors

ejoerns:
The only valid `BOOT_PARTITION_ENABLE` bits for boot partition handling are:

    0x0: Device not boot enabled
    0x1: Boot partition 1 enabled for boot
    0x2: Boot partition 2 enabled for boot

All other values either mean boot partitions are not selected or are 'Reserved' values. Make the method `r_emmc_read_bootpart()` fail for these values with an explicit error message.

This can prevent RAUC from performing eMMC boot partition operations on actually invalid configurations.

While at it, this PR also introduces dedicated error codes in `src/emmc.c` and adds some cleanup and simplification of code.

### DIFF
--- a/include/emmc.h
+++ b/include/emmc.h
@@ -3,6 +3,14 @@
 #include <glib.h>
 #include <linux/types.h>
 
+#define R_EMMC_ERROR r_emmc_error_quark()
+GQuark r_emmc_error_quark(void);
+
+typedef enum {
+	R_EMMC_ERROR_FAILED,
+	R_EMMC_ERROR_IOCTL,
+} REmmcError;
+
 #define EMMC_BOOT_PARTITIONS			2
 #define INACTIVE_BOOT_PARTITION(part_active)	((part_active + 1) % EMMC_BOOT_PARTITIONS)
 
--- a/include/emmc.h
+++ b/include/emmc.h
@@ -9,6 +9,8 @@ GQuark r_emmc_error_quark(void);
 typedef enum {
 	R_EMMC_ERROR_FAILED,
 	R_EMMC_ERROR_IOCTL,
+	R_EMMC_ERROR_BOOTPART_UDA,
+	R_EMMC_ERROR_BOOTPART_INVALID,
 } REmmcError;
 
 #define EMMC_BOOT_PARTITIONS			2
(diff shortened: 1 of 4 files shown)


src/emmc.c
```diff
@@ -30,42 +36,54 @@ static int r_emmc_read_extcsd(int fd, guint8 extcsd[512])
...
+		default:
+			g_set_error(error,
+					R_EMMC_ERROR,
+					R_EMMC_ERROR_BOOTPART_INVALID,
+					"Invalid (Reserved) eMMC boot part number.");
+			return FALSE;
+	}
 
-	return res;
+	return TRUE; /* unreachable */
```
jluebbe, line 86: This isn't the successful case.

```suggestion
	g_return_val_if_reached(FALSE);
```

include/emmc.h
```diff
@@ -39,9 +49,7 @@
  *
  * @param device eMMC /dev path (/dev/mmcblkX)
  * @param bootpart_active will contain the active boot partition
- * (0 for mmcblkXboot0, 1 for mmcblkXboot1, 6 for mmcblkX user partition and
- * -1 for no active boot partition)
- * for user partition)
+ *        (0 for partition 1, 1 for partition 2, -1 for no active boot partition)
```
jluebbe, line 52: Why not keep boot0/1? "partition 1/2" is less clear in the context of UDA and GP partitions.

src/update_handler.c
```diff
@@ -2208,21 +2208,17 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+		part_active = 1;
+	} else {
+		g_message("Found active eMMC boot partition %sboot%d", realdev, part_active);
 	}
-	g_message("Found active eMMC boot partition %s", part_active_str);
 
 	/* create a temporary RaucSlot with the actual (currently inactive) boot
-	 * partition as device; for simplicity reasons: in case the user partition
-	 * is active use mmcblkXboot1, in case no partition is active use mmcblkXboot0
+	 * partition as device;
```
jluebbe, line 2221:
```suggestion
	 * partition as device.
```

src/emmc.c
```diff
@@ -59,13 +57,7 @@ gboolean r_emmc_read_bootpart(const gchar *device, gint *bootpart_active, GError
...
 	*bootpart_active = active_partition - 1;
 
-	res = TRUE;
-
-out:
-	if (fd >= 0)
-		g_close(fd, NULL);
-
-	return res;
+	return TRUE; /* unreachable */
```
jluebbe, line 86: In the initial commit, the return is still reachable.

jluebbe (changes requested):
Looks good otherwise.

jluebbe (approved):

PR #1607, opened 2025-01-23
qemu-test: few fixes

gportay:
Hello,

I have tried to use `qemu-test` to run the whole test suite, and I faced some issues with my host (Arch Linux, sh -> bash, meson 1.6.1).

I have made a change that moves the mark-good before the interactive shell is started, and I am not sure about that one. I do not understand why it runs after the shell and before the tests :thinking: I have probably a misunderstanding of how to use that script.

Also, is there a good way to ensure mark-good runs after the rauc service is started?

Regards

### DIFF
--- a/qemu-test
+++ b/qemu-test
@@ -15,7 +15,7 @@
 #   passthrough       Enable passthrough of the host's CPU features
 #   shell             Setup qemu test system and provide an interactive shell
 #   system            Setup qemu dummy system that behaves like a target (without reboot!)
-#   service-backtrace Run service under gdb and show backtrace on error
+#   service-backtrace Run service under gdb and show backtrace on error (requires 'system')
 #   asan              Set environment variables to support address sanitizer
 #   test=<test>       Run specified test (for use with git bisect)
 #
--- a/qemu-test-init
+++ b/qemu-test-init
@@ -43,9 +43,6 @@ for x in $(cat /proc/cmdline); do
     SERVICE_BACKTRACE=1
   fi
 done
-if [ -n "$test" ]; then
-  MESON_TEST="$test"
-fi
 
 hostname qemu-test
 
@@ -227,7 +224,7 @@ if [ -n "$INTERACTIVE_SHELL" ]; then
   sleep 1
 fi
 
-if meson test -C $BUILD_DIR "$MESON_TEST" -v; then
+if meson test -C $BUILD_DIR $test -v; then
   touch qemu-test-ok
   echo "RESULT: PASSED"
 else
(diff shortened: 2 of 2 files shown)


qemu-test
```diff
@@ -15,7 +15,7 @@
 #   passthrough       Enable passthrough of the host's CPU features
 #   shell             Setup qemu test system and provide an interactive shell
 #   system            Setup qemu dummy system that behaves like a target (without reboot!)
-#   service-backtrace Run service under gdb and show backtrace on error
+#   service-backtrace Run service under gdb and show backtrace on error (require system)
```
jluebbe, line 18:
```suggestion
#   service-backtrace Run service under gdb and show backtrace on error (requires 'system')
```

qemu-test-init
```diff
@@ -227,7 +224,8 @@ if [ -n "$INTERACTIVE_SHELL" ]; then
   sleep 1
 fi
 
-if meson test -C $BUILD_DIR "$MESON_TEST" -v; then
+eval "set -- $line"
+if meson test -C $BUILD_DIR "$@" -v; then
```
jluebbe, line 228: Were is `$line` set? Perhaps it's easier to just avoid the quoting on $MESON_TEST? If there are spaces in the variable, meson should handle that fine.

jluebbe (changes requested):
Please mention in the commit message of https://github.com/rauc/rauc/pull/1607/commits/20a64c5e4dc24be1dac085dd49b316ef981aa5c5 that the fixed issue only occurs on systems which use `bash` as `sh`.

Also, there's another instance of `$SHELL` in `qemu-test-init`

qemu-test-init
gportay: arf, it is "$test" :facepalm: it is remaining artifact of testing.

Sure, it is another possibility. It is not [shellcheck](https://www.shellcheck.net/wiki/SC2086) compatible.

qemu-test-init
```diff
@@ -218,13 +221,11 @@ if [ -n "$SHELL" ]; then
...
+  sleep 1
 fi
 
-if [ -n "$SERVICE" ]; then
-  rauc status mark-good
-fi
-
-if meson test -C $BUILD_DIR "$MESON_TEST" -v; then
+eval "set -- $test"
+if meson test -C $BUILD_DIR "$@" -v; then
```
jluebbe, line 228: Wouldn't
```suggestion
if meson test -C $BUILD_DIR $test -v; then
```
behave the same?

gportay: Indeed.

As long as the variable is not be double-quoted (otherwise the meson command has an empty string parameter and would fail if no test is given).

The double-quoting prevents from globbing. However the globbing happens by the shell on the host (unless escaped).

And my solution does not prevent globbing, it does shut the shellcheck warning without adding a comment.

gportay:
I will squash the commit, and rebase on top of master and test it again.

PR #1608, opened 2025-01-23
src/context: fix leaking GError in getting custom bootslot

gportay:
Hello,

I am not very familiar with the glib, but I think I have found a possible leak in reading the code of the function `get_custom_bootname()`.

I guess it is safe to auto free `ierror` since it is used internally only, am I right?

Regards,
Gaël

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -146,8 +146,8 @@ static gchar* get_custom_bootname(void)
 	g_autoptr(GSubprocess) handle = NULL;
 	g_autoptr(GDataInputStream) datainstream = NULL;
 	g_autoptr(GPtrArray) args_array = NULL;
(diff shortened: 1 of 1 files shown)


ejoerns:
> I am not very familiar with the glib, but I think I have found a possible leak in reading the code of the function `get_custom_bootname()`.

Indeed.

> I guess it is safe to auto free `ierror` since it is used internally only, am I right?

Yes, it is. (One could also think of printing the `ierror->message` if that's already obtained, but this could also be added later..)

ejoerns (dismissed):

gportay:
> > 
> 
> Yes, it is. (One could also think of printing the `ierror->message` if that's already obtained, but this could also be added later..)

I have hesitated to do it. I will do it though.

Issue #1277, opened 2023-10-30
Detailed progress reporting not working with adaptive update bundles


BubuOT:
**Describe the bug**
The more detailed progress reporting currently seems not to be working with adaptive update bundles:

```
 70% Checking slot rootfs.1
 73% Checking slot rootfs.1 done.
 73% Copying image to rootfs.1
 99% Copying image to rootfs.1 done.
```

**Background information**
Rauc 1.10.1
System image/update bundles built with buildroot.

**To Reproduce**
1. Create a bundle with `adaptive=block-hash-index`
2. Install to a target

**Expected behavior**
Rauc produces intermediate progress reporting while updating a slot. Instead rauc only prints the next progress line after updating a single slot is done.

**Logs**
See above.

**Additional context**
-

jluebbe:
See also https://github.com/rauc/rauc/discussions/1564#discussioncomment-11301626

ejoerns:
FTR: We don't consider this a real bug, but a missing implementation.
(There are also other handlers in RAUC that do not support fine-grained progress, yet.)
However, having support for this would be great.

PR #1619, opened 2025-02-04
src/update_handler: add detailed progress when installing block hash index images

BubuOT:
Use the number of chunks in the source image for tracking installation progress.

Fixes: #1277

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -847,7 +847,7 @@ static gboolean copy_block_hash_index_image_to_dev(RaucImage *image, RaucSlot *s
 			target_old->invalid_below = c;
 		}
 
-		percent = c * 100 / chunk_count;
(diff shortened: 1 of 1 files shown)


src/update_handler.c
```diff
@@ -845,6 +846,13 @@ static gboolean copy_block_hash_index_image_to_dev(RaucImage *image, RaucSlot *s
 			target_written->invalid_from = c+1;
 			target_old->invalid_below = c;
 		}
+
+		percent = c * 100 / chunk_count;
```
ejoerns, line 850: @BubuOT Thank you for addressing this!

Since `c` counts up to `chunk_count` - 1 in the for loop, I guess we have to use

```suggestion
		percent = (c + 1) * 100 / chunk_count;
```

to reach 100% for the last chunk.

PR #1624, opened 2025-02-10
test: improve http tests

jluebbe:
Add some endpoints to the aiohttp-based HTTP server backend so that we can run tests while collecting additional info, such as request headers or ranges.

Also, support this using a class-based fixture and add some basic test cases.

### DIFF
--- a/test/nginx_backend.py
+++ b/test/nginx_backend.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python3
 
+import json
 import os
 import socket
 import sys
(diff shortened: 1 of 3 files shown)


test/nginx_backend.py
```diff
@@ -58,6 +59,56 @@ async def token_get(request):
...
+        return web.Response(text="OK")
+    except json.JSONDecodeError:
+        return web.Response(text="bad JSON", status=400)
+
+
+@routes.get("/get")
+async def get_handler(request):
+    config = request.app["rauc"].get("config", {})
+    if not config:
+        return web.Response(text="not setup yet", status=400)
```
ejoerns, line 89:
```suggestion
        return web.Response(text="not set up yet", status=400)
```

test/test_http.py
```diff
@@ -0,0 +1,67 @@
...
+    resp.raise_for_status()
+    assert resp.status_code == 206
+    assert resp.content == b"hsqs"
+
+    summary = http_server.get_summary()
+    assert summary["requests"] == 1
+    assert summary["range_requests"] == ["0:4"]
+
+
+def test_headers(http_server):
```
ejoerns, line 22: This is more like a test-internal sanity check, right?

jluebbe: I've added some comments.

ejoerns (approved):

PR #1625, opened 2025-02-11
Add support for directories in bundles

jkehne:
The FAQ says to contact you if someone uses subdirectories or symlinks. Unfortunately, we do: We have put some additional files into the bundle which are read by the target device before applying the update. We put those in the bundle because we wanted them to be signed along with the update, and we put them in a directory because it made deployment simpler on our side. At the time we did this, we were using Yocto Langdale/rauc 1.9 because our hardware vendor's BSP was not compatible with anything newer, so we did not know this was going to be problematic. Now, we have already shipped devices which expect the subdirectory to be present and will not accept bundles without a subdirectory.
I ran into the problem when I updated our build to Yocto Scarthgap/rauc 1.12. To fix this for the time being, I created the proposed patch which re-adds directory support. The patch recursively mirrors any directories found in the content directory to the work directory -- including permissions, but not including owner/group information as these are set to root:root anyway by mksquashfs -- and hardlinks any files found in those directories to the appropriate subdirectory in the workdir. 
We have been using this patch in our internal development version (rauc 1.12 on real hardware) for a few weeks now, without problems. I have also built this branch for our target, and it worked fine; both 1.12 and the build from this branch accept the bundles created by the patched version. I have also tested a few additional cases like deeper directory structures and no subdirectories, all of which worked as well.
We do not use symlinks or special files, so the patch does not touch the code handling those.


jkehne:
Excellent idea! Yes, that would work for us. We don't use anything hidden in our bundles.

jkehne:
I added an additional check that throws an error if a directory starting with . is found. Let me know what you think.

jluebbe:
I've force pushed to rebase on master and change the check to only reject top-level hidden subdirs. Would this work for you?

jkehne:
Sure, that works. Thanks you!

ejoerns (approved):
Looks fine for me, too. Thanks @jkehne for your contribution (and @jluebbe for the finalization)!

PR #1627, opened 2025-02-11
src: simplify and fix NFS (external) boot detection

ejoerns:
Since 2928ac06 ("src/main: fix and add support for boot detection with rauc.external"), we consider `"_external_"` in `r_context()->bootslot` as an indicator for an external boot.
While this works well for explicit 'rauc.external' boots, it does not for boots with 'root=/dev/nfs' in the commandline since `r_context()->bootslot` is '/dev/nfs' in this case.

Unify these cases by returning `"_external_"` from `get_cmdline_bootname()` both for the 'rauc.external' as well as the 'root=/dev/nfs' case.

To not loose the info printout for external or nfs, move these to get_cmdline_bootname(), too.

From here on it is now actually safe to use `"_external_"` as an indicator for all kinds of external boots.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -69,8 +69,10 @@ static gchar* get_cmdline_bootname(void)
 	else if (!g_file_get_contents("/proc/cmdline", &contents, NULL, NULL))
 		return NULL;
 
-	if (strstr(contents, "rauc.external") != NULL)
+	if (strstr(contents, "rauc.external") != NULL) {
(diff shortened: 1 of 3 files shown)


michaelolbrich:
I can confirm that this fixes booting with `root=/dev/nfs` but without `rauc.external` in the kernel command-line.

Without it, the rauc service failed to start with: `r_slot_status_load: assertion 'dest_slot' failed`

jluebbe (approved):

PR #1629, opened 2025-02-13
Fix handling for missing boot ID

ejoerns:
This fixes issues that occur when `/proc/sys/kernel/random/boot_id` (or procfs) is not present for some reason.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -426,9 +426,11 @@ static gboolean r_context_configure_target(GError **error)
 	g_clear_pointer(&context->machine_id, g_free);
 
 	context->boot_id = get_boot_id();
-	g_debug("Obtained system boot ID: '%s'", context->boot_id);
+	if (context->boot_id)
+		g_debug("Obtained system boot ID: '%s'", context->boot_id);
 	context->machine_id = get_machine_id();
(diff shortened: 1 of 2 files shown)


jluebbe (changes requested):
Should we document the requirement for `PROC_SYSCTL`? The option only shows up when `EXPERT` is enabled, which is documented as:
```
menuconfig EXPERT
        bool "Configure standard kernel features (expert users)"
        # Unhide debug options, to make the on-by-default options visible
        select DEBUG_KERNEL
        help
          This option allows certain base kernel options and settings
          to be disabled or tweaked. This is for specialized
          environments which can tolerate a "non-standard" kernel.
          Only use this if you really know what you are doing.
```
We probably also need other options that could be disabled in `EXPERT` mode.

ejoerns:
> Should we document the requirement for `PROC_SYSCTL`? 

Thought about this, too, but was quite unsure. In most systems, this option should either be on or even not be visible for deactivation. Thus maybe we can assume that if someone disables it, he knows about the potential impact.

I agree that otherwise, we'd maybe need to check for all basic kernel features we use and figure out if they can potentially be disabled.

jluebbe (approved):

PR #1632, opened 2025-02-20
require glib version 2.64.0

jluebbe:
This allows us to use g_ptr_array_copy()/_extend() from 2.62, g_canonicalize_file() from 2.58 and g_warning_once() from 2.64.

Versions newer than 2.64.0 are used by the oldest supported releases of major distributions.
- Debian bullseye: 2.66.8 (EOL: 2026-08-31)
- Ubuntu focal: 2.64.2 (EOL Standard Support: May 2025)
- Yocto kirkstone: 2.72.3 (EOL: April 2026)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -25,18 +25,18 @@ conf.set('bindir', bindir)
 libexecdir = join_paths(prefixdir, get_option('libexecdir'))
 conf.set('libexecdir', libexecdir)
 
-glibdep = dependency('glib-2.0', version : '>=2.56')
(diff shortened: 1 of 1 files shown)


jluebbe:
@ejoerns Perhaps we should merge this only after the next release.

jluebbe:
This depends on #1633.

ejoerns (approved):
I don't see anything that should prevent us from moving forward here.

PR #1633, opened 2025-02-20
avoid GDestroyNotify cast for g_clear_pointer

jluebbe:
With the cast, it would cause a warning:
```
../src/artifacts.c: In function ‘r_artifact_repo_free’: ../src/artifacts.c:85:43: warning: function called through a non-compatible type
   85 |         g_clear_pointer(&repo->artifacts, (GDestroyNotify)g_hash_table_destroy);
/usr/include/glib-2.0/glib/gmem.h:136:8: note: in definition of macro ‘g_clear_pointer’
  136 |       (destroy) (_ptr);                                  \
      |        ^~~~~~~
```
See also https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/269.

### DIFF
--- a/src/artifacts.c
+++ b/src/artifacts.c
@@ -82,7 +82,7 @@ void r_artifact_repo_free(gpointer value)
 	g_free(repo->path);
 	g_free(repo->type);
 	g_free(repo->data_directory);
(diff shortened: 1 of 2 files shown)


ejoerns (approved):
Should be fine if it still compiles without errors.

PR #1636, opened 2025-02-23
Split backends into dedicated files

gportay:
Hello,

As discussed, this PR splits the bootloader backends into dedicated files.

Regards,
Gaël


jluebbe:
I've split up the commits to make the main refactoring easier to review with `git show --color-moved`: https://github.com/rauc/rauc/compare/master...jluebbe:rauc:split-backends-into-dedicated-files

There is no change in the result. If you're happy with that, could you force push that to this PR branch?

Also, I think we should group the bootloader files by name. Perhaps `bl_barebox.c`, `bl_custom.c`, `bl_efi.c`, ...? @ejoerns

gportay:
> > Also, I think we should group the bootloader files by name. Perhaps `bl_barebox.c`, `bl_custom.c`, `bl_efi.c`, ...? @ejoerns
> 
> That was my first thought before having read your comment, thus agreed. I have no strong preference for the naming as long as it is consistent.

I do not really like the short prefix `bl_`. I would prefer a full prefix such as `bootloader_` or even `bootchooser_`.

Also, they could be move to a dedicated directory `bootloaders` as directories organizes things.

jluebbe:
Moving them to `src/bootloaders` sounds good.

obbardc:
This was always my biggest complaint reading the bootchoosers file, thanks for working on this @gportay !

jluebbe (approved):

jluebbe:
Thanks for the cleanup, @gportay!

PR #1638, opened 2025-02-24
src/context: introduce r_boot_get_current_bootname()

gportay:
This introduces the function r_boot_get_bootname() to retreive the bootname of the current booted slot.

### DIFF
--- a/include/bootchooser.h
+++ b/include/bootchooser.h
@@ -2,7 +2,7 @@
 
 #include <glib.h>
 
-#include "config_file.h"
+#include "context.h"
 
 #define R_BOOTCHOOSER_ERROR r_bootchooser_error_quark()
 GQuark r_bootchooser_error_quark(void);
@@ -19,6 +19,16 @@ GQuark r_bootchooser_error_quark(void);
  * @return TRUE if it is supported, otherwise FALSE
  */
 gboolean r_boot_is_supported_bootloader(const gchar *bootloader)
+G_GNUC_WARN_UNUSED_RESULT;
+
+/**
+ * Get current bootname slot.
+ *
+ * @param config the RaucConfig
+ *
+ * @return bootname, NULL if detection failed
+ */
+gchar *r_boot_get_current_bootname(RaucConfig *config)
 G_GNUC_WARN_UNUSED_RESULT;
 
 /**
--- a/include/bootchooser.h
+++ b/include/bootchooser.h
@@ -25,10 +25,11 @@ G_GNUC_WARN_UNUSED_RESULT;
  * Get current bootname slot.
  *
  * @param config the RaucConfig
+ * @param error return location for a GError, or NULL
  *
  * @return bootname, NULL if detection failed
  */
-gchar *r_boot_get_current_bootname(RaucConfig *config)
+gchar *r_boot_get_current_bootname(RaucConfig *config, GError **error)
 G_GNUC_WARN_UNUSED_RESULT;
 
 /**
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -1568,7 +1568,7 @@ static gboolean custom_backend_set(const gchar *cmd, const gchar *bootname, cons
 }
 
 /* Get current bootname */
-static gchar *custom_backend_get_bootname(RaucConfig *config, GError **error)
+static gchar *custom_get_current_bootname(RaucConfig *config, GError **error)
 {
 	g_autoptr(GSubprocessLauncher) launcher = NULL;
 	g_autoptr(GSubprocess) handle = NULL;
@@ -1728,7 +1728,7 @@ gchar *r_boot_get_current_bootname(RaucConfig *config, GError **error)
 	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
 
 	if (g_strcmp0(config->system_bootloader, "custom") == 0) {
-		res = custom_backend_get_bootname(config, &ierror);
+		res = custom_get_current_bootname(config, &ierror);
 	}
 
 	if (ierror) {
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -1567,6 +1567,53 @@ static gboolean custom_backend_set(const gchar *cmd, const gchar *bootname, cons
 	return TRUE;
 }
 
+static gchar *custom_backend_get_bootname(RaucConfig *config)
+{
+	g_autoptr(GSubprocessLauncher) launcher = NULL;
+	g_autoptr(GSubprocess) handle = NULL;
+	g_autoptr(GDataInputStream) datainstream = NULL;
+	g_autoptr(GPtrArray) args_array = NULL;
+	g_autoptr(GError) ierror = NULL;
+	g_autofree gchar *outline = NULL;
+	GInputStream *instream;
+	int res;
+
+	args_array = g_ptr_array_new();
+	g_ptr_array_add(args_array, config->custom_bootloader_backend);
+	g_ptr_array_add(args_array, (gchar *)("get-current"));
+	g_ptr_array_add(args_array, NULL);
+
+	launcher = g_subprocess_launcher_new(G_SUBPROCESS_FLAGS_STDOUT_PIPE);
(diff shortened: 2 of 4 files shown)


src/bootchooser.c
```diff
@@ -1654,6 +1718,30 @@ static gboolean custom_set_primary(RaucSlot *slot, GError **error)
 	return custom_backend_set("set-primary", slot->bootname, NULL, error);
 }
 
+/* Get current booted bootname */
+gchar *r_boot_get_bootname(RaucConfig *config, GError **error)
+{
+	GError *ierror = NULL;
+	gchar *res = NULL;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
jluebbe, line 1727: Add a check for the config parameter.

src/context.c
```diff
@@ -427,8 +377,12 @@ static gboolean r_context_configure_target(GError **error)
...
 	}
 
-	if (context->bootslot == NULL && (g_strcmp0(context->config->system_bootloader, "custom") == 0)) {
-		context->bootslot = get_custom_bootname();
+	if (context->bootslot == NULL) {
+		context->bootslot = r_boot_get_bootname(context->config, &ierror);
+		if (ierror) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
```
jluebbe, line 385: Aborting on an error here is a behavior change and so should be explained in the commit message. Perhaps it should be a separate commit?

jluebbe (changes requested):

src/bootchooser.c
gportay: ... and should return `NULL`.

gportay: also, I think there are issues with error handling and `g_return_val_if_fail()` in general.

if think most of the code does not check if an error if set to propagate or prefix error, after the function ends with an error, because the function `g_return_val_if_fail(..., FALSE)` does not set the `GError`.

``` c
if (func(&ierror) == FALSE) {
    return FALSE;
}
```

[g_propagate_error(dest, src)[https://docs.gtk.org/glib/func.propagate_error.html] says *src must be non-NULL.*

jluebbe: The `g_return_*` macros log a critical error, which we make always fatal by `g_log_set_always_fatal()` in `main()`.

jluebbe:
> > Hmm, I'm not sure if we should abort if the bootloader backend can't determine the booted slot. `rauc status` sould still work, but `rauc mark` or `rauc install` would fail because we don't know which slots are active. @ejoerns?
> 
> I guess the function should just warn if the script returns empty string. If the script fails for some reason, I guess it is fine to abort.

That would be stricter than what we do for for cmdline parsing. I'd stay with the current behavior of printing a warning if `r_boot_get_bootname()` returns an error.

gportay:
> > > Hmm, I'm not sure if we should abort if the bootloader backend can't determine the booted slot. `rauc status` sould still work, but `rauc mark` or `rauc install` would fail because we don't know which slots are active. @ejoerns?
> > 
> > 
> > I guess the function should just warn if the script returns empty string. If the script fails for some reason, I guess it is fine to abort.
> 
> That would be stricter than what we do for for cmdline parsing. I'd stay with the current behavior of printing a warning if `r_boot_get_bootname()` returns an error.

@jluebbe, is the change you expect?

``` diff
@@ -380,8 +381,8 @@ static gboolean r_context_configure_target(GError **error)
        if (context->bootslot == NULL) {
                context->bootslot = r_boot_get_bootname(context->config, &ierror);
                if (ierror) {
-                       g_propagate_error(error, ierror);
-                       return FALSE;
+                       g_warning("Failed to get bootname: %s", ierror->message);
+                       g_clear_error(&ierror);
                }
        }
```

gportay:
@jluebbe, I have reused `g_message()` (instead of `g_warning()`).

jluebbe:
> While adding the test to avoid the code coverage to fail, I figured out that if the custom bootloader can output an empty string.
> 
> Thus I wonder if this is valid situation, and if this change is something acceptable?
> 
> Note: I have not check yet what happens if the `bootslot` is set to an invalid/not-existent slot string value.

It will raise an error in https://github.com/rauc/rauc/blob/master/src/install.c#L151

> ```diff
> diff --git a/src/bootchooser.c b/src/bootchooser.c
> index 6bfee2f4..298de77e 100644
> --- a/src/bootchooser.c
> +++ b/src/bootchooser.c
> @@ -1616,7 +1616,7 @@ static gchar *custom_backend_get_bootname(RaucConfig *config, GError **error)
>                                 config->custom_bootloader_backend);
>                 return NULL;
>         }
> -       if (!outline) {
> +       if (!outline || *outline == 0) {
>                 g_set_error(
>                                 error,
>                                 R_BOOTCHOOSER_ERROR,
> diff --git a/test/context.c b/test/context.c
> index 12cc6c63..220419a5 100644
> --- a/test/context.c
> +++ b/test/context.c
> @@ -121,7 +121,7 @@ static void test_bootslot_custom_bootloader_no_bootslot(void)
>         r_context_conf()->mock.proc_cmdline = "quiet";
>         g_clear_pointer(&r_context_conf()->bootslot, g_free);
> 
> -       g_assert_cmpstr(r_context()->bootslot, ==, "");
> +       g_assert_null(r_context()->bootslot);
> 
>         r_context_clean();
>  }
> ```

That's probably a sensible fix (as a separate commit in this PR, though :).

src/bootchooser.c
```diff
@@ -1654,6 +1702,21 @@ static gboolean custom_set_primary(RaucSlot *slot, GError **error)
 	return custom_backend_set("set-primary", slot->bootname, NULL, error);
 }
 
+/* Get current booted bootname */
+gchar *r_boot_get_bootname(RaucConfig *config)
```
ejoerns: I guess `r_boot_get_current_bootname()` would be clearer here.

The other methods like `r_boot_get_state()` or `r_boot_get_primary()` all refer to a specific slot instance given in the parameters. `r_boot_get_bootname()` refers to the *current* bootname which is not given in any parameters and thus quite implicit.

src/bootchooser.c
```diff
@@ -1654,6 +1702,21 @@ static gboolean custom_set_primary(RaucSlot *slot, GError **error)
...
+	g_return_val_if_fail(config, NULL);
+
+	if (g_strcmp0(config->system_bootloader, "custom") == 0) {
+		res = custom_backend_get_bootname(config);
+	}
+
+	return res;
+}
+
+/* Get state of current slot */
```
ejoerns, line 1745: Unrelated change, but I would be fine with it if you replace 'current' with 'given' to avoid confusion with the above's 'current'.

include/bootchooser.h
```diff
@@ -21,6 +21,16 @@ GQuark r_bootchooser_error_quark(void);
...
 G_GNUC_WARN_UNUSED_RESULT;
 
+/**
+ * Get current booted bootname slot.
+ *
+ * @param config the RaucConfig
+ *
+ * @return bootname, NULL if detection failed
+ */
+gchar *r_boot_get_bootname(RaucConfig *config);
```
ejoerns: The trailing ";" will be removed in the follow-up commit anyway, but better not not introduce it at all ;)

ejoerns (changes requested):
Just a few minor things left from my side. Looks good for me, otherwise ! :+1:

include/bootchooser.h
gportay: good catch

src/bootchooser.c
gportay: Okay, I have split the commit in two.

gportay: and replace the word at two places.

gportay:
@ejoerns, thanks for your review.

I have added a commit to rename the function `custom_backend_get_bootname()` to `custom_get_current_bootname()` to match the name of the other bootchooser functions.

ejoerns (approved):
Nit: s/add few comments/add a few comments/ in the commit title, but shouldn't prevent us from merging, thus approved from my side.

jluebbe (approved):

Issue #1416, opened 2024-05-13
migrate pytest tests from pydbus to dasbus


jluebbe:
The pytest test suite currently uses [pydbus](https://github.com/LEW21/pydbus) to [access the RAUC service](https://github.com/rauc/rauc/blob/master/test/conftest.py#L10). As pydbus is no longer actively maintained, it should be replaces by something else. [dasbus](https://github.com/dasbus-project/dasbus) seems a likely choice.

- [x] add `python3-dasbus` to https://github.com/rauc/rauc/blob/master/test/Dockerfile
- [ ] update `conftest.py`

jluebbe:
It seems that dasbus isn't very active either, so perhaps https://github.com/python-sdbus/python-sdbus would be a good alternative in the longer term, but it needs explicit interface classes.

BubuOT:
You can autogenerate the interface classes from the rauc dbus xml. I'm accessing rauc's dbus interface with sdbus in our project. :)

jluebbe:
> You can autogenerate the interface classes from the rauc dbus xml. I'm accessing rauc's dbus interface with sdbus in our project. :)

Yes, I sway that, but I'm currently focused on a different part in RAUC and don't want to get sidetracked too far. If your code is public, could you link to it?

BubuOT:
Not public unfortunately, but I can  share the rauc parts here:

`python -m sdbus gen-from-file --block de.pengutronix.rauc.Installer.xml` generates the interface code (or without `--block` for the async variant, which you need for signal handling if I remember correctly)

You use it like this then:
```
class RaucDbusClient:
    def __init__(self) -> None:
        self._proxy = DePengutronixRaucInstallerInterface("de.pengutronix.rauc", "/")

    def get_rauc_operation(self) -> str:
        """Get current rauc operation via dbus"""
        return self._proxy.operation

    def get_rauc_boot_slot(self) -> str:
        """Get current boot slot (A or B)"""
        return self._proxy.boot_slot
```

Or the async variant:
```

class RaucDbusClientAsync:
    def __init__(self) -> None:
        self._proxy = DePengutronixRaucInstallerInterfaceAsync.new_proxy(
            "de.pengutronix.rauc", "/"
        )

    async def get_rauc_operation(self) -> str:
        """Get current rauc operation via dbus"""
        return await self._proxy.operation

    async def get_rauc_boot_slot(self) -> str:
        """Get current boot slot (A or B)"""
        return await self._proxy.boot_slot

[...]

PR #1641, opened 2025-02-25
test: migrate from pydbus to dasbus

jluebbe:
Fixes: #1416

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -300,9 +300,6 @@ def pkcs11(tmp_path_factory):
 
 @pytest.fixture(scope="session")
 def dbus_session_bus(tmp_path_factory):
-    if not have_service():
-        pytest.skip("No service")
-
     addr_r_fd, addr_w_fd = os.pipe()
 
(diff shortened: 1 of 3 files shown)


test/conftest.py
```diff
@@ -600,16 +602,17 @@ def running_service(self, bootslot):
...
+        bus = SessionMessageBus()
 
         # Wait for de.pengutronix.rauc to appear on the bus
         timeout = time.monotonic() + 5.0
         while True:
             time.sleep(0.1)
             try:
-                self.proxy = bus.get("de.pengutronix.rauc", "/")
+                self.proxy = bus.get_proxy("de.pengutronix.rauc", "/")
+                self.proxy.Operation
```
ejoerns, line 613: What's the purpose of this line? I couldn't find it described anywhere.

jluebbe: dasbus seems to communicate with the service only when accessing a property for the first time. So fetch the current Operation and ignore the resut.

PR #1645, opened 2025-02-27
add support for requiring a specific manifest hash during installation

jluebbe:
Check that the hash of the to be installed bundle's manifest matches the expected value. This can be used to when explicit confirmation is needed before installing a bundle by first reviewing the information returned by InspectBundle and then passing the hash here. If the bundle was replaced by a different (but correctly signed) bundle, this is detected by comparing the manifest hashes.

This is also useful with the upcoming polling support if the user wants to automate the polling only and still trigger the installation explicitly.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -1653,6 +1653,15 @@ IN a{sv} *args*:
         identifying the (installation) transaction.
         If not given, RAUC will generate a random one.
 
+    :STRING 'require-manifest-hash', VARIANT 's' <HASH>:
+       Check that the manifest hash of the to-be-installed bundle's matches the
+       expected value.
+       This can be used when explicit confirmation is needed before installing a
+       bundle by first reviewing the information returned by ``InspectBundle``
+       and then passing the hash here.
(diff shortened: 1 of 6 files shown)


docs/reference.rst
```diff
@@ -1639,6 +1639,15 @@ IN a{sv} *args*:
         identifying the (installation) transaction.
         If not given, RAUC will generate a random one.
 
+    :STRING 'require-manifest-hash', VARIANT 's' <HASH>:
+       Check that the hash of the to be installed bundle's manifest matches the
```
ejoerns, line 1643:
```suggestion
       Check that the manifest hash of the to-be-installed bundle matches the
```

docs/reference.rst
```diff
@@ -1639,6 +1639,15 @@ IN a{sv} *args*:
...
+       expected value.
+       This can be used to when explicit confirmation is needed before
```
ejoerns, line 1645:
```suggestion
       This can be used when explicit confirmation is needed before
```

ejoerns (changes requested):

ejoerns (approved):

PR #1650, opened 2025-03-03
src/manifest: ignore the rollout section for now

jluebbe:
The main polling configuration is set in the system.conf. In the future, we may want to further delay or restrict polling/installation, e.g. to specific time windows or for phased roll-outs.

By ignoring unknown options in this section until then, it will be easier to use them later as an intermediate update can be avoided.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -327,6 +327,10 @@ static gboolean parse_manifest(GKeyFile *key_file, RaucManifest **manifest, GErr
 		}
 	}
 
+	/* ignore [rollout] section for now, so that we can add hints/overrides
+	 * for rollout/polling behaviour later */
+	g_key_file_remove_group(key_file, "rollout", NULL);
+
 	if (!check_remaining_groups(key_file, &ierror)) {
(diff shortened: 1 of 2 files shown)


ejoerns:
To give a little more context: this is related to the polling support, described in the simple update server support concept #1114. Noted there as:

> * Add simple roll-out handling by giving each target a random (or fixed via config) value in the range 0.0-1.0 and defining a start time and duration in the manifest. The target can then calculate if it should install the update ((current time - start time)/duration < configured value).

My minor concern is that maybe `[poll]` could be too generic or not descriptive enough. Just thought about `[polling]` or `[rollout]`, or depending on what it should also cover in the future, something like `[update-policy]`?

With `[poll]`, the manifest section would also have the same name as the `system.conf` section proposed in #1114. I'm not sure if that's particularly good or particularly bad.

jluebbe:
We discussed this internally and I'll update it to use ``[rollout]``.

ejoerns (approved):

PR #1653, opened 2025-03-05
utils: add function to format a duration as a human readable string

jluebbe:
This will be useful to print the time until the next poll for the upcoming polling support.


ejoerns (commented):
With a brief note on what that will be useful for, I'd be fine with it.

jluebbe:
@ejoerns I've updated the commit message.

ejoerns (approved):

PR #1621, opened 2025-02-10
context: add support for a system version

jluebbe:
The system version is provided by the system-info handler in the RAUC_SYSTEM_VERSION variable. In a later commit, we'll support exposing it expose as a request header in the initial streaming request.

PR #1654, opened 2025-03-05
send headers for InspectBundle and add support for a system-version HTTP header

jluebbe:
Previously, they were sent only during installation. By moving the header setup to the bundle access level, we'll also be able to use them when polling.

Also, add support for a `RAUC-System-Version` header which uses the information collected from the `system-info` handler (since #1621).


src/main.c
```diff
@@ -1307,6 +1307,9 @@ static gboolean info_start(int argc, char **argv)
 	if (no_check_time)
 		check_bundle_params |= CHECK_BUNDLE_NO_CHECK_TIME;
 
+	g_assert(access_args.http_info_headers == NULL);
+	access_args.http_info_headers = assemble_info_headers(NULL);
```
ejoerns, line 1311: What's the scenario this would be useful for? When using the CLI, the context information will not be ready. When using RAUC without the service, this might not be relevant anymore (since e.g. polling will not work).

test/test_http.py
```diff
@@ -60,9 +63,13 @@ def is_uptime(value):
     return True
 
 
+@pytest.mark.parametrize("api", ["cli", "dbus"])
 @have_json
-def test_info_headers(create_system_files, system, http_server):
+def test_info_headers(create_system_files, system, http_server, api):
```
ejoerns, line 68: This is only indirectly related to the patch series, isn't it?

docs/reference.rst
```diff
@@ -363,6 +363,7 @@ For more information about using the streaming support of RAUC, refer to
 
     .. note:: The machine ID should be considered "confidential" and thus not
        be used over unauthenticated connections or with untrusted servers!
+  * ``system-version``: Enables sending the *system version* as ``RAUC-System-Version`` header field.
```
ejoerns, line 366: In the corresponding commit message: s/disabiguate/disambiguate (or 'distinguish')?

ejoerns (commented):
I thought about the need to mention in the docs that these headers are sent on InspectBundle calls, too, now. But I'm unsure since we obviously did not describe exactly in which cases (installation only) they were sent so far.

src/main.c
jluebbe: For `rauc install`, `r_context_configure_target()` is called even without the service, as it uses `R_CONTEXT_CONFIG_MODE_REQUIRED`. For `rauc info` (which uses `_AUTO`), the target information (mainly the system-info handler) is not loaded. So we probably only get the uptime.

We could change that by calling `r_context_configure_target()` in more cases, but that can be left for later.

test/test_http.py
jluebbe: The DBus and CLI code paths for info are slightly different, so we should test them both.

src/main.c
ejoerns: Documenting this clearly might be sufficient for now.

test/test_http.py
jluebbe: With the last test suite fixes, the patch coverage looks good now.

ejoerns (approved):

PR #1658, opened 2025-03-10
nbd: add support for ETag based caching

jluebbe:
This is useful to avoid fetching the manifest unless it has changed. For now, we don't expose the received ETag over D-Bus.


jluebbe:
This is missing some required commits. Will resubmit in a later.

PR #1659, opened 2025-03-12
src/install: ensure converted pointer is valid before dereferencing

aly-ebee:
fix segmentation fault caused by checking `converted->len` without ensuring `converted` is valid in `pre_install_check`

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -994,7 +994,7 @@ static gboolean pre_install_checks(gchar* bundledir, GPtrArray *install_plans, G
 		}
 
 		if (!g_file_test(plan->image->filename, G_FILE_TEST_EXISTS)) {
-			if (plan->image->converted->len == 0) {
+			if (!plan->image->converted || plan->image->converted->len == 0) {
 				g_set_error(error, G_FILE_ERROR, G_FILE_ERROR_NOENT,
 						"Source image '%s' not found in bundle", plan->image->filename);
 				return FALSE;


jluebbe:
How did you trigger this issue?

The fix itself looks correct.

aly-ebee:
> How did you trigger this issue?

It was a weird set of co-incidences. We created a bundle that was read-only for root but we are running RAUC as a non-root user (I understand this is not the recommended usage but we require it for legal Eichrecht compliance in Germany). We discovered, instead of reporting bundle not found, the service would crash.

jluebbe:
> > How did you trigger this issue?
> 
> It was a weird set of co-incidences. We created a bundle that was read-only for root but we are running RAUC as a non-root user (I understand this is not the recommended usage but we require it for legal Eichrecht compliance in Germany). We discovered, instead of reporting bundle not found, the service would crash.

Ah. This might be because the bundle's root directory [is created as mode 0700](https://github.com/rauc/rauc/blob/master/src/bundle.c#L1179). You may need to patch that or perhaps use a user namespace for the RAUC service.

aly-ebee:
> > > How did you trigger this issue?
> > 
> > 
> > It was a weird set of co-incidences. We created a bundle that was read-only for root but we are running RAUC as a non-root user (I understand this is not the recommended usage but we require it for legal Eichrecht compliance in Germany). We discovered, instead of reporting bundle not found, the service would crash.
> 
> Ah. This might be because the bundle's root directory [is created as mode 0700](https://github.com/rauc/rauc/blob/master/src/bundle.c#L1179). You may need to patch that or perhaps use a user namespace for the RAUC service.

user namespace is a good idea. We will evaluate that when we are able to but since we are currently on a very old kernel, we need to wait a bit.

We were able to get around it for now by using `mksquashfs-args` during bundle generation and change the mode there and it seems to work well. The specific bundle that led to us discovering this issue was created without the extra args.

jluebbe:
@aly-ebee What did you use for `mksquashfs-args`?

aly-ebee:
We used `--mksquashfs-args='-root-mode 755'`

jluebbe:
I've updated the commit message with some background info and to align it to our conventions:
https://github.com/jluebbe/rauc/commit/1bfa826515074f85a21eb3ea431f60a2b8fd88a1

As I don't have write permissions to this PR's repo, could you force-push that commit to your branch?

jluebbe (approved):

PR #1662, opened 2025-03-14
src/nbd: avoid a false positive in coverity

jluebbe:
It doesn't seem to understand that g_strchomp returns the same pointer, so it thinks that the memory allocated by g_strndup is leaked.


ejoerns (approved):
I also saw this and considered just ignoring it, but if coverity is happy with this version, that'd be fine for me too :wink:

PR #1664, opened 2025-03-18
docs/integration: change old override syntax in example

antznin:
The FILESEXTRAPATHS_prepend syntax is no longer valid since honister[1], switch to the new one.

Also change it for IMAGE_INSTALL, and add the missing space after the quote.

[1]: https://docs.yoctoproject.org/migration-guides/migration-3.4.html#override-syntax-changes


jluebbe (approved):

jluebbe:
Thanks!

PR #1665, opened 2025-03-18
docs/integration: fix rauc bbappend

antznin:
We should append `rauc-conf.bb`, not `rauc_<version>.bb`.


ejoerns (approved):
@antznin Thank you for fixing this!

Since `scarthgap`, it's actually `rauc-conf.bb`. I'd assume (i.e. hope) that no new integrations are started on releases before scartghap, thus it should be fine to just change it without referring to the old variant.

ejoerns:
@antznin Not sure if it actually needs to be a warning, but the intro for the Yocto integration section (where also the reference to the README can be found) could be improved at least by giving more credits to the README and adding the note about older releases in a (info-level?) box.

PR #1669, opened 2025-03-20
src/context: priorize bootchooser to select bootslot (for custom bootloader backend)

gportay:
The commit 9af0613c7527ebadc855c1a49a4d10757150194a introduced the new bootchooser API r_boot_get_current_bootname() to allow the backend to determine the current bootslot.

Since then, the bootchooser is able determine the bootslot if it cannot be guessed from the kernel cmdline; it is used by the custom backend only at that moment.

This gives the priority to the bootchooser over the kernel cmdline.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -374,15 +374,15 @@ static gboolean r_context_configure_target(GError **error)
 		g_warning("Ignoring surrounding whitespace in system variant: %s", context->config->system_variant);
 
 	if (context->bootslot == NULL) {
-		context->bootslot = get_cmdline_bootname();
-	}
-
-	if (context->bootslot == NULL) {
 		context->bootslot = r_boot_get_current_bootname(context->config, &ierror);
 		if (ierror) {
 			g_message("Failed to get bootname: %s", ierror->message);
 			g_clear_error(&ierror);
 		}
+	}
+
+	if (context->bootslot == NULL) {
+		context->bootslot = get_cmdline_bootname();
 	}
 
 	g_clear_pointer(&context->boot_id, g_free);
--- a/src/context.c
+++ b/src/context.c
@@ -59,44 +59,41 @@ static gchar* get_boot_id(void)
 	return g_strchomp(contents);
 }
 
-static gchar* get_cmdline_bootname(void)
+static gchar *get_cmdline(void)
 {
-	g_autofree gchar *contents = NULL;
+	gchar *contents = NULL;
+	g_autoptr(GError) ierror = NULL;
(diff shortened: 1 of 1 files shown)


ejoerns:
Sounds reasonable to me to give the custom backend priority for setting the booted slot.

`get-current` remains optional since missing output or any failure in get-current (e.g. due to strict input checking) will not lead to an error and just fall back to the  `root=`-based  bootname.

I guess it would make sense to clarify the detection order in the description in https://rauc.readthedocs.io/en/latest/integration.html#identification-via-custom-backend

jluebbe:
@ejoerns I've not checked the code yet, but we need to ensure that external and NFS override results from the bootloader backend.

ejoerns:
@jluebbe Indeed, that's one corner case I overlooked. Thus maybe it makes sense to just move the call to `r_boot_get_current_bootname()` into `get_cmdline_bootname()` (and rename this?).

It's also a bit inconsistent currently that we check `rauc.external` first, then `rauc.slot` and `bootchooser.active`, THEN `root=/dev/nfs`, and finally other `root=` options.
I'd suggest we check `rauc.external` and `root=/dev/nfs` first, then handle the custom backend, and then the rest.

oli-ben:
> @jluebbe Indeed, that's one corner case I overlooked. Thus maybe it makes sense to just move the call to `r_boot_get_current_bootname()` into `get_cmdline_bootname()` (and rename this?).
> 
> It's also a bit inconsistent currently that we check `rauc.external` first, then `rauc.slot` and `bootchooser.active`, THEN `root=/dev/nfs`, and finally other `root=` options. I'd suggest we check `rauc.external` and `root=/dev/nfs` first, then handle the custom backend, and then the rest.

I ran into the same issue as @gportay, I think this would be a great path forward!

gportay:
@jluebbe, @ejoerns, I have split the `get_cmdline_bootname()` logic in two. I have moved the explicit `rauc slot` before calling the bootloader backend (and the nfs case). I have kept the rauc slot guessing from the root device after.

Also, I am thinking about moving the `barebox` thing in the bootloader backend by adding the cmdline in parameter to the `r_boot_get_current_bootname(..., cmdline, ...)`.

Is that looking good to you?

jluebbe:
I've created #1712 to make the refactoring easier to review. What do you think, @gportay?

gportay:
> I've created #1712 to make the refactoring easier to review. What do you think, @gportay?

I saw, let's close that one.

PR #1672, opened 2025-03-21
add polling support

jluebbe:
This is based on the concept described in https://github.com/rauc/rauc/issues/1114.

Documentation:
- https://rauc--1672.org.readthedocs.build/en/1672/advanced.html#update-polling
- https://rauc--1672.org.readthedocs.build/en/1672/reference.html#polling-section

TODO:
- [x] https://github.com/rauc/rauc/pull/1654 needs to be merged first.
- [x] add test case for inhibit files
- [x] add test case for HTTP 204 status code

PR #1674, opened 2025-03-27
contrib: polling-server: example on bundle HTTP headers and HTTP redirects

hnez:
RAUC can be configured to send custom HTTP headers using the `send-headers` parameter in the `[streaming]` section and `RAUC_HTTP_` keys set by the `system-info` handler script.

These are intended for telemetry collection about devices in the field and controlled rollouts of devices in the wild by providing different bundles based on information in the header.

This will become more relevant with the upcoming polling feature https://github.com/rauc/rauc/pull/1672, which enables the automatic installation of updates using RAUC alone.

This script provides a basic example on how to use these features.

The table of previous polling attempts looks something like this:

rauc-poll

### DIFF
--- a/contrib/polling-server.py
+++ b/contrib/polling-server.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python3
+
+import argparse
+from datetime import datetime
+
+from aiohttp import web
+
+
+class PollingServer:
+    HEADERS = {
+        "boot-id": "RAUC-Boot-ID",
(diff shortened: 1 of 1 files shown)


hnez:
I've used this script in the last few days to test out the polling integration that I've added to the LXA TAC's tacd https://github.com/linux-automation/tacd/pull/90 and it did what I needed.

I think this is ready for review.

jluebbe (approved):

PR #1680, opened 2025-03-28
Allow fallback to bundle download if streaming install failed.

HiFiPhile:
Add an option `bundle-fallback-download`, if set to `true`, allows a streaming install fallback to bundle download if nbd setup failed, which could be caused by server range request handling issue.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -260,6 +260,10 @@ Example configuration:
   mark-bad other' after marking the currently booted slot as good.
   This means that the other slot(s) is/are no longer eligible for fallback.
 
+``bundle-fallback-download`` (optional)
+  If set to ``true``, allow fallback to bundle download if streaming install initialization
+  is failed, for example server failed to response ranged request.
+
 .. _keyring-section:
 
 ``[keyring]`` Section
@@ -1369,8 +1373,8 @@ Command Line Tool
     RAUC_PKCS11_MODULE  Library filename for PKCS#11 module (signing only)
     RAUC_PKCS11_PIN     PIN to use for accessing PKCS#11 keys (signing only)
 
-.. note:: 
-  Using -C / --confopt can not only override settings of the config file but also 
+.. note::
+  Using -C / --confopt can not only override settings of the config file but also
   set new values that haven't been present before.
 
 .. _sec-handler-interface:
--- a/include/config_file.h
+++ b/include/config_file.h
@@ -77,6 +77,8 @@ typedef struct {
 	guint bundle_formats_mask;
 	/* enable complete read before mount */
 	gboolean perform_pre_check;
+	/* Fallback to download if streaming failed */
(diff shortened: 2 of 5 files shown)


jluebbe:
Could you give some details which HTTP server doesn't support range requests? And (if you can) why it's easier to change RAUC than to update/fix the HTTP server? :)

I'm skeptical if we want to have this kind of fallback. As we don't have the privilege separation for the full bundle download, I'd actually prefer to deprecate that mode at some point instead of using it more widely.

HiFiPhile:
> Could you give some details which HTTP server doesn't support range requests? And (if you can) why it's easier to change RAUC than to update/fix the HTTP server? :)

Our products are deployed in buildings, depending on the site the network condition can be very limited even without network to reaching our Internet upgrade server.

To make an upgrade we need a subcontractor technician (who are generally have low IT skills and limited equipments) on site and maybe use something like a FTP(!) server.

> As we don't have the privilege separation for the full bundle download

You mean the file downloading procedure or prevent bundle modification ?

jluebbe:
> > Could you give some details which HTTP server doesn't support range requests? And (if you can) why it's easier to change RAUC than to update/fix the HTTP server? :)
> 
> Our products are deployed in buildings, depending on the site the network condition can be very limited even without network to reaching our Internet upgrade server.
> 
> To make an upgrade we need a subcontractor technician (who are generally have low IT skills and limited equipments) on site and maybe use something like a FTP(!) server.

If you have a web interface on your product, the approach described in https://github.com/rauc/rauc/discussions/1440 would let you avoid the need to reserve space for a temporary download.

> > I'm skeptical if we want to have this kind of fallback
> 
> It's disabled by default, anyone enable it should know the downside.

Still, we'd need to add and maintain test cases for the additional fallback code paths. If there are better solutions than the fallback, I'd like to avoid that.

> > As we don't have the privilege separation for the full bundle download
> 
> You mean the file downloading procedure or prevent bundle modification ?

With streaming, the network access is done via an unprivileged helper process. That's not supported for the full download mode.

If you need to support FTP servers, you must have space for the temporary download anyway. So you could disable steaming and do a full download in every case (if the browser approach above can't work for you).

HiFiPhile:
> If you have a web interface on your product, the approach described in https://github.com/rauc/rauc/discussions/1440 would let you avoid the need to reserve space for a temporary download.

We don't have an UI, we can only set the url from where the bundle is downloaded. Thanks to curl backend the server is not limited to HTTP.

> Still, we'd need to add and maintain test cases for the additional fallback code paths.
> If you need to support FTP servers, you must have space for the temporary download anyway.

Currently once compiled with `STREAMING=1` rauc loss the capability to install outside from a HTTP server. So in order to support FTP server I have to give up all advantages of streaming install when the network condition allows.

Instead of a fallback it could an more explicit argument like `rauc install --download-mode`

PR #1678, opened 2025-03-28
Cleanup dangling bundle mount point before install.

HiFiPhile:
If rauc install process is terminated, the next run will fail with the following error: 
`Failed mounting bundle: Failed to create dm device: Device or resource busy`

The cause is the bundle from previous run is still mounted. This patch address this issue by check bundle mount point and unmount dangling bundle.

Fix #1676

PR #1681, opened 2025-03-30
Do slot external mount check after pre-install handler

HiFiPhile:
Currently if the target slot is already mounted, installation will fail even if unmount is handled by pre-install handler. A 2nd launch is needed to make it work.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -1594,14 +1594,6 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
 
 	log_event_installation_started(args);
 
-	r_context_begin_step("determine_slot_states", "Determining slot states", 0);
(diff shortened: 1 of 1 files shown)


jluebbe:
Normally, RAUC should only update inactive slots. One exception is [[slot.foo.0]allow-mounted=true](https://rauc.readthedocs.io/en/latest/reference.html#slot-slot-class-idx-sections), which can be used with to modify mounted slots via `install` (not `pre-install`) hooks.

Please describe your use-case. How do you synchronize use of that slot with the rest of the system?

HiFiPhile:
> How do you synchronize use of that slot with the rest of the system?

It's the inactive slot which is not mounted in normal case. Here pre-install handler do the cleanup job if previous installation is aborted, as you said in #1678.

PR #1687, opened 2025-04-03
prepare for 1.15 development

jluebbe:


meson.build
```diff
@@ -1,7 +1,7 @@
 project(
   'rauc',
   'c',
-  version : '1.14',
+  version : '1.15',
```
ejoerns, line 4: In the previous iteration, we used something like:

```suggestion
  version : '1.15-dev',
```

PR #1689, opened 2025-04-04
docs/integration: clarify the RAUC_BUNDLE_FORMAT variable

ejoerns:
And link to the class variable description.

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -1645,7 +1645,8 @@ For using the built-in bundle generation, you need to specify some variables:
 
 ``RAUC_BUNDLE_FORMAT``
   Use this to choose the :ref:`sec_ref_formats` for the generated bundle.
-  It currently defaults to ``plain``, but you should use ``verity`` if possible.
+  It currently defaults to ``""`` which will result in a 'plain' bundle, but
+  you should use ``"verity"`` if possible.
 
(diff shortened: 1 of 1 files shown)


jluebbe (changes requested):
The commit message says:
> docs/integration: be more precise on the default RAUC_BUNDLE_FORMAT
> The class actually sets "" instead of "plain" as the default.
> The difference is that "plain" will also generate a [plain] manifest
> section (which is then incompatible with RAUC versions before 1.5).

As I read the class source, `"plain"` doesn't generate a `[plain]` manifest section, but adds a `[bundle]` section with `format=${RAUC_BUNDLE_FORMAT}`.

ejoerns:
> The commit message says:
> 
> > docs/integration: be more precise on the default RAUC_BUNDLE_FORMAT
> > The class actually sets "" instead of "plain" as the default.
> > The difference is that "plain" will also generate a [plain] manifest
> > section (which is then incompatible with RAUC versions before 1.5).
> 
> As I read the class source, `"plain"` doesn't generate a `[plain]` manifest section, but adds a `[bundle]` section with `format=${RAUC_BUNDLE_FORMAT}`.

Typo, will fix.

jluebbe (approved):

Issue #1688, opened 2025-04-03
OpenSSL 3.0.0 has deprecated the ENGINE API


baxeno:
OpenSSL 3.0.0 was released Sep 21, 2024 and it has deprecated the engine API.
`#include <openssl/engine.h>` is used in `src/signature.c`

Release note snippets:

- Deprecated the ENGINE API.
- Added providers, a new pluggability concept that will replace the ENGINE API and ENGINE implementations.

Suggested updates:

- Make get_pkcs11_engine optional behind a compile flag.
- Migrate from engine to providers.

References:

- https://github.com/openssl/openssl/releases/tag/openssl-3.0.0
- https://docs.openssl.org/3.0/man1/openssl-engine/
- https://docs.openssl.org/3.0/man7/migration_guide/#providers-are-a-replacement-for-engines-and-low-level-method-overrides
- https://www.redhat.com/en/blog/openssl-3-providers-rhel-10
- https://blog.hansenpartnership.com/converting-engines-to-openssl-3-providers/

jluebbe:
OpenSSL 3.0.0 was actually released Sep 7, 2021 (the GitHub release page was created much later). For us the main blocker was [support for PKCS#11](https://github.com/latchset/pkcs11-provider) via the providers API, which has only recently reached 1.0. That version is not yet part of stable releases for widely used distributions, so we can't rely on it yet.

As I understand it, the engine API will only be removed with OpenSSL 4.0, which is not under development yet.

So, for now, we've preferred to stay on the engine API to keep the testing effort low. We'd probably want remove engine support and switch the test suite to providers for a release later this year, though, when pkcs11-provider is more widely available. If you have some time to work on this, a PR  would be very appreciated (either making it optional at compile-time if that can be done cleanly or otherwise removing it). :)

Do you have a specific reason to prefer the providers API (beyond the deprecation notices)?

baxeno:
I have been working on adding rauc package to Fedora, since I use rauc (outside of ptxdist) in a firmware signing setup where the developer signature is replaced with an official signature before field deployment. In F41 (release 2024H2) a change proposal was accepted that would _prevent creating new packages dependent on OpenSSL ENGINE API and remove ENGINE dependencies from the existing packages_.

In the latest rauc RPM spec I have added a quick-and-dirty patch that `#ifdef` remove the engine support:
https://github.com/baxeno/fedora-copr-spec/blob/main/rauc_no_openssl_engine.patch

[...]

PR #1690, opened 2025-04-05
meson: add support for building without OpenSSL engine support

baxeno:
Adding -DDISABLE_OPENSSL_ENGINE will disable engine support and remove openssl-devel-engine dependency on Fedora.
OpenSSL engine support will also be removed in 4.0 in favour of providers API.
I have tested an earlier version of the patch in Fedora Copr RPM build.
https://copr.fedorainfracloud.org/coprs/bax/rauc/build/8859137/

Fixes: #1688

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -172,6 +172,10 @@ if composefsdep.found()
   sources_rauc += files('src/artifacts_composefs.c')
 endif
 
+# To allow building against OpenSSL 3.0 and 4.0 without engine support
+# as they are deprecated in favor of providers API
+conf.set10('ENABLE_OPENSSL_PKCS11_ENGINE', get_option('pkcs11_engine'))
+
 gnome = import('gnome')
 dbus_ifaces = files('src/de.pengutronix.rauc.Installer.xml')
 dbus_sources = gnome.gdbus_codegen(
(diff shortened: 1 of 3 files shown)


jluebbe:
Does this allow using PKCS#11 URIs with the provider as described in https://github.com/latchset/pkcs11-provider/blob/main/HOWTO.md? Ideally the tests would work as is in either configuration. For that to work we probably need to add the provider to the CI images first (and configure it?).

I'd slightly prefer to use positive logic for the option, so something like "enable pkcs11 engine". Enabled by default for now.

baxeno:
No, this PR does not add PKCS11 provider support, it only makes PKCS11 engine support optional. I have changed the wording so it's now positive logic (`ENABLE_OPENSSL_PKCS11_ENGINE`).

jluebbe (approved):

ejoerns:
I'd propose to squash these two commits since the first is broken without the second.

Not sure if we should also indicate in the error message how to re-enable engine support or why it's disabled, but that's probably not as relevant as in options that are default false or more common to be deactivated.

Fine for me otherwise.

baxeno:
@ejoerns I will squash the commits as I only did it this way so it was easier to review the change to the existing change.

PR #1693, opened 2025-04-09
README.rst: use explict code block for shell lines

jluebbe:
This also adds a copy button to these commands on GitHub.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -144,19 +144,19 @@ Host (Build) Prerequisites
 -  libcurl3-dev
 -  libssl-dev
 
-::
+.. code-block:: sh
 
    sudo apt-get install build-essential meson libtool libdbus-1-dev libglib2.0-dev libcurl3-dev libssl-dev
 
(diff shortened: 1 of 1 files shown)


jluebbe:
You can see the result here: https://github.com/jluebbe/rauc/tree/readme-copy?tab=readme-ov-file#host-build-prerequisites

jluebbe:
> Neat! However, `console` is maybe even more appropriate (which we use for the docs, too)?

As I understand it, `console` is meant to include output as well (not only the shell command itself as we have here).

ejoerns:
> > Neat! However, `console` is maybe even more appropriate (which we use for the docs, too)?
> 
> As I understand it, `console` is meant to include output as well (not only the shell command itself as we have here).

Not sure how output should be highlighted. Since we don't prepend any prompt character like we do in the docs, I'm fine with both for now, so let's just merge it.

ejoerns (approved):

PR #1695, opened 2025-04-11
src/nbd:set groups of child process

omrisarig13:
When rauc is streaming a bundle during an installation operation, it creates a sub-process, and switches the user of this subprocess, to create e sandbox environment where this user have access only to the required resources in the system.

Until now, when this subprocess received the UID and GID of the streaming user, however, the groups of this user were not set accordingly.
This can cause problems when resources needed for the installation (e.g., installation keys) are available only through group permissions of the sandbox user.

This commit adds relevant logic to the child preparation and setup functions, where the code finds out which groups the user is a part of and sets them accordingly.

Setting the groups of the process must happen before changing the UID of the process, otherwise, the group setting operation will fail with permissions errors.

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -2,6 +2,7 @@
 #define G_LOG_DOMAIN "rauc-nbd"
 
 #include <unistd.h>
+#include <grp.h>
 #include <fcntl.h>
 #include <errno.h>
 #include <string.h>
@@ -1139,13 +1140,28 @@ static gboolean nbd_configure(RaucNBDServer *nbd_srv, GError **error)
 typedef struct {
 	uid_t uid;
 	gid_t gid;
+	gid_t *groups;
+	int ngroups;
 } child_setup_args;
+
+static void clear_child_setup_args(child_setup_args *child_args)
+{
+	g_clear_pointer(&child_args->groups, g_free);
+}
+G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC(child_setup_args, clear_child_setup_args);
 
 static void nbd_server_child_setup(gpointer user_data)
 {
 	/* see signal-safety(7) for functions which can be used here */
 	child_setup_args *args = user_data;
 
+	if (args->groups) {
+		if (setgroups(args->ngroups, args->groups) == -1) {
+			const char *msg = "setgroups failed\n";
+			write(STDOUT_FILENO, msg, strlen(msg));
+			exit(1);
+		}
+	}
 	if (args->gid) {
 		if (setgid(args->gid) == -1) {
 			const char *msg = "setgid failed\n";
@@ -1168,8 +1184,10 @@ static gboolean nbd_server_child_prepare(child_setup_args *args, GError **error)
 	struct passwd passwd = {0};
 	struct passwd *result = NULL;
 	g_autofree gchar *buf = NULL;
+	g_autofree gid_t *groups = NULL;
 	long bufsize;
 	int err;
+	int ngroups;
 
 	g_return_val_if_fail(args != NULL, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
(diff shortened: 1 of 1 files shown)


omrisarig13:
In contrast to the uid and gid values of the child-process, the groups require allocation of dynamic memory (to avoid allocating the maximum value, which is unreasonably high). This means that the groups values of child_setup_args need to be freed. Currently, this is not done very elegantly.
The code is freeing the pointer in the parent process between the forking and the error validation, and is not being freed in the child (which we expect to be okay, as the child continue to "exec" shortly after).
This maybe can be solved in a better problem, but I'm not familiar enough with the code-base to understand what the proper way to do it in rauc is.

src/nbd.c
```diff
@@ -1168,8 +1178,10 @@ static gboolean nbd_server_child_prepare(struct child_setup_args *args, GError *
 	struct passwd passwd = {0};
 	struct passwd *result = NULL;
 	g_autofree gchar *buf = NULL;
+	g_autofree pid_t *groups = NULL;
```
jluebbe, line 1181:
```suggestion
	g_autofree gid_t *groups = NULL;
```

src/nbd.c
```diff
@@ -1205,8 +1217,29 @@ static gboolean nbd_server_child_prepare(struct child_setup_args *args, GError *
...
+			R_NBD_ERROR, R_NBD_ERROR_STARTUP,
+			"cannot get groups for user %s", user);
+		return FALSE;
+	}
+
 	args->uid = result->pw_uid;
 	args->gid = result->pw_gid;
+	/* Transfer ownership, nullified the internal var, to avoid auto-free. */
+	args->groups = groups;
+	groups = NULL;
```
jluebbe, line 1242: You can use `g_steal_pointer()` here.

jluebbe:
> In contrast to the uid and gid values of the child-process, the groups require allocation of dynamic memory (to avoid allocating the maximum value, which is unreasonably high). This means that the groups values of child_setup_args need to be freed. Currently, this is not done very elegantly. The code is freeing the pointer in the parent process between the forking and the error validation, and is not being freed in the child (which we expect to be okay, as the child continue to "exec" shortly after). This maybe can be solved in a better way, but I'm not familiar enough with the code-base to understand what the proper way to do it in rauc is.

You could define a G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC function for `child_setup_args`. Then you can use g_auto with it and don't need to handle the contents seprately.

src/nbd.c
omrisarig13: Good catch, fixed.

omrisarig13:
> > In contrast to the uid and gid values of the child-process, the groups require allocation of dynamic memory (to avoid allocating the maximum value, which is unreasonably high). This means that the groups values of child_setup_args need to be freed. Currently, this is not done very elegantly. The code is freeing the pointer in the parent process between the forking and the error validation, and is not being freed in the child (which we expect to be okay, as the child continue to "exec" shortly after). This maybe can be solved in a better way, but I'm not familiar enough with the code-base to understand what the proper way to do it in rauc is.
> 
> You could define a G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC function for `child_setup_args`. Then you can use g_auto with it and don't need to handle the contents seprately.

Thanks, I didn't know about this option, I've implemented it accordingly now.  
If I haven't missed anything, to use this macros, the object must be referred to as a single word, so I've prepended a commit that adds this typedef.

src/nbd.c
jluebbe: The Yocto build probably doesn't use -Werror.

jluebbe:
Force-pushed to align the commit titles with out convention (no capitalization).

jluebbe:
In the commit message, you write:

> Until now, when this subprocess received the UID and GID of the
> streaming user, however, the groups of this user were not set
> accordingly.
> This can cause problems when resources needed for the installation
> (e.g., installation keys) are available only through group permissions
> of the sandbox use

With installation keys, you mean private keys for TLS client authentication, right?

omrisarig13:
> In the commit message, you write:
> 
> > Until now, when this subprocess received the UID and GID of the
> > streaming user, however, the groups of this user were not set
> > accordingly.
> > This can cause problems when resources needed for the installation
> > (e.g., installation keys) are available only through group permissions
> > of the sandbox use
> 
> With installation keys, you mean private keys for TLS client authentication, right? The subprocess shouldn't really need to access much else...

Yes.

In my specific case, I have a private key (for TLS client authentication) that is stored in a TPM, so the user needs to be part of the 'tss' group to be able to access the relevant devices, so it's a bit more than just a file, but in essence, it is only for the TLS client private-key access.

jluebbe (approved):

jluebbe:
Thanks for your contribution!

PR #1696, opened 2025-04-15
bootloader: efi: add support for getting current bootname

loblik:
When booting Unified kernel image (UKI) with baked-in kernel command line, arguments passed from firmware boot entry are often ignored. This is also behavior of systemd stub, which ignores passed arguments if secure boot is active.

This patch extends EFI bootloader backend with function to get current bootname from UEFI variables, in case there is nothing on command line.

Tested on Jetson Orin with NVIDIA EDK2 port, and systemd stub and efibootmgr v18.


jluebbe:
Could there be cases where the value `BootCurrent` is not correct?

jluebbe:
Do you also use `BootCurrent` to select the corresponding Root-FS partition in the initramfs or do you have a different solution for that?

loblik:
> Could there be cases where the value `BootCurrent` is not correct?

I guess most UEFI ports go thru all active boot entries in  `BootOrder` until they find something which boots. If that is an entry which is not not mentioned in `system.conf`, rauc will fail with [this error](https://github.com/loblik/rauc/blob/ef6a8066b186e03129a07f191413c24c380987f3/src/bootloaders/efi.c#L524). I believe this is the same thing, like booting an UEFI entry with no `rauc.slot` on command line.

> Do you also use `BootCurrent` to select the corresponding Root-FS partition in the initramfs or do you have a different solution for that?

I use the same mechanism. I have rauc without dbus in ramdisk for recovery purposes. So I simply call `rauc status --output-format=shell` in ramdisk script to find root device.

src/bootloaders/efi.c
```diff
@@ -482,3 +496,32 @@ gboolean r_efi_get_state(RaucSlot *slot, gboolean *good, GError **error)
...
+gchar *r_efi_get_current_bootname(RaucConfig *config, GError **error)
+{
+	g_autoptr(GList) bootorder_entries = NULL;
+	g_autolist(efi_bootentry) all_entries = NULL;
+	GError *ierror = NULL;
+	efi_bootentry *bootcurrent = NULL;
+	RaucSlot *slot = NULL;
+	GHashTableIter iter;
+
+	if (!efi_bootorder_get(&bootorder_entries, &all_entries, NULL, &bootcurrent, &ierror)) {
```
jluebbe, line 509: You could use `NULL` instead of `bootorder_entries`.

jluebbe (changes requested):
Please also use `src/bootloaders/efi:` as the commit message prefix for consistency.

jluebbe:
You could also add a test for `r_boot_get_current_bootname` in `bootchooser_efi` in `test/bootchooser.c` to improve the test coverage of the new code.

hnez:
Hi,

I think the feature makes sense and is something we want to have.

I justed wanted to tune in and share an alternative way of solving the question "which slot is currently running" on a EFI system with signed UKIs, because I have also implemented that very recently. Just in case someone stumbles over this pull request in search for an answer to that question.

I have used [Multi-Profile UKIs](https://www.freedesktop.org/software/systemd/man/latest/systemd-stub.html#Multi-Profile%20UKIs) to include two different command lines in one UKI.

# Generating the profiles

```bash
$ ukify build --profile="TITLE=Boot system0" --cmdline="root=/dev/sda3 rauc.slot=system0" \
    --output=system0-profile.efi
$ ukify build --profile="TITLE=Boot system1" --cmdline="root=/dev/sda4 rauc.slot=system1" \
    --output=system1-profile.efi
```

# Generating the signed UKI

```bash
$ ukify build \
    --initrd="…" --linux="…" --cmdline="" --output="uki.efi" \
    --signtool=systemd-sbsign --no-sign-kernel \
    --secureboot-private-key="key.pem" \
    --secureboot-certificate="cert.pem" \
    --join-profile=system0-profile.efi \
    --join-profile=system1-profile.efi
```

This means the (implicit) profile `@0` will have an empty command line set, which likely makes it non-bootable.
This is deliberate, since we have no way of knowing which slot it corresponds to.

# Adding boot entries

```bash
$ efibootmgr --create --disk "/dev/sda1" --part 1 --label "system0" \
    --loader "\\EFI\\LINUX\\UKI.EFI"  --unicode "@1"
$ efibootmgr --create --disk "/dev/sda2" --part 2 --label "system1" \
    --loader "\\EFI\\LINUX\\UKI.EFI"  --unicode "@2"
```

The `--unicode @1` and `--unicode @2` parameters configure the command line passed to the `systemd-stub` by the EFI firmware and are used to select the boot profile.

---

Being able to pass different command lines based on the active boot slot (while still making sure that only signed command lines may be used) can simplify the logic in the initramfs.

jluebbe (approved):

PR #1697, opened 2025-04-15
fix OpenSSL 3.5 compatibility

jluebbe:
In the test suite, we sometimes recreate the context, which caused
multiple initialization of the network and signature dependencies. With
OpenSSL 3.5, this leads to purpose ID collisions, as we try to add the
same purpose multiple times.

Avoid this by remembering if the have initialized our dependencies
already.

Also handle possible gaps in the number space for OpenSSL 3.5.

Fixes: 295fa3a ("src/context.c: fully free context in r_context_clean()")

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -780,7 +780,9 @@ void r_context_register_progress_callback(progress_callback progress_cb)
 
 RaucContext *r_context_conf(void)
 {
-	if (context == NULL) {
+	static gboolean initialized = FALSE;
+
+	if (!initialized) {
 		GError *ierror = NULL;
 
 		// let us handle broken pipes explicitly
@@ -797,6 +799,10 @@ RaucContext *r_context_conf(void)
 			return NULL;
 		}
 
+		initialized = TRUE;
+	}
+
+	if (context == NULL) {
 		context = g_new0(RaucContext, 1);
 		context->progress = NULL;
 		context->install_info = g_new0(RContextInstallationInfo, 1);
--- a/src/signature.c
+++ b/src/signature.c
@@ -93,14 +93,12 @@ gboolean signature_init(GError **error)
 		return FALSE;
 	}
 
-	id = X509_PURPOSE_get_count() + 1;
-	if (X509_PURPOSE_get_by_id(id) >= 0) {
-		g_set_error_literal(
-				error,
-				R_SIGNATURE_ERROR,
-				R_SIGNATURE_ERROR_CRYPTOINIT_FAILED,
-				"Failed to calculate free OpenSSL X509 purpose id");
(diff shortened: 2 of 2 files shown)


jluebbe:
Thanks to Debian for finding this issue: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1102969

jluebbe:
Note that this bug is only triggered in the test suite.

src/signature.c
```diff
@@ -93,14 +93,10 @@ gboolean signature_init(GError **error)
...
 
-	id = X509_PURPOSE_get_count() + 1;
-	if (X509_PURPOSE_get_by_id(id) >= 0) {
-		g_set_error_literal(
-				error,
-				R_SIGNATURE_ERROR,
-				R_SIGNATURE_ERROR_CRYPTOINIT_FAILED,
-				"Failed to calculate free OpenSSL X509 purpose id");
-		return FALSE;
+	/* OpenSSL 3.5 warns that there may be gaps, so we need to search. */
```
ejoerns, line 96: Could we use `X509_PURPOSE_get_unused_id()` optionally with `#if OPENSSL_VERSION_PREREQ(3, 0)` ?

src/signature.c
```diff
@@ -93,14 +93,10 @@ gboolean signature_init(GError **error)
...
+	id = X509_PURPOSE_MAX + 1;
+	while (X509_PURPOSE_get_by_id(id) >= 0) {
```
ejoerns, line 98:
```suggestion
	while (X509_PURPOSE_get_by_id(id) != -1) {
```

would be more appropriate since the method is documented to return `-1` if not found.

src/context.c
```diff
@@ -780,7 +780,9 @@ void r_context_register_progress_callback(progress_callback progress_cb)
 
 RaucContext *r_context_conf(void)
 {
-	if (context == NULL) {
+	static gboolean initialized = FALSE;
```
ejoerns, line 783: Change itself looks appropriate, but I'm unsure how that's a fix for 295fa3a5059784ef65651dd1f31e6d88ae64a2f8, which only adds a missing free call.

In the commit message: 's/the have/we have/'

ejoerns (changes requested):

src/signature.c
jluebbe: I've added a comment.

src/context.c
jluebbe: Resetting context to NULL triggered the reinitialization of OpenSSL during the tests.

ejoerns (approved):

ejoerns (approved):
:see_no_evil:

PR #1699, opened 2025-04-16
some minor cleanups

jluebbe:


include/utils.h
```diff
@@ -45,6 +45,8 @@ G_DEFINE_AUTO_CLEANUP_FREE_FUNC(filedesc, close_preserve_errno, -1)
 
 static inline GSubprocess* r_subprocess_newv(GPtrArray *args, GSubprocessFlags flags, GError **error)
 {
+	g_assert(args->len);
```
ejoerns, line 48: For input arguments, maybe `g_return_val_if_fail()` fits better here?
Could also be combined with the usual GError check, then.

jluebbe: I thought I had some reason to use `g_assert()`. Will change.

ejoerns (approved):

PR #1700, opened 2025-04-16
nbd: prepare for ETag based caching

jluebbe:
This is useful to avoid fetching the manifest unless it has changed. For now, we don't expose the received ETag over D-Bus. We'll use this later in for the native polling mechanism.


ejoerns (commented):
Maybe rename it to 'nbd: *prepare* for ETag-based caching' to make it clearer that this has no functionality, yet? While at it: s/in for/for/ in the commit message.
Fine otherwise.

ejoerns (approved):

PR #1702, opened 2025-04-25
docs/examples: fix grub.cfg example

ecksun:
Prior to this patch if the A side failed to boot the B side would never have been tried.

This is because default=0 is used as a sentinel value, while in fact it is also the case where the A side is selected

I think this happened when someone took the grub.conf from the repo [0] and tried to remove the rescue menuentry. If the rescue menuentry is present using default=0 as a sentinel value works, as it will never be selected unless both sides are broken.

[0]: https://github.com/rauc/rauc/blob/master/contrib/grub.conf

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -77,8 +77,11 @@ out of the box. Instead we use a script to implement it
 
 .. code-block:: sh
 
+  # set default menuentry (Slot A) and timeout (3s)
   default=0
   timeout=3
+
+  any_ok=0
 
   set ORDER="A B"
   set A_OK=0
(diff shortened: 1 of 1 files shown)


jluebbe:
Force-pushed to adjust the commit title to our conventions. Also changed `ANY_OK` to `any_ok` to keep it separate from the persistent variables.

jluebbe (dismissed):

ejoerns:
> Force-pushed to adjust the commit title to our conventions. Also changed `ANY_OK` to `any_ok` to keep it separate from the persistent variables.

Force-pushed to adjust `any_ok` to also not look like one of the built-in variables like `default` or `timeout` (and added a short description for these while at it).

ejoerns (approved):

jluebbe (approved):

Issue #1120, opened 2023-05-09
do ioctl(BLKDISCARD) in copy_raw_image_to_dev


Villemoes:
**Is your feature request related to a problem? Please describe.**
When we install a new squashfs image to a partition, if the new image happens to be (much) smaller than the old, the underlying block device has lots of unused blocks that might as well be scheduled for erasing and possibly put to use elsewhere.

**Describe the solution you'd like**
I think we should do ioctl(BLKDISCARD) on the whole partition in copy_raw_image_to_dev() before we start writing the data. We obviously cannot and should not do this earlier, i.e. not if adaptive updating is enabled.

**Describe alternatives you've considered**
I suppose this could be done in a pre-install hook, but it seems like the kind of thing that one could expect RAUC to handle transparently. And I suppose it's not really a problem with writable file systems that will be expanded to the full partition, since a periodic fstrim would then soonish deal with the unused blocks.

**Additional context**
The above assumes an eMMC with support for async erase. If the block device is not eMMC (say ordinary spinning rust), I don't know what BLKDISCARD would do. Also, we probably don't want to do the BLKDISCARD if it ends up synchronously waiting for the device to actually erase all the sectors. So perhaps this should be opt-in (via system.conf), or we should figure out if there's some way to detect that BLKDISCARD doesn't fall back to a sync erase.

jluebbe:
> **Is your feature request related to a problem? Please describe.** When we install a new squashfs image to a partition, if the new image happens to be (much) smaller than the old, the underlying block device has lots of unused blocks that might as well be scheduled for erasing and possibly put to use elsewhere.

I've also thought about this feature some time ago, but didn't find time to implement it. Depending on the device, this can also make writing the image faster, as it might be able to skip read-modify-write cycles or avoid wear leveling.
 
> **Describe the solution you'd like** I think we should do ioctl(BLKDISCARD) on the whole partition in copy_raw_image_to_dev() before we start writing the data. We obviously cannot and should not do this earlier, i.e. not if adaptive updating is enabled.

Note that, when using tar images in RAUC, we create a new FS. Some mkfs implementations (such as ext4) will then use discard internally.

For adaptive updates, we could discard any empty space after the image to the end of the partition (after the new data is written).

[...]

PR #1706, opened 2025-05-02
src/update_handler: fix memory leak on G_IO_ERROR_NO_SPACE in clear_slot()

Jo3Zhou:
## Summary
This PR improves the robustness of clear_slot() by addressing a potential memory leak when G_IO_ERROR_NO_SPACE occurs

## Changes
- Added explicit g_clear_error() when the error is expected and recoverable.

## Question
Using `ioctl(fd, BLKDISCARD)` to discard blocks rather than zero-write may be more efficient for flash-based block devices?

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -82,11 +82,15 @@ static gboolean clear_slot(RaucSlot *slot, GError **error)
 		 * G_IO_ERROR_NO_SPACE is expected here, because the block
 		 * device is cleared completely
 		 */
-		if (write_count == -1 &&
-		    !g_error_matches(ierror, G_IO_ERROR, G_IO_ERROR_NO_SPACE)) {
-			g_propagate_prefixed_error(error, ierror,
-					"failed clearing block device: ");
-			return FALSE;
+		if (write_count == -1) {
+			if (g_error_matches(ierror, G_IO_ERROR, G_IO_ERROR_NO_SPACE)) {
+				g_clear_error(&ierror);
+				break;
+			} else {
+				g_propagate_prefixed_error(error, ierror,
+						"failed clearing block device: ");
+				return FALSE;
+			}
 		}
 	}
 


jluebbe:
How did you find this leak? :)

a3f:
> Using ioctl(fd, BLKDISCARD) to discard blocks rather than zero-write may be more efficient for flash-based block devices?

At least for eMMCs, this doesn't guarantee reading back zeroes per reading of the code and checking the eMMC v5.1 spec. We want it to translate to a TRIM command, not a DISCARD, so we would need to do BLKZEROOUT instead.

See for more information: https://github.com/barebox/barebox/commit/91a11c7d50df911a3c7bf328e8c4ccbfde201dff

jluebbe:
See also https://github.com/rauc/rauc/issues/1120 for some discussion regarding discard and zeroing.

Jo3Zhou:
> How did you find this leak? :)

Jo3Zhou:
> How did you find this leak? :)

I found it while analysing error-handling paths in the block device clearing logic. Specifically, I noticed that when g_error_matches() returns FALSE, the GError (ierror) isn't cleared or propagated in all cases — which could lead to a memory leak if left unhandled.

Glad to help clean it up! :)

Jo3Zhou:
> > Using ioctl(fd, BLKDISCARD) to discard blocks rather than zero-write may be more efficient for flash-based block devices?
> 
> At least for eMMCs, this doesn't guarantee reading back zeroes per reading of the code and checking the eMMC v5.1 spec. We want it to translate to a TRIM command, not a DISCARD, so we would need to do BLKZEROOUT instead.
> 
> See for more information: [barebox/barebox@91a11c7](https://github.com/barebox/barebox/commit/91a11c7d50df911a3c7bf328e8c4ccbfde201dff)

Thanks for explaining that, I take a look of it

Jo3Zhou:
> See also #1120 for some discussion regarding discard and zeroing.

Thanks for reference, I will have a look

jluebbe (approved):

jluebbe:
Force-pushed to align the commit title with our conventions.

PR #1708, opened 2025-05-12
signature: simplify cleanup of STACK_OF(X509) with g_autoptr

jluebbe:
Ownership-wise, there are two variants of `STACK_OF(X509)`:
- one which does not own the contents and is produced by `CMS_get0_signers()`
- one which does and is produced by `CMS_get1_certs()` Each needs to be freed in the appropriate way.

After increasing the minimum OpenSSL version to 3.2, we could switch to `OSSL_STACK_OF_X509_free` and avoid our helper function.

### DIFF
--- a/include/signature.h
+++ b/include/signature.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <openssl/cms.h>
+#include <openssl/safestack.h>
 #include <glib.h>
 
 #include "manifest.h"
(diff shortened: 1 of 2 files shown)


ejoerns (commented):
Looks fine for me!

I'd just suggest adding some hints about the autoptr typedefs in the code, too.

ejoerns (approved):
:+1:

PR #1710, opened 2025-05-13
Add allowed-signer-cns parameter to keyring section

zentax-dev:
This PR adds a parameter `allowed-signer-cns` to the `keyring` section.
The parameter contains a list of semicolon separated values and checks these values against  the `CommonName` field of the bundle's signer certificates.
This way multiple rauc bundles can be signed with the same root certificate. And then the eligible target devices will just be distinguished by the parameter and the corresponding `CommonName` field in the certificate.


src/signature.c
```diff
@@ -1031,6 +1031,67 @@ static gchar *cms_get_signers(CMS_ContentInfo *cms, GError **error)
...
+
+	// config option not used, so the cert is valid for all devices and we can skip everything
+	if (!allowed_cns) {
+		success = TRUE;
+		goto out;
+	}
+
+	signers = CMS_get0_signers(cms);
+	if (signers == NULL) {
+		g_set_error_literal(
```
jluebbe, line 1052: This will be overwritten after the goto.

We should merge https://github.com/rauc/rauc/pull/1708 first and then simplify this function with the `g_autoptr` helper.

jluebbe (changes requested):

src/signature.c
zentax-dev: rebased and refactored code to use respective autoptr from #1708

src/signature.c
```diff
@@ -1033,6 +1033,56 @@ static gchar *cms_get_signers(CMS_ContentInfo *cms, GError **error)
...
+				R_SIGNATURE_ERROR,
+				R_SIGNATURE_ERROR_GET_SIGNER,
+				"Failed to obtain signer info");
+		return FALSE;
+	}
+
+	for (int i = 0; i < sk_X509_num(signers); i++) {
+		X509_NAME *current_signer = X509_get_subject_name(sk_X509_value(signers, i));
+		int lastpos = -1;
+		X509_NAME_ENTRY *cn;
```
jluebbe, line 1063: Move this to the assignment and make it `const`.

docs/reference.rst
```diff
@@ -302,6 +302,19 @@ signature.
   can not be authenticated.
   If CRL checking is needed, the PKI needs to be structured with this in mind.
 
+``allowed-signer-cns=Name 1;Other Name`` (optional)
+  If this config parameter is set, RAUC will check the ``CommonName`` field
+  of the bundle's signer certificates against this semicolon-separated list.
+  If a certificate with a matching ``CommonName`` is found,
+  the target device is eligible for the update bundle and the update process
+  will begin.
```
jluebbe, line 310:
```suggestion
  If no signature from a certificate with a matching ``CommonName`` is found,
  the target device is not eligible for the update bundle and the update process
  is aborted.
```

docs/reference.rst
```diff
@@ -302,6 +302,19 @@ signature.
...
+  This parameter can be used to sign multiple RAUC bundles
+  with the same root certificate.
+  The eligible target devices are then distinguished only by this parameter
+  and the corresponding leaf certificate on the target device.
+  This parameter can contain multiple values, separated by semicolons.
+  Values can contain whitespaces.
```
jluebbe, line 316:
```suggestion

  This parameter can be used instead of the system compatible if only
  specific certificates in a shared PKI are trusted for specific target devices,
  as the system compatible check can be overridden by a bundle via the
  `install-check` hook.
  If this check fails, no code from the bundle is executed.
  
  This parameter can contain multiple values, separated by semicolons.
  Values can contain whitespace.
```

jluebbe (changes requested):
Seems almost done.

docs/reference.rst
zentax-dev: I think the info that the system compatible check can be overridden would be a good information to have at the compatible parameter in the system secton, so I would move it there and then rephrase your suggestion accordingly

zentax-dev: okay, used your suggestion and changed it slightly

docs/reference.rst
```diff
@@ -302,6 +302,23 @@ signature.
...
+  of the bundle's signer certificates against this semicolon-separated list.
+  If no signature from a certificate with a matching ``CommonName`` is found,
+  the target device is not eligible for the update bundle and the update process
+  is aborted.
+
+  This parameter can be used instead of checking the rauc system compatible
+  if only specific certificates in a shared PKI are trusted
+  for specific target devices.
+  The system compatible check could be overridden by a bundle
+  via the `install-check` hook.
```
jluebbe, line 316:
```suggestion
  This parameter can be used instead of relying on the RAUC system compatible check
  if only specific certificates in a shared PKI are trusted
  for specific target devices,
  as system compatible check could be overridden by a bundle
  via the `install-check` hook.
```

jluebbe (dismissed):

src/signature.c
```diff
@@ -1033,6 +1033,55 @@ static gchar *cms_get_signers(CMS_ContentInfo *cms, GError **error)
...
+		return FALSE;
+	}
+
+	for (int i = 0; i < sk_X509_num(signers); i++) {
+		X509_NAME *current_signer = X509_get_subject_name(sk_X509_value(signers, i));
+		int lastpos = -1;
+		for (;;) {
+			lastpos = X509_NAME_get_index_by_NID(current_signer, NID_commonName, lastpos);
+			// no matching CN entry in any leaf cert, but required by config, so don't continue update
+			if (lastpos == -1) {
```
ejoerns, line 1066: I guess the handling is not consistent, yet.

Having multiple signers is something we do not really support, yet.
However, some of the implementations already potentially allow it, while others explicitly fail.

Having multiple CNs in a single certificate is, to my understanding, technically possible but a very rare case.

The way the code is structured, we can only have a single iteration through the 'signers' stack and thus check only a single certificate (not all, as the comment above suggests).

We should decide if we actually want to make this code multiple-signer and multiple-CN-ready or not.
If not, we can drop the iterations. If yes, then we should fix the handling and abort only when having checked all signers.

ejoerns (changes requested):

src/signature.c
zentax-dev: After discussion, we've decided to just support a single CN per signer. Changes made.

jluebbe (approved):

Issue #1679, opened 2025-03-28
Better bootname parsing priority handling when custom bootloader is used


HiFiPhile:
**Is your feature request related to a problem? Please describe.**
Custom bootloader handler knows better the running system and is more capable about determine the boot device, if `get-current` is available it should be prioritized.

Currently if root device resolution is failed `get-current` won't be called, resulting following error:
```
rauc status
rauc-Message: 22:41:28.188: Using central status file /cal/ota/central.raucs
rauc-Message: 22:41:28.209: Using system config file /etc/rauc/system.conf
rauc-Message: 22:41:28.288: Failed to resolve realpath for 'ubi0:rootfs'
rauc-Message: 22:41:28.296: Failed to resolve realpath for '/dev/ubi3_0'
Failed to determine slot states: Did not find booted slot (matching 'ubi0:rootfs')
```

**Describe the solution you'd like**

A simple solution is:
```diff
diff --git a/src/context.c b/src/context.c
index 1fb65e63..a4f3f3cb 100644
--- a/src/context.c
+++ b/src/context.c
+        context->bootslot = r_boot_get_current_bootname(context->config, &ierror);
+        if (ierror) {
+            g_message("Failed to get bootname: %s", ierror->message);
+            g_clear_error(&ierror);
+        }
+    }
+
     if (context->bootslot == NULL) {
         context->bootslot = get_cmdline_bootname();
     }
 
-    if (context->bootslot == NULL) {
+    if (context->bootslot == NULL && g_strcmp0(context->config->system_bootloader, "custom") != 0) {
         context->bootslot = r_boot_get_current_bootname(context->config, &ierror);
         if (ierror) {
             g_message("Failed to get bootname: %s", ierror->message);
```

**Describe alternatives you've considered**
I'm thinking about explicit boot name (`rauc.external`, `rauc.slot`) vs custom priority:

- `rauc.slot` > custom handler > parse `root=*` 
or
-  custom handler > `rauc.slot` > parse `root=*`

HiFiPhile:
> That's _very_ closely related to [#1669 (comment)](https://github.com/rauc/rauc/pull/1669#issuecomment-2760723355), isn't it?

Yes indeed, sorry didn't check out the PR.

> r_boot_get_current_bootname() into get_cmdline_bootname() (and rename this?).

I agree with your idea.

PR #1599, opened 2025-01-16
bootchooser: add Raspberry Pi firmware initial support

gportay:
This adds an initial backend support for the Raspberry Pi firmware.

PR #1712, opened 2025-05-15
prioritize bootloader backend over root= when detecting booted slot

jluebbe:
We've had a few cases where the bootloader backend should be able to identify the booted slot before looking at the `root=` option in there kernel commandline.

Raspberry Pi: #1599 and #1669
EFI: #1696
UBI: #1679

This PR refactors the commandline parsing to make it easier to understand and then increases the priority of the bootloader backend step. It also updates the related documentation.


gportay:
I have not read the documentation yet.

Code speaking, you took my changes, you split them into dedicated commits, you fixed them, you terminated my work. Well, let's say that I couldn't agree more 😜

And you just fixed the issue I found with the missing read for `root=` before verity.

src/context.c
```diff
@@ -59,44 +44,51 @@ static gchar* get_boot_id(void)
...
-	/* For barebox, we check if the bootstate code set the active slot name
-	 * in the command line */
-	if (g_strcmp0(context->config->system_bootloader, "barebox") == 0) {
-		bootname = regex_match(
-				"(?:bootstate|bootchooser)\\.active=(\\S+)",
-				contents);
-		if (bootname)
-			return bootname;
-	}
+	return bootname;
```
ejoerns, line 81:
```suggestion
	return NULL;
```

might be more explicit here.

src/context.c
```diff
@@ -160,15 +160,15 @@ static gchar* get_bootname(void)
...
-
 	bootname = r_boot_get_current_bootname(context->config, &ierror);
 	if (ierror) {
 		g_message("Failed to get bootname: %s", ierror->message);
 		g_clear_error(&ierror);
 	}
+	if (bootname)
+		return bootname;
+
+	bootname = get_cmdline_bootname_root(cmdline);
```
ejoerns, line 161: For the prioritization task, I'd prefer to have at least a little background description in the commit message. And maybe this could be an option to also add a "Suggested-by" line? ;)

docs/integration.rst
```diff
@@ -576,106 +576,102 @@ are implemented in RAUC.
...
 
-Giving the plain device name is supported, of course.
+Automatic Identification via Kernel Commandline
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. note::
+If none of the approaches above are able to find the booted slot, the generic
+parameters in the commandline are evaluated by RAUC.
+
+The ``root=`` entry contains the device from which device the kernel (or
```
ejoerns, line 647:
```suggestion
The ``root=`` entry contains the device from which the kernel (or
```
wrong in the original documentation, too. But maybe worth fixing while at it.

ejoerns (changes requested):
Looks great overall; just a few points from my side.

ejoerns (approved):

PR #1721, opened 2025-05-28
clarify the write-slot use-case and simplify control flow

jluebbe:


jluebbe:
This was split off from #1717 to merge the uncontroversial changes first.

PR #1722, opened 2025-05-28
src/slot: fix mount point detection for symlinks

amuetzel:
In multiple use cases, it is desirable to use a symlink to a device in the `device=/dev/<device>` attribute of a slot in `system.conf` instead of the actual path of the device.
For example, one might want to identify the slots using attributes such as the disk id or path by using udev-provided symlinks such as `/dev/disk/by-id/...` or `/dev/disk/by-path`.
Or, on devices that support booting from multiple media, dynamically create a symlink such as
`/dev/disk/rootfs-0` that points to the appropriate partition on the currently-active boot device.

This currently makes the mount point detection fail because RAUC cannot identify that such a device is actually mounted to a slot. (`rauc status` does not identify such a slot as `mounted: /<mountpoint>`.)

The reason seems to be the way that `update_external_mount_points()` matches mount points with slots:

To get the currently-mounted devices, `g_unix_mounts_get()` is used and returns the real paths of all mounted devices.
Then, `find_config_slot_by_device()` is used and calls `r_slot_find_by_device()` to match active mounts with config entries.

This iterates over all slots and checks whether the `slot->device` is equal to the given `device` name. In the symlink case, this will fail because the path of the `slot->device` is never resolved. This MR adds handling for the symlink case to `r_slot_find_by_device()` by using `realpath` to resolve the slot's device path.

I could imagine two alternative approaches:
 - Always and unconditionally checking the `realpath()` instead of only when the direct comparison fails. This would introduce a small additional cost in the regular case.
 Also, it would make the function not identify a slot as mounted if the `slot->device` does not point to an existing file that is listed as  a mount point nonetheless - which is relevant at least in one of the `bootchooser` unit tests (but most likely not in real-world use-cases because non-existing devices should not be returned by `g_unix_mounts_get()`).
 - Resolving the paths early after loading the config - I am not sure whether this would be desirable, so I went with the minimally-invasive route.

### DIFF
--- a/src/slot.c
+++ b/src/slot.c
@@ -68,6 +68,19 @@ RaucSlot *r_slot_find_by_device(GHashTable *slots, const gchar *device)
 	while (g_hash_table_iter_next(&iter, NULL, (gpointer*) &slot)) {
 		if (g_strcmp0(slot->device, device) == 0) {
 			return slot;
+		} else {
+			/* The device mentioned in the slot definition may be a symlink (e.g. /dev/disk/by-id/...),
+			 * but the mount point is always passed to this function as the real device path.
+			 * So try to resolve the slot's device path and check if this path matches the device.
+			 */
+			char* resolved = realpath(slot->device, NULL);
+			if (resolved) {
+				if (g_strcmp0(resolved, device) == 0) {
+					g_clear_pointer(&resolved, free);
+					return slot;
+				}
+				g_clear_pointer(&resolved, free);
+			}
 		}
 	}
 


jluebbe:
Hmm. I would have thought that we already have something like this...

We could resolve the path during config loading and store it in a new field of `RaucSlot`. This would allow us to print both in `rauc status` if they differ, which could be helpful.

Resolving during config loading would rely on the assumption that the symlinks don't change at runtime, but I think that's reasonable. If `realpath()` fails, we could print a warning and simply copy the existing string to be backwards compatible.

amuetzel:
This is somewhat similar to #406, which was supposed to fix #388 but never got finished.

The major difference seems to be that #406 also normalizes the path of the mounted device (not just of the mount point specified in the config). That might indeed lead to somewhat strange behaviour if a symlink called like a special filesystem exists in the current RAUC working directory.

For example (not tested, just deduced from the related discussions):
 - Put a symlink `tmpfs -> /dev/example` into the RAUC working directory
 - Set `device=/dev/example` in a slot
 - If any "device" `tmpfs` is returned by `g_unix_mounts_get()`, the slot might incorrectly get identified as mounted.

This whole situation is avoided in this PR because only the mount point specified in the config is resolved/normalized, not the mount point returned by `g_unit_mounts_get()`.

Resolving this at runtime and storing it as an additional field (`normalized_path`?) has some minor pros and cons compared to this PR:

**Pro:**
 - Only need to normalize the path once.
 - More easily accessible, should it be needed in other use cases.
 - Possibly easier to test?
 - 
**Con:**
 - Distributes the related code more widely around the codebase.
 - More integration and test effort.
 - Will not handle possibly-changing symlinks correctly, e.g. when updating slots located on hot-pluggable devices that are plugged in after RAUC is started.

I think the last point is the most important point. Are such use-cases relevant?

PR #1726, opened 2025-06-02
clean up error handling in several functions

jluebbe:
This avoids the use of `goto` and `gboolean res` in several functions, simplifying the control flow.


src/bundle.c
```diff
@@ -2898,8 +2897,6 @@ static gboolean read_complete_dm_device(gchar *dev, GError **error)
 		}
 	}
 
-	g_close(fd, NULL);
-
 	return ret;
```
dwntri: You can get rid of the `ret` variable and return `TRUE` here
It is only used once a few lines above at `ret = FALSE;`
As you break directly afterwards you can use return there.

PR #1727, opened 2025-06-03
src/signature: refactor CMS helper functions

jluebbe:
This gets rid of gotos and prepares for handling of multiple signing certificates with custom policies.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -1001,7 +1001,7 @@ gchar* format_cert_chain(STACK_OF(X509) *verified_chain)
 	return bio_mem_unwrap(text);
 }
 
-static gchar *cms_get_signers(CMS_ContentInfo *cms, GError **error)
+static STACK_OF(X509) *cms_get_signer_certs(CMS_ContentInfo *cms, GError **error)
 {
 	g_return_val_if_fail(cms != NULL, NULL);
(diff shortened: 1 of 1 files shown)


zentax-dev:
looks good to me

zentax-dev (approved):

src/signature.c
```diff
@@ -959,7 +959,7 @@ static gchar* get_cert_time(const ASN1_TIME *time)
 	size = BIO_get_mem_data(mem, &data);
 	ret = g_strndup(data, size);
 
-	BIO_set_close(mem, BIO_CLOSE);
+	g_assert(BIO_set_close(mem, BIO_CLOSE));
```
ejoerns, line 964: manpage notes:

    BIO_set_close() returns 1 on success or <=0 for failure.

but maybe this is just since `BIO_ctrl()` may fail in general?

ejoerns (approved):

src/signature.c
jluebbe: At least for mem BIOs it's basically just an assignment. Looking for the `BIO_CTRL_SET_CLOSE` in the OpenSSL code, I only found similar cases which cannot fail. OpenSLL internal code usually just uses `(void)BIO_set_close()` do silence the warning. I don't see any risk with this change.

Issue #1684, opened 2025-04-02
Crash when mounting a corrupt bundle


graemi:
An attempt to mount a corrupt verity bundle results in a crash of rauc:

Steps to reproduce:

1. Create a bundle in verity format. 
2. Manually delete a few bytes in the bundle payload with an editor, so that payload size is no longer a multiple of 4KiB.
3. Attempt to mount the bundle.

Expected: Mount fails with an error message, rauc exits with error.
Observed: Crash in rauc
```
$  rauc mount /path/to/bundle.raucb
(rauc:976): rauc-DEBUG: 11:43:40.053: No value for key "max-bundle-download-size" in [system] defined - using default value of 8388608 bytes.
(rauc:976): rauc-DEBUG: 11:43:40.055: No mount prefix provided, using /mnt/rauc/ as default
rauc-Message: 11:43:40.056: No data directory or status file set, falling back to per-slot status.
Consider setting 'data-directory=<path>' or 'statusfile=<path>/per-slot' explicitly.
rauc-Message: 11:43:40.056: Using per-slot statusfile. System status information not supported!
rauc-Message: 11:43:40.058: Using system config file /etc/rauc/system.conf
(rauc:976): rauc-DEBUG: 11:43:40.059: Obtained system boot ID: 'f7a11aeb-e8e0-446d-b39f-7c22c1840440'
(rauc:976): rauc-DEBUG: 11:43:40.059: Obtained system machine ID: '8e0859162b3b434ab6bcee067e7c1f23'
(rauc:976): rauc-DEBUG: 11:43:40.059: input bundle: /path/to/bundle.raucb
rauc-Message: 11:43:40.060: Reading bundle: /path/to/bundle.raucb
rauc-Message: 11:43:40.064: Payload size (259505939) is not a multiple of 4KiB. See https://rauc.readthedocs.io/en/latest/faq.html#what-causes-a-payload-size-that-is-not-a-multiple-of-4kib
(rauc:976): rauc-DEBUG: 11:43:40.079: Found valid CMS data
rauc-Message: 11:43:40.083: Verifying bundle signature...
rauc-Message: 11:43:40.093: Verified inline signature by '[manually removed]'
(rauc:976): rauc-DEBUG: 11:43:40.096: Got 2 elements for trust chain
(rauc:976): rauc-DEBUG: 11:43:40.097: bundle payload size: 259505939
rauc-Message: 11:43:40.098: Mounting bundle '/path/to/bundle.raucb'
[...]

PR #1728, opened 2025-06-05
src/dm: replace data size assertion by explicit gerror

jfgreenmt:
When using a modified raucb bundle with payload that is not a multiple of 4096, the following assertion happens:

"rauc-CRITICAL **: r_dm_setup: assertion 'dm->data_size > 0 && dm->data_size % 4096 == 0' failed"

Repalce the assertion with a proper gerror failure.

Fixes: #1684
Reported-by: Tobias Graemer <tobias.graemer@mt.com>
Signed-off-by: Janek Filus <janek.filus@mt.com>

### DIFF
--- a/src/dm.c
+++ b/src/dm.c
@@ -99,8 +99,19 @@ gboolean r_dm_setup(RaucDM *dm, GError **error)
 	g_return_val_if_fail(dm->uuid != NULL, FALSE);
 	g_return_val_if_fail(dm->lower_dev != NULL, FALSE);
 	g_return_val_if_fail(dm->upper_dev == NULL, FALSE);
-	g_return_val_if_fail(dm->data_size > 0 && dm->data_size % 4096 == 0, FALSE);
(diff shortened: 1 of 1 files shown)


src/dm.c
```diff
@@ -99,9 +99,19 @@ gboolean r_dm_setup(RaucDM *dm, GError **error)
 	g_return_val_if_fail(dm->uuid != NULL, FALSE);
 	g_return_val_if_fail(dm->lower_dev != NULL, FALSE);
 	g_return_val_if_fail(dm->upper_dev == NULL, FALSE);
-	g_return_val_if_fail(dm->data_size > 0 && dm->data_size % 4096 == 0, FALSE);
```
ejoerns, line 102: This removes the `dm->data_size > 0` check.

Maybe keep this one and just check for the 4096 below.

While at it: s/Repalce/Replace/ in your commit message

jfgreenmt: Agree, will update asap

jluebbe (approved):

PR #1732, opened 2025-06-15
Extend prevent-late-fallback by lock-counter

zentax-dev:
This feature allows to lock the the ``remaining_attempts`` counter.
When ``remaining_attempts`` is locked, the bootloader should not decremented and  incremented the variable anymore during each boot.
It is active when a slot is _marked good_ and inactive when a slot is _marked active_.
This way it prevents fallback to an earlier version, whilst inhibiting additional write cycles to the target medium. 
The status can be printed out with `barebox-state` and `rauc status`.
In a prelimary talk with @ejoerns the decision was made to not add this to the _D-Bus_ interface yet. It will be added in a future pull request.
This feature also needs to be supported by the bootloader. 
So far, a patch has been handed in for _barebox_ to support this feature, see [](https://lists.infradead.org/pipermail/barebox/2025-June/051393.html)

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -890,6 +890,31 @@ boot target is required:
           };
   };
 
+.. _integration-barebox-lock-counter:
+Optionally, if ``prevent-late-fallback=lock-counter`` is set,
+another single variable must be added.
+
+.. code-block:: Devicetree
+
+  bootstate {
+
+          [...]
+
+          attempts_locked@14 {
+                  reg = <0x14 0x4>;
+                  type = "uint32";
+          };
+  };
+
+This makes barebox aware of the counter locking state due to the
+the ``remaining_attempts`` counter.
+Boot counter locking is activated when 'rauc status mark-good' is executed
+implicitly during a RAUC update or explicitly through manual commands.
+The current value of the ``attempts_locked`` state variable can be read via the commands
+``barebox-state -d`` and ``rauc status``.
+It is not yet available via the D-Bus interface and thus cannot be displayed
+when calling ``rauc status`` with the service enabled.
+
 .. warning::
   This example shows only a highly condensed excerpt of setting up Barebox
   state for bootchooser.
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -256,13 +256,32 @@ Example configuration:
   It has no effect for ``plain`` bundles, as the signature verification already checks the
   whole bundle.
 
-``prevent-late-fallback=<true/false>`` (optional)
-  In some use-cases, fallback to an older version must be prevented after the
+``prevent-late-fallback=<disable/mark-bad/lock-counter>`` (optional)
+  .. note:: The former existing Values *<true/false>* for this option are deprecated.
+     They will be removed in a future release, Please use the new values.
+
+  In order to get the same behaviour set ``prevent-late-fallback=mark-bad``.
+  For some use-cases, fallback to an older version must be prevented after the
   update is completed successfully (``rauc status mark-good`` executed from the
-  new version).
-  If this option is enabled, RAUC will execute the equivalent of ``rauc status
-  mark-bad other`` after marking the currently booted slot as good.
-  This means that the other slot(s) is/are no longer eligible for fallback.
+  new version). This setting is slot independent.
+
+  Supported values are:
+
+  * ``disable`` Default if not set. Do not prevent fallback to older version.
+  * ``mark-bad`` If this option is enabled, RAUC will execute the equivalent of
+    'rauc status mark-bad other' after marking the currently booted slot as good.
+    This means that the other slot(s) is/are no longer eligible for fallback.
+    It behaves exactly like the *deprecated* ``prevent-late-fallback=false``.
+  * ``lock-counter`` This options allows RAUC to maintain awareness of
+    their respective states.
+    The ``slots-locking`` state is stored in a separate variable that must be
+    recognized by the ``bootloader``.
+    Currently, this feature is only supported when ``bootloader``
+    is set to ``barebox``, and the ``&bootstate`` in the devicetree must
+    include a new variable, which is described in the
+    :ref:`integration chapter <integration-barebox-lock-counter>`.
+    In case the variable is not set in the ``bootloader``, a warning is emitted
+    after an update in order to keep the system in an updatable state.
 
 .. _keyring-section:
 
--- a/include/bootchooser.h
+++ b/include/bootchooser.h
@@ -56,6 +56,26 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
 G_GNUC_WARN_UNUSED_RESULT;
 
 /**
+ * Enable or disable locking of the boot counter.
+ * This variable is not set per slot, but is valid for all slots.
+ *
+ * @param locked TRUE to lock counting, FALSE to unlock it
+ * @param error return location for a GError, or NULL
+ *
+ * @return TRUE if successful, FALSE if failed
+ */
+gboolean r_boot_set_counters_lock(gboolean locked, GError **error);
+
+/**
+ * Check current status of boot counter locking in the bootloader.
+ *
+ * @param locked return location for the related value in the bootloader
+ * @param error return location for a GError, or NULL
+ *
+ * @return TRUE if successful, FALSE if failed
+ */
+gboolean r_boot_get_counters_lock(gboolean *locked, GError **error);
+/**
  * Mark slot as primary boot option of its slot class.
  *
  * @param slot Slot to mark
--- a/src/bootchooser.c
+++ b/src/bootchooser.c
@@ -155,6 +155,62 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
 	return res;
 }
 
+gboolean r_boot_set_counters_lock(gboolean locked, GError **error)
+{
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	gboolean res = FALSE;
+	GError *ierror = NULL;
+	if (g_strcmp0(r_context()->config->system_bootloader, "barebox") == 0) {
+		res = r_barebox_set_lock_counter(locked, &ierror);
+	} else {
+		g_set_error(
+				error,
+				R_BOOTCHOOSER_ERROR,
+				R_BOOTCHOOSER_ERROR_NOT_SUPPORTED,
+				"Using boot lock-counter for bootloader '%s' not supported yet", r_context()->config->system_bootloader);
+		return FALSE;
+	}
+
+	if (!res) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"%s backend: ", r_context()->config->system_bootloader);
+	}
+
+	return res;
+}
+
+gboolean r_boot_get_counters_lock(gboolean *locked, GError **error)
+{
+	GError *ierror = NULL;
+	gboolean res = FALSE;
+
+	g_return_val_if_fail(locked != NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (g_strcmp0(r_context()->config->system_bootloader, "barebox") == 0) {
+		res = r_barebox_get_lock_counter(locked, &ierror);
+	} else {
+		g_set_error(
+				error,
+				R_BOOTCHOOSER_ERROR,
+				R_BOOTCHOOSER_ERROR_NOT_SUPPORTED,
+				"Counter locking not yet supported for bootloader '%s'", r_context()->config->system_bootloader);
+		return FALSE;
+	}
+
+	if (!res) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"%s backend: ", r_context()->config->system_bootloader);
+	}
+
+	return res;
+}
+
 /* Get slot marked as primary one */
 RaucSlot *r_boot_get_primary(GError **error)
 {
--- a/src/bootloaders/barebox.c
+++ b/src/bootloaders/barebox.c
@@ -216,6 +216,105 @@ gboolean r_barebox_set_state(RaucSlot *slot, gboolean good, GError **error)
 	return TRUE;
 }
 
+/* This freezes the remaining attempts counters in barebox by setting the attempts_locked variable */
+gboolean r_barebox_set_lock_counter(gboolean locked, GError **error)
+{
+	GError *ierror = NULL;
+	g_autoptr(GPtrArray) pairs = g_ptr_array_new_full(1, g_free);
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	g_ptr_array_add(pairs, g_strdup_printf(BOOTSTATE_PREFIX ".attempts_locked=%i",
+			locked ? 1 : 0));
+
+	if (!barebox_state_set(pairs, &ierror)) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+/* This returns the value of the attempts_locked variable in barebox which
+ * controls the locking of the remaining_attempts counter */
+gboolean r_barebox_get_lock_counter(gboolean *locked, GError **error)
+{
+	GError *ierror = NULL;
+	g_autoptr(GSubprocess) sub = NULL;
+	GInputStream *instream = NULL;
+	g_autoptr(GDataInputStream) datainstream = NULL;
+	g_autofree gchar *outline = NULL;
+	gchar *endptr = NULL;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	g_autoptr(GPtrArray) args = g_ptr_array_new_full(4, g_free);
+	g_ptr_array_add(args, g_strdup(BAREBOX_STATE_NAME));
+	/* does not support -n here as it is also called during context setup and will cause recursion */
+	g_ptr_array_add(args, g_strdup("-g"));
+	g_ptr_array_add(args, g_strdup_printf(BOOTSTATE_PREFIX ".attempts_locked"));
+	g_ptr_array_add(args, NULL);
+
+	sub = r_subprocess_newv(args, G_SUBPROCESS_FLAGS_STDOUT_PIPE, &ierror);
+	if (!sub) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"Failed to start " BAREBOX_STATE_NAME ": ");
+		return FALSE;
+	}
+
+	instream = g_subprocess_get_stdout_pipe(sub);
+	datainstream = g_data_input_stream_new(instream);
+
+	outline = g_data_input_stream_read_line(datainstream, NULL, NULL, &ierror);
+	if (!outline) {
+		/* Having no error set means no content to read */
+		if (ierror == NULL)
+			g_set_error(
+					error,
+					R_BOOTCHOOSER_ERROR,
+					R_BOOTCHOOSER_ERROR_PARSE_FAILED,
+					"No content to read");
+		else
+			g_propagate_prefixed_error(
+					error,
+					ierror,
+					"Failed parsing " BAREBOX_STATE_NAME " output: ");
+		return FALSE;
+	}
+
+	guint64 lock_value = g_ascii_strtoull(outline, &endptr, 10);
+	if (lock_value == 0 && outline == endptr) {
+			g_set_error(
+					error,
+					R_BOOTCHOOSER_ERROR,
+					R_BOOTCHOOSER_ERROR_PARSE_FAILED,
+					"Failed to parse value: '%s'", outline);
+		return FALSE;
+	} else if (lock_value == G_MAXUINT64 && errno != 0) {
+		g_set_error(
+				error,
+				R_BOOTCHOOSER_ERROR,
+				R_BOOTCHOOSER_ERROR_PARSE_FAILED,
+				"Return value overflow: '%s', error: %d", outline, errno);
+		return FALSE;
+	}
+
+	if (!g_subprocess_wait_check(sub, NULL, &ierror)) {
+		g_propagate_prefixed_error(
+				error,
+				ierror,
+				"Failed to run " BAREBOX_STATE_NAME ": ");
+		return FALSE;
+	}
+
+	if (lock_value > 0)
+		*locked = TRUE;
+
+	return TRUE;
+}
+
 /* Get slot marked as primary one */
 RaucSlot *r_barebox_get_primary(GError **error)
 {
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -131,6 +131,29 @@ out:
 	return res;
 }
 
+static gboolean parse_late_fallback_mode(const gchar *mode_str, RConfigLateFallback *mode, GError **error)
+{
+	g_return_val_if_fail(mode != NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (g_strcmp0(mode_str, "disable") == 0) {
+		*mode = R_CONFIG_FALLBACK_DISABLE;
+		return TRUE;
+	} else if (g_strcmp0(mode_str, "mark-bad") == 0) {
+		*mode = R_CONFIG_FALLBACK_MARK_BAD;
+		return TRUE;
+	} else if (g_strcmp0(mode_str, "lock-counter") == 0) {
+		*mode = R_CONFIG_FALLBACK_LOCK_COUNTER;
+		return TRUE;
+	}
+	g_set_error(error,
+			G_KEY_FILE_ERROR,
+			G_KEY_FILE_ERROR_INVALID_VALUE,
+			"Invalid prevent-late-fallback value '%s'. Must be one of: disable, mark-bad, lock-counter, true, false",
+			mode_str);
+	return FALSE;
+}
+
 #define RAUC_LOG_EVENT_CONF_PREFIX "log"
 
 static gboolean r_event_log_parse_config_sections(GKeyFile *key_file, RaucConfig *config, GError **error)
@@ -465,13 +488,33 @@ static gboolean parse_system_section(const gchar *filename, GKeyFile *key_file, 
 	if (dtbvariant)
 		c->system_variant_type = R_CONFIG_SYS_VARIANT_DTB;
 
+	/* Boolean value is deprecated. To keep it simple and easy to remove in the future,
+	 * we assign the old boolean value to their respective new enum values.
+	 * Both are internally mapped to integers so:
+	 * FALSE=R_CONFIG_FALLBACK_DISABLE=0
+	 * TRUE=R_CONFIG_FALLBACK_MARK_BAD=1
+	 */
 	c->prevent_late_fallback = g_key_file_get_boolean(key_file, "system", "prevent-late-fallback", &ierror);
 	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
-		c->prevent_late_fallback = FALSE;
+		c->prevent_late_fallback = R_CONFIG_FALLBACK_DISABLE;
 		g_clear_error(&ierror);
+	} else if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE)) {
+		/* Boolean is deprecated, so it's probably the new string format */
+		g_clear_error(&ierror);
+		g_autofree gchar *mode_str = g_key_file_get_string(key_file, "system", "prevent-late-fallback", &ierror);
+		if (ierror) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+		if (!parse_late_fallback_mode(mode_str, &c->prevent_late_fallback, &ierror)) {
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
 	} else if (ierror) {
 		g_propagate_error(error, ierror);
 		return FALSE;
+	} else {
+		g_info("Using bool for 'prevent-late-fallback' is deprecated. Use one of 'disable', 'mark-bad' or 'lock-counter' instead.");
 	}
 	g_key_file_remove_key(key_file, "system", "prevent-late-fallback", NULL);
 
--- a/src/context.c
+++ b/src/context.c
@@ -4,6 +4,7 @@
 #include <string.h>
 
 #include "bootchooser.h"
+#include "bootloaders/barebox.h"
 #include "config_file.h"
 #include "context.h"
 #include "event_log.h"
@@ -404,6 +405,31 @@ static gboolean r_context_configure_target(GError **error)
 
 	if (context->bootslot == NULL) {
 		context->bootslot = get_bootname();
+	}
+
+	/* If prevent_late_fallback=lock-counter, it needs a variable available in barebox-state.
+	 * We can not use r_boot_get_lock_counter here, as it uses r_context() to get the variable
+	 * and would cause a recursion. But we can access context directly here. */
+	if (context->config->prevent_late_fallback == R_CONFIG_FALLBACK_LOCK_COUNTER) {
+		if (g_strcmp0(context->config->system_bootloader, "barebox") == 0) {
+			if (context->config->system_bb_statename || context->config->system_bb_dtbpath) {
+				g_set_error_literal(error, R_BOOTCHOOSER_ERROR,
+						R_BOOTCHOOSER_ERROR_NOT_SUPPORTED,
+						"Providing custom name, state or path not yet supported for locking");
+				return FALSE;
+			}
+			gboolean locked = FALSE;
+			if (!r_barebox_get_lock_counter(&locked, &ierror)) {
+				/* If we would throw an error here, RAUC would fail and it might not be possible to execute updates anymore. */
+				g_warning("Failed to read barebox lock counter: %s", ierror->message);
+				g_clear_error(&ierror);
+			}
+		} else {
+			g_set_error(error, R_BOOTCHOOSER_ERROR, R_BOOTCHOOSER_ERROR_NOT_SUPPORTED,
+					"prevent-late-fallback is set to 'lock-counter', but not supported by selected bootloader: %s",
+					context->config->system_bootloader);
+			return FALSE;
+		}
 	}
 
 	g_clear_pointer(&context->boot_id, g_free);
--- a/src/main.c
+++ b/src/main.c
@@ -1429,6 +1429,7 @@ typedef struct {
 	gchar *bootslot;
 	GHashTable *slots;
 	GVariant *artifacts;
+	gboolean lock_counter;
 } RaucStatusPrint;
 
 static void free_status_print(RaucStatusPrint *status)
@@ -1600,7 +1601,9 @@ static gchar* r_status_formatter_readable(RaucStatusPrint *status)
 	g_string_append(text, "=== System Info ===\n");
 	g_string_append_printf(text, "Compatible:  %s\n", status->compatible);
 	g_string_append_printf(text, "Variant:     %s\n", status->variant);
-	g_string_append_printf(text, "Booted from: %s (%s)\n\n", bootedfrom ? bootedfrom->name : NULL, status->bootslot);
+	g_string_append_printf(text, "Booted from: %s (%s)\n", bootedfrom ? bootedfrom->name : NULL, status->bootslot);
+	if (!ENABLE_SERVICE && status->lock_counter)
+		g_string_append_printf(text, "Counter lock: enabled");
 
 	g_string_append(text, "=== Bootloader ===\n");
 	if (!status->primary)
@@ -1679,6 +1682,9 @@ static gchar* r_status_formatter_shell(RaucStatusPrint *status)
 	slotstring = g_strjoinv(" ", (gchar**) slotnames->pdata);
 	r_ptr_array_add_printf(entries, "RAUC_SYSTEM_SLOTS=%s", slotstring);
 	g_free(slotstring);
+
+	r_ptr_array_add_printf(entries, "RAUC_LOCK_COUNTER=%s", status->lock_counter ? "enable" : "disabled");
+
 	slotstring = g_strjoinv(" ", (gchar**) slotnumbers->pdata);
 	r_ptr_array_add_printf(entries, "RAUC_SLOTS=%s", slotstring);
 	g_free(slotstring);
@@ -1826,6 +1832,9 @@ static gchar* r_status_formatter_json(RaucStatusPrint *status, gboolean pretty)
 
 	json_builder_set_member_name(builder, "boot_primary");
 	json_builder_add_string_value(builder, status->primary ? status->primary->name : NULL);
+
+	json_builder_set_member_name(builder, "lock-counter");
+	json_builder_add_string_value(builder, status->lock_counter ? "enable" : "disabled");
 
 	json_builder_set_member_name(builder, "slots");
 	json_builder_begin_array(builder);
@@ -2239,6 +2248,19 @@ static gboolean status_start(int argc, char **argv)
 			g_printerr("Failed getting primary slot: %s\n", ierror->message);
 			g_clear_error(&ierror);
 		}
+
+		/* This checks for inconsistency between configured option in rauc config
+		 * and the current state of the required barebox variable.
+		 * So users get notified about it early */
+		gboolean locked = FALSE;
+		if (r_context()->config->prevent_late_fallback == R_CONFIG_FALLBACK_LOCK_COUNTER) {
+			if (!r_boot_get_counters_lock(&locked, &ierror)) {
+				g_printerr("Failed to determine boot counter lock state: %s\n", ierror->message);
+				g_clear_error(&ierror);
+				g_printerr("Inconsistency in 'lock-counter' handling: enabled, but not supported by the bootloader\n");
+			}
+		}
+		status_print->lock_counter = locked;
 
 		status_print->compatible = g_strdup(r_context()->config->system_compatible);
 		status_print->variant = g_strdup(r_context()->config->system_variant);
--- a/src/mark.c
+++ b/src/mark.c
@@ -1,6 +1,8 @@
 #include "bootchooser.h"
+#include "config_file.h"
 #include "event_log.h"
 #include "context.h"
+#include "glib.h"
 #include "install.h"
 #include "mark.h"
 #include "slot.h"
@@ -141,6 +143,22 @@ gboolean r_mark_active(RaucSlot *slot, GError **error)
 	r_slot_status_load(slot);
 	slot_state = slot->status;
 
+	/* Unlock boot counting before marking the slot primary. Reversing the order would create
+	 * an unintended corner case where a power cut could leave the system configured to boot from a new,
+	 * untested slot without the ability to fall back if that slot fails to boot properly, as it would
+	 * still be locked. So it is better to prepare the fallback mechanisms first. */
+	if (r_context()->config->prevent_late_fallback == R_CONFIG_FALLBACK_LOCK_COUNTER) {
+		if (!r_boot_set_counters_lock(FALSE, &ierror)) {
+			if (g_error_matches(ierror, R_BOOTCHOOSER_ERROR, R_BOOTCHOOSER_ERROR_NOT_SUPPORTED)) {
+				g_propagate_error(error, ierror);
+				return FALSE;
+			}
+			/* If we would throw an error here, RAUC would fail and it might not be possible to execute updates anymore. */
+			g_warning("Unable to set lock-counter in bootloader. Check your bootloader setup: %s", ierror->message);
+			g_clear_error(&ierror);
+		}
+	}
+
 	if (!r_boot_set_primary(slot, &ierror)) {
 		g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_MARK_BOOTABLE,
 				"failed to activate slot %s: %s", slot->name, ierror->message);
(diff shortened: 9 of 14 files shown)


docs/reference.rst
```diff
@@ -252,13 +252,53 @@ Example configuration:
   It has no effect for ``plain`` bundles, as the signature verification already checks the
   whole bundle.
 
-``prevent-late-fallback=<true/false>`` (optional)
+``prevent-late-fallback=<true/false>`` (deprecated, optional)
+  .. note:: Values *<true/false>* for this option are deprecated. They will be removed in a future release
+    The syntax has changed. Please use the new values.
+    In order to get the same behaviour set ``prevent-late-fallback=mark-bad``.
+
+``prevent-late-fallback=<disabled/mark-bad/lock-counter>`` (optional)
```
ejoerns: What's the benefit of starting a new section here? I'd just mention the deprecated keys in the list below.

docs/reference.rst
```diff
@@ -252,13 +252,53 @@ Example configuration:
...
+  * ``lock-counter`` This is a slot independent setting that preserves
+    the status of individual slots, allowing RAUC to maintain awareness of
+    their respective states.
+    The ``slots-locking`` state is stored in a separate variable that must be
+    recognized by the ``bootloader``.
+    Currently, this feature is only supported when ``bootloader``
+    is set to ``barebox``, and the ``&bootstate`` in the devicetree must
+    include a new variable, for example:
+
+  .. code-block:: Devicetree
```
ejoerns: This isn't indented correctly. It might also make sense to move these details into the barebox integration chapter where we have the other dtb examples anyway.

include/bootchooser.h
```diff
@@ -55,6 +55,28 @@ G_GNUC_WARN_UNUSED_RESULT;
...
+/**
+ * Enable or disable locking of boot counter.
+ * This variable is not set per slot, but is valid for all slots.
+ *
+ * @param locked value to set for lock-counter in bootloader
+ * @param error return location for a GError, or NULL
+ *
+ * @return TRUE if successful, FALSE if failed
+ */
+gboolean r_boot_set_lock_counter(gboolean locked, GError **error);
```
ejoerns: For me this reads as 'set the counter that counts locking'.
To stick close to this one, maybe use 
```suggestion
gboolean r_boot_set_counters_lock(gboolean locked, GError **error);
```
or similar.

include/bootchooser.h
```diff
@@ -55,6 +55,28 @@ G_GNUC_WARN_UNUSED_RESULT;
 gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
 G_GNUC_WARN_UNUSED_RESULT;
 
+/**
+ * Enable or disable locking of boot counter.
+ * This variable is not set per slot, but is valid for all slots.
+ *
+ * @param locked value to set for lock-counter in bootloader
```
ejoerns: This isn't really clear.

```suggestion
 * @param TRUE to lock counting, FALSE to unlock
```

include/bootchooser.h
```diff
@@ -55,6 +55,28 @@ G_GNUC_WARN_UNUSED_RESULT;
...
+/**
+ * Check current status of lock-counter in bootloader
+ * This variable is not set per slot, but is valid for all slots.
+ *
+ * @param locked return location for value of lock-counter
+ * @param error return location for a GError, or NULL
+ *
+ * @return TRUE if successful, FALSE if failed
+ */
+gint r_boot_get_lock_counter(gboolean *locked, GError **error);
```
ejoerns: See notes on the set method.

src/bootchooser.c
```diff
@@ -155,6 +156,47 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
 	return res;
 }
 
+gboolean r_boot_set_lock_counter(gboolean locked, GError **error)
+{
+	gboolean res = FALSE;
+	GError *ierror = NULL;
+	if (g_strcmp0(r_context()->config->system_bootloader, "barebox") == 0) {
+		res = r_barebox_set_lock_counter(locked, &ierror);
```
ejoerns, line 165: ierror is set but never propagated.

Like all other r_boot_* methods, this needs a 

``` c
        if (!res) {
                g_propagate_prefixed_error(
                                error,
                                ierror,
                                "%s backend: ", r_context()->config->system_bootloader);
        }
```

just before `return res;`.

src/bootchooser.c
```diff
@@ -155,6 +156,47 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
 	return res;
 }
 
+gboolean r_boot_set_lock_counter(gboolean locked, GError **error)
+{
```
ejoerns, line 159: Add input guards here as you did in the get method alreay.

```suggestion
{
        g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```

src/bootchooser.c
```diff
@@ -155,6 +156,47 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
...
+}
+
+gboolean r_boot_get_lock_counter(gboolean *locked, GError **error)
+{
+	GError *ierror = NULL;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (g_strcmp0(r_context()->config->system_bootloader, "barebox") == 0) {
+		if (!r_barebox_get_lock_counter(locked, &ierror)) {
```
ejoerns: Better handle errors consistently. I'd prefer to also use `res =` here and add a generic error propagation at the method's end, as I proposed for the set method.

src/bootchooser.c
```diff
@@ -155,6 +156,47 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
...
+			g_propagate_error(error, ierror);
+			return FALSE;
+		}
+	} else {
+		g_set_error(
+				error,
+				R_BOOTCHOOSER_ERROR,
+				R_BOOTCHOOSER_ERROR_NOT_SUPPORTED,
+				"Using boot lock-counter for bootloader '%s' not supported yet", r_context()->config->system_bootloader);
```
ejoerns: `boot lock-counter` sounds a bit imprecise. Maybe:

```suggestion
				"Counter locking not yet supported for bootloader '%s'", r_context()->config->system_bootloader);
```

src/bootloaders/barebox.c
```diff
@@ -120,7 +120,13 @@ static gboolean barebox_state_get(const gchar *bootname, BareboxSlotState *bb_st
 	return TRUE;
 }
 
-/* names: list of gchar, values: list of gint */
+/**
+ * @brief This function forwards a list of key->value pairs to barebox
```
ejoerns: 'forwards [..] to barebox' is too broad. If we add documentation for this, we should be more precise here.

```suggestion
 * @brief Sets barebox state variables based on a key->value list.
```

src/bootloaders/barebox.c
```diff
@@ -120,7 +120,13 @@ static gboolean barebox_state_get(const gchar *bootname, BareboxSlotState *bb_st
...
+ *
+ * @param pairs GPtrArray of key: list of gchar -> value: list of gint
```
ejoerns: The original comment was actually invalid since 93bc4f79895fc91d2d22e52ad6a27a4519a0541a, but the correct description is more like 
```suggestion
 * @param pairs GPtrArray of "var=value" strings to set as barebox state variables.
```

src/bootloaders/barebox.c
```diff
@@ -120,7 +120,13 @@ static gboolean barebox_state_get(const gchar *bootname, BareboxSlotState *bb_st
...
+ * @param error	Error to propagate to the calling function
```
ejoerns:
```suggestion
 * @param error	return location for a GError, or NULL
 
```

src/bootloaders/barebox.c
```diff
@@ -199,6 +205,104 @@ gboolean r_barebox_set_state(RaucSlot *slot, gboolean good, GError **error)
...
+	g_autoptr(GSubprocess) sub = NULL;
+	GInputStream *instream = NULL;
+	g_autoptr(GDataInputStream) datainstream = NULL;
+	g_autofree gchar *outline = NULL;
+	gchar *endptr = NULL;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	g_autoptr(GPtrArray) args = g_ptr_array_new_full(6, g_free);
+	g_ptr_array_add(args, g_strdup(BAREBOX_STATE_NAME));
```
ejoerns, line 252: This wouldn't support `config->system_bb_statename`.

It might make sense to rework the barebox state *get* handling to be analogous to the state *set* handling (with passing items to get as GPtrArray key/values.).

This should also save a majority of the code duplication here.

src/main.c
```diff
@@ -1540,7 +1541,8 @@ static gchar* r_status_formatter_readable(RaucStatusPrint *status)
 	g_string_append(text, "=== System Info ===\n");
 	g_string_append_printf(text, "Compatible:  %s\n", status->compatible);
 	g_string_append_printf(text, "Variant:     %s\n", status->variant);
-	g_string_append_printf(text, "Booted from: %s (%s)\n\n", bootedfrom ? bootedfrom->name : NULL, status->bootslot);
+	g_string_append_printf(text, "Booted from: %s (%s)\n", bootedfrom ? bootedfrom->name : NULL, status->bootslot);
+	g_string_append_printf(text, "Lock counter: %s\n\n", status->lock_counter ? "enabled" : "disabled");
```
ejoerns:
```suggestion
	g_string_append_printf(text, "Counter lock: %s\n\n", status->lock_counter ? "enabled" : "disabled");
```

meson_options.txt
```diff
@@ -2,22 +2,22 @@
 option(
   'service',
   type : 'boolean',
-  value : 'true',
+  value : true,
```
ejoerns, line 5: As this is unrelated, maybe move it into a separate PR.

src/config_file.c
```diff
@@ -965,8 +987,17 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
 	c->prevent_late_fallback = g_key_file_get_boolean(key_file, "system", "prevent-late-fallback", &ierror);
 	if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND)) {
-		c->prevent_late_fallback = FALSE;
+		c->prevent_late_fallback = R_CONFIG_FALLBACK_DISABLE;
+		g_clear_error(&ierror);
+	} else if (g_error_matches(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE)) {
+		/* Boolean is deprecated, so it's probably the new string format */
 		g_clear_error(&ierror);
+		g_autofree gchar *mode_str = g_key_file_get_string(key_file, "system", "prevent-late-fallback", &ierror);
+		// TODO[lsc]: error handling here
```
ejoerns: Yes, error handling is still missing here.

src/context.c
```diff
@@ -398,6 +399,21 @@ static gboolean r_context_configure_target(GError **error)
...
 	}
 
+	/* If prevent_late_fallback=lock-counter, it needs a variable available in barebox-state.
+	 * We can not use r_boot_get_lock_counter here, as it uses r_context() to get the variable
+	 * and would cause a recursion. But we can access context directly here. */
+	if (context->config->prevent_late_fallback == R_CONFIG_FALLBACK_LOCK_COUNTER) {
+		if (g_strcmp0(context->config->system_bootloader, "barebox") == 0) {
+			gboolean locked = FALSE;
+			if (!r_barebox_get_lock_counter(&locked, &ierror)) {
+				g_propagate_error(error, ierror);
```
ejoerns:
```suggestion
				g_propagate_error(error, ierror);
				return FALSE;
```

src/main.c
```diff
@@ -2166,6 +2174,19 @@ static gboolean status_start(int argc, char **argv)
 			g_clear_error(&ierror);
 		}
 
+		/* This checks for inconsistency between configured option in rauc config
+		 * and the current state of the required barebox variable. So  */
```
ejoerns: Should the sentence continue here?

src/main.c
```diff
@@ -2166,6 +2174,19 @@ static gboolean status_start(int argc, char **argv)
...
+		gboolean locked = FALSE;
+		if (r_context()->config->prevent_late_fallback == R_CONFIG_FALLBACK_LOCK_COUNTER) {
+			if (!r_boot_get_lock_counter(&locked, &ierror)) {
+				g_printerr("Failed to determine prevent-late-fallback state: %s\n", ierror->message);
```
ejoerns:
```suggestion
				g_printerr("Failed to determine boot counter lock state: %s\n", ierror->message);
```

src/main.c
```diff
@@ -2166,6 +2174,19 @@ static gboolean status_start(int argc, char **argv)
...
+				g_clear_error(&ierror);
+			} else if (!locked) {
+				g_printerr("Inconsistency lock-counter state: enabled in rauc but not in bootloader");
```
ejoerns:
```suggestion
				g_printerr("Inconsistency in 'lock-counter' handling: enabled, but not supported by the bootloader");
```

src/mark.c
```diff
@@ -153,8 +155,15 @@ gboolean r_mark_active(RaucSlot *slot, GError **error)
 	g_free(slot_state->activated_timestamp);
 	now = g_date_time_new_now_utc();
 	slot_state->activated_timestamp = g_date_time_format(now, "%Y-%m-%dT%H:%M:%SZ");
-	slot_state->activated_count++;
 
+	if (r_context()->config->prevent_late_fallback == R_CONFIG_FALLBACK_LOCK_COUNTER) {
```
ejoerns: Is there a specific reason for placing it between two `slot_state` variable updates?

Another potential issue might be that if `r_boot_set_primar()` succeeds but `r_boot_set_lock_counter()` fails for some reason, you might end in a system that is locked to the new slot, i.e. which does not provide a fallback anymore. So maybe changing the order of the  `r_boot_set_lock_counter()` and `r_boot_set_primar()` call might make sense here?

This would also be consistent with the order used in `r_mark_good()` below because the locking then surrounds the other bootloader operations.

docs/reference.rst
```diff
@@ -252,13 +252,53 @@ Example configuration:
...
-  This means that the other slot(s) is/are no longer eligible for fallback.
+
+  Supported values are:
+
+  * ``disable`` Default if not set. Do not prevent fallback to older version.
+  * ``mark-bad`` If this option is enabled, RAUC will execute the equivalent of
+    'rauc status mark-bad other' after marking the currently booted slot as good.
+    This means that the other slot(s) is/are no longer eligible for fallback.
+    It behaves exactly like the *deprecated* ``prevent-late-fallback=false``.
+  * ``lock-counter`` This is a slot independent setting that preserves
```
ejoerns: `prevent-late-fallback` is always slot-independent, no matter which value you give.

docs/reference.rst
```diff
@@ -252,13 +252,53 @@ Example configuration:
...
+
+            [...]
+
+            attempts_locked@14 {
+                    reg = <0x14 0x4>;
+                    type = "uint32";
+            };
+    };
+
+  With this configuration, the bootloader is aware of the locking state due to the
```
```suggestion
  With this configuration, barebox is aware of the locking state due to the
```

docs/reference.rst
```diff
@@ -252,13 +252,53 @@ Example configuration:
...
+  usage of the variable in the devicetree. So the bootloader stops decrementing
+  the ``remaining_attempts`` counter.
+  Slots locking is activated when 'rauc status mark-good' is executed
```
```suggestion
  Boot counter locking is activated when 'rauc status mark-good' is executed
```

docs/reference.rst
```diff
@@ -252,13 +252,53 @@ Example configuration:
...
+  and deactivated when 'rauc status mark-active' is executed. This can happen either
+  implicitly during a RAUC update or explicitly through manual commands.
+  The current value of the ``slots_locked`` variable can be read via the commands
```
```suggestion
  The current value of the ``attempts_locked`` state variable can be read via the commands
```

docs/reference.rst
```diff
@@ -252,13 +252,53 @@ Example configuration:
...
+  ``barebox-state -d`` or and ``rauc status``.
```
```suggestion
  ``barebox-state -d`` and ``rauc status`` (without service).
```

docs/reference.rst
```diff
@@ -252,13 +252,53 @@ Example configuration:
...
+  It is not yet available via die D-Bus interface.
```
```suggestion
  It is not yet available via the D-Bus interface and thus cannot be displayed when calling `rauc status` with the service enabled.
```

src/config_file.c
```diff
@@ -965,8 +987,17 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
 
 	c->prevent_late_fallback = g_key_file_get_boolean(key_file, "system", "prevent-late-fallback", &ierror);
```
ejoerns: This now assigns a boolean to an enum; if this is intentional, it might at least deserve a comment in code.

And, if we deprecate the boolean values, we should at least print some kind of deprecation info/warning.
I guess this can easily be added in an `else` statement below.

ejoerns (changes requested):
@zentax-dev Thank you for the contribution!

I had a look at the code and left some comments where I think the feature still deserves some rework. I guess most are just about proper wording/naming and error handling.

You should also point out in the documentation when and why boot counter locking might be preferable to the existing 'mark bad other' solution we have.

docs/reference.rst
dwntri: This way, I think it is more obvious, that the old keys are replaced. And it keeps streamlined with deprecated options like ``statusfile (deprecated, optional)``

ejoerns: But in our case, no option is deprecated. The values are.

include/bootchooser.h
dwntri: That's exactly what I've mentioned in the discussion already. 
And that's the problem with the name of the flag ``lock-counter`` overall. It has a double meaning. Maybe we should aim for my suggestion ``freeze-counter`` ? :thinking:

ejoerns: I don't see how this changes the situation. Then we have a counter that counts freezes rather than locks.

dwntri: Okay, phrased differently then, what would change if we rename this method?
If we swap "counter" and "lock" you do not know, to which variable it belongs to, and you'd have to look that up and probably what it exactly is for. 
In the current state, you can map it to the variable it belongs to easier, but you'd have to look up what it exactly does

ejoerns: > If we swap "counter" and "lock" you do not know, to which variable it belongs to, and you'd have to look that up and probably what it exactly is for.

I can't see how this belongs to any variable. Could you point out which variable you refer to?

src/bootloaders/barebox.c
ejoerns: FTR: We discussed this and agreed on adding support for `system_bb_statename` and leaving the rest for possible later refactoring

docs/reference.rst
dwntri: Okay, changed it.

src/mark.c
```diff
@@ -141,6 +143,13 @@ gboolean r_mark_active(RaucSlot *slot, GError **error)
 	r_slot_status_load(slot);
 	slot_state = slot->status;
 
+	if (r_context()->config->prevent_late_fallback == R_CONFIG_FALLBACK_LOCK_COUNTER) {
```
a3f, line 150: Please add a comment explaining why locking removal happens before set_primary here.

We can set multiple state variables at once atomically, but we don't. This might very well not be a problem, but it's not immediately clear why. A comment should help explaining why a powercut between removal of lock and set_primar below doesn't harm correctness.

src/mark.c
```diff
@@ -177,6 +185,13 @@ gboolean r_mark_good(RaucSlot *slot, GError **error)
 		return FALSE;
 	}
 
+	if (r_context()->config->prevent_late_fallback == R_CONFIG_FALLBACK_LOCK_COUNTER) {
```
a3f, line 202: Likewise, a comment is needed here, why this order was chosen.

a3f (changes requested):

src/mark.c
```diff
@@ -141,6 +143,17 @@ gboolean r_mark_active(RaucSlot *slot, GError **error)
 	r_slot_status_load(slot);
 	slot_state = slot->status;
 
+	/* It is more sensible to unlock the counter, before marking it primary. Reversing the order would create
+	 * a possible vulnerability where a power cut could leave the system configured to boot from a new,
```
a3f:
```suggestion
	 * an unintended corner case where a power cut could leave the system configured to boot from a new,
```

Vulnerability is something with a security impact IMO.

src/mark.c
```diff
@@ -170,13 +182,23 @@ gboolean r_mark_good(RaucSlot *slot, GError **error)
 	g_return_val_if_fail(slot, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
+	/* It is more sensible to mark the slot as good before locking the counter. As in case of a power fail between
+	 * marking and locking, we'd have a system with a working fallback to the old stot, so in case of problems
```
a3f:
```suggestion
	 * marking and locking, we'd have a system with a working fallback to the old slot, so in case of problems
```

I suggest pointing out that on next boot, the mark good still has a chance to lock, so it's eventually consistent of sorts. Also I think comment style used differs from the usual RAUC convention. Please align.

src/bootloaders/barebox.c
ejoerns: FTR: Since calling `r_context()` currently doesn't work if this method is used inside context setup for sanity checking, we agreed on adding an explicit comment and error for the so-far-unsupported use case of combining custom state name or path with boot counter locking

src/config_file.c
ejoerns: @zentax-dev Thanks for adding the comment.

However, the suggested depreciation warning has not been implemented yet, correct?

include/bootchooser.h
ejoerns: FTR: We have clarified this in chat and it's implemented as proposed now (as `r_boot_set_counters_lock()`)

docs/reference.rst
ejoerns: This paragraph/hint seems to have been lost with the current version of the commits. Is that intentional?

src/config_file.c
ejoerns: The line is removed, but error handling for `g_key_file_get_string()` is still missing. `ierror` would be overridden by `parse_late_fallback_mode()`

src/mark.c
ejoerns: I guess this was changed and discussed in the latest version. Will mark it as 'resolved', though.

docs/reference.rst
dwntri: nope wasn't intentional

src/config_file.c
dwntri: Done. used a `g_info` for now. Tests fail when I use a `g_warning`.
And I think, an info is enough as a starter. Once a decision is made when the deprecated old parameters will be deleted, this should be changed into a warning.

docs/integration.rst
```diff
@@ -877,6 +877,31 @@ boot target is required:
...
+  };
+
+This makes barebox aware of the counter locking state due to the
+the ``remaining_attempts`` counter.
+Boot counter locking is activated when 'rauc status mark-good' is executed
+implicitly during a RAUC update or explicitly through manual commands.
+The current value of the ``attempts_locked`` state variable can be read via the commands
+``barebox-state -d`` and ``rauc status``.
+It is not yet available via the D-Bus interface and thus cannot be displayed
+when calling `rauc status` with the service enabled.
```
ejoerns:
```suggestion
when calling ``rauc status`` with the service enabled.
```

src/mark.c
```diff
@@ -170,13 +182,24 @@ gboolean r_mark_good(RaucSlot *slot, GError **error)
 	g_return_val_if_fail(slot, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
+	/* It is more sensible to mark the slot as good before locking the counter. As in case of a power fail between
+	 * marking and locking, the boot lock counter would be set during the next boot.
```
a3f:
```suggestion
	 * marking and locking, the boot lock counter could be set by a mark-good on a later boot.
```

src/mark.c
```diff
@@ -170,13 +182,24 @@ gboolean r_mark_good(RaucSlot *slot, GError **error)
...
+	 * In case of a problem during the next boot we'd have a system with a working fallback to the old slot, so
+	 * the system would still all back to the other slot if the current one fails, as no slot is locked yet. */
```
a3f:
```suggestion
	 * the system would still fall back to the other slot if the current one fails, as the bootcounter has not been locked yet. */
```

a3f (changes requested):

a3f (commented):
I have no further comments on the series and haven't looked deep enough to put a checkmark on it.

zentax-dev:
> I have no further comments on the series and haven't looked deep enough to put a checkmark on it.

I'll mark your remarks as resolved then

zentax-dev:
@ejoerns would be nice if you could review this again

js731ca:
gave this PR + the corresponding barebox changes a spin - nice work!
sth i stumbled into: rauc fails when configured with `prevent-late-fallback=lock-counter`, but without the barebox-state being available for $reasons
-> could this be handled more gracefully by maybe falling back to 'disabled'?

js731ca:
is this a minor bug/regression? - the 'rauc status' always reports `Counter lock: disabled`, or what is the meaning of that status?
<details>

```
$> barebox-state -d
bootstate.systemA.remaining_attempts=3
bootstate.systemA.priority=10
bootstate.systemB.remaining_attempts=3
bootstate.systemB.priority=20
bootstate.last_chosen=2
bootstate.attempts_locked=1
```
```
$> rauc status
=== System Info ===
Compatible:  Test System
Variant:     abc,xyz-rev-c
Booted from: fit.1 (systemB)
Counter lock: disabled

=== Bootloader ===
Activated: fit.1 (systemB)
```
</details>

src/main.c
```diff
@@ -1540,7 +1541,8 @@ static gchar* r_status_formatter_readable(RaucStatusPrint *status)
 	g_string_append(text, "=== System Info ===\n");
 	g_string_append_printf(text, "Compatible:  %s\n", status->compatible);
 	g_string_append_printf(text, "Variant:     %s\n", status->variant);
-	g_string_append_printf(text, "Booted from: %s (%s)\n\n", bootedfrom ? bootedfrom->name : NULL, status->bootslot);
+	g_string_append_printf(text, "Booted from: %s (%s)\n", bootedfrom ? bootedfrom->name : NULL, status->bootslot);
+	g_string_append_printf(text, "Counter lock: %s\n\n", status->lock_counter ? "enabled" : "disabled");
```
js731ca:
```suggestion
//	g_string_append_printf(text, "Counter lock: %s\n\n", status->lock_counter ? "enabled" : "disabled");
```
a `rauc status` depends on `retrieve_status_via_dbus` - which isn't implemented yet
so the status would always say "disabled" - which is misleading
(and mis-lead me to debug into the code... sth we could save others from ;-)

js731ca:
> But I think you mean the case where you have A/B boot slots, but only a single root slot?

when everything is A/B (bootloader, fitimage/kernel, rootfs); during an update rauc only installs a new bootloader and does an atomic operation (to switch the emmc boot-part select bit). Note that neither rauc nor barebox flip that switch back when one slot depletes attempts.
The "problem"/cornercase we would run into if the boot-slot is switched back externally.

js731ca:
> @ejoerns would be nice if you could review this again

and @a3f : gentle(?) ping :-)
the barebox parts where merged with 2025.06.01 - but since both parts are involved: the more reviewers the better

zentax-dev:
> > @ejoerns would be nice if you could review this again
> 
> and @a3f : gentle(?) ping :-) the barebox parts where merged into master - but since both parts are involved: the more reviewers the better

Just a quick heads up: I've briefly talked to @ejoerns about this and he will look into it.
As far as I know @a3f was already alright with everything, but I'll see him the next days, so I'll have a chat with him :)

src/main.c
ejoerns: I agree that it's misleading and should not be printed. As it works for the non-service case, however, using

``` diff
-       g_string_append_printf(text, "Counter lock: %s\n\n", status->lock_counter ? "enabled" : "disabled");
+       if (!ENABLE_SERVICE) {
+               g_string_append_printf(text, "Counter lock: %s\n\n", status->lock_counter ? "enabled" : "disabled");
+       }
```

should work until the D-Bus part is implemented.

ejoerns:
> > But I think you mean the case where you have A/B boot slots, but only a single root slot?
> 
> when everything is A/B (bootloader, fitimage/kernel, rootfs); during an update rauc only installs a new bootloader and does an atomic operation (to switch the emmc boot-part select bit). Note that neither rauc nor barebox flip that switch back when one slot depletes attempts. The "problem"/cornercase we would run into if the boot-slot is switched back externally - and happens to be out-of-sync with the proper grouping

Is that an actual use case or a theoretical consideration?

For most cases, the update of the bootloader is considered atomic, but without any fallback.

A proper fallback would require an entity before the bootloader that can decide that a fallback is necessary.

A design that actually expects the bootloader to be part of the A/B setup would also need to ensure that the bootloader boots the corresponding slot. Thus, if there is actually a fallback, the system with the older RAUC version should be booted instead. This sounds like what #1750 targets. For good reasons, the boot partition switching is not performed by RAUC here (but by the bootloader itself instead).

ejoerns:
> > gave this PR + the [corresponding barebox changes](https://lore.kernel.org/barebox/20250619-bootchooser-lock-v3-0-fc4ad92c3866@pengutronix.de/) a spin - nice work! sth i stumbled into: rauc fails when configured with `prevent-late-fallback=lock-counter`, but without the barebox-state being available for $reasons
> > ```
> > rauc[839]: Using central status file /mnt/data/rauc/central.raucs
> > rauc[839]: Using system config file /etc/rauc/system.conf
> > rauc[845]: no such variable: bootstate.attempts_locked
> > rauc[839]: Failed to initialize context: No content to read
> > systemd[1]: rauc.service: Failed with result 'exit-code'.
> > systemd[1]: Failed to start RAUC Update Service.
> > ```
> >   
> > -> could this be handled more gracefully by maybe falling back to 'disabled'?
> 
> The intention behind this is to make it clear, that the system is misconfigured and not working as intended. If rauc would fall back to disabled, the service would be running, and the user probably wouldn't notice it

At least the error checking and reporting could be improved here, since the message actually does not give a good indication about what's going wrong. The part that could be addressed in this PR is a change in `src/context.c` to add something like

``` diff
                        if (!r_barebox_get_lock_counter(&locked, &ierror)) {
-                               g_propagate_error(error, ierror);
+                               g_propagate_prefixed_error(error, ierror, "Failed to read barebox lock counter: ");
                                return FALSE;
```

The other part is that output parsing and error code checking for barebox-state happen in an unoptimal order currently. But that's the case for existing methods, too.

ejoerns:
> The other part is that output parsing and error code checking for barebox-state happen in an unoptimal order currently. But that's the case for existing methods, too.

For this, I've created https://github.com/rauc/rauc/pull/1781 now.

src/config_file.c
```diff
@@ -143,6 +143,28 @@ gboolean parse_bundle_formats(guint *mask, const gchar *config, GError **error)
 	return res;
 }
 
+static gboolean parse_late_fallback_mode(const gchar *mode_str, RConfigLateFallback *mode, GError **error)
+{
+	g_return_val_if_fail(mode != NULL, FALSE);
```
ejoerns, line 136:
```suggestion
	g_return_val_if_fail(mode != NULL, FALSE);
	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```

src/config_file.c
```diff
@@ -143,6 +143,28 @@ gboolean parse_bundle_formats(guint *mask, const gchar *config, GError **error)
...
+
+	if (!mode_str || g_strcmp0(mode_str, "disable") == 0) {
```
ejoerns: Can `mode_str` be NULL? From the calling code, it should at least be `''`.

src/config_file.c
```diff
@@ -143,6 +143,28 @@ gboolean parse_bundle_formats(guint *mask, const gchar *config, GError **error)
 	return res;
 }
 
+static gboolean parse_late_fallback_mode(const gchar *mode_str, RConfigLateFallback *mode, GError **error)
```
ejoerns, line 134: In the corresponding commit message:
Better use 'lock-counter' in the title (than just 'lock').

Despite the text suggesting that locking is global, the following text refers to "counter of the slot", which seems confusing.

Since the commit also does not introduce the feature, it might make sense to better point out that only the parsing of the option is implemented.

src/bootloaders/barebox.c
```diff
@@ -120,7 +120,14 @@ static gboolean barebox_state_get(const gchar *bootname, BareboxSlotState *bb_st
...
 
-/* names: list of gchar, values: list of gint */
+/**
+ * @brief Sets barebox state variables based on a key->value list.
+ *
+ * @param pairs GPtrArray of "var=value" strings to set as barebox state variables.
+ * @param error	return location for a GError, or NULL
+
+ * @return TRUE in case of success, FALSE in case of failure
+ */
```
ejoerns: This is a good improvement/fix, but unrelated to 'barebox: add counter locking to barebox backend'.
Thus, please move it into a separate commit or PR.

src/bootloaders/barebox.c
```diff
@@ -199,6 +206,105 @@ gboolean r_barebox_set_state(RaucSlot *slot, gboolean good, GError **error)
 	return TRUE;
 }
 
+/* This freezes the remaining attempts counter in barebox by setting the attempts_locked variable */
```
ejoerns:
```suggestion
/* This freezes the remaining attempts counters in barebox by setting the attempts_locked variable */
```

src/bootloaders/barebox.c
```diff
@@ -199,6 +206,105 @@ gboolean r_barebox_set_state(RaucSlot *slot, gboolean good, GError **error)
...
+gboolean r_barebox_get_lock_counter(gboolean *locked, GError **error)
+{
+	GError *ierror = NULL;
+	g_autoptr(GSubprocess) sub = NULL;
+	GInputStream *instream = NULL;
+	g_autoptr(GDataInputStream) datainstream = NULL;
+	g_autofree gchar *outline = NULL;
+	gchar *endptr = NULL;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
ejoerns, line 249:
```suggestion
    g_return_val_if_fail(locked != NULL, FALSE);
	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```

src/bootchooser.c
```diff
@@ -6,6 +6,7 @@
 #include "bootloaders/uboot.h"
 #include "config_file.h"
 #include "context.h"
+#include "glib.h"
```
ejoerns: This should be included indirectly via `bootchooser.h`. And if actually required, it should be part of a separate system include section as (`#include <glib.h>`).

include/bootchooser.h
```diff
@@ -55,6 +55,28 @@ G_GNUC_WARN_UNUSED_RESULT;
 gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
 G_GNUC_WARN_UNUSED_RESULT;
 
+/**
+ * Enable or disable locking of the boot counter.
+ * This variable is not set per slot, but is valid for all slots.
```
ejoerns, line 60: 'valid for all slots' sounds a bit misleading here and 'this variable' is not defined.
Also, the bootloader does not have a concept of slots, thus maybe:

```suggestion
 * Lock or unlock boot counting in the bootloader.
 * For supported bootloaders, it sets a flag that instructs the bootloader
 * to globally enable/disable its boot counting logic.
```

include/bootchooser.h
```diff
@@ -55,6 +55,28 @@ G_GNUC_WARN_UNUSED_RESULT;
...
+/**
+ * Check current status of boot counter locking in the bootloader.
+ * This variable is not set per slot, but is valid for all slots.
+ *
+ * @param locked return location for value of lock-counter
+ * @param error return location for a GError, or NULL
+ *
+ * @return TRUE if successful, FALSE if failed
+ */
+gint r_boot_get_counters_lock(gboolean *locked, GError **error);
```
ejoerns:
```suggestion
gboolean r_boot_get_counters_lock(gboolean *locked, GError **error);
```

include/bootchooser.h
```diff
@@ -55,6 +55,28 @@ G_GNUC_WARN_UNUSED_RESULT;
...
+ *
+ * @return TRUE if successful, FALSE if failed
+ */
+gboolean r_boot_set_counters_lock(gboolean locked, GError **error);
+
+/**
+ * Check current status of boot counter locking in the bootloader.
+ * This variable is not set per slot, but is valid for all slots.
+ *
+ * @param locked return location for value of lock-counter
```
ejoerns: What's `lock-counter` here? So far this is only the name of the 'prevent-late-fallback' strategy.

include/bootchooser.h
```diff
@@ -55,6 +55,28 @@ G_GNUC_WARN_UNUSED_RESULT;
...
+ * @param locked TRUE to lock counting, FALSE to unlock it
+ * @param error return location for a GError, or NULL
+ *
+ * @return TRUE if successful, FALSE if failed
+ */
+gboolean r_boot_set_counters_lock(gboolean locked, GError **error);
+
+/**
+ * Check current status of boot counter locking in the bootloader.
+ * This variable is not set per slot, but is valid for all slots.
```
ejoerns: Maybe just drop this line.

src/bootchooser.c
```diff
@@ -155,6 +156,61 @@ gboolean r_boot_set_state(RaucSlot *slot, gboolean good, GError **error)
...
+
+	return res;
+}
+
+gboolean r_boot_get_counters_lock(gboolean *locked, GError **error)
+{
+	GError *ierror = NULL;
+	gboolean res = FALSE;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
ejoerns, line 191:
```suggestion
    g_return_val_if_fail(locked != NULL, FALSE);
	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```

src/mark.c
```diff
@@ -1,6 +1,8 @@
 #include "bootchooser.h"
+#include "config_file.h"
 #include "event_log.h"
 #include "context.h"
+#include "glib.h"
```
ejoerns, line 5: See above.

src/mark.c
```diff
@@ -141,6 +143,17 @@ gboolean r_mark_active(RaucSlot *slot, GError **error)
 	r_slot_status_load(slot);
 	slot_state = slot->status;
 
+	/* It is more sensible to unlock the counter, before marking it primary. Reversing the order would create
```
ejoerns:
```suggestion
	/* It is more sensible to unlock the counter before marking it primary. Reversing the order would create
```

For users not aware of the previous implementation attempt, the 'It is more sensible' sounds a bit unexpected. We also don't mark the counter primary.
So maybe just start with `Unlock boot counting before marking the slot primary...`.

src/mark.c
```diff
@@ -141,6 +143,17 @@ gboolean r_mark_active(RaucSlot *slot, GError **error)
...
+	 * an unintended corner case where a power cut could leave the system configured to boot from a new,
+	 * untested slot without the ability to fall back if that slot fails to boot properly, as it would
+	 * still be locked. So we prepary the safety mechanisms first and then do the potentially risky operation. */
```
ejoerns:
```suggestion
	 * still be locked. So we prepare the safety mechanisms first and then do the potentially risky operation. */
```

'safe' and 'risky' sounds a bit inappropriate for the individual operations.

src/mark.c
```diff
@@ -170,13 +182,24 @@ gboolean r_mark_good(RaucSlot *slot, GError **error)
 	g_return_val_if_fail(slot, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
+	/* It is more sensible to mark the slot as good before locking the counter. As in case of a power fail between
```
ejoerns: Above, this section is attached to the `R_CONFIG_FALLBACK_LOCK_COUNTER` check, which also seems more appropriate here since for 90% of the cases, the `r_boot_set_state()` will not have anything to do with slot locking.

src/mark.c
```diff
@@ -230,7 +253,7 @@ gboolean mark_run(const gchar *state,
 			return FALSE;
 		}
 
-		if (r_context()->config->prevent_late_fallback
+		if (r_context()->config->prevent_late_fallback == R_CONFIG_FALLBACK_MARK_BAD
```
ejoerns, line 269: This should be part of the commit changing the type of `prevent_late_fallback`.

src/main.c
```diff
@@ -1540,7 +1541,8 @@ static gchar* r_status_formatter_readable(RaucStatusPrint *status)
 	g_string_append(text, "=== System Info ===\n");
 	g_string_append_printf(text, "Compatible:  %s\n", status->compatible);
 	g_string_append_printf(text, "Variant:     %s\n", status->variant);
-	g_string_append_printf(text, "Booted from: %s (%s)\n\n", bootedfrom ? bootedfrom->name : NULL, status->bootslot);
+	g_string_append_printf(text, "Booted from: %s (%s)\n", bootedfrom ? bootedfrom->name : NULL, status->bootslot);
```
ejoerns, line 1604: In the commit message:

```
    src/main: add prevent-late-fallback to userspace tools
```

Since RAUC only runs in 'userspace', I'd leave that term out. The commit also doesn't add 'prevent-late-fallback', it just adds output for counter locking. Essentially, you `add counter locking information to rauc status output`.

```
    This adds the option to print the current state of prevent-late-fallback
    and also to enable and disable sets the lock-counter via userspace.
```

The sentence looks inconsistent. Also, the corresponding commit does not add any functionallity to *enable/disable/set* something.

```
    The possibility to get/set it via D-Bus is not yet
    implemented and will be added separately.
```

src/main.c
ejoerns: And of course, this should only be printed when counter locking is enabled!

test/bootchooser.c
```diff
@@ -274,7 +284,7 @@ static void bootchooser_barebox_conf_attempts(BootchooserFixture *fixture,
 	RaucSlot *rootfs0 = NULL;
 	GError *error = NULL;
 	gboolean res;
-
+	// TODO[lsc]: add test here
```
ejoerns: This is probably still open?

ejoerns (changes requested):

src/bootloaders/barebox.c
dwntri: Technically there are counters for each slot, yes, but it just freezes the currently active one, as only that is getting decremented.
So, a more precise formulation could be 
```suggestion
/* This freezes the active remaining attempts counter in barebox by setting the attempts_locked variable */
```

dwntri: or maybe even 
```suggestion
/* This freezes the remaining attempts counter of the currently activated slot in barebox by setting the attempts_locked variable */
```

ejoerns: My understanding of the concept was that it is a global lock that simply deactivates the bootloader *boot counting* in general. Writing "of the currently activated slot" would give me the impression that other slots' attempt counters can still be modified by the bootloader.
Maybe that's a matter of what we understand as 'currently activated'. 'currently activated' when performing the freeze/lock, or 'currently activated' from within the bootloader's individual boot.

src/main.c
dwntri: `r_status_formatter_readable` is called via `print_status` -> `status_start`
And there `print_status` is just called if `!ENABLE_SERVICE`.
So I thought this is not necessary here.
Or is there another way to call it, which I didn't see?

src/mark.c
dwntri: you're right, and also the information probably is also a bit redundant when taking the previous sentence into account.
```suggestion
	 * still be locked. So it is better to prepare the fallback mechanisms first. */
```

include/bootchooser.h
dwntri: hmhm, to keep it generic and not just to match barebox: it is a related bootloader value
so maybe something like
```suggestions
 * @param locked return location for the related value in the bootloader
```
or maybe use `current` instead of related

src/main.c
ejoerns: Running `rauc status` needs to be possible for both the 'service' and the 'non-service' case.
In the code, `print_status` is called independently of `ENABLE_SERVICE`!
The `status_print` information is filled by either `retrieve_status_via_dbus()` (service case) or by filling in the information based on config and status (non-service case).

include/bootchooser.h
ejoerns: Not sure if we always have a 1:1 mapping to the gboolean, but better, yes.

An alternative would be something like.
```suggestion
 * @param locked return location for a gboolean. Value is set to TRUE if boot counting is locked or FALSE otherwise.
```

src/bootloaders/barebox.c
dwntri: Yeah, probably I've got too much of a whitebox point of view, knowing what barebox is doing in this case.

a3f:
We can't necessarily expect that an older bundle is never installed.
FOTA is not the only one way to do an update, USB recovery (with a bundle that often lags behind the online version) is a possibility in my project. The problematic sequence could look as follows:

- FOTA installation of new RAUC bundle with this feature here
- System is partially broken: boot-complete is reached, but it's not possible to do further FOTA
- USB recovery stick is inserted, update to old system: booth rootfs and bootloader are at old state now
- Recovered system fails to start up
- As old system doesn't have support for switching eMMC boot partition on attempt depletion, we end up booting the new rootfs with the old bootloader
- RAUC refuses to startup: online USB recovery no longer possible

With some more thought put into this contrived example, one might be able to arrive at a simpler sequence of events.

With a recovery mode implemented directly into the bootloader we side step this particular issue, but it doesn't sit well with me that there are sequences that have the potential for RAUC to fail to startup not because of prior misconfiguration, but because of the bootloader version in the field. I'd rather have RAUC warn very loudly about this, but then do some fallback behavior, either set the attempts to a high number or just ignore the locking.

Interested in your thoughts.

src/context.c
```diff
@@ -404,6 +405,30 @@ static gboolean r_context_configure_target(GError **error)
...
+			if (context->config->system_bb_statename || context->config->system_bb_dtbpath) {
+				g_set_error_literal(error, R_BOOTCHOOSER_ERROR,
+						R_BOOTCHOOSER_ERROR_NOT_SUPPORTED,
+						"Providing custom name, state or path not yet supported for locking");
+				return FALSE;
+			}
+			gboolean locked = FALSE;
+			if (!r_barebox_get_lock_counter(&locked, &ierror)) {
+				g_propagate_prefixed_error(error, ierror, "Failed to read barebox lock counter: ");
+				return FALSE;
```
js731ca:
```suggestion
				/* If we would throw an error here, RAUC would fail and it might not be possible to execute updates anymore. */
				g_warning("Failed to read barebox lock counter: ");
				g_clear_error(&ierror);
```
this should also be a "warn very loudly, but proceed" in the case where the state variable is not present

dwntri: Thanks for the hint, I've changed this and added a test for it

PR #1734, opened 2025-06-16
Add progress handling during hash index generation for adaptive mode 'block-hash-index'

ejoerns:
Depending on the CPU and storage speed, generating the hash index for the adaptive mode 'block-hash-index' might take a notable amount of time.

So far, we don't emit any progress information during these steps, which could give the impression that the installation progress stalls.

With this approach, we reserve 10% of the step progress for both the 'target_slot' and the 'active_slot' generation and increment the progress evenly while generating the hash index.

Also, add some testing adaptations and improvements related to this feature.

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -157,6 +157,13 @@ void r_context_end_step(const gchar *name, gboolean success);
 void r_context_set_step_percentage(const gchar *name, gint percentage);
 
 /**
+ * Increases step percentage by one.
+ *
+ * @param name identifying the step. Must be a step with no explicit substeps.
+ */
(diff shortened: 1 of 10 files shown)


jluebbe:
I've force pushed (https://github.com/rauc/rauc/compare/a87e8bd17478..80a8852a573d) to fix the test case when running it separately. The `callback_counter` must be reset for this test case as well.

jluebbe (approved):
I don't like working around the nested step structure by calculating ranges explicitly, but it better than not having process. We'll have to refactor this later.

zentax-dev (approved):

PR #1735, opened 2025-06-16
src/install.c: Add details to the log messages

leon-anavi:
Enhance the log messages during RAUC Bundle installation to provide more clarity on the process:

- For non-bootable, indicate that the slot is being marked as bad by calling r_mark_bad. For example:

```
Jun 16 14:58:12 raspberrypi5 rauc[925]: Marking target slot rootfs.1 as non-bootable (bad)...
Jun 16 14:58:12 raspberrypi5 rauc[925]: Marked slot rootfs.1 as bad
```

- For bootable, clarify that r_mark_active is being executed to mark the slot as active and to make it primary. For example:

```
Jun 16 14:58:49 raspberrypi5 rauc[925]: Marking target slot rootfs.1 as bootable (active and primary)...
Jun 16 14:58:49 raspberrypi5 rauc[925]: installing http://XXX.XXX.X.XXX:8000/update-bundle-raspberrypi5.raucb: All slots updated
Jun 16 14:58:49 raspberrypi5 rauc[925]: Marked slot rootfs.1 as active
```

Hopefully, the added details will make the log messages more helpful for debugging. According to me, currently, it is not very clear what "non-bootable" and "bootable" actually signify in the RAUC context.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -1427,10 +1427,10 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 
 	if (boot_mark_slot) {
 		/* Mark boot slot non-bootable */
-		g_message("Marking target slot %s as non-bootable...", boot_mark_slot->name);
+		g_message("Marking target slot %s as non-bootable (bad)...", boot_mark_slot->name);
 		if (!r_mark_bad(boot_mark_slot, &ierror)) {
 			g_set_error(error, R_INSTALL_ERROR, R_INSTALL_ERROR_MARK_NONBOOTABLE,
-					"Failed marking slot %s non-bootable: %s", boot_mark_slot->name, ierror->message);
+					"Failed marking slot %s non-bootable (bad): %s", boot_mark_slot->name, ierror->message);
 			g_clear_error(&ierror);
 			return FALSE;
 		}
@@ -1472,10 +1472,10 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 	if (boot_mark_slot) {
 		if (r_context()->config->activate_installed) {
 			/* Mark boot slot bootable */
-			g_message("Marking target slot %s as bootable...", boot_mark_slot->name);
+			g_message("Marking target slot %s as bootable (active/primary)...", boot_mark_slot->name);
 			if (!r_mark_active(boot_mark_slot, &ierror)) {
 				g_propagate_prefixed_error(error, ierror,
-						"Failed marking slot %s bootable: ", boot_mark_slot->name);
+						"Failed marking slot %s bootable (active/primary): ", boot_mark_slot->name);
 				return FALSE;
 			}
 		} else {


src/install.c
```diff
@@ -1472,10 +1472,10 @@ static gboolean launch_and_wait_default_handler(RaucInstallArgs *args, gchar* bu
 	if (boot_mark_slot) {
 		if (r_context()->config->activate_installed) {
 			/* Mark boot slot bootable */
-			g_message("Marking target slot %s as bootable...", boot_mark_slot->name);
+			g_message("Marking target slot %s as bootable (active and primary)...", boot_mark_slot->name);
```
ejoerns:
```suggestion
			g_message("Marking target slot %s as bootable (active)...", boot_mark_slot->name);
```

This suggests 'active' and 'primary' are two different properties, but actually, the 'mark active' operation sets the slot as 'primary'. Thus maybe first of all stick to the term that's used for the 'mark' operation.

ejoerns (commented):
As you've shown. the follow-up log messages already indicate the actual action that's performed.

But maybe having this hint in the install.c's message helps a little to make the caller clearer.

src/install.c
leon-anavi: This additional word "primary" in the log is particularly helpful when debugging a [custom bootloader handler](https://rauc.readthedocs.io/en/latest/integration.html#custom-bootloader-backend-interface), as the handler is required to implement `set-primary`, but not a specific method for marking a slot as active or bootable.

In this case, after RAUC completes the installation, the slot is not explicitly marked as good. Instead, the `set-primary` implementation must ensure that the remaining boot attempts for the slot are reset to a value greater than zero. Having `primary` in the log message will lead to association with `set-primary`.

jluebbe:
The [custom bootloader backend interface docs](https://rauc.readthedocs.io/en/latest/integration.html#custom-bootloader-backend-interface) should also be updated to mention that `set-primary` needs to implicitly also set the state to "good" or an equivalent.

src/install.c
ejoerns: @leon-anavi Thank you for providing some background on this!
I'd still prefer to leave the boot selection implementation details to the bootloader layer and instead better document the expectations behind 'set-primary' as @jluebbe suggested.

An alternative could be to have some sort of debug messages in the boot selection methods (`r_boot_set_primary()`, `r_boot_get_state()`, etc.) that can be enabled with `-d` option and will log which actual boot selection method was called.

What do you think?

leon-anavi: Yes, absolutely, improving the documentation with more information is a great idea.

However, in scenarios where a custom bootloader handler is needed, it is up to the integrating engineer to implement the required boot selection methods (`set-primary`, `get-primary`, etc.). Because of this, I kind of prefer more generic logs that are unified across all supported bootloaders with these minor but important details that aid in debugging.

I am not familiar with all covered scenarios so may be I am missing something. Why having both 'active' and 'primary' in the log bothers you?

leon-anavi:
Hi @ejoerns, @jluebbe,

Sorry for bothering you, but I am not sure how to proceed with this pull request. Is a change required?

Thanks,
Leon

src/install.c
ejoerns: > However, in scenarios where a custom bootloader handler is needed, it is up to the integrating engineer to implement the required boot selection methods (set-primary, get-primary, etc.). Because of this, I kind of prefer more generic logs that are unified across all supported bootloaders with these minor but important details that aid in debugging.

A debug log from the bootloader layer would also aid in debugging, wouldn't it?
I also don't see why this option wouldn't be generic to all supported bootloaders.

The actual difference between 'active' and 'primary' is that 'active' is the wording for the 'mark' layer, while 'primary' is the wording for the 'bootloader' layer.  The idea was that the mark layer is the higher-level interface and that the installation only deals with that. In the current implementation, the `r_mark_active()` calls `r_boot_set_primary()`, but also sets slot status information, like the activation timestamp or the activation counter. More could be added later.

Long story short, for now I'd be happy with each alternative that doesn't say 'and' since these are no independent operations, but one implies the other. Simplest solution:


```suggestion
			g_message("Marking target slot %s as bootable (active/primary)...", boot_mark_slot->name);
```

Alternative:

```suggestion
			g_message("Marking target slot %s as active (primary)...", boot_mark_slot->name);
```

Maybe the first one has less impact and is closer to what you suggested.

leon-anavi: @ejoerns thank you for the provided suggestions. Based on them I updated the patch and force pushed the new version.

ejoerns (approved):
@leon-anavi Thank you for updating the PR and for the aim to improve RAUC's usability! For me, it would be fine now.

jluebbe (approved):

PR #1736, opened 2025-06-17
Error handling and usability improvements around (eMMC) bootloader updates

ejoerns:
* Adds missing bootloader update option bits to reference.rst
* size check for eMMC bootloader updates
* better error handling or `boot-emmc` device misconfiguration
* more clear error printout for `write-slot`

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -545,7 +545,9 @@ hierarchical separator.
 
 ``type=<type>`` (optional, recommended)
   The type describing the slot. Currently supported ``<type>`` values are ``raw``,
-  ``nand``, ``nor``, ``ubivol``, ``ubifs``, ``ext4``, ``vfat``.
+  ``nand``, ``nor``, ``ubivol``, ``ubifs``, ``ext4``, ``vfat`` for normal slots
+  and ``boot-emmc``, ``boot-mbr-switch``, ``boot-gpt-switch``, and ``boot-raw-fallback``
+  for atomically updatable bootloader slots.
   See table :ref:`sec-slot-type` for a more detailed list of these different types.
   Defaults to ``raw`` if none given.
 
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -604,6 +604,18 @@ hierarchical separator.
   Allows to specify custom mount options that will be passed to the slot's
   ``mount`` call as ``-o`` argument value.
 
+``region-start`` (mandatory for specific types, invalid for others)
+  Defines the start of the disk region used for atomic bootloader updates.
+  Valid for slot types ``boot-mbr-switch``, ``boot-gpt-switch``,
+  ``boot-raw-fallback`` only!
+  See :ref:`sec-mbr-partition` and the following for more details.
+
+``region-size`` (mandatory for specific types, invalid for others)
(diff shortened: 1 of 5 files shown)


docs/reference.rst
```diff
@@ -602,6 +604,18 @@ hierarchical separator.
   Allows to specify custom mount options that will be passed to the slot's
   ``mount`` call as ``-o`` argument value.
 
+``region-start`` (type-specific, mandatory)
```
```suggestion
``region-start`` (mandatory for specific types, invalid for others)
```

docs/reference.rst
```diff
@@ -602,6 +604,18 @@ hierarchical separator.
...
+  Defines the start of the disk region used for atomic bootloader updates.
+  Valid for slot types ``boot-mbr-switch``, ``boot-gpt-switch``,
+  ``boot-raw-fallback`` only!
+  See :ref:`sec-mbr-partition` and the following or more details.
```
```suggestion
  See :ref:`sec-mbr-partition` and the following for more details.
```

docs/reference.rst
```diff
@@ -602,6 +604,18 @@ hierarchical separator.
...
+
+``region-size`` (type-specific, mandatory)
+  Defines the size of the disk region used for atomic bootloader updates.
+  Valid for slot types ``boot-mbr-switch``, ``boot-gpt-switch``,
+  ``boot-raw-fallback`` only!
+  See :ref:`sec-mbr-partition` and the following or more details.
```
```suggestion
  See :ref:`sec-mbr-partition` and the following for more details.
```

src/main.c
```diff
@@ -522,7 +522,7 @@ static gboolean write_slot_start(int argc, char **argv)
 
 	/* call update handler */
 	if (!update_handler(image, slot, NULL, &ierror)) {
-		g_printerr("%s\n", ierror->message);
+		g_printerr("Failed writing slot: %s\n", ierror->message);
```
jluebbe:
```suggestion
		g_printerr("Failed to write slot: %s\n", ierror->message);
```

src/update_handler.c
```diff
@@ -2172,14 +2172,22 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
-		goto out;
+		return FALSE;
+	}
+
+	if (g_str_has_suffix(realdev, "boot0") || g_str_has_suffix(realdev, "boot1")) {
+		g_set_error(error,
+				R_UPDATE_ERROR,
+				R_UPDATE_ERROR_FAILED,
+				"Device path '%s' refers to boot partition, but base device is expected", realdev);
+		return FALSE;
```
jluebbe: We should check this during config loading. See for example `R_CONFIG_ERROR_INVALID_DEVICE` as used for `jffs2` slots.

jluebbe (changes requested):

src/update_handler.c
ejoerns: That was my first impulse, too. But I was unsure how much of the handler-specific constraints and logic we should put into the config parsing. But maybe it's more a question of the slot type, anyway. I'll move the check.

jluebbe: At some point we should split it from the parsing itself and perform the checks in a separate step, similar to what we have for the manifest.

Issue #780, opened 2021-10-14
Allow more fine-grained control about clearing eMMC boot slots


Ablu:
It looks like `clear_slot` always clears the entire "device" before writing the new slot. This can be problematic for devices which "abuse" the boot0 partition by storing additional data there.

For example, Toradex stores their hardware configuration block at the end of boot0 and also, by default, places the u-boot environment there.
One can argue that this is not really a nice location for such data, but we found ourselves in the need to keep that data there for backwards-compatibility reasons.

For now I simply patched out the clearing of the slot (even if some old data is left over it does not really matter in our case since hopefully nothing jumps or reads from there :P), but we would be interested into seeing this in mainline.

Would you be interested into accepting a patch which makes the clearing configurable? Or do maybe need some more fine-grained control about the area which should be cleared?

I would try to find time to submit a patch if we can agree on a solution that would be mainline-able :)

UVV-gh:
@Ablu I solved this issue by moving Toradex factory config block during my module bring-up procedure out of eMMC boot partition and placing it before the first GPT partition and after reserved GPT header space.

jluebbe:
> > RAUC has some tradition of trying to make it harder to make design errors, so I'm skeptical. One approach might be to reuse the `region-start/-size` options from https://rauc.readthedocs.io/en/latest/advanced.html#update-boot-partition-in-gpt.
> > This would mean that RAUC always erases this region completely and also writes only to the region. It would need to check that the image is not larger than the `region-size`. Also, it should be documented that this method is only intended for backwards compatibility and not a recommended design.
> 
> Sounds good to me! Let me know if this approach would be fine to implement in a patch.
> 
> > The change would need to place emphasis on keeping the additional complexity for the normal case low.
> 
> If the region-start/-size thing is too complex code wise, I personally would be fine with a simple flag (which is documented with a big warning describing that no clearing happens) to disable the clearing too... If in doubt I can still over-provision the .img file that I am flashing in order to make sure enough stuff is overwritten.

We thought about this again. `region-size` has a different meaning, so we wouldn't want to reuse it. Perhaps `size-limit`?

To be predictable and consistent to the existing behavior (and what would happen without this option), RAUC should still clear the partition up to the limit. The result is that everything below the limit is completely predictable. RAUC should also check that the limit does not exceed the actual device size.

loblik:
[...]

PR #1740, opened 2025-06-19
add new 'size-limit' option for type 'boot-emmc'

ejoerns:
For special cases, where e.g. the manufacturer decided to store data at the end of the boot partition and where this information is required and cannot be migrated, we allow specifying a `size-limit` option for the `boot-emmc` slot type that allows limiting the cleared/written size.

Fixes: #780

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -294,3 +294,23 @@ need to split the payload and CMS data and then use `openssl cms -verify` with
 the `-content` option.
 As this is more involved, we recommend using either `rauc extract` or switching
 to verity bundles.
+
+How can I protect manufacturer data at the end of the boot partition when using ``boot-emmc``?
+----------------------------------------------------------------------------------------------
+
+When using RAUC's boot-emmc slot type for bootloader updates, the entire eMMC
+boot partition is cleared before the new image is written.
+If this partition contains manufacturer-specific data (e.g., calibration data
+at the end), that data will be lost unless special precautions are taken.
+
+The recommended solution is to migrate this data to a safe location during the
+first boot of the device. This ensures future bootloader updates can proceed
+safely.
+
+If migration is not feasible (for example, on already deployed devices) RAUC
+provides a ``size-limit`` :ref:`slot option <_slot.slot-class.idx-section>` for
+the ``boot-emmc`` slot.
+This restricts the writable area to avoid overwriting critical data.
+
+.. warning:: The ``size-limit`` option is intended only for backwards
+   compatibility and should not be used in new designs!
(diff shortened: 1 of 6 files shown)


docs/faq.rst
```diff
@@ -294,3 +294,23 @@ need to split the payload and CMS data and then use `openssl cms -verify` with
 the `-content` option.
 As this is more involved, we recommend using either `rauc extract` or switching
 to verity bundles.
+
+How can I protect manufacturer data at the end of the boot partition when using ``boot-emmc``?
+----------------------------------------------------------------------------------------------
+
+When using RAUC's boot-emmc slot type for bootloader updates, the entire eMMC
+boot partition is overwritten.
```
jluebbe:
```suggestion
boot partition is cleared before the new image is written.
```

src/update_handler.c
```diff
@@ -92,8 +130,14 @@ static gboolean clear_slot(RaucSlot *slot, GError **error)
 				return FALSE;
 			}
 		}
+
+		total_written += write_count;
+		if (slot->size_limit > 0 && total_written >= slot->size_limit)
+			break;
 	}
 
+	g_debug("Cleared %"G_GOFFSET_FORMAT " bytes", total_written);
```
jluebbe: Perhaps add the slot/device name?

src/update_handler.c
```diff
@@ -32,21 +32,52 @@ GQuark r_update_error_quark(void)
...
+ *
+ * Checks if the image size fits into the block device referred to by the
+ * provided file descriptor.
+ * If the size cannot be determined, the image is assumed to fit.
+ *
+ * Additionally, if a slot is given and the slot has a 'size-limit' set, the
+ * image is checked against this limit first.
+ *
+ * @param fd file descriptor of device to check
+ * @param slot slot for optional size-limit check, or NULL
```
jluebbe: We could pass the slot in all cases and avoid the related conditionals, as the size-limit is already restricted to `boot-emmc` during config parsing.

jluebbe (changes requested):
Looks good otherwise, but test coverage would be nice.

src/update_handler.c
ejoerns: Thought about this, too. But decided on this solution with the least impact. Can change it, however.

ejoerns:
v2:

* rebased onto master
* added suggestions
* `slot` parameter of `check_image_size()` is now mandatory and always set
* `g_message()` log level for `clear_slot()` information if `size-limit` is set. This improves visibility and helps testing.
* two `size-limit` test cases in `test_write_slot.py`

src/update_handler.c
ejoerns: Chose the device name, since the slot name isn't set for the eMMC pseudo slot.

PR #1741, opened 2025-06-19
nbd: avoid race with udev during setup

michaelolbrich:
With the NBD_CFLAG_DISCONNECT_ON_CLOSE flag set, the nbd device will be
automatically removed when there are no more users after it was was
opened the first time.

The idea here is, that the nbd device is created, and then immediately
mounted (possibly with dm-verity or dm-crypt in between). Wenn the
bundle ist later unmounted then the nbd device is automatically removed.

Unfortunately this is racy: udev will see the new device and open it
immediately. If it closes the device before it is mounted (or dm-verity
or dm-crypt is configured) then the nbd device is removed immediately
and mounting will fail.

So instead create the nbd device without NBD_CFLAG_DISCONNECT_ON_CLOSE,
open it to keep it busy and then reconfigure the device to set
NBD_CFLAG_DISCONNECT_ON_CLOSE.
The file descriptor is kept until the bundle is mounted.

Note that with this change, the nbd device will not be fully removed if
rauc dies between creating the nbd device and reconfiguring it, or if
reconfiguring fails.

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -142,7 +142,7 @@ out:
 	return nl;
 }
 
-gboolean r_nbd_setup_device(RaucNBDDevice *nbd_dev, GError **error)
+gboolean r_nbd_setup_device(RaucNBDDevice *nbd_dev, int *devicefd, GError **error)
 {
 	GError *ierror = NULL;
 	gboolean res = FALSE;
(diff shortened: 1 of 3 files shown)


src/nbd.c
```diff
@@ -203,6 +200,27 @@ gboolean r_nbd_setup_device(RaucNBDDevice *nbd_dev, GError **error)
 
 	nbd_dev->dev = g_strdup_printf("/dev/nbd%"G_GUINT32_FORMAT, nbd_dev->index);
 
+	*devicefd = open(nbd_dev->dev, O_RDWR|O_CLOEXEC);
```
jluebbe: Perhaps open read-only?

src/nbd.c
```diff
@@ -203,6 +200,27 @@ gboolean r_nbd_setup_device(RaucNBDDevice *nbd_dev, GError **error)
...
+	NLA_PUT_U64(msg, NBD_ATTR_CLIENT_FLAGS,
+			NBD_CFLAG_DISCONNECT_ON_CLOSE
+			);
+
+	nl_socket_modify_cb(nl, NL_CB_VALID, NL_CB_CUSTOM, NULL, NULL);
+	if (nl_send_sync(nl, msg) < 0) {
+		res = FALSE;
+		g_set_error(error, G_IO_ERROR, G_IO_ERROR_FAILED, "netlink send_sync failed");
+		goto out;
+	}
```
jluebbe, line 229: Don't we need to wait/process the response here?

jluebbe (commented):
Does this introduce dependencies on specific kernel versions?

src/nbd.c
jluebbe: `nl_send_sync` already waits for a response/error.

ejoerns (approved):

PR #1743, opened 2025-06-20
meson_options: make use of boolean values

zentax-dev:
Using string literals for bool values is
deprecated, as stated in [1].
So change them to proper boolean values.

[1] https://mesonbuild.com/Release-notes-for-1-1-0.html#coercing-values-in-the-option-function-is-deprecated

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -2,22 +2,22 @@
 option(
   'service',
   type : 'boolean',
-  value : 'true',
+  value : true,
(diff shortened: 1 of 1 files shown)


ejoerns (approved):

jluebbe (approved):

jluebbe:
https://mesonbuild.com/Build-options.html#build-option-types looks like this doesn't need any specific meson version.

Issue #1530, opened 2024-09-26
Support Image Type Option in Manifest


ejoerns:
**Is your feature request related to a problem? Please describe.**

So far, RAUC uses only the image's file name extension (like `.ext4`) to determine the *image type*.
The *image type* however is important for the correct update handler selection in RAUC.

Different tools tend to use different or custom file endings for the images they create.
This led to several requests in the past that aimed to add more matching patterns in https://github.com/rauc/rauc/blob/7f74ea372a3afffb495ff929b7a032bca3f042a2/src/update_handler.c#L2583.

For example, see #1009, #1177, #1201, #1453.

The table is already complex and it's impossible to handle all potential file endings that image generation tools would emit.

**Describe the solution you'd like**

The solution brought up in https://github.com/rauc/rauc/pull/1201#issuecomment-1663668792 would be to add an explicit (new) option to the manifest where one can override the auto-determined file type.
The option could simply be named `type` as this is not yet defined in the context of a manifest image.
It must be optional.

An example configuration would then look as follows:

``` ini
[image.rootfs]
filename=root-file-system.ext4.verity
type=img
```

It should be clarified if all image types currently listed in `updatepairs[]` should be supported or if a filtered/explicit list is preferable.

PR #1744, opened 2025-06-20
Support bin files for eMMC boot partition

rm5248:
Some devices(such as the Allwinner A64 chip) have a uboot that is just a .bin file.  This .bin file gets written to the eMMC boot partition directly, not as a partition.

**What do you use the feature for?**
We use casync to deliver firmware updates to devices.  Without this, the `.bin` file that is generated will not install when using casync.  We use yocto to create the OS.

**How did you verify the feature works?**
Without this, a cacync upgrade will not install correctly.  In addition, the normal rauc upgrade would fail because the `.bin` file that is uboot is not extracted properly from the upgrade file.

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -2253,7 +2253,18 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot, 
 	/* copy */
 	g_message("Copying image to slot device partition %s",
 			part_slot->device);
-	res = copy_raw_image(image, outstream, 0, &ierror);
+	/* Handle casync index file */
+	if (g_str_has_suffix(image->filename, ".caibx")) {
+		g_message("Extracting %s to %s", image->filename, part_slot->device);
+
+		/* Extract caibx to device */
+		if (!casync_extract_image(image, part_slot->device, -1, &ierror)) {
+			res = FALSE;
+		}
+	}else{
+		res = copy_raw_image(image, outstream, 0, &ierror);
+	}
+
 	if (!res) {
 		g_propagate_error(error, ierror);
 		goto out;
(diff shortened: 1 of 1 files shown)


rm5248:
> @rm5248 Thank you for your contribution.
> 
> Please note that this is mainly a duplicate of #1201.
> 
> The suggested generic solution to this should be what's proposed in #1530.

While I don't disagree, it would be nice to have this(or something similar) available in the meantime.  It's unclear to me as to if and when #1530 would be implemented, so this is a stopgap measure.

rm5248:
> Adding the `.bin` match here would lead to the proliferation of matches in RAUC that we want to avoid (see [#1201 (comment)](https://github.com/rauc/rauc/pull/1201#issuecomment-1663668792)). Due to backwards compatibility, we could never remove it again, so I don't see how merging this would be a stopgap measure.

My assumption is that whatever the long-term change that happens in the future would effectively do what this patch is doing, turning this into a stopgap solutiojn.  This may not be a valid assumption though.

> Why do you want to support casync for bootloader images? They are so small that you usually don't save much. You can use `CONVERT_ARGS=--ignore-image=<your bootloader class>` to avoid converting the bootloader image when using the `bundle.bbclass`.

The casync for the bootloader images doesn't work properly at the moment anyway(at least as far as I can tell), it always downloads it.  For our purposes the devices that we are updating may be on a cell connection which is very limited in the amount of data per month that is allocated to each device, so if we can avoid downloading the new bootloader(which is the same as the old bootloader) that would definitely help.

As to why I don't just do `CONVERT_ARGS=--ignore-image=<your bootloader class>`, the answer is that it was not clear to me that this would be needed.  It is unexpected to me that the bootloader image does not support casync, while the other image types do.

rm5248:
> @rm5248 Is this still needed for you? If so, do you have a test setup that you could share where the download differences could be reproduces?

For the download size issue, yes.  I started on something but I haven't had a chance to finish it up.

For this particular issue, it looks like RAUC 1.15 fixes the bin files for eMMC; I'll see about testing and making sure that it is sufficient.

PR #1746, opened 2025-06-26
src/bundle: add g_auto-based helper for temporary files and use it to simplify control flow

jluebbe:


src/utils.c
```diff
@@ -881,6 +881,21 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
 	return rm_tree(tmpdir, error);
 }
 
+void r_tempfile_cleanup(gchar *filename)
+{
+	//g_message("removing %s", filename);
```
ejoerns: Is this intentionally left here?

test/utils.c
```diff
@@ -518,6 +518,37 @@ static void regex_match_test(void)
...
 }
 
+static void tempfile_cleanup_test(void)
+{
+	g_autofree gchar *tmpdir = g_dir_make_tmp("rauc-XXXXXX", NULL);
+
+	g_auto(RTempFile) filename = g_build_filename(tmpdir, "file", NULL);
+	g_assert_false(g_file_test(filename, G_FILE_TEST_EXISTS));
+
+	{
```
ejoerns, line 529: When combining multiple tests into one test function, maybe add a note on each test what it intends to test.

src/utils.c
```diff
@@ -881,6 +881,21 @@ gboolean r_fakeroot_cleanup(const gchar *envpath, GError **error)
...
+
+	if (!filename)
+		return;
+
+	if (g_file_test(filename, G_FILE_TEST_EXISTS)) {
+		if (g_unlink(filename) != 0)
+			g_warning("failed to remove %s", filename);
```
ejoerns, line 891:
```suggestion
			g_warning("failed to remove temporary file %s", filename);
```

src/bundle.c
```diff
@@ -1273,6 +1251,8 @@ gboolean create_bundle(const gchar *bundlename, const gchar *contentdir, GError
 		return FALSE;
 	}
 
+	g_auto(RTempFile) bundletmp = g_strdup(bundlename);
```
ejoerns: Not sure if defining `bundletmp` here isn't too confusing if we continue using `bundlename` for writing..

src/bundle.c
```diff
@@ -1416,7 +1399,7 @@ static gboolean truncate_bundle(const gchar *inpath, const gchar *outpath, goffs
 	}
 
 	res = g_seekable_truncate(G_SEEKABLE(outstream), size, NULL, &ierror);
-	if (!res) {
+	if (!res)
```
ejoerns: That's in the wrong hunk/commit.

src/utils.c
jluebbe: No, it was removed with the force push.

src/utils.c
jluebbe: It could also be the output file, which is only "temporary" from the perspective of the calling function.

ejoerns: Yes, but then the method maybe shouldn't be named 'tempfile', too? It also makes some existing messages less precise. Anyway, the file name might be sufficient to distinguish tempfiles from non-tempfiles.

src/bundle.c
jluebbe: This is just to keep the resource open, so we keep the existing code the same.

ejoerns: It's clear to me what the intention is.
Exactly because it has no other meaning than implicit refcounting, I say this might be confusing. At least when reading this out of this PRs context.
I'd be fine with a comment. (As you already did for the `g_clear_pointer()` calls, where you already seemed to have the impression that this might be too implicit. :smirk: )

jluebbe: I've added more comments (to the declaration and g_clear_pointer calls), please take a look again.

PR #1750, opened 2025-06-30
Link eMMC boot slots with parent slots

zentax-dev:
This feature can be used if it is not possible to avoid incompatibilities between
the bootloader (and it's components) and the operating system
During each update RAUC will ensure that the active boot partition
matches the active unsuccessful slots.
This feature will need to be supported by the bootloader, as RAUC will not write the *ext_csd registers* itself as in the already existing eMMC boot method, but will rely on the bootloader to sort handle it.

Depends On #1752


zentax-dev:
Just rebased onto current mastter

src/emmc.c
```diff
@@ -206,3 +206,39 @@ gboolean r_emmc_force_part_rw(const gchar *device, GError **error)
...
+	gboolean retval = FALSE;
+
+	g_return_val_if_fail(device_path != NULL, FALSE);
+	g_return_val_if_fail(base_device != NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	/* Pattern to find valid MMC/eMMC base device
+	 * Matches: /dev/mmcblk[0-9]+
+	 */
+	regex = g_regex_new("/dev/mmcblk?[0-9]+", 0, 0, error);
```
ejoerns: The `?` should not be required here.
Then it also matches the second part of the comment, which might be dropped then.

(The `error` would have also been overridden. But like in `r_regex_match_simple()`, there is no user input involved in the pattern anyway. Thus it should not fail.).

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
...
     <https://lore.kernel.org/all/20180308140811.6966-1-bst@pengutronix.de/>`_.
 
+.. _sec-emmc-boot-linked:
+
+eMMC Boot Linked
+^^^^^^^^^^^^^^^^
+
+.. note:: This feature does need to be supported by the bootloader. Currently this is the case for `barebox`.
+
+For eMMCs there is an option with fallback support to make the update of
```
ejoerns:
```suggestion
For eMMCs, there is an option with fallback support to make the update of
```

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
...
+the bootloader more consistent.
+This method can be used, if it is not possible to avoid incompatibilities between
```
ejoerns:
```suggestion
This method can be used if it is not possible to avoid incompatibilities between
```

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
...
+the bootloader (and it's components) and the operating system
```
ejoerns:
```suggestion
the bootloader (and its components) and the operating system
```

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
     `important register cache fix
     <https://lore.kernel.org/all/20180308140811.6966-1-bst@pengutronix.de/>`_.
 
+.. _sec-emmc-boot-linked:
+
+eMMC Boot Linked
+^^^^^^^^^^^^^^^^
+
+.. note:: This feature does need to be supported by the bootloader. Currently this is the case for `barebox`.
```
ejoerns:
```suggestion
.. note:: This feature requires support from the bootloader. Currently, this is the case for `barebox`.
```

src/config_file.c
```diff
@@ -1255,6 +1275,71 @@ static GHashTable *parse_artifact_repos(const char *filename, const char *data_d
 	return g_steal_pointer(&repos);
 }
 
+/*
+ * Checks if the emmc-boot-linked slots are configured correctly.
+ * Criteria: - All emmc-boot-linked slots refer to unique boot devices
+ *			 - All emmc-boot-linked slots are on the same device
+ *
+ * @param config RaucConfig object
+ * Returns: TRUE if all criteria pass or no emmc-boot-linked exists, FALSE otherwise
```
ejoerns:
```suggestion
 * @param config RaucConfig object
 * @param error Return location for a GError, or NULL
 *
 * @return TRUE if all criteria pass or no emmc-boot-linked exists, FALSE otherwise
```

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
...
+
+For eMMCs there is an option with fallback support to make the update of
+the bootloader more consistent.
+This method can be used, if it is not possible to avoid incompatibilities between
+the bootloader (and it's components) and the operating system
+During each update RAUC will ensure that the active boot partition
```
ejoerns:
```suggestion
During each update, RAUC will ensure that the active boot partition
```

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
...
+matches the active unsuccessful slots.
```
ejoerns: What does 'unsuccessful' mean here?

zentax-dev:
Got rid of unneeded function call and function parameter. And improved some code comments.
Just left the fixup commit for now to make changes more ovbious. Will squash this before merge

src/update_handler.c
```diff
@@ -2387,40 +2386,137 @@
...
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean img_to_emmc_boot_linked_handler(RaucImage *image, RaucSlot *dest_slot, const gchar *hook_name, GError **error)
+{
+	g_autofree gchar *real_dest = NULL;
```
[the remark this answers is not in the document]
ejoerns: @zentax-dev This should be fixed.

zentax-dev:
> @zentax-dev Note that the GH code scanning still reports an "Unused local variable".

Removed now :)

src/emmc.c
```diff
@@ -206,3 +206,74 @@ gboolean r_emmc_force_part_rw(const gchar *device, GError **error)
 
 	return ret;
 }
+
+gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **base_device, GError **error)
```
ejoerns, line 210: The method could also directly return the path or NULL, but I'd be fine with the current implementation, too.

Note that the method lacks autopotr handling.

However, it might save a notable lines of code when leveraging our `r_regex_match_simple()` helper. (Note that this would require using a match group).

src/utils.c
```diff
@@ -721,6 +721,33 @@ goffset get_device_size(gint fd, GError **error)
...
+		return 0;
+	}
+
+	g_return_val_if_fail(error == NULL || *error == NULL, 0);
+
+	GError *ierror = NULL;
+	g_auto(filedesc) fd = g_open(device, O_RDONLY);
+	if (fd == -1) {
+		g_set_error(error, R_UTILS_ERROR, R_UTILS_ERROR_FAILED,
+				"Failed to open device %s: %s", device, g_strerror(errno));
```
ejoerns: In newer code, we would use

```suggestion
	if (fd == -1) {
	    int err = errno;
		g_set_error(error, G_FILE_ERROR, g_file_error_from_errno(err),
				"Failed to open device %s: %s", device, g_strerror(err));
```

src/utils.c
```diff
@@ -721,6 +721,33 @@ goffset get_device_size(gint fd, GError **error)
...
 }
 
+goffset get_device_size_from_dev(const gchar *device, GError **error)
+{
+	if (device == NULL) {
+		g_set_error(error, R_UTILS_ERROR, R_UTILS_ERROR_FAILED,
+				"Device path is NULL");
+		return 0;
+	}
+
```
ejoerns: Is there a use case for passing a null pointer? Otherwise I'd switch this to

```suggestion
    g_return_val_if_fail(device != NULL, FALSE);
```

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
-			part_slot->device,
-			INACTIVE_BOOT_PARTITION(part_active),
-			&ierror);
+	if (!r_emmc_toggle_active_bootpart(realdev, part_active, error))
+		return FALSE;
+
+	return TRUE;
+}
+
+static gboolean emmc_boot_linked_migration_helper(RaucImage *image, RaucSlot *dest_slot, const gchar *hook_name, GError **error)
```
ejoerns: What's the purpose of the `hook_name` argument? Could this be dropped?

```suggestion
static gboolean emmc_boot_linked_migration_helper(RaucImage *image, RaucSlot *dest_slot, GError **error)
```

I also wouldn't call it 'two handlers' in the commit message, as this might suggest they are both update handlers.

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+{
+	gboolean res = FALSE;
+	g_autofree gchar *real_dest = NULL;
+	g_autofree gchar *boot_suffix = NULL;
+	g_autofree gchar *base_device = NULL;
+	GError *ierror = NULL;
+	g_autoptr(RaucSlot) inactive_slot = NULL;
+	g_autoptr(RaucImage) source_image = NULL;
```
ejoerns: For newer code, there's no need anymore to place the declarations at the very beginning of a method.
Placing them where they are first initialized should slim down the code a bit.

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+
+	g_return_val_if_fail(image, FALSE);
+	g_return_val_if_fail(dest_slot, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	gint active_partition = -1;
```
ejoerns: Is this intentionally defined at this location?

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
-			&ierror);
+	if (!r_emmc_toggle_active_bootpart(realdev, part_active, error))
+		return FALSE;
+
+	return TRUE;
+}
+
+static gboolean emmc_boot_linked_migration_helper(RaucImage *image, RaucSlot *dest_slot, const gchar *hook_name, GError **error)
+{
+	gboolean res = FALSE;
```
ejoerns: No need to use `res` in this flow. Dropping it would save some lines.

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+	g_autoptr(RaucSlot) inactive_slot = NULL;
+	g_autoptr(RaucImage) source_image = NULL;
+
+	g_return_val_if_fail(image, FALSE);
+	g_return_val_if_fail(dest_slot, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	gint active_partition = -1;
+
+	g_message("Checking if the current layout needs to be migrated");
```
ejoerns: This message is bit generic and I wouldn't print it in 'message' log level, since a migration is only a very special case that should happen only once in the device's lifetime.

We should however print a message **if** a migration is actually required.

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+	/* Device is not boot enabled at all, so there is nothing to switch */
+	if (active_partition == -1) {
+		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_EMMC_MIGRATION,
+				"Device '%s' is not boot enabled", base_device);
+		return FALSE;
+	}
+
+	/* Check if the currently active boot partition is the same as the target boot partition.
+	 * Overwriting the active boot partition has to be avoided so there's still a fallback in
+	 * case anything anything goes wrong */
```
ejoerns:
```suggestion
	 * case anything goes wrong */
```

src/config_file.c
```diff
@@ -1027,6 +1028,25 @@ static GHashTable *parse_slots(const char *filename, const char *data_directory,
...
 
+			if (g_str_equal(slot->type, "emmc-boot-linked")) {
+				if (!(g_str_has_suffix(slot->device, "boot0") || g_str_has_suffix(slot->device, "boot1"))) {
+					g_set_error(error,
+							R_CONFIG_ERROR,
+							R_CONFIG_ERROR_INVALID_DEVICE,
+							"%s: 'device' must refer to the specific eMMC boot partitions, not the base device", groups[i]);
+					return NULL;
+				}
+				slot->size_limit = key_file_consume_binary_suffixed_string(key_file, groups[i],
```
ejoerns: Below, we already implemented parsing `size-limit`. Just add `emmc-boot-linked` to the cases where this is done.

Also, move the `emmc-boot-linked` parsing to the `boot-emmc` parsing, where it better fits.

src/config_file.c
```diff
@@ -1255,6 +1275,71 @@ static GHashTable *parse_artifact_repos(const char *filename, const char *data_d
 	return g_steal_pointer(&repos);
 }
 
+/*
+ * Checks if the emmc-boot-linked slots are configured correctly.
+ * Criteria: - All emmc-boot-linked slots refer to unique boot devices
+ *			 - All emmc-boot-linked slots are on the same device
```
ejoerns: Use spaces for alignment here.

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
-	 * for simplicity reasons: in case the user partition is active use
-	 * mmcblkXboot1, in case no partition is active use mmcblkXboot0
-	 */
-	g_debug("Toggling active eMMC boot partition %sboot%d -> %s", realdev, part_active,
-			part_slot->device);
-	res = r_emmc_write_bootpart(
-			part_slot->device,
-			INACTIVE_BOOT_PARTITION(part_active),
-			&ierror);
+	if (!r_emmc_toggle_active_bootpart(realdev, part_active, error))
```
ejoerns: For consistency, better use `g_propagate_error(error, ierror);` here, too:

```suggestion
	if (!r_emmc_toggle_active_bootpart(realdev, part_active, &ierror))
	        g_propagate_error(error, ierror);
```

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+		return TRUE;
+	}
+
+	/* Create a temporary RaucSlot for the inactive boot partition */
+	inactive_slot = g_new0(RaucSlot, 1);
+	inactive_slot->device = g_strdup_printf("%sboot%d", base_device, INACTIVE_BOOT_PARTITION(active_partition));
+	/* We need to deactivate the size limit for migration, as we're just copying the image
+	 * to set everything right for the new layout. So we just need to know
+	 * if source device fits into target device */
+	inactive_slot->size_limit = get_device_size_from_dev(inactive_slot->device, &ierror);
```
ejoerns: The `size-limit` exists to ensure that no pre-existing data on the eMMC is accidentally overridden.
This data is typically available on only one of the eMMC boot partitions.
Simply ignoring the `size-limit` undermines the whole point of the feature.

Setting `size_limit` for a pseudo slot also doesn't 'disable' the size limit, since it has never been set.
Setting it to the size of the target device just mimics the default behavior of omitting the size limit.

src/update_handler.c
```diff
@@ -2387,40 +2386,136 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+	if (ierror != NULL) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
+	/* Create a temporary RaucImage using the active boot partition device as filename */
+	source_image = g_new0(RaucImage, 1);
+	source_image->filename = g_strdup_printf("%sboot%d", base_device, active_partition);
+	source_image->checksum.size = get_device_size_from_dev(source_image->filename, &ierror);
```
ejoerns, line 2461: For proper `size-limit` handling, we would also make sure we only copy `size-limit` bytes from the original eMMC boot partition as we otherwise copy data that should remain untouched for RAUC.

The current copy behavior, however, does not support setting the `image->checksum.size` to something smaller than the data referred to by the `image->filename`.

At this point, I am unsure if we should support the (quite special) `size-limit` option for the (quite special) use case `emmc-boot-linked` at all.

src/config_file.c
```diff
@@ -6,9 +6,10 @@
 #include "config_file.h"
 #include "context.h"
 #include "event_log.h"
+#include "glibconfig.h"
```
ejoerns: What's the purpose of this? This does not seem to be a local header file, nor actually required.

src/config_file.c
```diff
@@ -6,9 +6,10 @@
...
 #include "install.h"
 #include "manifest.h"
-#include "mount.h"
```
ejoerns, line 11: Removing `mount.h` seems unrelated here.

test/test_write_slot.py
```diff
@@ -148,3 +148,627 @@ def test_write_boot_emmc_size_limit_too_large(system):
     assert f"Boot partition {device}boot0 is now active" in err
 
     assert "The size-limit (10485760 bytes) exceeds actual device size" in err
+
+
+@needs_emmc
+def test_write_emmc_boot_linked_no_active_boot(system):
```
ejoerns: Just a high-level feedback here first of all: The test cases largely extend the purpose of this file and make `test_write_slot` primarily an eMMC test suite. While using `write-slot` for most of the eMMC testing seems a simple solution (as I leveraged this for my tests, too, already), they should at least be moved to a distinct test file now.

I am also unsure if all the test cases for `boot-emmc` need to be replicated since most code paths are actually identical.

The code also suffers from a load of duplication, especially for the test system config. This should be moved to a common method or directly integrated into the `System` class. Most slot `*_device` variables are only used a single time, so they could be replaced by hard-coding the string (with keeping `{device}` as a parameter).

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
...
+
+This feature requires two boot slots, each one of type ``emmc-boot-linked``.
```
ejoerns: 'boot slots' is a term that has a slightly different meaning in RAUC already (a slot with a `bootname` set).
It would be more precise to say something like `requires each of the two eMMC boot partitions to be described as a slot of type..`

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
...
+They need to be uniquely linked and on the same eMMC.
+
+When this mechanism is used, RAUC will not write the *ext_csd registers*
+itself as in the :ref:`default mechanism <sec-emmc-boot>`,
+but will hand that over to the bootloader.
+Instead RAUC just maintains the boot state, which is then used by
+the bootloader to activate the correct boot partition from the eMMC,
+reset and start the OS with its matching bootloader.
+
+During migration from any mechanism not maintaining compatibility between
```
ejoerns: Aren't we migrating **to** a mechanism that doesn't require maintaining compatibility between bootloader and OS?

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
...
+bootloader and OS to the *eMMC Fallback Mechanism* the current state of
+the bootloader slot and the system slot can differ from what is expected of
+the grouped slots in the update.
```
ejoerns: It is too imprecise what 'state of the bootloader slot' means here. RAUC ensures that the state and the slot grouping are consistent. However, the activated eMMC boot partition might not yet match this.

docs/advanced.rst
```diff
@@ -1188,11 +1188,16 @@ SoC/firmware and storage medium.
...
 
-.. _sec-emmc-boot:
+
 
 Update Bootloader in eMMC Boot Partitions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. _sec-emmc-boot:
+
+Default Mechanism without Fallback
```
ejoerns: I wouldn't frame it like this is the mechanism that lacks functionality, since for most use cases, this is still the way to go. Maybe something like:

```suggestion
Default Mechanism (Atomic Update)
```

docs/advanced.rst
```diff
@@ -1229,6 +1234,70 @@ A ``system.conf`` could look like this:
     `important register cache fix
     <https://lore.kernel.org/all/20180308140811.6966-1-bst@pengutronix.de/>`_.
 
+.. _sec-emmc-boot-linked:
+
+eMMC Boot Linked
```
ejoerns: This might deserve a better title than just the deconstructed slot type.

ejoerns (changes requested):

ejoerns:
Applied some minor fixes.

test/slot.c
```diff
@@ -164,6 +164,52 @@ static void test_slot_get_root_classes(void)
...
+	RaucSlot *datafs_0 = NULL;
+	RaucSlot *datafs_1 = NULL;
+	RaucSlot *boot_0 = NULL;
+	g_autoptr(GList) type_slots = NULL;
+
+	slots = g_hash_table_new_full(g_str_hash, g_str_equal, NULL, r_slot_free);
+
+	rootfs_0 = g_new0(RaucSlot, 1);
+	rootfs_0->name = g_intern_string("rootfs.0");
+	rootfs_0->type = g_strdup("rootfs");
```
ejoerns: I guess it would be less confusing if you'd use valid types here.

test/config_file.c
```diff
@@ -729,6 +730,219 @@ type=boot-emmc\n";
...
+	gboolean res;
+
+	const gchar *contents = "\
+[system]\n\
+compatible=FooCorp Super BarBazzer\n\
+bootloader=barebox\n\
+\n\
+[slot.bootloader.0]\n\
+device=/dev/mmcblk0boot0\n\
+parent=rootfs.0\n\
```
ejoerns: I am unsure whether this semantically better fits or if we should make the bootloader the parent (and assign it a `bootname`). Maybe @jluebbe has an opinion on this. Leave it as-is for now.

ejoerns: Changed it as proposed.

src/emmc.c
```diff
@@ -206,3 +206,61 @@ gboolean r_emmc_force_part_rw(const gchar *device, GError **error)
...
+{
+	g_return_val_if_fail(device_path != NULL, FALSE);
+	g_return_val_if_fail(base_device != NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	/* Pattern to find valid MMC/eMMC base device */
+	*base_device = r_regex_match_simple("(/dev/mmcblk[0-9]+)", device_path);
```
jluebbe, line 217: The regex should be anchored at the start:
```suggestion
	*base_device = r_regex_match_simple("^(/dev/mmcblk[0-9]+)", device_path);
```

docs/advanced.rst
```diff
@@ -1270,6 +1274,64 @@ A ``system.conf`` could look like this:
...
+
+Each of the two eMMC boot partitions must be described as a slot of type ``emmc-boot-linked``.
+The boot slots need to be :ref:`grouped <sec-grouping-slots>` with their
+corresponding system slots.
+
+When this mechanism is used, RAUC will not switch the eMMC boot partitions
+itself (like the :ref:`default mechanism <sec-emmc-boot>` would)
+but will leave that to the bootloader.
+RAUC simply maintains the boot state, which the bootloader then uses
+to activate the correct boot partition on the eMMC,  perform a reset,
```
jluebbe, line 1297:
```suggestion
to activate the correct boot partition on the eMMC, perform a reset,
```

docs/advanced.rst
```diff
@@ -1270,6 +1274,64 @@ A ``system.conf`` could look like this:
...
+and start the OS with its matching bootloader.
+This reduces the effort required to maintain bootloader and OS consistency.
```
jluebbe, line 1299:
```suggestion
This can reduce the effort required to maintain bootloader and OS compatibility,
as each OS version is only started by a specific bootloader version.
On the other hand, protecting against bootloader downgrades becomes more complex.
```

docs/advanced.rst
```diff
@@ -1270,6 +1274,64 @@ A ``system.conf`` could look like this:
...
+
+When migrating to this mechanism, the currently used eMMC boot partition and system slot
+may not yet match those which are linked.
+This case is handled by RAUC in order to avoid inconsistent results and unbootable systems.
```
jluebbe, line 1303:
```suggestion
In order to avoid inconsistent results and unbootable systems, this case is handled by RAUC.
```

docs/advanced.rst
```diff
@@ -1270,6 +1274,64 @@ A ``system.conf`` could look like this:
...
+During the update process, RAUC compares the bootloader update target with
+the currently activated bootloader.
```
jluebbe, line 1305:
```suggestion
During the update process, RAUC checks if the bootloader update target is
the currently activated bootloader partition.
```

docs/advanced.rst
```diff
@@ -1270,6 +1274,64 @@ A ``system.conf`` could look like this:
...
+If the bootloader currently active is also the one intended for the bootloader target slot,
+it will be copied to the slot that is inactive, which is then activated by RAUC.
+The update procedure will then continue.
```
jluebbe, line 1308:
```suggestion
If the currently active bootloader partition is the intended update target slot,
it's contents will first be copied to the inactive slot, which is then activated by RAUC.
The update procedure will then continue as normal.
```

docs/reference.rst
```diff
@@ -567,6 +567,9 @@ separator.
   ``ubivol``, ``ubifs``, ``ext4``, ``vfat``, ``jffs2`` for normal slots and
   ``boot-emmc``, ``boot-mbr-switch``, ``boot-gpt-switch``, and
   ``boot-raw-fallback`` for atomically updatable bootloader slots.
+  The type ``emmc-boot-linked`` can be used to group two eMMC boot slots
+  on the same device with other partitions, when incompatibilities
+  during updates cannot be resolved differently.
```
jluebbe, line 572:
```suggestion
  The type ``emmc-boot-linked`` can be used to group two eMMC boot slots
  on the same device with other partitions, when incompatibilities
  during updates cannot be resolved in another way.
```

src/emmc.c
```diff
@@ -224,3 +224,43 @@ gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **base_device,
 
 	return TRUE;
 }
+
+gboolean r_emmc_toggle_active_bootpart(const gchar *device, gint active_partition, GError **error)
+{
+	GError *ierror = NULL;
+
+	g_return_val_if_fail(device != NULL, FALSE);
+	g_return_val_if_fail(active_partition < EMMC_BOOT_PARTITIONS, FALSE);
```
jluebbe, line 233: Check that `active_partition` is not negative.

src/emmc.c
```diff
@@ -224,3 +224,43 @@ gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **base_device,
...
+	g_debug("Toggling active eMMC boot partition %sboot%d", device, active_partition);
+	if (!r_emmc_write_bootpart(device, INACTIVE_BOOT_PARTITION(active_partition), &ierror)) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
+	/* sanity check: read active boot partition from ext_csd
+	 *
+	 * Read explicitly from root device (this forces another kernel
+	 * partition switch and should trigger the ext_csd bug more reliably).
```
jluebbe, line 250: This comment is no longer correct as both calls now access `device`. We should discuss how best to fix this.

src/config_file.c
```diff
@@ -1165,6 +1176,72 @@ static GHashTable *parse_artifact_repos(const char *filename, const char *data_d
...
+ *
+ * @param config RaucConfig object
+ * @param error Return location for a GError, or NULL
+ *
+ * @return TRUE if all criteria pass or no emmc-boot-linked exists, FALSE otherwise */
+
+static gboolean check_emmc_linked_slots(RaucConfig *config, GError **error)
+{
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+	g_return_val_if_fail(config != NULL, FALSE);
```
jluebbe, line 1192: Sort in the same order as the parameters.

src/update_handler.c
```diff
@@ -2396,6 +2396,114 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
 
 	return TRUE;
 }
+static gboolean emmc_boot_linked_migration_helper(RaucImage *image, RaucSlot *dest_slot, GError **error)
```
jluebbe, line 2399: Add empty line. Also, the `image` parameter is unused.

src/update_handler.c
```diff
@@ -2396,6 +2396,114 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
+				"Device '%s' is not boot enabled", base_device);
+		return FALSE;
+	}
+
+	/* Check if the target slot is the currently active eMMC boot partition */
+	g_message("Active eMMC boot partition for %s: boot%d", real_dest, active_partition);
+	g_autofree gchar *boot_suffix = g_strdup_printf("boot%d", active_partition);
+	if (!g_str_has_suffix(dest_slot->device, boot_suffix)) {
+		return TRUE;
+	}
```
jluebbe, line 2442: The comment is wrong. This checks if the the target slot is not the active one and returns successfully in that case, as no copying is needed.

test/conftest.py
```diff
@@ -667,6 +692,13 @@ def system(tmp_path, dbus_session_bus):
     yield system
 
 
+class EMMC:
+    def __init__(self):
+        self.base_dev = os.environ["RAUC_TEST_EMMC"]
+        self.boot0 = f"{self.base_dev}boot0"
+        self.boot1 = f"{self.base_dev}boot1"
```
jluebbe, line 699: The tests could be simplified a lot by using Path objects, but that could be a follow-up PR.

test/test_emmc.py
```diff
@@ -107,4 +107,226 @@ def test_write_boot_emmc_size_limit_too_large(system):
...
+    force_ro_path = f"/sys/block/{emmc.boot0[5:]}/force_ro"
+    assert os.path.exists(force_ro_path)
+
+    # Make the force_ro file read-only to simulate permission failure
+    os.chmod(force_ro_path, 0o444)
+
+    out, err, exitcode = run(f"{system.prefix} write-slot bootloader.0 install-content/rootfs.img")
+
+    # Restore permissions
+    os.chmod(force_ro_path, 0o644)
```
jluebbe, line 274: Make sure to restore permissions by using `try:`/`finally:`.

include/emmc.h
```diff
@@ -89,3 +89,31 @@ gboolean r_emmc_force_part_ro(const gchar *device, GError **error);
...
+ */
+gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **base_device, GError **error)
+G_GNUC_WARN_UNUSED_RESULT;
+
+/**
+ * Toggles the active eMMC boot partition.
+ *
+ * This function sets the specified boot partition as active for the given
+ * eMMC device. It ensures that the specified partition is correctly set
+ * as the active boot partition, allowing the system to boot from it.
```
jluebbe, line 110: The comment is wrong. The function sets the **other** partition as active, which is leads to confusing usage of this function. Perhaps a better name would be `r_emmc_activate_bootpart` and passing the intended new state.

jluebbe (changes requested):

ejoerns:
Update of the PR with what I had in the queue from the last rework iteration.


<details><summary>range-diff</summary>
<p>

``` range-diff
 1:  928040c5 !  1:  b7781fa2 src/emmc: add function to get base dev of partitioned eMMC
    @@ Commit message
         if two partitions are targeting the same eMMC.
     
         Signed-off-by: Lars Schmidt <l.schmidt@pengutronix.de>
    +    Signed-off-by: Enrico Jörns <ejo@pengutronix.de>
     
      ## include/emmc.h ##
     @@ include/emmc.h: gboolean r_emmc_force_part_ro(const gchar *device, GError **error);
    @@ include/emmc.h: gboolean r_emmc_force_part_ro(const gchar *device, GError **erro
     + * Extracts the base device from the given eMMC boot partition.
     + *
     + * @param device_path eMMC boot partition /dev path (/dev/mmcblkXbootY)
    -+ * @param base_device return location for the output base device (/dev/mmcblkX)
     + * @param error return location for a GError, or NULL
     + *
    -+ * @return True if succeeded, False if failed
    ++ * @return base device if succeeded, NULL if failed
     + */
    -+gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **base_device, GError **error)
    ++gchar* r_emmc_extract_base_dev(const gchar *device_path, GError **error)
     +G_GNUC_WARN_UNUSED_RESULT;
     
      ## src/emmc.c ##
    @@ src/emmc.c: gboolean r_emmc_force_part_rw(const gchar *device, GError **error)
      	return ret;
      }
     +
    -+gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **base_device, GError **error)
    ++gchar* r_emmc_extract_base_dev(const gchar *device_path, GError **error)
     +{
    -+	g_return_val_if_fail(device_path != NULL, FALSE);
    -+	g_return_val_if_fail(base_device != NULL, FALSE);
    -+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
    ++	g_return_val_if_fail(device_path != NULL, NULL);
    ++	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
     +
     +	/* Pattern to find valid MMC/eMMC base device */
    -+	*base_device = r_regex_match_simple("(/dev/mmcblk[0-9]+)", device_path);
    ++	gchar *base_device = r_regex_match_simple("^(/dev/mmcblk[0-9]+)", device_path);
     +
    -+	if (*base_device == NULL) {
    ++	if (base_device == NULL) {
     +		g_set_error(error, R_EMMC_ERROR, R_EMMC_ERROR_FAILED,
     +				"Device path '%s' does not contain valid MMC device pattern", device_path);
    -+		return FALSE;
    ++		return NULL;
     +	}
     +
    -+	return TRUE;
    ++	return base_device;
     +}
 2:  78d7e90e =  2:  2569e863 src/utils: add function to get size from dev
 3:  63006b63 =  3:  8a9de392 src/slot: add function to get all slots of a type
 4:  e58556b0 !  4:  72dbb69e src/emmc: refactor toggle the active boot partition of an eMMC
    @@ Metadata
     Author: Lars Schmidt <l.schmidt@pengutronix.de>
     
      ## Commit message ##
    -    src/emmc: refactor toggle the active boot partition of an eMMC
    +    src/emmc: refactor eMMC boot partition toggling
     
    -    Move toggling and checking of active partition to its own function .
    -    The function takes the currently active boot partition from its parameters
    -    and changes the active boot partition to the currently inactive boot partition.
    -    It also validates if the activation was successful, in order to check for a known
    -    ioctl ext_csd bug, see [1].
    +    Move boot partition activation and validation into a helper function.
     
         This encapsulates already existing code into a function and makes it easier to reuse
         and avoids duplication in the future.
     
    -    [1]: https://rauc.readthedocs.io/en/latest/advanced.html#update-bootloader-in-emmc-boot-partitions
    -
         Signed-off-by: Lars Schmidt <l.schmidt@pengutronix.de>
     
      ## include/emmc.h ##
     @@ include/emmc.h: G_GNUC_WARN_UNUSED_RESULT;
       */
    - gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **base_device, GError **error)
    + gchar* r_emmc_extract_base_dev(const gchar *device_path, GError **error)
      G_GNUC_WARN_UNUSED_RESULT;
     +
     +/**
    -+ * Toggles the active eMMC boot partition.
    ++ * Active the specified eMMC boot partition.
     + *
     + * This function sets the specified boot partition as active for the given
     + * eMMC device. It ensures that the specified partition is correctly set
     + * as the active boot partition, allowing the system to boot from it.
     + *
     + * @param device eMMC device path
    -+ * @param active_partition the currently active partition index (0 or 1)
    ++ * @param activate_partition the partition index to activate (0 or 1)
     + * @param error return location for a GError, or NULL
     + *
     + * @return True if the operation succeeded, False if it failed
     + */
    -+gboolean r_emmc_toggle_active_bootpart(const gchar *device, gint active_partition, GError **error)
    ++gboolean r_emmc_activate_bootpart(const gchar *device, gint activate_partition, GError **error)
     +G_GNUC_WARN_UNUSED_RESULT;
     
      ## src/emmc.c ##
    -@@ src/emmc.c: gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **base_device,
    +@@ src/emmc.c: gchar* r_emmc_extract_base_dev(const gchar *device_path, GError **error)
      
    - 	return TRUE;
    + 	return base_device;
      }
     +
    -+gboolean r_emmc_toggle_active_bootpart(const gchar *device, gint active_partition, GError **error)
    ++gboolean r_emmc_activate_bootpart(const gchar *device, gint activate_partition, GError **error)
     +{
     +	GError *ierror = NULL;
     +
     +	g_return_val_if_fail(device != NULL, FALSE);
    -+	g_return_val_if_fail(active_partition < EMMC_BOOT_PARTITIONS, FALSE);
    ++	g_return_val_if_fail(activate_partition >= 0 && activate_partition < EMMC_BOOT_PARTITIONS, FALSE);
     +
    -+	/* toggle active boot partition in ext_csd register; do this explicitly on
    -+	 * determined boot partition to force the kernel to switch to the partition;
    -+	 * for simplicity reasons: in case the user partition is active use
    -+	 * mmcblkXboot1, in case no partition is active use mmcblkXboot0
    -+	 */
    -+	g_debug("Toggling active eMMC boot partition %sboot%d", device, active_partition);
    -+	if (!r_emmc_write_bootpart(device, INACTIVE_BOOT_PARTITION(active_partition), &ierror)) {
    ++	if (!r_emmc_write_bootpart(device, activate_partition, &ierror)) {
     +		return FALSE;
     +	}
    @@ src/emmc.c: gboolean r_emmc_extract_base_dev(const gchar *device_path, gchar **b
     +		return FALSE;
     +	}
     +
    -+	if (active_partition == part_active_after) {
    ++	if (activate_partition != part_active_after) {
    -+				"Toggling the boot partition failed! Your kernel is most-likely affected by the ioctl ext_csd bug: see https://rauc.readthedocs.io/en/latest/advanced.html#update-bootloader-in-emmc-boot-partitions");
    ++				"Activating the boot partition failed! Your kernel is most-likely affected by the ioctl ext_csd bug: see https://rauc.readthedocs.io/en/latest/advanced.html#update-bootloader-in-emmc-boot-partitions");
     +		return FALSE;
     +	}
     +
    @@ src/update_handler.c: static gboolean img_to_boot_emmc_handler(RaucImage *image,
      	g_autoptr(RaucSlot) part_slot = NULL;
      	g_autoptr(GHashTable) vars = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
     @@ src/update_handler.c: static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
    - 		return FALSE;
    - 	}
    - 
    --	/* toggle active boot partition in ext_csd register; do this explicitly on
    --	 * determined boot partition to force the kernel to switch to the partition;
    --	 * for simplicity reasons: in case the user partition is active use
    --	 * mmcblkXboot1, in case no partition is active use mmcblkXboot0
    --	 */
    --	g_debug("Toggling active eMMC boot partition %sboot%d -> %s", realdev, part_active,
    --			part_slot->device);
    + 	 */
    + 	g_debug("Toggling active eMMC boot partition %sboot%d -> %s", realdev, part_active,
    + 			part_slot->device);
     -	res = r_emmc_write_bootpart(
     -			part_slot->device,
     -			INACTIVE_BOOT_PARTITION(part_active),
     -			&ierror);
     -	if (!res) {
    -+	if (!r_emmc_toggle_active_bootpart(realdev, part_active, &ierror)) {
    - 		g_propagate_error(error, ierror);
    - 		return FALSE;
    - 	}
    - 
    +-		g_propagate_error(error, ierror);
    +-		return FALSE;
    +-	}
    +-
     -	/* sanity check: read active boot partition from ext_csd
     -	 *
     -	 * Read explicitly from root device (this forces another kernel
    @@ src/update_handler.c: static gboolean img_to_boot_emmc_handler(RaucImage *image,
     -	 */
     -	res = r_emmc_read_bootpart(realdev, &part_active_after, &ierror);
     -	if (!res) {
    --		g_propagate_error(error, ierror);
    --		return FALSE;
    --	}
    --
    ++	if (!r_emmc_activate_bootpart(realdev, INACTIVE_BOOT_PARTITION(part_active), &ierror)) {
    + 		g_propagate_error(error, ierror);
    + 		return FALSE;
    + 	}
    + 
     -	if (part_active == part_active_after) {
     -		g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_FAILED,
     -				"Toggling the boot partition failed! Your kernel is most-likely affected by the ioctl ext_csd bug: see https://rauc.readthedocs.io/en/latest/advanced.html#update-bootloader-in-emmc-boot-partitions");
 5:  dc5ce6e0 !  5:  003b8fcc src/config_file: Add emmc-boot-linked to config
    @@ Commit message
         gets detected early and the user is notified accordingly.
     
         Signed-off-by: Lars Schmidt <l.schmidt@pengutronix.de>
    +    Signed-off-by: Enrico Jörns <ejo@pengutronix.de>
     
      ## src/config_file.c ##
     @@
    @@ src/config_file.c: static GHashTable *parse_artifact_repos(const char *filename,
     +
     +static gboolean check_emmc_linked_slots(RaucConfig *config, GError **error)
     +{
    -+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
     +	g_return_val_if_fail(config != NULL, FALSE);
    ++	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
     +
     +	GError *ierror = NULL;
     +
    @@ src/config_file.c: static GHashTable *parse_artifact_repos(const char *filename,
     +	}
     +
     +	/* And all slots need to be on same base device */
    -+	g_autofree gchar *base_dev1 = NULL;
    -+	g_autofree gchar *base_dev2 = NULL;
    -+	if (!r_emmc_extract_base_dev(slot1->device, &base_dev1, &ierror) ||
    -+	    !r_emmc_extract_base_dev(slot2->device, &base_dev2, &ierror)) {
    ++	g_autofree gchar *base_dev1 = r_emmc_extract_base_dev(slot1->device, &ierror);
    ++	if (base_dev1 == NULL) {
    ++		g_propagate_error(error, ierror);
    ++		return FALSE;
    ++	}
    ++	g_autofree gchar *base_dev2 = r_emmc_extract_base_dev(slot2->device, &ierror);
    ++	if (base_dev2 == NULL) {
     +		return FALSE;
     +	}
 6:  877c98fd !  6:  7cf79194 src/update_handler: add handler for emmc-boot-linked
    @@ src/update_handler.c: static gboolean img_to_boot_emmc_handler(RaucImage *image,
      
      	return TRUE;
      }
    -+static gboolean emmc_boot_linked_migration_helper(RaucImage *image, RaucSlot *dest_slot, GError **error)
    ++
    ++static gboolean emmc_boot_linked_migration_helper(RaucSlot *dest_slot, GError **error)
     +{
    -+	g_return_val_if_fail(image, FALSE);
     +
    @@ src/update_handler.c: static gboolean img_to_boot_emmc_handler(RaucImage *image,
     +	/* Get the base device for out destination partition so we can check which boot partition
     +	 * is actually activated */
     +	GError *ierror = NULL;
    -+	g_autofree gchar *base_device = NULL;
    -+	if (!r_emmc_extract_base_dev(real_dest, &base_device, &ierror)) {
    ++	g_autofree gchar *base_device = r_emmc_extract_base_dev(real_dest, &ierror);
    ++	if (base_device == NULL) {
     +		return FALSE;
     +	}
    @@ src/update_handler.c: static gboolean img_to_boot_emmc_handler(RaucImage *image,
     +		return FALSE;
     +	}
     +
    -+	/* Check if the target slot is the currently active eMMC boot partition */
    ++	/* Skip migration if the target slot is not the currently active eMMC boot partition */
     +	g_message("Active eMMC boot partition for %s: boot%d", real_dest, active_partition);
     +	g_autofree gchar *boot_suffix = g_strdup_printf("boot%d", active_partition);
     +	if (!g_str_has_suffix(dest_slot->device, boot_suffix)) {
    @@ src/update_handler.c: static gboolean img_to_boot_emmc_handler(RaucImage *image,
     +		return FALSE;
     +	}
     +
    -+	if (!r_emmc_toggle_active_bootpart(base_device, active_partition, &ierror)) {
    ++	g_debug("Toggling active eMMC boot partition %sboot%d", base_device, INACTIVE_BOOT_PARTITION(active_partition));
    ++	if (!r_emmc_activate_bootpart(base_device, INACTIVE_BOOT_PARTITION(active_partition), &ierror)) {
     +		return FALSE;
     +	}
    @@ src/update_handler.c: static gboolean img_to_boot_emmc_handler(RaucImage *image,
     +	 * the boot partitions might not yet be aligned as expected from the
     +	 * new linked config. So we have to check how they are aligned and
     +	 * ensure the target boot partition is not the currently activated one. */
    -+	if (!emmc_boot_linked_migration_helper(image, dest_slot, &ierror)) {
    ++	if (!emmc_boot_linked_migration_helper(dest_slot, &ierror)) {
     +		return FALSE;
     +	}
 7:  364a1314 =  7:  3f6eba63 test/emmc: separate emmc tests from write_slots
 8:  f0a710f8 !  8:  0e466b33 test/emmc: Add tests for emmc-boot-linked slot types
    @@ test/conftest.py: class System:
     +    def prepare_emmc_boot_linked_config(self):
     +        device_emmc = EMMC()
     +        self.config["slot.rootfs.0"] = {
    -+            "device": f"{device_emmc.base_dev}.root0",
    ++            "device": f"{device_emmc.base_dev}p1",
     +            "type": "ext4",
     +        }
     +
    @@ test/conftest.py: class System:
     +        }
     +
     +        self.config["slot.rootfs.1"] = {
    -+            "device": f"{device_emmc.base_dev}.root1",
    ++            "device": f"{device_emmc.base_dev}p2",
     +            "type": "ext4",
     +        }
     +
    @@ test/test_emmc.py: def test_write_boot_emmc_size_limit_too_large(system):
     +
     +    out, err, exitcode = run(f"{system.prefix} write-slot bootloader.0 install-content/rootfs.img")
     +    assert exitcode == 1
    -+    assert "No data directory or status file set, falling back to per-slot status."
    -+    assert "Consider setting 'data-directory=<path>' or 'statusfile=<path>/per-slot' explicitly."
    -+    assert "Using per-slot statusfile. System status information not supported!"
    ++    assert "No data directory or status file set, falling back to per-slot status." in err
    ++    assert "Consider setting 'data-directory=<path>' or 'statusfile=<path>/per-slot' explicitly." in err
    ++    assert "Using per-slot statusfile. System status information not supported!" in err
     +    assert (
     +        f"emmc-boot-linked slots 'bootloader.1' and 'bootloader.0' cannot use the same boot device '{emmc.boot0}'"
     +        in err
    @@ test/test_emmc.py: def test_write_boot_emmc_size_limit_too_large(system):
     +    # Make the force_ro file read-only to simulate permission failure
     +    os.chmod(force_ro_path, 0o444)
     +
    -+    out, err, exitcode = run(f"{system.prefix} write-slot bootloader.0 install-content/rootfs.img")
    -+
    -+    # Restore permissions
    -+    os.chmod(force_ro_path, 0o644)
    ++    try:
    ++        out, err, exitcode = run(f"{system.prefix} write-slot bootloader.0 install-content/rootfs.img")
    ++    finally:
    ++        # Restore permissions
    ++        os.chmod(force_ro_path, 0o644)
     +
     +    assert exitcode != 0
     +    assert (
 9:  5e7ab431 !  9:  12d280f7 docs/advanced: describe new emmc-boot-linked slot type
    @@ docs/advanced.rst: A ``system.conf`` could look like this:
     +itself (like the :ref:`default mechanism <sec-emmc-boot>` would)
     +but will leave that to the bootloader.
     +RAUC simply maintains the boot state, which the bootloader then uses
    -+to activate the correct boot partition on the eMMC,  perform a reset,
    ++to activate the correct boot partition on the eMMC, perform a reset,
     +and start the OS with its matching bootloader.
    -+This reduces the effort required to maintain bootloader and OS consistency.
    ++This can reduce the effort required to maintain bootloader and OS compatibility.
     +
     +When migrating to this mechanism, the currently used eMMC boot partition and system slot
     +may not yet match those which are linked.
    -+This case is handled by RAUC in order to avoid inconsistent results and unbootable systems.
    -+During the update process, RAUC compares the bootloader update target with
    -+the currently activated bootloader.
    -+If the bootloader currently active is also the one intended for the bootloader target slot,
    -+it will be copied to the slot that is inactive, which is then activated by RAUC.
    -+The update procedure will then continue.
    ++In order to avoid inconsistent results and unbootable systems, this case is handled by RAUC.
    ++During the update process, RAUC checks if the bootloader update target is
    ++the currently activated bootloader partition.
    ++If the currently active bootloader partition is the intended update target slot,
    ++it's contents will first be copied to the inactive slot, which is then activated by RAUC.
    ++The update procedure will then continue as normal.
     +This ensures that there is still a working fallback in case of an unsuccessful update.
     +
     +.. code-block:: cfg
    @@ docs/reference.rst: separator.
        ``boot-raw-fallback`` for atomically updatable bootloader slots.
     +  The type ``emmc-boot-linked`` can be used to group two eMMC boot slots
     +  on the same device with other partitions, when incompatibilities
    -+  during updates cannot be resolved differently.
    ++  during updates cannot be resolved in another way.
        Defaults to ``raw`` if none given.
```

src/emmc.c
ejoerns: It might also make sense to wait for the rework in #1901 to have settled before touching this code part.

PR #1753, opened 2025-07-17
src/context: parse the case cmdline contain root=None

BenjaminMiss:
In the case of the cmdline contain root=None, rauc service fail. It is because the function get_cmdline_bootname_root() does not provide for this scenario and continue to the end before crashing. If cmdline contain root=None, the function get_cmdline_bootname_root() should give the same result as for case root=xxx is not define, namely returning NULL to avoid rauc failures.
root=None on cmdline can occur on Yocto images for x86_64 system.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -96,7 +96,7 @@ static gchar* get_cmdline_bootname_root(const gchar *cmdline)
 	if (!bootname)
 		bootname = r_regex_match_simple("systemd\\.verity_root_data=(\\S+)", cmdline);
 
-	if (!bootname)
+	if (!bootname || strncmp(bootname, "None", 4) == 0)
 		return NULL;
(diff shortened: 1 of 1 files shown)


jluebbe:
Wouldn't the kernel also fail to boot with `root=None`, as that's not a valid device? Where does this string come from in your case?

ejoerns:
@BenjaminMiss  `None` really looks like a bitbake value that is empty. I don't think we should check for values resulting only from invalid build system results. I have found rare cases with `root=none` on the internet, but I'm also unsure if this is a proper string for the cmdline.
If you can describe a proper use case where `None` is *intentionally* used as a `root=` value, then we might consider handling this; otherwise, I'd prefer to close this PR.

PR #1756, opened 2025-07-25
Feat add image type to manifest

zentax-dev:


zentax-dev:
Just added a few more tests

src/update_handler.c
```diff
@@ -2709,6 +2709,47 @@ RaucUpdatePair updatepairs[] = {
...
+};
+
+static gboolean derive_image_type_from_filename_pattern(const gchar *filename, const gchar **type)
+{
+	g_return_val_if_fail(filename, FALSE);
+	g_return_val_if_fail(type, FALSE);
+
+	for (unsigned long i = 0; i < G_N_ELEMENTS(ext_type_map); i++) {
+		if (g_pattern_match_simple(ext_type_map[i].fileext, filename)) {
+			*type = ext_type_map[i].type;
```
jluebbe: This should just return the pointer or `NULL` on error instead of a output parameter.

src/update_handler.c
```diff
@@ -2709,6 +2709,79 @@ RaucUpdatePair updatepairs[] = {
...
+	{"image", "nand", img_to_nand_handler},
+	{"image", "ubifs", img_to_ubifs_handler},
+	{"image", "ubivol", img_to_ubivol_handler},
+	{"image", "vfat", img_to_fs_handler},
+	{"image", "raw", img_to_raw_handler},
+	/* archive */
+	{"tar", "ext4", archive_to_ext4_handler},
+	{"tar", "ubifs", archive_to_ubifs_handler},
+	{"tar", "vfat", archive_to_vfat_handler},
+	/* casync mappings from test matrix */
```
jluebbe: Leftover comment?

src/update_handler.c
```diff
@@ -2709,6 +2709,79 @@ RaucUpdatePair updatepairs[] = {
 	{0}
 };
 
+typedef struct {
+	const gchar *type;
+	const gchar *slottype;
+	img_to_slot_handler handler;
+} RaucImageTypeMap;
+
+/* New image type to handler mapping */
```
jluebbe: Please add a comment how these are sorted.

src/update_handler.c
```diff
@@ -2823,6 +2823,22 @@ static gboolean derive_image_type_from_filename_pattern(const gchar *filename, c
...
 }
 
+static img_to_slot_handler get_handler_from_type(const gchar *image_type, const gchar *slot_type)
+{
+	g_return_val_if_fail(image_type, NULL);
+	g_return_val_if_fail(slot_type, NULL);
+
+	for (RaucImageTypeMap *map = image_type_map; map->type != NULL; map++) {
+		if (g_strcmp0(map->type, image_type) == 0 &&
+		    (g_strcmp0(map->slottype, "*") == 0 ||
```
jluebbe: `slottype` = `"*"` seems unused?

src/update_handler.c
```diff
@@ -2852,21 +2776,24 @@ img_to_slot_handler get_update_handler(RaucImage *mfimage, RaucSlot *dest_slot,
...
+			g_message("Found handler for image type '%s' and slot type '%s'", mfimage->type, dest);
+			return handler;
+		}
 	}
 
-out:
-	return handler;
+	g_set_error(error, R_UPDATE_ERROR, R_UPDATE_ERROR_NO_HANDLER,
+			"Unsupported image %s for slot type %s", src, dest);
+	return NULL;
```
jluebbe: Keep the good case as the normal path by checking the error case using `if (!mfimage->type)` or `if (!handler)`.

include/manifest.h
```diff
@@ -13,6 +13,7 @@ GQuark r_manifest_error_quark(void);
 #define R_MANIFEST_PARSE_ERROR		3
 #define R_MANIFEST_EMPTY_STRING		4
 #define R_MANIFEST_CHECK_ERROR		5
+#define R_MANIFEST_ERROR_INVALID_TYPE 6
```
jluebbe: Keep the alignment and use `R_MANIFEST_ERROR_INVALID_IMGE_TYPE` to be more explict.

docs/reference.rst
```diff
@@ -921,12 +921,40 @@ The following fields are supported for image sections:
 ``filename`` (required)
   Name of the image file (relative to bundle content).
 
+  .. note::
+    While RAUC can automatically detect the image type from filename extensions
+    for backward compatibility, it is recommended to explicitly specify the
+    ``type`` field for new configurations. The filename extension-based detection
+    is considered deprecated and may be removed in future versions.
```
jluebbe: Use https://sembr.org/ formatting. Also I wouldn't consider the detection to be deprecated yet.

docs/reference.rst
```diff
@@ -921,12 +921,40 @@ The following fields are supported for image sections:
...
 
+``type`` (required)
+  Specifies the type of the image content. This determines how RAUC will process
+  the image during installation. If not specified, the type is automatically
+  detected based on the filename extension and target slot type.
+
+  Supported values are:
+
+  * ``caidx`` - Casync directory tree index format
+  * ``catar`` - Casync directory tree archive format
```
jluebbe: `casync` should not be capitalized.

docs/reference.rst
```diff
@@ -921,12 +921,40 @@ The following fields are supported for image sections:
...
+  * ``ext4`` - ext[234] file system image
+  * ``ext4-caibx`` - ext4 file system image in casync blob index format
+  * ``vfat`` - vfat/EFI file system image
```
jluebbe: That EFI can use vfat filesystems is not relevant here.

docs/reference.rst
```diff
@@ -921,12 +921,40 @@ The following fields are supported for image sections:
...
+  * ``vfat-caibx`` - vfat file system image in casync blob index format
+  * ``tar`` - TAR archive (compressed or uncompressed)
+  * ``archive`` - Generic archive format (TAR-based)
```
jluebbe: We should only use the explicit `tar` type.

docs/reference.rst
```diff
@@ -921,12 +921,40 @@ The following fields are supported for image sections:
...
     extract the image file content to the slot.
     Make sure to only use :ref:`supported file name extensions
     <sec-ref-supported-image-types>`!
 
+``type`` (required)
+  Specifies the type of the image content. This determines how RAUC will process
+  the image during installation. If not specified, the type is automatically
+  detected based on the filename extension and target slot type.
+
+  Supported values are:
```
jluebbe: We should refer to the "Supported Image Types" instead of duplicating the list here.

jluebbe (changes requested):
With the current ordering, some patches use the `image->type` field before it is introduced, breaking `git bisect`.

The type detection should be introduced first, then the handler selection should be changed to used the explicit or detected type.

zentax-dev:
I think I've adressed everything with the latest changes

src/update_handler.c
dwntri: In the image type map, all mappings are unique. So actually it does not matter how this map is sorted.
But I've added a comment to `RaucFileExtTypeMap ext_type_map[]`. As the order of that map might be important

test/manifest.c
```diff
@@ -359,6 +361,89 @@ filename=rootfs-var2.ext4\n\
...
+	GError *error = NULL;
+	const gchar *mffile = "\
+[update]\n\
+compatible=FooCorp Super BarBazzer\n\
+version=2015.04-1\n\
+\n\
+[image.rootfs]\n\
+type=invalid\n\
+filename=rootfs-default.ext4\n\
+";
```
jluebbe, line 427: Also add a test case for a file name for which no image type exists. That would improve the test coverage for manifest parsing as well.

src/update_handler.c
```diff
@@ -2633,85 +2633,142 @@ static gboolean hook_install_handler(RaucImage *image, RaucSlot *dest_slot, cons
...
+	{"*.img.caibx", "img-caibx"},
+	{"*.squashfs.caibx", "squashfs-caibx"},
+	{"*.squashfs-*.caibx", "squashfs-caibx"},
+	{"*.catar", "catar"},
+	{"*.caidx", "caidx"},
+	{"*.tar*", "tar"},
+	{"*.tgz", "tar"},
+	{"*.ext4", "ext4"},
+	{"*.vfat", "vfat"},
+	{"*.img", "image"},
```
jluebbe: Should this be `raw` instead (also above as `img-caibx`)? Having the "Image Type" `image` seems a bit redundant.

docs/reference.rst
ejoerns: FTR: This text originates from the existing documentation.

src/update_handler.c
```diff
@@ -2633,85 +2633,142 @@ static gboolean hook_install_handler(RaucImage *image, RaucSlot *dest_slot, cons
...
-	{"*.squashfs-lzo", "ubivol", img_to_ubivol_handler},
-	{"*.squashfs-xz", "ubivol", img_to_ubivol_handler},
-	{"*.squashfs-zst", "ubivol", img_to_ubivol_handler},
+} RaucImageTypeMap;
+
+/* Image type to handler mapping */
+static RaucImageTypeMap image_type_map[] = {
+	/* casync - caibx */
+	{"ext4-caibx", "ext4", img_to_fs_handler},
+	{"vfat-caibx", "ext4", img_to_fs_handler},
```
jluebbe: Writing a `vfat` image to an `ext4` slot is wrong. Remove this.

src/update_handler.c
```diff
@@ -2633,85 +2633,142 @@ static gboolean hook_install_handler(RaucImage *image, RaucSlot *dest_slot, cons
...
+	{"img-caibx", "ext4", img_to_fs_handler},
```
jluebbe: The old code used `img_to_raw_handler` via the fallback.

src/update_handler.c
```diff
@@ -2633,85 +2633,142 @@ static gboolean hook_install_handler(RaucImage *image, RaucSlot *dest_slot, cons
...
 };
 
+typedef struct {
+	const gchar *fileext;
+	const gchar *type;
+} RaucFileExtTypeMap;
+
+/* For compatibility reasons, this mapping is needed to map the former file extension based approach to the
+ * image types. The file name extensions are processed in the order of more specific ones first (with double endings)
+ * down to the more general ones to avoid probable misdetection */
```
jluebbe: Use "The mapping must be ordered with more specific patterns first (e.g. '*.squashfs-*.caibx' before '*.squashfs-*') to avoid selecting the wrong image type." for the second sentence.

src/update_handler.c
```diff
@@ -2633,85 +2633,142 @@ static gboolean hook_install_handler(RaucImage *image, RaucSlot *dest_slot, cons
...
 #endif
-	{"*", "boot-gpt-switch", NULL},
-	{"*.img", "boot-raw-fallback", img_to_boot_raw_fallback_handler},
-	{"*", "boot-raw-fallback", NULL},
-	{"*.img.caibx", "*", img_to_raw_handler}, /* fallback */
-	{"*.img", "*", img_to_raw_handler}, /* fallback */
-	{0}
+	{"image", "boot-raw-fallback", img_to_boot_raw_fallback_handler},
+	{NULL, NULL, NULL}
 };
```
jluebbe, line 2704: Some image types which were handled via the fallback to `img_to_raw_handler` are no longer handled:

- `img-caibx` to `jffs2`
- `img-caibx` to `nand` (should be explicitly unsupported, but was mapped to the `img_to_raw_handler` anyway)
- `img-caibx` to `nor`
- `img-caibx` to `vfat`
- `img` to `jffs2`

As this commit shouldn't change behavior, they should now be mapped explicitly (still to `img_to_raw_handler`). This inconsistency can then be fixed in a separate PR.

To make sure that we don't change which combinations are supported, you should first add cases which call `test_get_update_handler()` for these combinations (in a separate PR).

docs/reference.rst
```diff
@@ -923,12 +923,26 @@ The following fields are supported for image sections:
 ``filename`` (required)
   Name of the image file (relative to bundle content).
 
+  .. note::
+    While RAUC can automatically detect the image type from filename extensions
+    for backward compatibility, it is recommended to explicitly specify the
+    ``type`` field for new configurations.
```
```suggestion
    ``type`` field for new projects or in cases where all target systems are running a new enough version.
```

docs/reference.rst
```diff
@@ -923,12 +923,26 @@ The following fields are supported for image sections:
...
     RAUC uses the file name extension and the slot type to decide how to
     extract the image file content to the slot.
     Make sure to only use :ref:`supported file name extensions
     <sec-ref-supported-image-types>`!
 
+``type`` (optional, recommended)
+  Specifies the type of the image content.
+  This determines how RAUC will process the image during installation.
+  If not specified, the type is automatically detected
+  based on the filename extension and target slot type.
```
```suggestion
  If not specified, the image type is automatically detected
  based on the filename extension.

  This was added in RAUC v1.15.
```

docs/reference.rst
```diff
@@ -1027,36 +1041,67 @@ In future releases, they will be accessible in other hooks, as well.
...
-To keep the number of combinations manageable, only a limited amount of file
-name extensions is supported by RAUC.
+Supported Image Types
+~~~~~~~~~~~~~~~~~~~~~
 
-The ``*.img`` extension is valid for all slot types.
-It should be used when no specific file name extension (and handler) is
-supported.
+.. note::
+  Newer versions of RAUC support setting an explicit ``type`` field
```
```suggestion
  RAUC v1.15 or newer support setting an explicit ``type`` field
```

docs/reference.rst
```diff
@@ -1027,36 +1041,67 @@ In future releases, they will be accessible in other hooks, as well.
...
+
+  * ``caidx``: casync directory tree index
+  * ``ext4-caibx``: ext4 file system image in casync blob index format
+  * ``vfat-caibx``: vfat file system image in casync blob index format
+  * ``ubifs-caibx``: UBIFS file system image in casync blob index format
+  * ``squashfs-caibx``: SquashFS image in casync blob index format
+  * ``img-caibx``: Generic image in casync blob index format
+  * ``catar``: casync directory tree archive
+
+The following image file name extensions can be auto-mapped by RAUC:
```
```suggestion
If no image `type` is given, the following file name extensions are be mapped automatically by RAUC:
```

docs/reference.rst
jluebbe: Ah, then perhaps something like `vfat file system image (sometimes used for the EFI System Partition)`?

src/update_handler.c
dwntri: added #1788 and cherry picked it into here for now

docs/reference.rst
dwntri: okay, rephrased it

zentax-dev:
This was just two minor spelling / naming changes

src/update_handler.c
jluebbe: I'll fix two small typos myself.

src/update_handler.c
jluebbe: I've mentioned this explicitly in the commit message.

PR #1760, opened 2025-08-01
refactor test/test_resign and check signature contents

jluebbe:
With the upcoming support for multiple signatures, we'll need to inspect
the generated CMS structures more closely. As there doesn't seem to be a
easy-to-use Python module for this and parsing 'openssl cms' output is
too fragile, we build a minimal decoder on top of pyasn1.

### DIFF
--- a/test/test_resign.py
+++ b/test/test_resign.py
@@ -189,6 +189,11 @@ def test_resign_extend_expired_no_verify(tmp_path):
     assert len(cms["signerInfos"]) == 1
     assert cms["signerInfos"][0]["issuer"] == SIGNERS["release-2018"][0]
     assert cms["signerInfos"][0]["serial"] == SIGNERS["release-2018"][1]
+    signing_time = cms["signerInfos"][0]["signedAttrs"]["signingTime"]
+    assert signing_time[0] == "utcTime"
+    assert signing_time[1].year == 2018
+    assert signing_time[1].month == 1
+    assert signing_time[1].day == 1
 
     out2_bundle = tmp_path / "out2.raucb"
 
@@ -208,3 +213,8 @@ def test_resign_extend_expired_no_verify(tmp_path):
     assert len(cms["signerInfos"]) == 1
     assert cms["signerInfos"][0]["issuer"] == SIGNERS["release-1"][0]
     assert cms["signerInfos"][0]["serial"] == SIGNERS["release-1"][1]
+    signing_time = cms["signerInfos"][0]["signedAttrs"]["signingTime"]
+    assert signing_time[0] == "utcTime"
+    assert signing_time[1].year == 2020
+    assert signing_time[1].month == 10
+    assert signing_time[1].day == 1


test/decode_cms.py
```diff
@@ -0,0 +1,166 @@
...
+        "contentType": resolve_oid(content_type),
+        **result,
+    }
+
+
+if __name__ == "__main__":
+    import sys
+    from pprint import pprint
+
+    der_data = open(sys.argv[1], "rb").read()
```
ejoerns: We should have minimal usage information somewhere if we fail that implicitly for a missing argument.
Ignoring error handling for open might be acceptable for a helper script.

test/decode_cms.py
```diff
@@ -0,0 +1,166 @@
...
+        else:
+            issuer = None
+            serial = None
+        digest_algorithm = signer_info.getComponentByName("digestAlgorithm")
+        sig_algorithm = signer_info.getComponentByName("signatureAlgorithm")
+        signature = signer_info.getComponentByName("signature")
+        result["signerInfos"].append(
+            {
+                "issuer": name_to_string(issuer),
+                "serial": int(serial),
```
ejoerns, line 174: `serial` can be `None`, but an `int(None)` will fail

test/decode_cms.py
```diff
@@ -0,0 +1,166 @@
...
+            return str(decoded)
+        except Exception:
+            continue
+    # fallback: return hex
+    return value_any.prettyPrint()
+
+
+def name_to_string(name):
+    parts = []
+    seq = name.getComponentByName("rdnSequence")
```
ejoerns, line 42: Actually a `CHOICE`, but simplification seems appropriate here since it never gained more than a single member.

test/test_resign.py
```diff
@@ -7,79 +6,85 @@
...
         " --cert openssl-ca/dev/autobuilder-1.cert.pem"
         " --key openssl-ca/dev/private/autobuilder-1.pem"
         " --keyring openssl-ca/rel-ca.pem"
-        f" resign {tmp_path}/good-bundle.raucb {tmp_path}/out.raucb"
+        f" resign {in_bundle} {out_bundle}"
         " --signing-keyring openssl-ca/dev-only-ca.pem"
     )
-
     assert exitcode == 0
+    assert out_bundle.exists()
```
ejoerns, line 58: A simple hint in the commit message that the refactoring is based on the fact that pytest's `tmp_path` fixture already is a `pathlib.Path` object would make it easier for others to follow.

Also slightly mixes variable declarations, internal path renames, and changes of `os.path.exists()` to `pathlib.Path. exists()` which isn't ideal but acceptable since it avoids touching some lines twice.
Some notes in the commit message might make it easier to get this, too.

test/test_resign.py
```diff
@@ -1,85 +1,125 @@
...
+
+
+def get_cms(bundle_path):
+    with open(bundle_path, "rb") as bundle_file:
+        bundle_file.seek(-8, 2)
+        cms_length_bytes = bundle_file.read(8)
+        cms_length = int.from_bytes(cms_length_bytes, "big")
+        cms_start = bundle_file.tell() - cms_length - 8
+        bundle_file.seek(cms_start)
+        cms_data = bundle_file.read(cms_length)
```
ejoerns, line 27: Maybe with some guidance:

```suggestion
        # read CMS/signature size from bundle (8 bytes)
        bundle_file.seek(-8, 2)
        cms_length_bytes = bundle_file.read(8)
        # calculate CMS offset and length
        cms_length = int.from_bytes(cms_length_bytes, "big")
        cms_start = bundle_file.tell() - cms_length - 8
        # read CMS data from bundle file
        bundle_file.seek(cms_start)
        cms_data = bundle_file.read(cms_length)
```

test/test_resign.py
```diff
@@ -2,12 +2,44 @@
...
+
+    return decode_cms(cms_data)
+
+
+def get_signers(bundle_path):
+    cms = get_cms(bundle_path)
+    assert cms["contentType"] == "signedData"
+
+    cms_signers = {(si["issuer"], si["serial"]) for si in cms["signerInfos"]}
```
ejoerns, line 37: Maybe add a short note that this already prepares for multiple signers? Since in `SIGNERS` (and in general) we only have single signers and the `{}` wrapping looks a bit artificial otherwise.

test/decode_cms.py
```diff
@@ -0,0 +1,217 @@
...
+
+            oid_str = str(type_oid)
+            name = oid_map.get(oid_str, oid_str)
+
+            parts.append(f"{name}={value_str}")
+
+    return ", ".join(parts)
+
+
+def utctime_to_datetime(utctime):
```
ejoerns, line 64: Maybe add a note that this decodes `YYMMDDHHMMSSZ` and the specialty of `YY` to ease following the code, e.g.

```suggestion
def utctime_to_datetime(utctime):
    """
    Decode an ASN.1 UTCTime string in YYMMDDHHMMSSZ format to a datetime object.
    Assumes:
      - YY in [00,49] => 2000..2049
      - YY in [50,99] => 1950..1999 (ASN.1 UTCTime rule)
    """
```

test/decode_cms.py
```diff
@@ -1,9 +1,11 @@
 #!/usr/bin/env python3
 
 import inspect
+from datetime import datetime
+
 from pyasn1.codec.der import decoder
 from pyasn1.type import univ, char
-from pyasn1_modules import rfc2315, rfc5280, rfc5480, rfc5652
+from pyasn1_modules import rfc2315, rfc5280, rfc5480, rfc5652, rfc5751
```
ejoerns, line 8: A short line of motivation in the commit message (e.g., which information we intend to obtain here for which future checks) would help readers.

ejoerns (changes requested):
A few comments, but looks good otherwise! :+1:

test/decode_cms.py
jluebbe: Yes, but all our certificates have serials. If this triggers, we'll have a backtrace which points to this location. I don't know if explicit error handling would make this easier to read or debug.

ejoerns (approved):

PR #1761, opened 2025-08-01
docs: convert manual page to rst

jluebbe:
Convert the existing manual page (rauc.1) to rst using pandoc and some manual fixes. Also update the meson configuration to build it if the manpages option is enabled.


jluebbe:
The output is visible here: https://rauc--1761.org.readthedocs.build/en/1761/man.html

docs/meson.build
```diff
@@ -31,3 +32,13 @@ custom_target(
...
   install: get_option('htmldocs'),
 )
+
+custom_target(
+  'manpages',
+  output: ['rauc.1'],
+  depend_files: sources_doc,
+  command: [sphinx, '-b', 'man', meson.current_source_dir(), '@OUTDIR@'],
+  build_by_default: get_option('manpages'),
+  install: true,
```
ejoerns:
```suggestion
  install: get_option('manpages'),
```

Required as we otherwise attempt to install a file we did not generate.

meson_options.txt
```diff
@@ -80,6 +80,11 @@ option(
   type : 'boolean',
   value : false,
   description : 'Enable/Disable HTML documentation')
+option(
+  'manpages',
+  type : 'boolean',
+  value : false,
```
ejoerns, line 86: Combined with the necessary install change, this changes the default for installing man pages.
Thus, maybe we should enable it by default?
```suggestion
  value : true,
```

meson_options.txt
```diff
@@ -80,6 +80,11 @@ option(
...
+  description : 'Enable/Disable manual pages')
```
ejoerns:
```suggestion
  description : 'Enable/Disable building and installing manual pages')
```

ejoerns (changes requested):
The visual comparison between the hard-coded and the sphinx-generated man page looks quite similar.

Allowing to write man page in rst and embedding it in the documentation is really a real added value :+1: 

Just two notes about the installation handling. Otherwise it looks fine to me.

meson_options.txt
jluebbe: As building the manual page now requires sphinx, didn't want to enable it by default. Distributions can enable it explicitly, when needed.

docs/meson.build
jluebbe: I get `File 'docs/rauc.1' not found, skipping`, which seems non-fatal. I've changed it to make it cleaner.

meson_options.txt
ejoerns: Ok. But we should add an explicit hint in the release notes about that later.

Issue #1704, opened 2025-04-29
rauc info ignores --confopt=keyring:check-purpose=codesign


airtower-luna:
**Describe the bug**

I'm trying to read bundle metadata with verification on the host that *built* the bundle, not the target system, so there's no system RAUC config. The signer certificate has the "code signing" purpose set. Verification fails with "unsuitable certificate purpose". So far, this is #1427.

However, setting the expected purpose on the command line does not work either:
```
$ rauc info --confopt=keyring:check-purpose=codesign --keyring=test-ca/updates/ca-chain.pem --output-format=json-pretty images/upgrade.rauc
rauc-Message: 17:47:20.053: Reading bundle: /.../images/upgrade.rauc
rauc-Message: 17:47:20.054: Verifying bundle signature... 
signature verification failed: Verify error: unsuitable certificate purpose
```

Using `-c` instead with a config file that contains
```
[keyring]
check-purpose=codesign
```
works, so the bug is specifically in the override. I'd like to avoid creating the config file because it requires additional `system` section settings that are meaningless on the build host.

**Background information**
I'm using RAUC 1.14, from the Buildroot package (current Git `master`).

**To Reproduce**
1. Create a signer CA and certificate with the "code signing" key purpose.
2. Create an upgrade bundle signed with the certificate.
3. Try to parse the bundle with `rauc info`, without disabling signature verification, e.g.: `rauc info --confopt=keyring:check-purpose=codesign --keyring=ca.pem upgrade.rauc`

**Expected behavior**
Verification should succeed (assuming the certificate isn't invalid for other reasons).

ejoerns:
@airtower-luna The way `--confopt` is currently designed, it allows **overriding** settings from the config file (or **adding** new ones), but it only works when a config file is present since the handling is part of the config file parsing.

Thus, the solution of #1542 satisfies the use case described in https://github.com/rauc/rauc/issues/953#issuecomment-2346256501 but does not really satisfy the 'host' use case brought up in #729 or would still require a dummy `system.conf`, which I also find quite unexpected from the user experience.

I guess there is no good way to allow setting single `system.conf` parameters without providing at least a base system.conf, thus I would propose adding explicit parameters like `--check-purpose` for options that might be relevant for 'host' tool use cases. And maybe we should warn or error when using `--confopt` without having a config to patch.

Maybe @jluebbe has some opinions on this, too

airtower-luna:
So to confirm, the idea would be to:

[...]

PR #1763, opened 2025-08-08
config_file: implement default_config using parse_config

jluebbe:
This lets us apply overrides set using -C even when no explicit config file is given.
To avoid printing confusing warnings for CLI commands where no is usually used, make them conditional on having an actual config file.

To make this PR easier to review, I've split out parsing of the different `system.conf` sections to separate functions.
In the future, we should simplify the parsing with better helper functions. Also we should move the default and warning handling out of the parsing functions and into dedicated steps which only need to know if a given value was set explicitly.

Fixes: #1704


airtower-luna:
I've tested this PR fixes the issue described in #1704. Using `rauc info` and with `--keyring={keyring} --confopt=keyring:check-purpose=codesign --confopt=keyring:allow-partial-chain=true` worked fine, as well as `rauc bundle` with `--confopt=keyring:path={keyring} --confopt=keyring:check-purpose=codesign --confopt=keyring:allow-partial-chain=true`. Update tests on my target systems (using a verity bundle, U-Boot integration, post install handler, marking a partition as good or bad) passed without issues as well. Thank you!

include/config_file.h
```diff
@@ -144,11 +144,15 @@ gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
...
 
 /**
- * Creates a default rauc system configuration.
+ * Creates a default RAUC system configuration for usage on the host.
+ *
+ * As this processes config overrides from the command line, errors can be
+ * returned.
  *
  * @param config a location to place the new config
+ * @param error a GError, or NULL
```
ejoerns, line 153: Please also document the new return value here.

src/config_file.c
```diff
@@ -1317,36 +1317,22 @@ void r_config_file_modified_check(void)
 	}
 }
 
-gboolean load_config(const gchar *filename, RaucConfig **config, GError **error)
+static RaucConfig *parse_config(const gchar *filename, const gchar *data, gsize length, GError **error)
```
ejoerns, line 1314: Since (in contrast to `load_config()`) for `parse_config()`, the `fiename` parameter is intentionally optional, it might make sense to note this. Either in the commit message or by providing a function documentation.

src/config_file.c
```diff
@@ -623,8 +605,9 @@ static gboolean parse_system_section(const gchar *filename, GKeyFile *key_file,
 		if (c->data_directory) {
 			c->statusfile_path = g_build_filename(c->data_directory, "central.raucs", NULL);
 		} else {
-			g_message("No data directory or status file set, falling back to per-slot status.\n"
-					"Consider setting 'data-directory=<path>' or 'statusfile=<path>/per-slot' explicitly.");
+			if (filename)
```
ejoerns, line 608: The commit message lacks a word, probably:

> To avoid printing confusing warnings for CLI commands where no **config file** is usually used, ..

src/config_file.c
```diff
@@ -687,6 +687,91 @@ static gboolean parse_system_section(const gchar *filename, GKeyFile *key_file,
...
+static gboolean parse_keyring_section(const gchar *filename, GKeyFile *key_file, RaucConfig *c, GError **error)
+{
+	GError *ierror = NULL;
+	gsize entries;
+
+	g_return_val_if_fail(key_file, FALSE);
+	g_return_val_if_fail(c, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	if (!g_key_file_has_group(key_file, "keyring"))
```
ejoerns, line 683: This optimization is new (and not part of the move) and should at least be noted in the first of these commits.

src/config_file.c
```diff
@@ -805,6 +805,48 @@ static gboolean parse_casync_section(GKeyFile *key_file, RaucConfig *c, GError *
 	return TRUE;
 }
 
+static gboolean parse_streaming_section(GKeyFile *key_file, RaucConfig *c, GError **error)
```
ejoerns, line 792: Nit: In the commit title: s/separte/separate

src/config_file.c
```diff
@@ -1410,6 +1394,24 @@ static RaucConfig *parse_config(const gchar *filename, const gchar *data, gsize
...
 
+void default_config(RaucConfig **config)
+{
+	GError *ierror = NULL;
+
+	g_return_if_fail(config && *config == NULL);
+
+	/* Use an empty system section to honor defaults from parse_system_section.
+	 * After we have implemented explicit defaults there, we can use an empty
+	 * string here. */
```
ejoerns, line 1416: Taking the defaults for host tool usage from the `system` section, which I understand as 'explicitly for the target', feels a bit wrong for me.

If we actually look at the defaults we talk about, these are just four. From which

* one is just for the deprecated download mode (`max_bundle_download_size`)
* one is unused in host tools as far as I can see (`mount_prefix`)
* two are probably misplaced in the `system` section since they are no system properties (`max_bundle_signature_size`, `bundle_formats_mask`) 

So maybe we can find another way here in the future.

src/config_file.c
```diff
@@ -1410,6 +1394,24 @@ static RaucConfig *parse_config(const gchar *filename, const gchar *data, gsize
...
+	const gchar *data = "[system]";
+	RaucConfig *c = parse_config(NULL, data, strlen(data), &ierror);
```
ejoerns, line 1418: In contrast to the original 'defaults', this also sets `statusfile` to `per-slot`. But that should be fine for now.

src/config_file.c
jluebbe: Yes, but I suspect moving them somewhere else now would only create more confusion than leaving them as-is...

ejoerns: Yes. I was only unsure if we should suggest it in the comment, though. Anyway, we might remember we can do better here when touching this again.

ejoerns (approved):

PR #1766, opened 2025-08-11
run test suite for various distributions

jluebbe:
Instead of running the test suite only on the Debian release used to build our CI container, also run them on the different distros and releases in the stable matrix. I'll build on this with a patched OpenSSL to test the multiple signature handling next.

Also fix some minor issues found while preparing this PR.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -140,8 +140,18 @@ jobs:
         - "ubuntu:22.04"
         - "ubuntu:24.04"
         - "debian:bullseye"
+        - "debian:bookworm"
         - "debian:trixie"
         - "debian:testing"
+        include:
+        - release: "ubuntu:24.04"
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -140,8 +140,18 @@ jobs:
         - "ubuntu:22.04"
         - "ubuntu:24.04"
         - "debian:bullseye"
+        - "debian:bookworm"
         - "debian:trixie"
         - "debian:testing"
+        include:
```
ejoerns, line 146: Maybe add a one-line description in the commit message why we run only a selection of stable builds.

jluebbe: I've added this to https://github.com/rauc/rauc/pull/1767.

ejoerns (approved):

PR #1768, opened 2025-08-12
.github/workflows/tests: don't try to print old logfile locations

jluebbe:
These log file locations are no longer valid, so remove them.


zentax-dev (approved):
I guess we do not need non-existing log file locations, so that looks fine.

PR #1779, opened 2025-08-26
Feature: OpenSSL providers support

jussike:
This change introduces new build option: pkcs11_providers
When the option is enabled (instead of pkcs11_engine), Rauc uses newer Openssl providers API for PKCS#11 operations (requires Openssl >= 3) instead of legacy Openssl engine API which is already deprecated. Engine API will likely be removed on next major Openssl 4.

https://openssl-communities.org/d/rxYpMNVn/api-deprecation-announcement-for-openssl4-0

This change also updates the test environment so that tests can also be run with the new option.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -823,12 +823,12 @@ RaucContext *r_context_conf(void)
 		// let us handle broken pipes explicitly
 		signal(SIGPIPE, SIG_IGN);
 
-		if (!network_init(&ierror)) {
+		if (!signature_init(&ierror)) {
 			g_warning("%s", ierror->message);
 			g_error_free(ierror);
 			return NULL;
 		}
-		if (!signature_init(&ierror)) {
+		if (!network_init(&ierror)) {
 			g_warning("%s", ierror->message);
 			g_error_free(ierror);
 			return NULL;
--- a/src/signature.c
+++ b/src/signature.c
@@ -8,6 +8,9 @@
 #if ENABLE_OPENSSL_PKCS11_ENGINE
 #include <openssl/engine.h>
 #endif
+#if ENABLE_OPENSSL_PKCS11_PROVIDERS
+#include <openssl/store.h>
+#endif
 #include <openssl/x509.h>
 #include <string.h>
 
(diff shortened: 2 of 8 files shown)


test/Dockerfile
```diff
@@ -82,6 +89,16 @@ RUN test "$(uname -m)" != "x86_64" && exit 0; \
...
+# Install the same libsofthsm2 version with important segfault fixes (Ubuntu repo)
+WORKDIR /tmp
+RUN test "$(uname -m)" = "armv7l" && exit 0; \
+    wget https://archive.ubuntu.com/ubuntu/project/ubuntu-archive-keyring.gpg -O /etc/apt/trusted.gpg.d/ubuntu.gpg && \
+    echo "deb-src http://archive.ubuntu.com/ubuntu noble universe" > /etc/apt/sources.list.d/ubuntu-src.list && \
+    apt-get update -q && apt-get source libsofthsm2 -q && apt-get build-dep libsofthsm2 -y && \
+    chown -R _apt:root softhsm2-2* && cd softhsm2-2* && dpkg-buildpackage -us -uc && cd - && \
+    dpkg -i libsofthsm2_*deb
+WORKDIR /
+
```
jluebbe: This should not be needed if you add `pkcs11-module-quirks = no-deinit` to `openssl-pkcs11.cnf`. See https://github.com/latchset/pkcs11-provider/blob/main/tests/softhsm-init.sh#L64

jluebbe (commented):
Thanks for your PR!

Do you know if there is a clean way to support the existing [environment variables](https://rauc.readthedocs.io/en/latest/man.html#environment) (`RAUC_PKCS11_MODULE` and `RAUC_PKCS11_PIN`)?

test/Dockerfile
jussike: Thanks, that is much more cleaner.

src/signature.c
```diff
@@ -98,6 +98,15 @@ gboolean signature_init(GError **error)
 {
 	int ret, id;
 
+#if ENABLE_OPENSSL_PKCS11_PROVIDERS
+	const gchar *env;
+
+	env = g_getenv("RAUC_PKCS11_MODULE");
+	if (env != NULL && env[0] != '\0') {
+		g_setenv("PKCS11_PROVIDER_MODULE", env, TRUE);
+	}
```
jluebbe, line 107: Should we warn the user to switch to using `PKCS11_PROVIDER_MODULE` (or the openssl config) here?

Also, if `PKCS11_PROVIDER_MODULE` is already set, we shouldn't overwrite it and warn as well.

src/signature.c
```diff
@@ -266,7 +295,9 @@ static EVP_PKEY *load_key_pkcs11(const gchar *url, GError **error)
 	OSSL_STORE_CTX *store_ctx = NULL;
 	OSSL_STORE_INFO *info = NULL;
 
-	store_ctx = OSSL_STORE_open(url, NULL, NULL, NULL, NULL);
+	const gchar* url_with_pin = apply_env_rauc_pkcs11_pin(url);
```
jluebbe, line 298: Use `g_autofree` here and below as well, as we do for other temporary strings.

src/signature.c
```diff
@@ -235,6 +244,26 @@ static EVP_PKEY *load_key_file(const gchar *keyfile, GError **error)
...
 }
 
+#if ENABLE_OPENSSL_PKCS11_PROVIDERS
+static gchar* apply_env_rauc_pkcs11_pin(const gchar *url)
+{
+	const gchar *env;
+
+	env = g_getenv("RAUC_PKCS11_PIN");
+	if (env != NULL && env[0] != '\0' && strstr(url, "pin-value") == NULL) {
+		const gchar* pin_value = ";pin-value=";
```
jluebbe, line 254: From https://datatracker.ietf.org/doc/html/rfc7512#section-3 it seems to me that `pin-value` should be a query parameter (so after the '?' and separated by '&').

With this change, OpenSSL is called with a different URL as passed in by the user. The errors still contain the original URL, which might be a good idea to avoid logging the pin. 

Perhaps this code would be easier to follow with `g_string_append_printf()`?

src/signature.c
```diff
@@ -235,6 +244,26 @@ static EVP_PKEY *load_key_file(const gchar *keyfile, GError **error)
 	return res;
 }
 
+#if ENABLE_OPENSSL_PKCS11_PROVIDERS
```
jluebbe, line 247: You can use `G_GNUC_UNUSED` to avoid the `#if` and always have it compile-checked.

test/meson.build
```diff
@@ -44,6 +44,7 @@ extra_test_sources = files([
 
 test_env = environment()
 test_env.set('MESON_BUILD_DIR', meson.build_root())
+test_env.set('OPENSSL_CONF', join_paths(meson.build_root(), '..', 'test', 'openssl-pkcs11.cnf'))
```
jluebbe, line 47: This should be configured in `test/conftest.py`, as only the pytest-bases test-cases use PKCS#11 and can also be run without meson.

src/context.c
```diff
@@ -823,12 +823,12 @@ RaucContext *r_context_conf(void)
...
 		signal(SIGPIPE, SIG_IGN);
 
-		if (!network_init(&ierror)) {
+		if (!signature_init(&ierror)) {
 			g_warning("%s", ierror->message);
 			g_error_free(ierror);
 			return NULL;
 		}
-		if (!signature_init(&ierror)) {
+		if (!network_init(&ierror)) {
```
jluebbe, line 831: Why do you need to change the order here? Perhaps add a comment if the order is important.

src/signature.c
jluebbe: Alternatively, what do you think about using a UI method as in https://github.com/latchset/pkcs11-provider/blob/main/tests/ccerts.c#L14?

jluebbe (changes requested):

UVV-gh:
Is there anything else to do here apart from addressing the review comments?

jluebbe:
@UVV-gh Nothing else, it's "just" reworking the PR to address the feedback. :)

PR #1782, opened 2025-08-30
test/test_mark: add minimal check for calling barebox backend via mark

ejoerns:
So far, we have no test for invoking RAUC's barebox boot selection backend through 'mark' methods.

While this is not an ideal solution since it only leverages the 'barebox-state' dummy binary, it at least allows to test if mark operations would alter the barebox state variable.

We test 'mark-bad' and 'mark-primary' here, which is essentially what would happen during a normal update.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -81,6 +81,14 @@ def _have_grub():
 have_grub = pytest.mark.skipif(not _have_grub(), reason="Have no grub-editenv")
 
 
+def _have_barebox():
+    out, err, exitcode = run("which barebox-state")
+    return exitcode == 0
+
+
+have_barebox = pytest.mark.skipif(not _have_barebox(), reason="Have no barebox-state")
+
+
 def _have_openssl():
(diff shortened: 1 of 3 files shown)


test/conftest.py
```diff
@@ -81,6 +81,14 @@ def _have_grub():
 have_grub = pytest.mark.skipif(not _have_grub(), reason="Have no grub-editenv")
 
 
+def _have_barebox():
+    out, err, exitcode = run("which barebox-state")
+    return exitcode == 0
+
+
+have_barebox = pytest.mark.skipif(not _have_barebox(), reason="Have no barebox-state")
```
jluebbe: This would also find a real `barebox-state`, which wouldn't match what the tests expect. Check if we're running under QEMU instead.

ejoerns: Admittedly, this originates from my initial optimism to potentially use barebox-state for testing :wink: Checking for our specific qemu test env sounds reasonable.

PR #1783, opened 2025-09-01
test/confest: make PEM loading compatible with OpenSC 0.26.0

zentax-dev:
OpenSC 0.26.0 includes https://github.com/OpenSC/OpenSC/pull/3174. As a result, OpenSC does not support reading the (DER-converted) object data from stdin anymore.

However, OpenSC/pkcs11-tool also supports reading PEM files directly, which can be used to replace the stdin piping. Instead of piping, the converted PEM file in the temp directory is stored in the temp directory and then read later on.

See: https://git.openembedded.org/meta-openembedded/commit/meta-oe/classes/signing.bbclass?id=2d1d128a41abb698874e2d0b8e59cb5ae0416937


test/conftest.py
```diff
@@ -159,29 +159,41 @@ def have_service():
 needs_composefs = pytest.mark.skipif(not string_in_config_h("ENABLE_COMPOSEFS 1"), reason="Missing composefs support")
 
 
-def softhsm2_load_key_pair(cert, privkey, label, id_, softhsm2_mod):
+def softhsm2_load_key_pair(cert, privkey, label, id_, softhsm2_mod, tmp_path=None):
```
ejoerns: Do we have a case where `tmp_path` is unset? Otherwise, I'd drop this case.

ejoerns (commented):
In the commit message

```
However, OpenSC/pkcs11-tool also supports reading PEM files directly, which
can be used to replace the stdin piping. Instead of piping, the converted PEM file
in the temp directory is stored in the temp directory and then read later on.
```

it might make sense to clarify why even if reading from PEM is supported, we need to convert/write the temporary PEM file.

test/conftest.py
dwntri: The `tmp_path` originates from
https://github.com/rauc/rauc/pull/1783/files#diff-7dae78a00d8d69422a3bfbde96c7c5d794eede504a0df763853896a1ba1b5ff2R322
And I wasn't completely sure, if that could result in a failure.

ejoerns: You mean if the `tmp_path` isn't created correctly? This should result in a failure, yes.
Thus, I would not introduce fallback handling.
Also note that other usages of `tmp_path` happen unconditionally, too.

test/conftest.py
```diff
@@ -159,29 +159,38 @@ def have_service():
...
-        f"openssl rsa -in {privkey} -inform pem -outform der "
-        f"| pkcs11-tool --module {softhsm2_mod} -l --pin 1111 -y privkey -w /proc/self/fd/0 "
-        f"--label {label} --id {id_}",
+        f"pkcs11-tool --module {softhsm2_mod} -l --pin 1111 -y privkey -w {privkey} --label {label} --id {id_}",
         shell=True,
     )
     assert proc.returncode == 0
 
+    try:
+        pubkey_file.unlink()
```
ejoerns: How can `pubkey_file` not exist? And should we ignore this?

ejoerns: Maybe this can be dropped anyway for now since `tmp_path`'s session scope should be sufficient here.

PR #1791, opened 2025-09-16
src/mark: document function mark_run()

gportay:
This documents the function mark_run() as every other functions in that header.

### DIFF
--- a/include/mark.h
+++ b/include/mark.h
@@ -43,6 +43,19 @@ G_GNUC_WARN_UNUSED_RESULT;
 gboolean r_mark_bad(RaucSlot *slot, GError **error)
 G_GNUC_WARN_UNUSED_RESULT;
 
+/**
+ * Mark a bootname slot.
+ *
+ * @param state String representation of slot state
+ *              (e.g. "good", "bad", or "active")
+ * @param slot_identifier String representation of slot identifier
+ *                        (e.g. "booted", "other", or the slot name)
+ * @param slot_name Return location for the resulting slot name
+ * @param message Return location for message describing what has been done
+ *                (e.g. "marked slot(s) rootfs.0 as good")
+ *
+ * @return Return TRUE on success and FALSE on error
+ */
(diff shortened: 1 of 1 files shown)


include/mark.h
```diff
@@ -43,6 +43,17 @@ G_GNUC_WARN_UNUSED_RESULT;
 gboolean r_mark_bad(RaucSlot *slot, GError **error)
 G_GNUC_WARN_UNUSED_RESULT;
 
+/**
+ * Mark a bootname slot.
+ *
+ * @param state Slot to mark as bad
```
ejoerns: This is not the slot, but the 'state' the slot should be marked as, i.e. one of `"good"`, `"bad"`, or `"active"`.

include/mark.h
```diff
@@ -43,6 +43,17 @@ G_GNUC_WARN_UNUSED_RESULT;
...
+ * @param slot_identifier Slot identifier
```
ejoerns: Could be described a bit more detailed (`.. like "booted", "other", or the slot's name`).

include/mark.h
```diff
@@ -43,6 +43,17 @@ G_GNUC_WARN_UNUSED_RESULT;
...
+ * @param slot_name Return slot name
```
ejoerns:
```suggestion
 * @param slot_name Return location for the resulting slot name
```

gportay: done.

include/mark.h
gportay: Oops, good catch. I have reworded as follow, tell me if this looks correct to you.
 
```
 * @param state String representation of slot state
 *              (e.g. "good", "bad", or "active")
```

include/mark.h
gportay: I have reworded as follow, tell if this looks correct to you:
```
 * @param slot_identifier String representation of slot identifier
 *                        (e.g. "booted", "other", or the slot name)
```

PR #1793, opened 2025-09-18
Fix documentation Advanced Topics page

gduvauchelle:
This PR fixes the "Advanced Topics" page of the documentation.
When describing `[log.install-log]` section of the system configuration file, the page mentions the `output-format` field.
The right field name is `format`.

See code here : https://github.com/rauc/rauc/blob/master/src/config_file.c#L199

And the reference page here : https://github.com/rauc/rauc/blob/master/docs/reference.rst?plain=1#L712

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -1684,14 +1684,14 @@ happened on the system, use::
   [log.install-log]
   filename=install.log
   events=install
-  output-format=short
+  format=short
 
 Or, if you want a json-based log of all events, limited to 1M per log file and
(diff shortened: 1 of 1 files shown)


ejoerns:
@gduvauchelle Thank you for fixing the event log documentation! This was actually wrong.

Could you still add your [Signed-off-by](https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin) (and probably prefix with the commit title with `docs/advanced: `) so we can merge this?

PR #1796, opened 2025-10-10
test/config_file: simplify config_file_invalid_items

jluebbe:
`pathname` uses `g_autofree`, so use `g_clear_error` for cleanup.
Also move the declaration of `unknown_key_cfg_file` to its use.

### DIFF
--- a/test/config_file.c
+++ b/test/config_file.c
@@ -217,14 +217,6 @@ mountprefix=/mnt/myrauc/\n\
 [unknown]\n\
 foo=bar\n\
 ";
-	const gchar *unknown_key_cfg_file = "\
-[system]\n\
-compatible=FooCorp Super BarBazzer\n\
-bootloader=barebox\n\
-mountprefix=/mnt/myrauc/\n\
-foo=bar\n\
-";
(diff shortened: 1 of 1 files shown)


test/config_file.c
```diff
@@ -233,16 +225,26 @@ foo=bar\n\
...
 	pathname = write_tmp_file(fixture->tmpdir, "unknown_key.conf", unknown_key_cfg_file, NULL);
 	g_assert_nonnull(pathname);
 
 	g_assert_false(load_config(pathname, &config, &ierror));
 	g_assert_error(ierror, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_PARSE);
 	g_assert_cmpstr(ierror->message, ==, "Invalid key 'foo' in group '[system]'");
 	g_clear_error(&ierror);
+
+	g_clear_pointer(&pathname, g_free);
+	g_clear_pointer(&config, free_config);
```
ejoerns: Why are these two lines necessary when using g_auto*?

Also, the commit doesn't mention touching `config`.

PR #1802, opened 2025-10-15
improve handling of bundles with multiple signatures

jluebbe:
The main use cases for multiple signatures on one message are:

* Simplify CA certificate roll-over by signing messages with keys in
  both the old and new hierarchies during a transition period.
  Recipients would have either the old or new CA installed (or
  temporarily even both). In either case, we'd like to accept any
  message which has at least one signature that can be verified.

* Support user-implemented verification policies which require
  signatures from multiple specific signers.

* Support user-implemented verification policies which require a minimum
  number of signatures from different signers under under a trusted CA.

On OpenSSL versions which support `CMS_VERIFY_PARTIAL` (which is
not yet merged), we can accept a bundle with multiple signatures even if
only one of them can be verified. We use the config option
`allow-single-signature=true` to enable this and keep the existing behavior
of requiring successful verification of all signatures as the default for
backwards compatibility.

This can be used to simplify a key rotation process: Instead of signing the
same bundle contents with the old and new keys separately and documenting
when to use each resulting bundle, a single bundle is enough.

In the future, we can build on the OpenSSL feature to actually require multiple signatures from different/specific signers.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -135,23 +135,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        release:
-        - "ubuntu:20.04"
-        - "ubuntu:22.04"
-        - "ubuntu:24.04"
-        - "debian:bullseye"
-        - "debian:bookworm"
-        - "debian:trixie"
-        - "debian:testing"
         include:
+        - release: "ubuntu:20.04"
+        - release: "ubuntu:22.04"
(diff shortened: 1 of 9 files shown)


jluebbe:
@ejoerns If you think it would simplify review, we could also merge the first three commits (appending additional signatures) in a separate PR.

jluebbe:
OpenSSL has merged my PR, so this feature should be part of their 4.1 release in October 2026.

PR #1805, opened 2025-10-16
refactor service/installer startup

jluebbe:
As a preparation for native polling, we need to refactor and
simplify a few things during service and installer startup.


ejoerns (approved):
:+1:

PR #1807, opened 2025-10-20
utils: add r_get_boottime for CLOCK_BOOTTIME

jluebbe:
Compared to CLOCK_MONOTONIC, this includes any time that the system is suspend, which will be useful for scheduling polling intervals in the future.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -10,6 +10,7 @@
 #include <string.h>
 #include <sys/ioctl.h>
 #include <unistd.h>
+#include <time.h>
 
 #include "utils.h"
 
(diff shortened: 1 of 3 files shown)


src/utils.c
```diff
@@ -1126,3 +1127,14 @@ gchar *r_regex_match_simple(const gchar *pattern, const gchar *string)
...
 	return NULL;
 }
+
+gint64 r_get_boottime(void)
+{
+	struct timespec ts;
+	gint result = clock_gettime(CLOCK_MONOTONIC, &ts);
+
+	if (result != 0)
+		g_error("RAUC needs CLOCK_BOOTTIME");
```
ejoerns: We could check errno here and use it for printing a helpful error.
Since getting the boot time should not fail, using `g_error()` seems ok.

src/utils.c
```diff
@@ -1126,3 +1127,14 @@ gchar *r_regex_match_simple(const gchar *pattern, const gchar *string)
 
 	return NULL;
 }
+
+gint64 r_get_boottime(void)
+{
+	struct timespec ts;
+	gint result = clock_gettime(CLOCK_MONOTONIC, &ts);
```
ejoerns:
```suggestion
	gint result = clock_gettime(CLOCK_BOOTTIME, &ts);
```

Issue #1522, opened 2024-09-13
Rauc install command fails with PKCS11 URL


RoboAsystent:
**Describe the bug**
rauc install from https server fails when pkcs11 url for private key is used,  local ssl certificate issue is reported in service logs

**Background information**
Yocto kirkstone with rauc 1.11.1
Atecc608 HSM module with libcryptoauth as provider

**To Reproduce**
1. Add [streaming] section into rauc system.conf :

[streaming]
sandbox-user=<user-name>
tls-cert=client.crt.pem
tls-key=pkcs11:token=00ABC;object=device
tls-ca=CA.pem

2. export RAUC_PKCS11_PIN=<pin>
3. systemctl restart rauc
4. rauc install https://localhost/bundle.raucb -d

**Expected behavior**
Bundle installation should finish with success

**Logs**
journal logs:

Sep 12 22:54:47  rauc-nbd[18047]: running as UID 0, GID 0
Sep 12 22:54:47 rauc-nbd[18047]: received: {'url': <'https://localhost/bundle.raucb'>, 'key': <'pkcs11:token=00ABC;object=device'>, 'info-headers': <@as []>, 'ca': <'CA.pem'>, 'cert': <'client.crt.pem'>}
Sep 12 22:54:47 rauc-nbd[18047]: configuring for URL: https://localhost/bundle.raucb
Sep 12 22:54:47 rauc-nbd[18047]: request failed: Problem with the local SSL certificate (retrying)
Sep 12 22:54:48 rauc-nbd[18047]: configuring for URL: https://localhost/bundle.raucb
Sep 12 22:54:48 rauc-nbd[18047]: request failed: Problem with the local SSL certificate (retrying)
Sep 12 23:34:44 rauc-nbd[21424]: request failed (no more retries)
Sep 12 23:34:44 rauc[21408]: received http error 0
Sep 12 23:34:44 rauc-nbd[21424]: exiting nbd server
Sep 12 23:34:44 rauc-nbd[21424]: nbd server failed with: finish_request failed, shutting down
Sep 12 23:34:44 rauc[21408]: nbd server stopping

rauc logs:

rauc-Message: 23:30:30.762: Debug log domains: 'rauc'
(rauc:21040): rauc-DEBUG: 23:30:30.765: install started
(rauc:21040): rauc-DEBUG: 23:30:30.765: input bundle: https://localhost/bundle.raucb
(rauc:21040): rauc-DEBUG: 23:30:30.776: Trying to contact rauc service
[...]

PR #1813, opened 2025-10-23
fix: set CURLOPT_SSLKEYTYPE when using curl and pkcs11 key

singds:
Set CURLOPT_SSLKEYTYPE when using curl and the key is a pkcs11 uri.
Now I'm setting only the "PROV" key type, but there should be the possibility to choose whether to set "ENG", to use legacy openssl engine interface, or "PROV" to use the new openssl provider interface.
From key url we can't tell so maybe we need to add some build flag check.

Fixes: #1522.

### DIFF
--- a/src/nbd.c
+++ b/src/nbd.c
@@ -465,8 +465,16 @@ static void prepare_curl(struct RaucNBDTransfer *xfer)
 	code |= curl_easy_setopt(xfer->easy, CURLOPT_URL, xfer->ctx->url);
 	if (xfer->ctx->tls_cert)
 		code |= curl_easy_setopt(xfer->easy, CURLOPT_SSLCERT, xfer->ctx->tls_cert);
-	if (xfer->ctx->tls_key)
+	if (xfer->ctx->tls_key) {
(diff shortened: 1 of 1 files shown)


singds:
> PR #1779 introduces the build option for provider support, so we could select the `CURLOPT_SSLKEYTYPE` based on that. That would mean that you couldn't use engine keys for bundle decryption but provider keys for curl (or the other way around). I'd hope nobody does this, though...

Seems reasonable. I'll fix that.

> Do you use a PIN for you key? If so, how to you pass it to the provider?

The pkcs11 module I'm using now ignores the PIN. But using libp11 you can set the pin in openssl.conf.

singds:
I'm not sure what should I do to fix the failing check.

jluebbe:
I've force-pushed to add a description to the commit message and to adjust the title to our conventions.

The check fails because we don't have a test case for pkcs11 keys yet, so that fine.

jluebbe (approved):

jluebbe (approved):

Issue #342, opened 2018-10-10
Support for per-installation post-install hook


ejoerns:
Discussion in #275 revealed that there is also a need for a post-install *install hook* comparable to what we have as a *handler* already.

This hook could perform operations *after* all slots were written successfully.

Issue #275, opened 2018-05-29
Handler/Hook after all slots are successfully updated


embeddedrocks:
Hi,

I'm using your updater since a while now and are very pleased with it. Please continue the great work.

I just came across an issue which is either not implemented or I just overlooked it.

The idea is to either have a handler or hook after all slots have been successfully updated. 
You offer a post-install handler within the system.conf file. Which is indeed executed after an update has been successfully installed. But it is still executed from the *old* system. So my *old* system need to know about things I would like to do in the future.
The slot hooks won't work either as I need to be sure all slots have been updated. Which is not the case if e.g. I use a hook in _appfs_ slot but don't know if _rootfs_ has already been updated.

The plan is to copy some files from the *new* _appfs_ into the *new* _rootfs_. Unfortunately I have to do it this way as I cannot modify the _rootfs_ during build time.

Many Thanks.

ejoerns:
Maybe this is something that should be handled in the newly booting system instead of during updating?

If you are using systemd, you might want to have a look at

https://www.freedesktop.org/software/systemd/man/systemd.offline-updates.html

or 

https://www.freedesktop.org/software/systemd/man/systemd-firstboot.html

Would that help in your case?

embeddedrocks:
Many thanks for your answer.

Unfortunately it doesn't suite my case, as it assumes a modifyable _rootfs_ system during build time. Otherwise I'm not able to add a systemd services, which could be done during the update by copying files to the _new rootfs_ and further execute _systemctl_ with the _--root_ option. 

When using the slot hook mechanism, is there a chance to validate that e.g. _slot A_ is already updated when _slot B_ is executed. Or in other words, are slots always updated in a fixed order?

ejoerns:
If I get it right, you have a system where one has control over the appfs image but not over the rootfs image, right?
Maybe one could prepare the rootfs to have a mechanism to copy the required data from its appfs when being booted and create the trigger not in the rootfs itself but in a rootfs post-install hook. This one could add for example the mentioned `system-install` symlink without the need to actually touch the rootfs image itself.

> When using the slot hook mechanism, is there a chance to validate that e.g. slot A is already updated when slot B is executed. Or in other words, are slots always updated in a fixed order?

Up to now, there is no guarantee about that. It will have a fixed order ATM, but this is only defined by the code and the way it uses the data containers. It might change if another handling is implemented thus it is unsafe ATM to rely on a specific order.
[...]

PR #1820, opened 2025-11-04
Support bundle-defined pre/post-install handlers

yosuke-tanimoto:
This PR implements support for defining pre/post-install handlers in a bundle’s manifest file. 

Currently, RAUC supports pre/post-install handlers only via system.conf defined in the old system. 
However, the old system must predefine any actions that should be executed in the future. 

With this PR, actions can be defined using pre/post-install handlers specified in the bundle itself, 
allowing actions to be executed before/after all slots are updated. This enables these handlers to be defined 
per installation, so the old system no longer needs to predefine future actions in its handlers. 

The need for per-installation post-install hook was previously discussed in #275 and #342 
(as a global post-install hook), but it had not been implemented until now.
This PR introduces that functionality.

## Example

In the manifest file, these handlers are defined as `preinstall` and `postinstall` entries in the `[handler]` section.

```
[handler]
pre-install=preinstall-handler.sh
post-install=postinstall-handler.sh
```

Handlers are executed in the following order:

1. System-defined pre-install handler (defined in `system.conf`)
2. **Bundle-defined pre-install handler (defined in the manifest file)**
3. Installation process for each slot
4. **Bundle-defined post-install handler**
5. System-defined post-install handler

## Tests

Both pytest and gtest passed successfully.  
For this implementation, I added tests for bundle-defined pre/post-install handlers similar 
to the existing tests for system-defined ones.  
These include tests for manifest file parsing and handler execution (`pytest-install`).

## Questions

There are two points I would like to discuss regarding this implementation.

### Question 1: Is it appropriate to define these handlers in the `[handler]` section?

The `[handler]` section of the manifest file was originally intended for defining full-custom handlers.  
Currently, a full-custom handler specifies the handler file path via `filename =` and its arguments via `args =`.  
However, adding new pre/post-install handler options to the same section might cause naming confusion.

I currently see three options:

1. Define them in a separate subsection, such as `[handler.preinstall]` or `[handler.postinstall]`.  
2. Rename the full-custom handler option to something else, such as `full-custom =`.  
3. Leave the implementation as is.

Which approach do you think is best? Or is there a better way?
I would appreciate your feedback.

### Question 2: Which private key should be used to create the test bundle?

Since I added new handlers, I also implemented tests for them.  
To do this, I defined these handlers in the manifest of `test/good-adaptive-meta-bundle.raucb` 
and used them in the installation test (`test/test_install.py`).  
Currently, I’m using the following private key to recreate the bundle:

- cert: `openssl-ca/dev/autobuilder-1.cert.pem`  
- key: `openssl-ca/dev/private/autobuilder-1.pem`  
- keyring: `openssl-ca/dev-ca.pem`

Please let me know if there is a more suitable key to use.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -853,20 +853,28 @@ The hooks section allows to provide a user-defined executable for
 ``[handler]`` Section
 ~~~~~~~~~~~~~~~~~~~~~
 
-The ``handler`` section refers to the
+The ``handler`` section in the bundle manifest allows you to customize the update process for each bundle.  
+It can be used either to specify a 
 `full custom handler <https://rauc.readthedocs.io/en/latest/using.html#full-custom-update>`_
-that allows to fully replace the default RAUC update process.
+that completely replaces the default RAUC update process,  
+or to define 
+`bundle-defined pre/post-install handlers <https://rauc.readthedocs.io/en/latest/using.html#pre-post-install-handlers>`_
+that execute custom scripts before or after the standard slot update steps.
 
 .. note:: This is not to be confused with the ``[handlers]`` section from the
    system.conf which defines e.g. pre- and post-install handlers!
+
+.. rubric:: Full Custom Update
 
 When the full custom handler is enabled in a bundle, it will be invoked during
 the bundle installation
 
 * **after** bundle signature verification
 * **after** slot state and target slots determination logic
-* **after** the ``pre-install`` system handler
-* **before** the ``post-install`` system handler
+* **after** the system's ``pre-install`` handler
+* **after** the bundle's ``pre-install`` handler
+* **before** the system's ``post-install`` handler
+* **before** the bundle's ``post-install`` handler
 
 Also, the bundle will be mounted at this point and thus all its content is
 available to the full custom handler.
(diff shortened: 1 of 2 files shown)


ejoerns:
@yosuke-tanimoto Thank you for the contribution!

Why did you decide against implementing this as an [install hook](https://rauc.readthedocs.io/en/latest/using.html#install-hooks)?

yosuke-tanimoto:
@ejoerns Thank you for your feedback.

I actually overlooked the (global) install hook...
My main intention was to enable defining pre/post-install handlers—similar to those already available in system.conf—in the bundle manifest.   
I am not entirely sure about the exact difference between a hook and a handler in the context of a bundle.  
In this case, which would be more appropriate?

ejoerns:
> I am not entirely sure about the exact difference between a hook and a handler in the context of a bundle.

The difference is:

> Unlike handlers, hooks are defined in the update bundle and must be specified in the bundle’s [Manifest](https://rauc.readthedocs.io/en/latest/reference.html#sec-ref-manifest) file. All hooks are handled by a common executable that must be included in the bundle. Hooks allow the author of a bundle to add or replace functionality for the installation of a specific bundle. This can be useful for performing additional migration steps, checking for specific previously installed bundle versions or for manually handling updates of images RAUC cannot handle natively.

(from https://rauc.readthedocs.io/en/latest/using.html#bundle-based-customization-hooks:)

> If implemented as hooks, the bundle-defined post-install handler would overlap in timing with the install-check hook and therefore become unnecessary.

That doesn't make sense.

> Implementing this as install hooks could confuse users: the script format would differ from the existing handler format in system.conf (e.g., install hooks require handling hook names passed as arguments). This inconsistency would increase the learning curve.

Defining all bundle-based customization as hooks and all system-based customization as 'handler' should not confuse users. Mixing it will.

> In this case, which would be more appropriate?

Implementing it as a hook is the way to go.

PR #1824, opened 2025-11-06
CHANGES: prepare for 1.15 release

jluebbe:


CHANGES
```diff
@@ -3,6 +3,13 @@ Release 1.15 (development)
 
 .. rubric:: Enhancements
 
+* Add an optional ``type=`` to images in the bunde manifest.
+  This can used to define the type explicitly, which is useful if the file name
```
ejoerns:
```suggestion
  This can be used to define the type explicitly, which is useful if the file name
```

OnkelUlla:
```suggestion
  This can be used to define the type explicitly, which is useful if the file name
```

CHANGES
```diff
@@ -3,6 +3,13 @@ Release 1.15 (development)
...
+  extension is not known to RAUC.
+  See the :ref:`documentation for details <sec-ref-supported-image-types>`.
+  (by Lars Schmidt in :pr:`1756`)
+* Add support for an ``emptyfs`` image type, which can be used instead of a
+  ``tar`` when installing into an ``ext4`` slot. (by Lars Schmidt in :pr:`1759`)
```
ejoerns: Not sure if this suggests that `emtpyfs` is an alternative to `tar` for the same purpose.

CHANGES
```diff
@@ -11,63 +18,109 @@ Release 1.15 (development)
...
-  generation.
+  generation. (in :pr:`1734`)
 * Support a ``size-limit`` configuration option for the ``boot-emmc`` slot
   type.
   This can be used in cases where unrelated data on the eMMC boot partitions
   must be preserved and migrating it to a dedicated location is not feasible.
   Also ensure that the image fits into the slot before attempting to write it.
+  (in :pr:`1740`)
+* Improve handling of A/B/C setups by selecting the slot group with the oldest
+  installation date. (in :pr:`1525`)
```
ejoerns, line 47: Should we mention that this also supports bootstrapping both slots' content from an external/factory system?

CHANGES
```diff
@@ -11,63 +18,109 @@ Release 1.15 (development)
...
+* Handle config overrides set via ``-C``/``--confopt`` even if no config file is
+  available. (in :pr:`1763`)
+* Implement support for appending signatures with resign.
```
ejoerns: Maybe

```suggestion
* Implement support for appending signatures to existing bundles with ``rauc resign``.
```

CHANGES
```diff
@@ -3,6 +3,13 @@ Release 1.15 (development)
 
 .. rubric:: Enhancements
 
+* Add an optional ``type=`` to images in the bunde manifest.
```
dwntri:
```suggestion
* Add an optional ``type=`` to images in the bundle manifest.
```

CHANGES
ejoerns: Decided that this is more a pleasant side-effect, but not something we should recommend.

PR #1826, opened 2025-11-09
new d-bus property for the last transaction and bundle

js731ca:
The Rauc service already exposes properties for `Operation`, `Progress`  and `LastError`.

But there is (currently) no way to correlate these to an input file/bundle.

This PR would add two additional properties to fill that gap:
```
    <!-- LastBundle: Holds the path and filename to the last bundle rauc consumed -->
    <property name="LastBundle" type="s" access="read"/>
    <!-- LastTransaction: the last (install) transaction Rauc performed -->
    <property name="LastTransaction" type="s" access="read"/>
```

This would come in handy for scenarios where multiple parties interact with rauc: one application feeding it a bundle, another monitoring its operation state and errors, ...

### DIFF
--- a/src/de.pengutronix.rauc.Installer.xml
+++ b/src/de.pengutronix.rauc.Installer.xml
@@ -93,6 +93,8 @@
 
     <!-- Operation: Represents the current (global) operation rauc performs -->
     <property name="Operation" type="s" access="read"/>
+    <!-- LastBundle: Holds the path and filename to the last bundle rauc consumed -->
+    <property name="LastBundle" type="s" access="read"/>
     <!-- LastTransaction: the last (install) transaction Rauc performed -->
     <property name="LastTransaction" type="s" access="read"/>
     <!-- LastError: Holds a message describing the last error that occurred -->
--- a/src/de.pengutronix.rauc.Installer.xml
+++ b/src/de.pengutronix.rauc.Installer.xml
@@ -93,6 +93,8 @@
 
     <!-- Operation: Represents the current (global) operation rauc performs -->
     <property name="Operation" type="s" access="read"/>
+    <!-- LastTransaction: the last (install) transaction Rauc performed -->
+    <property name="LastTransaction" type="s" access="read"/>
     <!-- LastError: Holds a message describing the last error that occurred -->
     <property name="LastError" type="s" access="read"/>
     <!-- Progress: Provides installation progress information in the form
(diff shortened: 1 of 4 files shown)


js731ca:
> I don't see a way to properly ensure that `LastTransaction` and `LastBundle` actually belong to the same installation.

currently they are "strongly co-related" since transaction uuids are only set in install.c = when a bundle is installed.
But granted: this is not future-proof

> Quickly thinking about this gives me the impression that one might want to have the Transaction ID as part of the `Completed` Signal. Since we cannot change this without breaking compatibility, it might require a new signal that provides more metadata.
> 
> You could then have another Method where you can query information about the transaction with the received ID. However, this would require more infrastructure in RAUC.

True, and then a journal of past transactions and their details/metadata would be required :-S

> I assume using the status information or the event logging is not an option for your use case?

what i'm actually after is a better solution for this:
```
        ##TODO: is there a better way to ask rauc for the last bundle it consumed?
        BUNDLE=$(journalctl -u rauc.service | grep 'input bundle' | sed 's,.*bundle: ,,' | tail -n 1)
```

So the primary target is to breakout the `args->name` into the `LastBundle` property - the LastTransaction was more of a side-thought ;-)

The current event-log also doesn't provide this information.
But what could work for our usecase is a1c2b85 + c1db8a0 (we could drop 80f4ee9 )
This could then work:
-  having a [log.install] event-log-file
-  listening to the d-bus-property-change on Operation == idle 
- get the  LastTransaction  id
- with grep+jq on the event-log: map the transaction to the bundle-name : `grep 7f8d5ed6-098c-4056-8b1b-6ab1c5568e64 /data/rauc/install.log | tail -n 1 | jq -r '.BUNDLE_NAME'`

js731ca:
thanks for getting back to this PR (-:

> This would be much simpler with an InstallCompleted D-Bus signal that contains the corresponding Transaction ID.
(Maybe also the Bundle Hash.)

True, IMHO rauc should just anounce both (transaction-id and bundle-hash) with all events. currently this is not the case...

> I don't see that a LastTransaction properties gives a benefit compared to looking up the latest installation transaction ID via event log.

I would argue runtime state (which is more reliable = up to date) vs persisted/logged state (which can be out of sync with the runtime, out of date, or fully unavailable)

> What do you need the bundle name for? This is totally interchangeable.

Yes it is arbitrary, but also more humane - and since there are still humans in the update loop (at least for us). There tracing back a name to the input bundle is easier, than looking for a hash :-S

How shall we proceed?
- I'll look into an overhaul/v2?
- You take it as "food for thought" to put the transaction-id "everywhere"? and we close this PR
- Are there already open plans on your side to build upon / extend the transaction ids?

PR #1827, opened 2025-11-10
Fix Documentation of New Image `type` Option

ejoerns:
Fix image type 'image' to be 'raw'
This got renamed during mainlining of this feature.
However, the documentation did not properly reflect this.


ejoerns:
> As adding explicit `type`s breaks installation on RAUC < v1.15, I wouldn't add it to all examples yet.

It was 'some' not 'all'. But maybe we should wait with this for one or two releases anyway.

jluebbe (approved):

PR #1828, opened 2025-11-11
Prevent Misinterpretation of `--keyring` and `--signing-keyring` Options 

ejoerns:
* limit scope/visibility of `--keyring` argument
* warn on using `--keyring` for `rauc install`
* adjust option test and entry for `--signing-keyring`

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2575,7 +2575,6 @@ static GOptionEntry entries_install[] = {
 };
 
 static GOptionEntry entries_bundle[] = {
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -253,6 +254,14 @@ static gboolean install_start(int argc, char **argv)
 		goto out;
 	}
 
+	if (ENABLE_SERVICE) {
+		if (keyring) {
+			g_printerr("Using the --keyring argument for 'rauc install' is not supported!\n");
```
jluebbe:
```suggestion
			g_printerr("The '--keyring' argument is not supported when using the D-Bus service. Reconfigure the service instead.\n");
```

jluebbe (changes requested):

PR #1829, opened 2025-11-12
src/manifest: properly handle missing 'filename' in handle_missing_type()`

ejoerns:
During manifest parsing, handle_missing_type() gets called before having checked that a filename is present. Thus the method must properly handle this to not run into an assertion when calling
derive_image_type_from_filename_pattern().

Also add a test case for this.

Fixes:
| (rauc:676850): rauc-CRITICAL **: 08:48:58.847: derive_image_type_from_filename_pattern: assertion 'filename' failed

Fixes: ec13c538 ("src/update_handler: add image type to manifest")

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -24,6 +24,12 @@ static gboolean handle_missing_type(RaucImage *image, GError **error)
 	if (image->hooks.install)
 		return TRUE;
 
+	if (!image->filename) {
+		g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE,
+				"Failed to detect image type: Neither 'type' nor 'filename' given.");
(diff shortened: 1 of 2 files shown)


src/manifest.c
```diff
@@ -24,6 +24,12 @@ static gboolean handle_missing_type(RaucImage *image, GError **error)
 	if (image->hooks.install)
 		return TRUE;
 
+	if (!image->filename) {
+		g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_INVALID_VALUE,
+				"Failed detecting image type: Neither 'type' nor 'filename' given.");
```
jluebbe:
```suggestion
				"Failed to detect image type: Neither 'type' nor 'filename' given.");
```

PR #1832, opened 2025-12-01
Improve casync seeding and slot detection (UBIFS support)

b4u-mw:
This PR improves the robustness of RAUC when working with UBIFS and casync, specifically addressing issues with seeding from the active rootfs.

It implements the approach suggested in the discussion of #774 by using ext_mount_point as the source for a read-only bind mount when seeding. This avoids issues with re-mounting already mounted filesystems (like UBIFS) and allows safe seeding from the active slot.

Additionally, it improves slot detection by:

Resolving symlinks when matching devices in r_slot_find_by_device.
Resolving UBI volume names (e.g. ubi0:volname) to their device nodes to correctly detect external mount points.

Replaces  #774
Fixes #772

Disclaimer: I'm neither a C developer nor do I have deep knowledge of rauc. So I only figured out what the problems are, and did the tests. The patches are written by AI and reviewed by myself with my limited knowledge.

### DIFF
--- a/src/install.c
+++ b/src/install.c
@@ -94,6 +94,57 @@ static gchar *resolve_loop_device(const gchar *devicepath, GError **error)
 	return g_strchomp(content);
 }
 
+/*
+ * Resolve UBI volume name notation (ubi<N>:<volname>) to device path.
+ * UBI volumes are mounted with notation like "ubi0:root" but the actual
+ * device is /dev/ubi0_0. This function resolves the name by looking up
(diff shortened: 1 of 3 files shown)


b4u-mw:
@ejoerns It's not a draft anymore, could you approve the workflows?

b4u-mw:
Rebased to maser

jluebbe:
I'm a bit hesitant as we're no longer using casync (or UBIFS) in any active customer projects, so we don't have good ways to test this beyond the CI. Are you currently using this change in production?

PR #1844, opened 2026-01-02
src/main: refactor handling of key/keyring options and support them for install and mount

jluebbe:
Several commands may need a key to decrypt the bundle, but should use a
common definition. The `install` and `mount` may need to access encrypted
bundles, for which the `--key` option is needed.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2606,24 +2606,18 @@ static GOptionEntry entries_convert[] = {
 };
 
 static GOptionEntry entries_extract_signature[] = {
-	{"key", '\0', G_OPTION_FLAG_NOALIAS, G_OPTION_ARG_FILENAME, &keypath, "decryption key file or PKCS#11 URL", "PEMFILE|PKCS11-URL"},
-	{"keyring", '\0', G_OPTION_FLAG_NOALIAS, G_OPTION_ARG_FILENAME, &keyring, "keyring file", "PEMFILE"},
 	{"trust-environment", '\0', 0, G_OPTION_ARG_NONE, &trust_environment, "trust environment and skip bundle access checks", NULL},
 	{0}
 };
 
 static GOptionEntry entries_extract[] = {
-	{"key", '\0', G_OPTION_FLAG_NOALIAS, G_OPTION_ARG_FILENAME, &keypath, "decryption key file or PKCS#11 URL", "PEMFILE|PKCS11-URL"},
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -2686,10 +2685,12 @@ static GOptionGroup *info_group;
 static GOptionGroup *status_group;
 static GOptionGroup *write_slot_group;
 static GOptionGroup *service_group;
+static GOptionGroup *mount_group;
 
 static void create_option_groups(void)
 {
 	install_group = g_option_group_new("install", "Install options:", "help dummy", NULL, NULL);
+	g_option_group_add_entries(install_group, entries_bundle_open);
```
ejoerns: This should be added for `!ENABLE_SERVICE` only.

src/main.c
```diff
@@ -2736,6 +2737,11 @@ static void create_option_groups(void)
 
 	service_group = g_option_group_new("service", "Service options:", "help dummy", NULL, NULL);
 	g_option_group_add_entries(service_group, entries_service);
+
+	mount_group = g_option_group_new("install", "Install options:", "help dummy", NULL, NULL);
+	g_option_group_add_entries(mount_group, entries_bundle_open);
+	if (ENABLE_STREAMING)
```
ejoerns: Adding streaming access options is not mentioned in the commit message and unrelated. Maybe move it to a distinct commit.

ejoerns (commented):
I'm not sure if `extract-signature` better fits *opening a bundle* than `resign`, `replace-signatures`, `convert`, which aren't added here. Admittedly, those don't support `--key` and I'm unsure if that's a use case for all.

I'm also unsure about `no-verify` inconsistencies between them, but that's maybe a different topic. Cleaning up and grouping is definitely a good idea.

src/main.c
jluebbe: Yes, this should be a different PR.

src/main.c
```diff
@@ -2568,7 +2568,6 @@ static GOptionEntry entries_install[] = {
 	{"progress", '\0', 0, G_OPTION_ARG_NONE, &install_progressbar, "show progress bar", NULL},
 #else
 	{"handler-args", '\0', 0, G_OPTION_ARG_STRING, &handler_args, "extra arguments for full custom handler", "ARGS"},
-	{"keyring", '\0', G_OPTION_FLAG_NOALIAS, G_OPTION_ARG_FILENAME, &keyring, "keyring file", "PEMFILE"},
```
pH5, line 2589: This is added back by `entries_bundle_open` below, but only if `-Dstreaming=false`. Is this correct?

Issue #1842, opened 2025-12-22
Not compatible with Linux v6.19-rc1


nunojpg:
rauc: master
linux: master

RAUC now gives error: (rauc:1066): rauc-WARNING **: 14:16:19.765: Failed mounting bundle: Unexpected dm-verity status 'V -' (instead of '\0')

This comes from dm.c line 249.

I suspected linux first, but they say likely RAUC is just confused by the new return parameter.

See https://lore.kernel.org/dm-devel/9027c5d6-3eb6-456a-a939-24639a3e5278@gmail.com/T/#t

jluebbe:
@nunojpg Can you check if #1846 fixes the issue for you?

PR #1846, opened 2026-01-02
src/dm: fix compatibility with new dm-verity output since v6.19-rc1

jluebbe:
In [1], the dm-verity status output was extend with the number of blocks
corrected by FEC (which we don't use). Split the status output on spaces
and check only the first field to fix compatibility.

[1] https://git.kernel.org/linus/ae97648e14f7907f4b0e0b295eb2fdcf43806f9d

Fixes: #1842


ejoerns (approved):

jacmet (approved):
LGTM

PR #1852, opened 2026-01-13
docs/requirements.txt: update to current versions

jluebbe:
Regenerated using './update-requirements.sh'.

Closes: #1851


jluebbe:
After updating Python as well, we get the results here: https://rauc--1852.org.readthedocs.build/en/1852/

ejoerns (approved):

Issue #1701, opened 2025-04-17
Support (re)creating UEFI boot entries


jluebbe:
**Is your feature request related to a problem? Please describe.**
When using a setup like the one suggested by @hnez in https://github.com/rauc/rauc/pull/1696#issuecomment-2809369764, some UEFI
implementations tend to remove boot entries if they point to unmountable
filesystems (which can happen due to a power cycle while installing).

**Describe the solution you'd like**
We should add the missing information to the slot sections in the `system.conf` so that RAUC can recreate entries if they are missing.

```
[slot.esp.0]
device=/dev/sda1
bootname=system0
efi-loader=\\EFI\\BOOT\\BOOTX64.EFI
efi-unicode="@1" # UKI profile 1

[slot.esp.1]
device=/dev/sda2
bootname=system1
efi-loader=\\EFI\\BOOT\\BOOTX64.EFI
efi-unicode="@2" # UKI profile 2
```

**Describe alternatives you've considered**
The current workaround is to use `efibootmgr` from a post-install hook.
As probably many UEFI implementations are affected, this should be moved to common code in RAUC.

loblik:
Seems UEFI (EDK2 with NVIDIA patches) I currently work with, is not doing this. 

But it could also simplify initial installation. Default profile cmdline could contain "rauc.external" and then each entry would be created automatically when slot is flashed for the first time.

loblik:
> * But how would we even know which of the slots contain the ESP if none have a `esp-*` entry?

Not sure I got this. Entry would be created for the disk specified in `device` parameter I guess? That is passing it to `efibootmgr --device`?

Also what is the difference between scenario 1 and 3?

hnez:
> > * But how would we even know which of the slots contain the ESP if none have a `esp-*` entry?
> 
> Not sure I got this. Entry would be created for the disk specified in `device` parameter I guess? That is passing it to `efibootmgr --device`?

> Also what is the difference between scenario 1 and 3?

I've re-read what I wrote and I did indeed not get that point across very well. It comes down to the question: "what happens if no slot in the `system.conf` has any `efi-*` parameters specified?".
If `efi-cmdline` and `efi-loader` were both optional and had sensible defaults, a user may not want to specify any of them explicitly.

What would happen in the three scenarios if no `efi-*` parameter is specified for any slot:

1) A boot entry is created anyways using some defaults, because if it weren't the next step of the installation fails (setting the boot entry as BootNext).
[...]

PR #1855, opened 2026-01-19
Recreate Missing EFI Boot Entries

Bastian-Krause:
Some UEFI implementations tend to remove boot entries if they point to unmountable filesystems (which can happen due to a power cycle while installing).

Allow RAUC to recreate missing EFI entries for slots with a bootname using the new slot options "efi-loader" and "efi-cmdline".

Implement this by wrapping `efi_bootorder_get()` for callers which will actually modify the boot entries. All other callers still use `efi_bootorder_get()` directly, so no unncecessary writes happen.

Add a variant to each of the existing EFI mark tests covering this new behavior. Also add a test for a missing slot without the new slot options during mark good/bad/active operations testing the failure case.
                                                        

---

Tested on x86 hardware whose UEFI implementation is not affected by the bug by dropping the inactive boot entry manually via `efibootmgr --delete-bootnum --bootnum 000x` and `rauc mark-good rootfs.1`.

RAUC system.conf used for testing:

```
[system]
compatible=foo
bootloader=efi
bundle-formats=-plain

[keyring]
path=/etc/rauc/ca.cert.pem

[slot.esp.0]
device=/dev/disk/by-partlabel/ESP-A
bootname=system0
efi-loader=\\EFI\\BOOT\\BOOTX64.EFI
efi-cmdline=@1

[slot.rootfs.0]
device=/dev/disk/by-partlabel/rootfs-A
type=raw
parent=esp.0

[slot.esp.1]
device=/dev/disk/by-partlabel/ESP-B
bootname=system1
efi-loader=\\EFI\\BOOT\\BOOTX64.EFI
efi-cmdline=@2

[slot.rootfs.1]
device=/dev/disk/by-partlabel/rootfs-B
type=raw
parent=esp.1
```

Implements first part of #1701 ("create if entries are found missing while we write bootnext/bootorder", as outlined in https://github.com/rauc/rauc/issues/1701#issuecomment-2812343799).

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -160,6 +160,19 @@ Example configuration:
   on watchdog resets.
   Behavior defaults to ``true`` if the option is not set.
 
+``efi-loader`` (optional)
+  Only valid when ``bootloader`` is set to ``efi``.
+  If set in combination with ``efi-cmdline``, an EFI boot entry for this slot
+  is (re)created with the given loader string via ``efibootmgr``'s ``--loader``
+  option during mark good/bad/active operations, if it is missing.
+
+``efi-cmdline`` (optional)
+  Only valid when ``bootloader`` is set to ``efi``.
+  If set in combination with ``efi-loader``, an EFI boot entry for this slot is
+  (re)created with the given command line string via ``efibootmgr``'s
+  ``--unicode`` option during mark good/bad/active operations, if it is
+  missing.
+
 .. _activate-installed:
 
 ``activate-installed`` (optional)
--- a/include/slot.h
+++ b/include/slot.h
@@ -59,6 +59,8 @@ typedef struct _RaucSlot {
 	guint64 region_size;
 	/** limit the writable size of the boot partition (for boot-emmc) */
 	guint64 size_limit;
+	gchar *efi_loader;
+	gchar *efi_cmdline;
 
 	/** current state of the slot (runtime) */
 	SlotState state;
--- a/src/bootloaders/efi.c
+++ b/src/bootloaders/efi.c
@@ -276,6 +276,9 @@ static gboolean efi_set_temp_primary(RaucSlot *slot, GError **error)
 	GError *ierror = NULL;
 	efi_bootentry *efi_slot_entry = NULL;
 
+	g_return_val_if_fail(slot, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
 	if (!efi_bootorder_get(NULL, &entries, NULL, NULL, &ierror)) {
 		g_propagate_error(error, ierror);
 		return FALSE;
--- a/src/bootloaders/efi.c
+++ b/src/bootloaders/efi.c
@@ -23,6 +23,73 @@ static void efi_bootentry_free(efi_bootentry *entry)
 
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(efi_bootentry, efi_bootentry_free);
 
+static gboolean efi_bootentry_create(RaucSlot *slot, GError **error)
+{
+	g_autoptr(GSubprocess) sub = NULL;
+	g_autofree gchar *realdev = NULL, *realdev_basename = NULL;
+	g_autofree gchar *sysfs_part_path = NULL, *part_num = NULL;
+	GError *ierror = NULL;
+
+	g_return_val_if_fail(slot->bootname, FALSE);
+	g_return_val_if_fail(slot->efi_loader, FALSE);
+	g_return_val_if_fail(slot->efi_cmdline, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	realdev = r_realpath(slot->device);
+	if (realdev == NULL) {
+		g_set_error(
+				error,
+				R_BOOTCHOOSER_ERROR,
+				R_BOOTCHOOSER_ERROR_FAILED,
+				"Can't resolve device %s for EFI boot entry",
+				slot->device);
+		return FALSE;
+	}
+
+	realdev_basename = g_path_get_basename(realdev);
+	/* use /sys/class/block because it contains partition entries */
+	sysfs_part_path = g_build_filename("/sys/class/block", realdev_basename, "partition", NULL);
+	if (!g_file_get_contents(sysfs_part_path, &part_num, NULL, &ierror)) {
+		g_propagate_error(error, ierror);
+		return FALSE;
+	}
+
+	/* File contains newline, modify in-place */
+	g_strchomp(part_num);
+
+	sub = r_subprocess_new(
+			G_SUBPROCESS_FLAGS_NONE,
+			&ierror,
+			EFIBOOTMGR_NAME,
+			"--create-only",
+			"--disk", realdev,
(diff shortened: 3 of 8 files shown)


src/bootloaders/efi.c
```diff
@@ -299,6 +385,50 @@
 		return FALSE;
 	}
 
+	/* Clear previously retrieved entries */
+	if (all_entries && *all_entries)
```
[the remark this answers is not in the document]
Bastian-Krause: Fixed.

Bastian-Krause:
Pushed one more small change that makes the parametrized test IDs better human readable:
```diff
-test/test_mark.py::test_status_mark_good_efi[False] PASSED
-test/test_mark.py::test_status_mark_good_efi[True] PASSED
-test/test_mark.py::test_status_mark_bad_efi[False] PASSED
-test/test_mark.py::test_status_mark_bad_efi[True] PASSED
-test/test_mark.py::test_status_mark_active_efi[False] PASSED
-test/test_mark.py::test_status_mark_active_efi[True] PASSED
+test/test_mark.py::test_status_mark_good_efi[efi_entry_exists] PASSED
+test/test_mark.py::test_status_mark_good_efi[efi_entry_missing] PASSED
+test/test_mark.py::test_status_mark_bad_efi[efi_entry_exists] PASSED
+test/test_mark.py::test_status_mark_bad_efi[efi_entry_missing] PASSED
+test/test_mark.py::test_status_mark_active_efi[efi_entry_exists] PASSED
+test/test_mark.py::test_status_mark_active_efi[efi_entry_missing] PASSED
```

test/test_mark.py
```diff
@@ -264,17 +264,26 @@ def efi_mock(monkeypatch):
...
+@pytest.fixture
+def create_efi_system_config(create_system_files, system):
+    def _create_efi_system_config():
+        system.prepare_minimal_config()
+        system.config["system"]["bootloader"] = "efi"
+        del system.config["system"]["grubenv"]
+
+        system.write_config()
+
+    yield _create_efi_system_config
```
jluebbe: Can't you just do the setup directly from the fixture instead of returning a function?

src/bootloaders/efi.c
```diff
@@ -299,6 +385,50 @@ static gboolean efi_set_temp_primary(RaucSlot *slot, GError **error)
 		return FALSE;
 	}
 
+	/* Clear previously retrieved entries */
+	if (*all_entries)
+		g_list_free_full(g_steal_pointer(all_entries), (GDestroyNotify)efi_bootentry_free);
+	if (bootorder_entries && *bootorder_entries)
+		g_list_free(g_steal_pointer(bootorder_entries));
+
```
jluebbe: Clear also bootnext/bootcurrent, as `efi_bootorder_get` should fail if they point to a non-NULL-pointer.

src/bootloaders/efi.c
```diff
@@ -270,27 +337,46 @@ static gboolean efi_bootorder_get(GList **bootorder_entries, GList **all_entries
...
-			break;
+			/* EFI boot entry exists, no further action needed */
+			g_debug("EFI boot entry for bootname '%s' already exists", slot->bootname);
+			return TRUE;
 		}
 	}
 
-	if (!efi_slot_entry) {
+	/* No efi-loader/efi-cmdline given, bail out */
+	if (!slot->efi_loader || !slot->efi_cmdline) {
```
jluebbe, line 382: Move this check *after* clearing the entries, as we otherwise return data even on failure.

src/bootloaders/efi.c
```diff
@@ -299,6 +385,50 @@ static gboolean efi_set_temp_primary(RaucSlot *slot, GError **error)
...
+
+	return TRUE;
+}
+
+static gboolean efi_set_temp_primary(RaucSlot *slot, GError **error)
+{
+	g_autolist(efi_bootentry) entries = NULL;
+	GError *ierror = NULL;
+	efi_bootentry *efi_slot_entry = NULL;
+
```
jluebbe, line 412: Add the guards:
``` C
```

docs/reference.rst
```diff
@@ -160,6 +160,19 @@ Example configuration:
   on watchdog resets.
   Behavior defaults to ``true`` if the option is not set.
 
+``efi-loader`` (optional)
+  Only valid when ``bootloader`` is set to ``efi``.
+  If set in combination with ``efi-cmdline``, a missing efi boot entry for this
+  slot is created with the given loader string via ``efibootmgr``'s
+  ``--loader`` option during mark good/bad/active operations.
```
```suggestion
  If set in combination with ``efi-cmdline``, an EFI boot entry for this
  slot is (re)created with the given loader string via ``efibootmgr``'s
  ``--loader`` option during mark good/bad/active operations,
  if it is missing.
```

docs/reference.rst
```diff
@@ -160,6 +160,19 @@ Example configuration:
...
+  Note that backslashes need to be escaped.
+
+``efi-cmdline`` (optional)
+  Only valid when ``bootloader`` is set to ``efi``.
+  If set in combination with ``efi-loader``, a missing efi boot entry for this
+  slot is created with the given command line string via ``efibootmgr``'s
+  ``--unicode`` option during mark good/bad/active operations.
```
```suggestion
  If set in combination with ``efi-loader``, an EFI boot entry for this
  slot is (re)created with the given command line string via ``efibootmgr``'s
  ``--unicode`` option during mark good/bad/active operations,
  if it is missing.
```

include/slot.h
```diff
@@ -59,6 +59,8 @@ typedef struct _RaucSlot {
 	guint64 region_size;
 	/** limit the writable size of the boot partition (for boot-emmc) */
 	guint64 size_limit;
+	gchar* efi_loader;
+	gchar* efi_cmdline;
```
jluebbe:
```suggestion
	gchar *efi_loader;
	gchar *efi_cmdline;
```

test/test_mark.py
```diff
@@ -264,85 +265,188 @@ def efi_mock(monkeypatch):
...
+
+        if configure_loader_cmdline:
+            system.config["slot.rootfs.0"]["efi-loader"] = r"\\EFI\\BOOT\\BOOTX64.EFI"
+            system.config["slot.rootfs.0"]["efi-cmdline"] = "@1"
+            system.config["slot.rootfs.1"]["efi-loader"] = r"\\EFI\\BOOT\\BOOTX64.EFI"
+            system.config["slot.rootfs.1"]["efi-cmdline"] = "@2"
+
+        system.write_config()
+
+    yield _create_efi_system_config
```
jluebbe: Wouldn't this be easier to read as a plain helper function to be called by the test case? Just the blk_dev_partitions fixture would need to be added to each test case.

jluebbe (changes requested):

src/bootloaders/efi.c
Bastian-Krause: I've added that as an extra commit since the `efi_set_temp_primary()` function existed before. The diff is a little confusing due to the changes in the function.

Bastian-Krause:
Removed the unused `bootnext` and `bootcurrent` parameters from `efi_bootorder_prepare()`.

jluebbe (approved):

PR #1857, opened 2026-01-20
test efibootmgr with pytest

jluebbe:
This prepares for more extensive tests which are needed for the entry creation support in #1855.

### DIFF
--- a/test/bin/efibootmgr
+++ b/test/bin/efibootmgr
@@ -1,33 +1,41 @@
-#!/bin/bash
+#!/usr/bin/env python3
 
-# Command line checking
-if [ $# -eq 0 ]; then
-	echo "\
-BootCurrent: 0002
-Timeout: 0 seconds
-BootOrder: 0001,0002,0003,0000
-Boot0000* invalid
-Boot0001* system0
-Boot0002* system1
-Boot0003  recovery"
-exit 0
-fi
+import argparse
+import sys
 
-if [ "$1" = "--bootorder" ]; then
-	# Test code must move system1 to primary position. Assert this!
-	echo "--bootorder: $2"
-	#if [ "$2" != "0002,0001,0003,0000" ]; then
-	#	exit 1
-	#fi
-elif [ "$1" = "--bootnext" ]; then
-	# Test code must move system1 to primary position. Assert this!
-	echo "--bootnext: $2"
-	if [ "$2" != "0002" ]; then
-		exit 1
-	fi
-else
-	echo "Invalid key: '$1'"
-	exit 1
-fi
 
-exit 0
+def main():
+    parser = argparse.ArgumentParser(description='efibootmgr mock for RAUC testing')
+
+    parser.add_argument("--bootorder", metavar="ORDER", help="Set boot order")
+    parser.add_argument("--bootnext", metavar="NEXT", help="Set next boot entry")
+
+    if len(sys.argv) == 1:
+        print("BootCurrent: 0002")
+        print("Timeout: 0 seconds")
+        print("BootOrder: 0001,0002,0003,0000")
+        print("Boot0000* invalid")
+        print("Boot0001* system0")
+        print("Boot0002* system1")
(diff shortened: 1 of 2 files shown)


test/bin/efibootmgr
```diff
@@ -1,27 +1,89 @@
...
 import sys
 
+from copy import deepcopy
 
-def main():
-    parser = argparse.ArgumentParser()
+DEFAULT_VARS = {
+    "boot_current": "0002",
+    "timeout": 0,
+    "boot_order": ["0001", "0002", "0000"],
```
ejoerns, line 11: Do you intentionally change the boot order her? (`0003` is missing)

test/bin/efibootmgr
```diff
@@ -1,27 +1,89 @@
...
-        sys.exit(0)
+def load_efi_vars(filepath):
+    """Load EFI variables from JSON file."""
+    if os.path.exists(filepath) and os.path.getsize(filepath) > 0:
+        try:
+            with open(filepath, "r") as f:
+                return json.load(f)
+        except (json.JSONDecodeError, IOError):
+            pass
+    # Return default structure if file doesn't exist or is invalid
```
ejoerns: Wouldn't it be safer to fail if the input file is invalid?

test/bin/efibootmgr
```diff
@@ -33,8 +95,31 @@ def main():
...
+    parser.add_argument("--bootnext", metavar="NEXT", help="Set next boot entry")
+    parser.add_argument("--create-only", action="store_true", help="Create new boot entry")
+    parser.add_argument("--disk", help="Disk device")
+    parser.add_argument("--part", help="Partition number")
+    parser.add_argument("--label", help="Boot entry label")
+    parser.add_argument("--loader", help="EFI loader path")
+    parser.add_argument("--unicode", help="Kernel parameters")
+
+    # Check if JSON mode is enabled
+    vars_file = os.environ.get("EFIBOOTMGR_VARS")
```
ejoerns: The name might be confusing. For the barebox state mock, the `BAREBOX_STATE_VARS_*` variable actually holds the variables. For the efi mock, it is the path to the storage file.

Thus maybe `EFIBOOTMGR_STATE_FILE` or, `EFIBOOTMGR_VAR_FILE`?

test/bin/efibootmgr
```diff
@@ -1,33 +1,126 @@
...
+        print(f"--bootnext: {args.bootnext}")
+        if args.bootnext != "0002":
+            sys.exit(1)
+    else:
+        # No arguments provided - print default output
+        show_efi_vars(DEFAULT_VARS)
+
+
+def main():
+    parser = argparse.ArgumentParser()
```
ejoerns: Since we already use argparse, it should be easy to add a minimal user output, too. Just to avoid confusion with an actual `efibootmgr`.

```suggestion
    parser = argparse.ArgumentParser(description='efibootmgr mock. Use EFIBOOTMGR_VARS for persiting data.')
```

test/bin/efibootmgr
```diff
@@ -33,8 +95,31 @@ def main():
...
+
+    parser.add_argument("--bootorder", metavar="ORDER", help="Set boot order")
+    parser.add_argument("--bootnext", metavar="NEXT", help="Set next boot entry")
+    parser.add_argument("--create-only", action="store_true", help="Create new boot entry")
```
ejoerns, line 101: If not a separate commit, the commit message should at least mention that this also extends the functionality by support for creating boot entries (and thus 6 additional arguments in total).

ejoerns (changes requested):

test/bin/efibootmgr
jluebbe: This was inconsistent with the active entries before.

jluebbe: Added an explanation to the commit message.

ejoerns (approved):
Looks really helpful for testing! :+1:

PR #1858, opened 2026-01-21
Support bundle-defined global-pre/post-install hooks

yosuke-tanimoto:
This PR replaces #1820 with an Install-Hook based implementation.

## Overview

This PR implements support for global-pre/post-install hooks in the bundle's manifest file.
This enables hooks to be defined per bundle, so the old system no longer needs to predefine future actions in its handlers.

The need for this functionality was previously discussed in #275 and #342, but was not implemented until now.

## Example

In the manifest file:
```cfg
[hooks]
filename=hook.sh
hooks=global-pre-install;global-post-install
```

In `hook.sh`:
```sh
#!/bin/sh

case "$1" in
        global-pre-install)
                # Log update start
                echo "$(date -Iseconds) Starting update to ${RAUC_MF_VERSION}" >> /data/update.log
                ;;
        global-post-install)
                # Log update completion
                echo "$(date -Iseconds) Updated to ${RAUC_MF_VERSION}" >> /data/update.log
                # Schedule reboot to apply update
                shutdown -r +1 "Rebooting to complete RAUC update" &
                ;;
        *)
                exit 1
                ;;
esac

exit 0
```

Hooks and handlers are executed in the following order:

1. System-defined pre-install handler (defined in system.conf)
2. Bundle-defined global-pre-install hook (defined in the manifest file)
3. Installation process for each slot
4. Bundle-defined global-post-install hook
5. System-defined post-install handler

## Tests

Both pytest and glib tests passed successfully.
I added tests for global-pre/post-install hooks:

- Hook execution test in install-test (`test/install.c` and `test/install-content/hook.sh`)
- Manifest save/load test with new hook flags in manifest-test (`test_save_load_manifest()` in `test/manifest.c`)
- Environment variable verification test in pytest-install (`test_install_hook_env()` in `test/test_install.py`)

## Changes from #1820

- Changed from handler implementation to Install Hooks implementation
- No manifest file format changes required, as the existing [hooks] section is used
- No need to recreate `good-adaptive-meta-bundle.raucb` for testing, as `test_install_hook_env` creates bundles at runtime
- The install-check hook replaces the compatible check, so it cannot serve as a simple global-pre-install hook. Therefore, I implemented both global-pre-install and global-post-install hooks.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -859,7 +859,7 @@ The hooks section allows to provide a user-defined executable for
   List of hooks enabled for this bundle.
   See :ref:`sec-install-hooks` for more details.
 
-  Valid items are: ``install-check``
+  Valid items are: ``install-check``, ``pre-install``, ``post-install``
 
 .. _sec-manifest-handler:
 
@@ -1050,7 +1050,8 @@ way to add metadata to the manifest which is not interpreted by RAUC in any
 way.
 They are accessible via ``rauc info``, the :ref:`"InspectBundle" D-Bus API
 <gdbus-method-de-pengutronix-rauc-Installer.InspectBundle>`, pre-/post-install
-:ref:`handlers <sec-handler-interface>` and the ``install-check`` hook.
+:ref:`handlers <sec-handler-interface>`, the ``install-check`` hook and the
+``pre-install``/``post-install`` hooks.
 In future releases, they will be accessible in other hooks, as well.
 
 ``<key>`` (optional)
@@ -1930,7 +1931,7 @@ OUT *info* ``a{sv}``:
             The hook filename
 
         *info.hooks.hooks* variant ``as`` <hooks>:
-            An array of enabled hooks (i.e. ``install-check``)
+            An array of enabled hooks (i.e. ``install-check``, ``pre-install`` or ``post-install``)
 
     *info.handler* variant ``a{sv}`` <handler-dict>:
         The bundle's ``[handler]`` section content
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -402,6 +402,47 @@ the hook executable as the rejection reason message and provide it to the user:
 
   exit 0
 
+.. rubric:: Pre-Install Hook
+
+.. code-block:: cfg
+
+  [hooks]
+  filename=hook
+  hooks=pre-install
+
+The pre-install hook will be called right before all slot installations are started.
+This hook is executed after the system-defined pre-install handler and the install-check hook (if defined).
+It allows to perform preparation tasks that should run once before any slot is updated.
+
+.. rubric:: Post-Install Hook
+
+.. code-block:: cfg
+
+  [hooks]
+  filename=hook
+  hooks=post-install
+
+The post-install hook will be called right after all slot installations have finished successfully.
+This hook is executed before the system-defined post-install handler.
+It allows to perform cleanup or finalization tasks that should run once after all slots are updated,
+for example to notify external services or to perform cross-slot validation.
+
+.. code-block:: sh
+
+  #!/bin/sh
+
+  case "$1" in
+          post-install)
+                  # Log update completion
+                  echo "$(date -Iseconds) Updated to ${RAUC_MF_VERSION}" >> /data/update.log
+                  ;;
+          *)
+                  exit 1
+                  ;;
+  esac
+
+  exit 0
+
 .. _sec-slot-hooks:
 
 Slot Hooks
--- a/src/install.c
+++ b/src/install.c
@@ -1652,6 +1652,18 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
 		goto umount;
 	}
 
+	if (bundle->manifest->hooks.pre_install) {
+		run_bundle_hook(bundle->manifest, bundle->mount_point, "pre-install", &ierror);
+		if (ierror) {
+			res = FALSE;
+			g_propagate_prefixed_error(
+					error,
+					ierror,
+					"Pre-install hook failed: ");
+			goto umount;
+		}
+	}
+
 	if (bundle->manifest->handler_name) {
 		g_message("Using custom handler: %s", bundle->manifest->handler_name);
 		res = launch_and_wait_custom_handler(args, bundle->mount_point, bundle->manifest, target_group, handler_env, &ierror);
@@ -1663,6 +1675,18 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
 	if (!res) {
 		g_propagate_prefixed_error(error, ierror, "Installation error: ");
 		goto umount;
+	}
+
+	if (bundle->manifest->hooks.post_install) {
+		run_bundle_hook(bundle->manifest, bundle->mount_point, "post-install", &ierror);
+		if (ierror) {
+			res = FALSE;
+			g_propagate_prefixed_error(
+					error,
+					ierror,
+					"Post-install hook failed: ");
+			goto umount;
+		}
 	}
 
 	if (r_context()->config->postinstall_handler) {
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -374,6 +374,10 @@ static gboolean parse_manifest(GKeyFile *key_file, RaucManifest **manifest, GErr
 	for (gsize j = 0; j < hook_entries; j++) {
 		if (g_strcmp0(bundle_hooks[j], "install-check") == 0) {
 			raucm->hooks.install_check = TRUE;
+		} else if (g_strcmp0(bundle_hooks[j], "pre-install") == 0) {
+			raucm->hooks.pre_install = TRUE;
+		} else if (g_strcmp0(bundle_hooks[j], "post-install") == 0) {
+			raucm->hooks.post_install = TRUE;
 		} else {
 			g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_PARSE,
 					"install hook type '%s' not supported", bundle_hooks[j]);
@@ -542,7 +546,7 @@ static gboolean check_manifest_common(const RaucManifest *mf, GError **error)
 
 	/* Check for hook file set if hooks are enabled */
 
-	if (mf->hooks.install_check == TRUE)
+	if ((mf->hooks.install_check == TRUE) || (mf->hooks.pre_install == TRUE) || (mf->hooks.post_install == TRUE))
 		have_hooks = TRUE;
 
 	for (GList *l = mf->images; l != NULL; l = l->next) {
@@ -953,6 +957,12 @@ static GKeyFile *prepare_manifest(const RaucManifest *mf)
 	if (mf->hooks.install_check == TRUE) {
 		g_ptr_array_add(hooks, g_strdup("install-check"));
 	}
+	if (mf->hooks.pre_install == TRUE) {
+		g_ptr_array_add(hooks, g_strdup("pre-install"));
+	}
+	if (mf->hooks.post_install == TRUE) {
+		g_ptr_array_add(hooks, g_strdup("post-install"));
+	}
 	g_ptr_array_add(hooks, NULL);
 	if (hooks->pdata && *hooks->pdata) {
 		g_key_file_set_string_list(key_file, "hooks", "hooks",
@@ -1126,6 +1136,10 @@ GVariant* r_manifest_to_dict(const RaucManifest *manifest)
 		g_variant_builder_init(&builder, G_VARIANT_TYPE("as"));
 		if (manifest->hooks.install_check)
 			g_variant_builder_add(&builder, "s", "install-check");
+		if (manifest->hooks.pre_install)
+			g_variant_builder_add(&builder, "s", "pre-install");
+		if (manifest->hooks.post_install)
+			g_variant_builder_add(&builder, "s", "post-install");
 		g_variant_dict_insert(&grp_dict, "hooks", "v", g_variant_builder_end(&builder));
 		g_variant_dict_insert(&root_dict, "hooks", "v", g_variant_dict_end(&grp_dict));
 	}
--- a/test/install.c
+++ b/test/install.c
@@ -110,6 +110,30 @@ filename=appfs.ext4";
 	fixture_helper_set_up_bundle(fixture->tmpdir, manifest_file, &data->manifest_test_options);
 }
 
+static void install_fixture_set_up_bundle_global_post_install_hook(InstallFixture *fixture,
+		gconstpointer user_data)
+{
+	InstallData *data = (InstallData*) user_data;
+	const gchar *manifest_file = "\
+[update]\n\
+compatible=Test Config\n\
+\n\
+[hooks]\n\
+filename=hook.sh\n\
+hooks=post-install\n\
+\n\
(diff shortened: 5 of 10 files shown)


src/install.c
```diff
@@ -1622,6 +1622,25 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
...
 
+	if (bundle->manifest->hooks.global_pre_install) {
+		run_bundle_hook(bundle->manifest, bundle->mount_point, "global-pre-install", &ierror);
+		if (ierror) {
+			res = FALSE;
+			if (g_error_matches(ierror, R_INSTALL_ERROR, R_INSTALL_ERROR_REJECTED)) {
+				g_propagate_prefixed_error(
+						error,
+						ierror,
+						"Bundle rejected: ");
```
ejoerns: The pre-install handle is not meant to reject bundles.

I am also not sure yet if the pre-install hook should run before the install check hook (neither am I for pre-install *handler*, but..)

docs/using.rst
```diff
@@ -402,6 +402,58 @@ the hook executable as the rejection reason message and provide it to the user:
 
   exit 0
 
+.. rubric:: Global Pre-Install Hook
+
+.. code-block:: cfg
+
+  [hooks]
+  filename=hook
+  hooks=global-pre-install
```
ejoerns: Since bundle hooks are 'global' by definition, I'd shorten the name and just call it `pre-install` / `post-install`.

src/install.c
```diff
@@ -1622,6 +1622,25 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
 		}
 	}
 
+	if (bundle->manifest->hooks.global_pre_install) {
+		run_bundle_hook(bundle->manifest, bundle->mount_point, "global-pre-install", &ierror);
+		if (ierror) {
+			res = FALSE;
+			if (g_error_matches(ierror, R_INSTALL_ERROR, R_INSTALL_ERROR_REJECTED)) {
```
ejoerns: The `run_bundle_hook()` error handling isn't as generic as it should be for this use case. At least from its error handling/naming. Maybe we can ignore this for now by just dropping the extra check for `R_INSTALL_ERROR_REJECTED` here.

docs/using.rst
```diff
@@ -402,6 +402,58 @@ the hook executable as the rejection reason message and provide it to the user:
...
+.. code-block:: sh
+
+  #!/bin/sh
+
+  case "$1" in
+         global-post-install)
+                  # Log update completion
+                  echo "$(date -Iseconds) Updated to ${RAUC_MF_VERSION}" >> /data/update.log
+                  # Schedule reboot to apply update
+                  shutdown -r +1 "Rebooting to complete RAUC update" &
```
ejoerns: The issue when using the `post-install` handler for rebooting is that it will interrupt the installation too early in the current design. We miss at least the event log noting a successful installation. But we should also wait for the hook to complete since it can make the installation fail..

ejoerns (commented):
@yosuke-tanimoto Thank you for converting the handler into a hook! I had a first look and some high level review/notes, first.

yosuke-tanimoto:
@ejoerns Thank you for your feedback. I've addressed your points.

### Bundle rejection handling (in `src/install.c`)

Done. Removed the extra check for `R_INSTALL_ERROR_REJECTED`.

> Maybe we can ignore this for now by just dropping the extra check for R_INSTALL_ERROR_REJECTED here.

I agree. If the exit code is 10 or higher, the error message from `run_bundle_hook()` changes, but there is no special handling for rejection in this context.

### Pre-install hook execution order (in `src/install.c`)

> I am also not sure yet if the pre-install hook should run before the install check hook (neither am I for pre-install _handler_, but..)

Yes, there are certainly use cases where we don't want pre-install to run if install-check fails the compatible check. So I changed the order to run the pre-install hook after install-check (leaving the pre-install _handler_ aside).

### Rename to pre-install/post-install (in `src/main.c`, etc.)

Done. Renamed to pre-install / post-install in the `[hooks]` section.

The hook names passed to the hook script remain 'global-pre-install' and 'global-post-install' to distinguish them from 'slot-pre-install' and 'slot-post-install'.

### Reboot example in post-install hook (in `docs/using.rst`)

Done. I removed the reboot example from the post-install hook section, as it may not be suitable as a recommended practice in the documentation.

docs/reference.rst
```diff
@@ -1167,7 +1050,8 @@ way to add metadata to the manifest which is not interpreted by RAUC in any
 way.
 They are accessible via ``rauc info``, the :ref:`"InspectBundle" D-Bus API
 <gdbus-method-de-pengutronix-rauc-Installer.InspectBundle>`, pre-/post-install
-:ref:`handlers <sec-handler-interface>` and the ``install-check`` hook.
+:ref:`handlers <sec-handler-interface>`, the ``install-check`` hook and the 
+global-pre-/post-install hooks.
```
```suggestion
`pre-install`/`post-install` hooks.
```

docs/using.rst
```diff
@@ -411,19 +402,68 @@ the hook executable as the rejection reason message and provide it to the user:
 
   exit 0
 
+.. rubric:: Global Pre-Install Hook
```
ejoerns:
```suggestion
.. rubric:: Pre-Install Hook
```

The scope should already be set by the chapter 'Install Hooks'.

docs/using.rst
```diff
@@ -411,19 +402,68 @@ the hook executable as the rejection reason message and provide it to the user:
...
+
+.. code-block:: cfg
+
+  [hooks]
+  filename=hook
+  hooks=pre-install
+
+The global-pre-install hook will be called right before the update procedure
```
ejoerns:
```suggestion
The pre-install hook will be called right before the update procedure
```

docs/using.rst
```diff
@@ -411,19 +402,68 @@ the hook executable as the rejection reason message and provide it to the user:
...
+for all slot installations will be started. This hook is executed after the 
+system-defined pre-install handler and the install-check hook (if defined).
+It allows to perform preparation tasks that should run once before any slot 
+is updated.
+
+To indicate that a bundle should be rejected, the script must return with an
+exit code >= 10.
+
+.. rubric:: Global Post-Install Hook
```
ejoerns:
```suggestion
.. rubric:: Post-Install Hook
```

docs/using.rst
```diff
@@ -411,19 +402,68 @@ the hook executable as the rejection reason message and provide it to the user:
...
+
+.. code-block:: cfg
+
+  [hooks]
+  filename=hook
+  hooks=post-install
+
+The global-post-install hook will be called right after all slot installations
```
ejoerns:
```suggestion
The post-install hook will be called right after all slot installations
```

test/install.c
```diff
@@ -1225,6 +1249,44 @@
 	args->cleanup(args);
 }
 
+static void install_test_bundle_hook_global_post_install(InstallFixture *fixture,
+		gconstpointer user_data)
+{
+	g_autofree gchar *bundlepath = NULL;
+	g_autofree gchar *mountdir = NULL;
+	g_autofree gchar *testfilepath = NULL;
```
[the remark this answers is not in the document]
ejoerns: This ist still open.

docs/using.rst
```diff
@@ -411,19 +402,68 @@ the hook executable as the rejection reason message and provide it to the user:
...
+.. rubric:: Global Pre-Install Hook
+
+.. code-block:: cfg
+
+  [hooks]
+  filename=hook
+  hooks=pre-install
+
+The global-pre-install hook will be called right before the update procedure
+for all slot installations will be started. This hook is executed after the
```
ejoerns: This does not make sense, yet. Did you mean 'before all slot installations will be started'?

docs/using.rst
```diff
@@ -411,19 +402,68 @@ the hook executable as the rejection reason message and provide it to the user:
...
+  filename=hook
+  hooks=pre-install
+
+The global-pre-install hook will be called right before the update procedure
+for all slot installations will be started. This hook is executed after the 
+system-defined pre-install handler and the install-check hook (if defined).
+It allows to perform preparation tasks that should run once before any slot 
+is updated.
+
+To indicate that a bundle should be rejected, the script must return with an
```
ejoerns: The rejection handling should also be dropped from the documentation here.

docs/using.rst
```diff
@@ -411,19 +402,68 @@ the hook executable as the rejection reason message and provide it to the user:
...
+  filename=hook
+  hooks=post-install
+
+The global-post-install hook will be called right after all slot installations
+have finished successfully. This hook is executed before the system-defined
+post-install handler. It allows to perform cleanup or finalization tasks that
+should run once after all slots are updated, for example to notify external
+services or to perform cross-slot validation.
+
+To indicate that the installation should be marked as failed, the script must
```
ejoerns: This is incorrect. There is no special handling for exit codes for this hook type.

ejoerns (changes requested):
@yosuke-tanimoto Thank you for addressing the comments. I guess we get close to what could work for RAUC now. Just some things left.

Currently, the documentation states:

> For install hooks, the hook call argument is just the hook name itself (e.g.
``install-check``).

I'm not sure if I would like to change this for the 'global' pre-install and post-install hooks. They can already be distinguished from the per-slot ones by the missing `slot-` prefix.

yosuke-tanimoto:
@ejoerns Thanks for your review! I've addressed the feedback:

- Renamed hooks to pre-install/post-install in the documentation
- Removed the error code description from the documentation
- Removed the unused testfilepath variable flagged by CodeQL

All changes have been squashed into the relevant commits.

ejoerns:
@yosuke-tanimoto Thank you for addressing my comments.

With a slightly clearer look at the actual changes now: Could you elaborate the benefits you see in having a `pre-install` hook in addition to the already-existing `install-check` hook?
In the current form, the two hooks would be executed one after the other, thus targeting the exact same state of the installation.
Do you have an actual use case for this? Or is this just for semantically separating the custom compatible check from other actions and to have the pre/post-duality? (The mentioned Also #342 / #275 only refer to the missing post-install handler, but not a missing pre-install handler.)

I'd also argue that if we would keep both, the `pre-install` hook should run *after* the `install-check` hook, since we should not start with performing preparatory step when not being sure the bundle is compatible with the system it runs on.


Also the commits still needs to be better squashed / cleaned up since some add things that others remove again, etc.

yosuke-tanimoto:
@ejoerns Thank you for your detailed feedback.

> Could you elaborate the benefits you see in having a `pre-install` hook in addition to the already-existing `install-check` hook?

Looking at the `do_install_bundle()` implementation, an install-check hook completely replaces the existing compatible check (`verify_compatible()`) with script execution. So using install-check hook for pre-install tasks would require users to implement the compatible check themselves in the install-check hook. The pre-install hook has no such side effect. That said, if you think pre-install hook is unnecessary, I'm open to removing it from this PR.

> I'd also argue that if we would keep both, the `pre-install` hook should run after the `install-check` hook

I agree. I have already changed pre-install hook to run after install-check hook.

> Also the commits still needs to be better squashed / cleaned up since some add things that others remove again, etc.

I have recreated the commits on a per-file basis.

docs/using.rst
ejoerns, line 366: This does not align with the `global-` prefix used for the `pre-install` and `post-install` hooks.

Since the per-slot hooks are already scoped with the `slot-` prefix, omitting `global-` should be fine and consistent with `install-check`.

docs/using.rst
```diff
@@ -412,19 +402,61 @@ the hook executable as the rejection reason message and provide it to the user:
...
 
+.. rubric:: Pre-Install Hook
+
+.. code-block:: cfg
+
+  [hooks]
+  filename=hook
+  hooks=pre-install
+
+The pre-install hook will be called right before all slot installations are
```
ejoerns: If you touch the documentation again, please consider using [semantic line breaks](https://rauc.readthedocs.io/en/latest/contributing.html#documentation).

If not, we could do it in a follow-up PR since the rest of the hook documentation doesn't use it yet, too.

ejoerns (changes requested):
@yosuke-tanimoto Thank you for addressing the comments!

> Looking at the do_install_bundle() implementation, an install-check hook completely replaces the existing compatible check (verify_compatible()) with script execution. So using install-check hook for pre-install tasks would require users to implement the compatible check themselves in the install-check hook. The pre-install hook has no such side effect. That said, if you think pre-install hook is unnecessary, I'm open to removing it from this PR.

This sounds reasonable to me and it is properly documented, I'd thus agree to keep both hooks.

The changes and commit separation look fine to me!

yosuke-tanimoto:
@ejoerns Thank you for your feedbacks!

I have removed the `global-` prefix from hook names and applied semantic line breaks to the documentation.
Now, each hook is invoked as `pre-install` and `post-install`.

These changes have been incorporated into the respective commits.
Please review when you have a chance.

src/manifest.c
```diff
@@ -547,6 +548,10 @@ static gboolean check_manifest_common(const RaucManifest *mf, GError **error)
 
 	if (mf->hooks.install_check == TRUE)
 		have_hooks = TRUE;
+	if (mf->hooks.pre_install == TRUE)
```
ejoerns: Would be more elegant and save some checks if these are combined with `||`.

But it works as-is, so could be a later cleanup.

src/install.c
```diff
@@ -1646,6 +1646,18 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
 		goto umount;
 	}
 
+	if (bundle->manifest->hooks.pre_install) {
```
ejoerns, line 1655: The reasoning in the corresponding commit seems to suggests that hooks are better than handlers and might replace them.
Just to be clear: Which one to pick (hooks or handlers) is actually quite use-case specific.

src/install.c
```diff
@@ -1653,6 +1646,18 @@ gboolean do_install_bundle(RaucInstallArgs *args, GError **error)
...
```
ejoerns, line 1655: The handler should also run **after** `check_version_limits()` (similar to install-check/compatible handling) to only run when the bundle is considered appropriate for the system.

docs/reference.rst
```diff
@@ -1167,7 +1050,7 @@ way to add metadata to the manifest which is not interpreted by RAUC in any
 way.
 They are accessible via ``rauc info``, the :ref:`"InspectBundle" D-Bus API
 <gdbus-method-de-pengutronix-rauc-Installer.InspectBundle>`, pre-/post-install
-:ref:`handlers <sec-handler-interface>` and the ``install-check`` hook.
+:ref:`handlers <sec-handler-interface>`, the ``install-check`` hook and the `pre-install`/`post-install` hooks.
```
```suggestion
:ref:`handlers <sec-handler-interface>`, the ``install-check`` hook and the ``pre-install``/``post-install`` hooks.
```

ejoerns (commented):
Looks quite good.

One issue left for the ordering, the rest are minor remarks.
I will quickly fix these to also allow re-triggering the CI build.

src/install.c
ejoerns: I've slightly adapted the commit message.

src/install.c
ejoerns: Fixed this.

ejoerns (approved):

PR #1864, opened 2026-01-26
src: store boot_state in enum to allow 'unknown' state.

### LINKED ISSUE #1751: rauc status warns that it couldn't determine all boot states but still reports boot_status = bad
**Describe the bug**
When running `rauc status` as a non-root user that cannot run `fw_printenv` rauc warns that it couldn't determine all boot states. This can be misleading when only looking at that boot status lines and perhaps missing the warnings as they are already scrolled out of view. It could also potentially be problematic in scripting usage as rauc still exits with exit_code = 0.

**Background information**
```
$ rauc --version
rauc 1.14
```
with uboot bootloader backend

**To Reproduce**
run `rauc status`

**Expected behavior**
Rauc not reporting bot status "bad" when it can't figure out the boot status. Maybe show "unknown" or "n/a"?

**Logs**
```
$ rauc status
rauc-Message: 12:53:44.867: Using central status file /data/rauc.status
rauc-Message: 12:53:44.868: Using system config file /etc/rauc/system.conf
Error opening lock file /run/fw_printenv.lock
rauc-Message: 12:53:44.875: Failed to get boot state of 'rootfs.1': uboot backend: fw_printenv failed with exit code: 1
Error opening lock file /run/fw_printenv.lock
rauc-Message: 12:53:44.880: Failed to get boot state of 'rootfs.0': uboot backend: fw_printenv failed with exit code: 1
Failed to determine boot states: Could not determine all boot states
Error opening lock file /run/fw_printenv.lock
Failed getting primary slot: uboot backend: fw_printenv failed with exit code: 1
=== System Info ===
[...] normal output continues 
```

**Additional context**
The rauc service is not used here.

ejoerns:
With the currently used boolean, RAUC has now way to differentiate between a failure during boot state determination and the boot state 'bad'.

Fixes: #1751

Reported-by: Marcus Hoffmann <marcus.hoffmann@othermo.de>

### DIFF
--- a/include/slot.h
+++ b/include/slot.h
@@ -11,6 +11,12 @@ typedef enum {
 	ST_BOOTED = 4 | ST_ACTIVE,
 } SlotState;
 
+typedef enum {
+	ST_BOOT_UNKNOWN = 0,
+	ST_BOOT_BAD = 1,
+	ST_BOOT_GOOD = 2,
+} SlotBootState;
+
 typedef struct {
 	gchar *bundle_compatible;
 	gchar *bundle_version;
@@ -64,7 +70,7 @@ typedef struct _RaucSlot {
 
 	/** current state of the slot (runtime) */
 	SlotState state;
-	gboolean boot_good;
+	SlotBootState boot_state;
 	struct _RaucSlot *parent;
 	/** the name of the parent as parsed by config (parsing-internal use only) */
 	gchar *parent_name;
@@ -117,6 +123,16 @@ G_GNUC_WARN_UNUSED_RESULT;
  * @return a RaucSlot pointer or NULL
  */
 RaucSlot *r_slot_get_booted(GHashTable *slots);
+
+/**
+ * Get string representation of boot state
+ *
+ * @param bootstate boot state to turn into string
+ *
+ * @return string representation of boot state
+ */
+const gchar* r_slot_bootstate_to_str(SlotBootState bootstate)
+G_GNUC_WARN_UNUSED_RESULT;
 
 /**
  * Get string representation of slot state
--- a/src/slot.c
+++ b/src/slot.c
@@ -94,6 +94,24 @@ RaucSlot *r_slot_get_booted(GHashTable *slots)
 }
 
 /* returns string representation of slot state */
+const gchar* r_slot_bootstate_to_str(SlotBootState bootstate)
+{
+	switch (bootstate) {
+		case ST_BOOT_UNKNOWN:
+			return "unknown";
+		case ST_BOOT_GOOD:
+			return "good";
+		case ST_BOOT_BAD:
+			return "bad";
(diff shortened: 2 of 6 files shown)


include/slot.h
```diff
@@ -62,7 +68,7 @@ typedef struct _RaucSlot {
 
 	/** current state of the slot (runtime) */
 	SlotState state;
-	gboolean boot_good;
+	SlotBootState boot_state;
```
jluebbe, line 73: Rename this to make it obvious that we adjusted all current users.
```suggestion
	SlotBootState bootstate;
```

jluebbe (changes requested):
Looks good to me otherwise.

src/slot.c
```diff
@@ -93,6 +91,25 @@ RaucSlot *r_slot_get_booted(GHashTable *slots)
...
+const gchar* r_slot_bootstate_to_str(SlotBootState bootstate)
+{
+	switch (bootstate) {
+		case ST_BOOT_UNKNOWN:
+			return "unknown";
+		case ST_BOOT_GOOD:
+			return "good";
+		case ST_BOOT_BAD:
+			return "bad";
+			break;
```
pH5: Superfluous `break;`

src/slot.c
```diff
@@ -93,6 +91,25 @@ RaucSlot *r_slot_get_booted(GHashTable *slots)
...
+		default:
+			g_error("invalid boot status %d", bootstate);
+			break;
```
pH5: Could `return NULL;` here ...

src/slot.c
```diff
@@ -93,6 +91,25 @@ RaucSlot *r_slot_get_booted(GHashTable *slots)
...
+	}
+
+	return NULL;
```
pH5: ... and drop this.

pH5 (commented):
Typo in commit message `s/now/no/`.

include/slot.h
ejoerns: I guess the rename from `boot_good` to `boot_state` should have been sufficient.

src/slot.c
ejoerns: replaced it by `g_assert_not_reached()`

ejoerns:
Note this now depends on #1904 being merged (for having a proper documentation place for the 'unknown' state).

include/slot.h
pH5: I agree.

pH5 (dismissed):
Looks good to me now.

ejoerns:
Rebased onto #1962 to fix unrelated CI test failure

include/slot.h
jluebbe: Sorry, it seems I misread this during review.

jluebbe (approved):

PR #1871, opened 2026-02-03
docs: fix and improve EFI bootloader handling

Bastian-Krause:
Some doc fixes and improvements targeting the EFI bootloader integration.

#1855

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -1277,10 +1277,10 @@ You can inspect and verify your settings by running:
 
 In your ``system.conf``, you have to list both the EFI partitions (each containing
 one kernel) as well as the rootfs partitions.
-Make the first EFI partition a child of the first rootfs partition and the
-second EFI partition a child of the second rootfs partition to have valid slot
+Make the first rootfs partition a child of the first EFI partition and the
+second rootfs partition a child of the second EFI partition to have valid slot
 groups.
-Set the rootfs slot bootnames to those we have defined with the ``--label``
+Set the EFI slot bootnames to those we have defined with the ``--label``
 argument in the ``efibootmgr`` call above:
 
 .. code-block:: cfg
@@ -1288,22 +1288,22 @@ argument in the ``efibootmgr`` call above:
   [slot.efi.0]
   device=/dev/sdX1
   type=vfat
-  parent=rootfs.0
+  bootname=system0
 
   [slot.efi.1]
   device=/dev/sdX2
   type=vfat
-  parent=rootfs.1
+  bootname=system1
 
   [slot.rootfs.0]
   device=/dev/sdX3
   type=ext4
-  bootname=system0
+  parent=efi.0
 
   [slot.rootfs.1]
   device=/dev/sdX4
   type=ext4
-  bootname=system1
+  parent=efi.1
 
 .. _sec-custom-bootloader-backend:
 
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -1242,7 +1242,8 @@ To enable EFI bootloader support in RAUC, write in your ``system.conf``:
 
 To set up a system ready for pure EFI-based redundancy boot without any further
 bootloader or initramfs involved, you have to create an appropriate
-partition layout and matching boot EFI entries.
+partition layout and either configure boot entries manually or let RAUC do
+that.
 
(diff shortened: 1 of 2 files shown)


docs/integration.rst
```diff
@@ -1237,15 +1237,38 @@ support in RAUC, write in your ``system.conf``:
...
+
+  [system]
+  ...
+  bootloader=efi
+
+  [slot.esp.0]
+  ...
+  bootname=system0
+  efi-loader=\\EFI\\BOOT\\BOOTX64.EFI
+  efi-cmdline=@1
```
hnez: I think the example should be kept consistent with the EFI stub kernel path mentioned above and the command line mentioned below, where the entries are added using `eifbootmgr`.

```suggestion
  efi-loader=\\EFI\\LINUX\\BZIMAGE.EFI
  efi-cmdline=root=PARTUUID=<partuuid-of-part-1>
```

hnez (changes requested):
With the last commit the EFI section contains two nearly identical `system.conf` sections, which feels a bit redundant.
All the information is there, so we can just leave it as is, but if we find a way to have just one `system.conf` example that documents both `efi-loader`/`efi-cmdline` and the `bootname`/`parent` relationships, I think that would be great.

docs/integration.rst
jluebbe: We should still mention the `root=` and UKI profiles as alternatives, either here or in the reference.

Bastian-Krause: Sounds good, I'll add that.

Bastian-Krause: Done.

hnez (dismissed):
The note about UKIs should be helpful for anyone wanting to build a system with RAUC and Secure Boot.

Still not a friend of the redundancy of the `system.conf` snippets, but I think it is good enough.

ejoerns (dismissed):

docs/integration.rst
```diff
@@ -1264,6 +1287,15 @@ boot, ``<partuuid-of-part-1>`` with the PARTUUID of the first rootfs
 partition and ``<partuuid-of-part-2>`` with the PARTUUID of the second rootfs
 partition.
 
+.. note:: When booting `UKIs
+   <https://uapi-group.org/specifications/specs/unified_kernel_image/>`_
+   from redundant ESPs (EFI system partitions), the EFI command line (RAUC's
```
jluebbe: As there should only be one ESP, there can't be 'redundant ESPs'.

```suggestion
   from redundant partitions, the EFI command line (RAUC's
```

docs/reference.rst
```diff
@@ -652,6 +639,19 @@ hierarchical separator.
   Accepts integer values in bytes.
   Supports optional size suffixes: ``K``, ``M``, ``G``, ``T`` (powers of 1024).
 
+``efi-loader`` (optional)
+  Only valid when ``bootloader`` is set to ``efi``.
+  If set in combination with ``efi-cmdline``, an EFI boot entry for this slot
+  is (re)created with the given loader string via ``efibootmgr``'s ``--loader``
+  option during mark good/bad/active operations, if it is missing.
```
jluebbe, line 665: Mention that backslashes must be escaped as `\\`, following the general glib keyfile format.

jluebbe (changes requested):

docs/integration.rst
ejoerns: I'm not sure how a *partition loaded by EFI* differs from an *ESP*?

From the spec, there is no such limitation on the number of ESPs:

> UEFI does not impose a restriction on the number or location of System Partitions that can exist on a system.

https://uefi.org/specs/UEFI/2.11/13_Protocols_Media_Access.html#number-and-location-of-system-partitions

docs/reference.rst
ejoerns: Not strictly related, since this just moves the existing documentation, but could be added in this context.

docs/integration.rst
jluebbe: Hmm, I was not aware that this was allowed.

ejoerns: However, it might make sense to not use ESPs (as defined by the GPT UUID type) to prevent the firmware from creating or changing boot entries. Will reflect this in a follow-up.

jluebbe (approved):

PR #1872, opened 2026-02-04
improve PKI docs

jluebbe:
We've seen these issues in the wild a few times, so add warnings and highlight the existing ones.


docs/reference.rst
```diff
@@ -280,6 +280,15 @@ Example configuration:
 
 The ``keyring`` section refers to the trusted keyring used for signature
 verification.
+
+.. important::
+   The keyring must only contain certificates which are used as trust anchors.
+   If intermediate certificates are needed to complete the chain from trust
```
ejoerns, line 290:
```suggestion
   If intermediate certificates are needed to complete the trust chain from the trust
```

docs/reference.rst
```diff
@@ -280,6 +280,15 @@ Example configuration:
...
+   anchor to bundle signing certificate, they must be :ref:`included when
```
ejoerns, line 291:
```suggestion
   anchor to the bundle signing certificate, they must be :ref:`included when
```

docs/advanced.rst
```diff
@@ -41,8 +41,14 @@ development and release is generated by the script
...
 it is intended to exercise many different code paths in RAUC.
 
-If you are unsure which PKI to create, carefully consider if a simple
-:ref:`single key setup <sec-pki-single-key>` matches your requirements.
+.. warning::
+   If you are unsure which PKI to create, carefully consider if a simple
+   :ref:`single key setup <sec-pki-single-key>` matches your requirements.
+
+   Avoid using the same PKI for other purposes besides signing RAUC bundles,
+   because configuring this is challaging and prone to subtle mistakes.
```
ejoerns:
```suggestion
   because configuring this is challenging and prone to subtle mistakes.
```

docs/examples.rst
```diff
@@ -29,11 +30,10 @@ To create a simple key pair for testing, we can use ``openssl``::
 
   > openssl req -x509 -newkey rsa:4096 -nodes -keyout demo.key.pem -out demo.cert.pem -subj "/O=rauc Inc./CN=rauc-demo"
 
-For actual usage, setting up a real PKI (with a CA separate from the signing
-keys and a revocation infrastructure) is *strongly* recommended. OpenVPN's
-easy-rsa_ is a good first step. See :ref:`sec-security` for more details.
-
-.. _easy-rsa: https://github.com/OpenVPN/easy-rsa
+For actual usage, setting up a PKI which matches your use-case (e.g. with a CA
+separate from the signing keys and a revocation infrastructure) is *strongly*
```
ejoerns: I'm not sure if this aligns well with 

> If you are unsure which PKI to create, carefully consider if a simple single key setup matches your requirements.

ejoerns (commented):
@jluebbe Thank you for these good changes! Just a few remarks.

Should we link from *Update Bundle* chapter in `basict.rst`, too?:

> A key design decision of RAUC is that signing a bundle is mandatory.
> For development purpose a self-signed certificate might be sufficient,
for production the signing process should be integrated with your PKI
infrastructure.


In `test/openssl-ca-create.sh`

> The PKI generated by this script is much more complex than what is needed for
> almost all use-cases, as it is intended to exercise many different code paths
> in RAUC.

 we note that this PKI is complex, but not that it is potentially inappropriate for real-world use cases (with how it organizes dev and prod keys). Should we note this more explicitly?

For the commit title maybe add 'only' to

>docs: emphasise that intermediate certificates should be supplied via the bundle

docs/reference.rst
jluebbe: I think the readability is better the way I have it now.

docs/reference.rst
jluebbe: See above.

docs/reference.rst
ejoerns: I don’t agree, but I also don’t see a strong need to change it.

docs/examples.rst
jluebbe: I've removed the specifics from the example page. It's better to explain this in a single place.

ejoerns (approved):
I really appreciate the additional changes!

PR #1875, opened 2026-02-09
 src/main: extract: strip trailing output bundle slash

ejoerns:
Calling `rauc extract` with a trailing slash as output directory can lead to:

> % rauc extract […] out.raucb out/
> […]
> rauc-Message: 13:27:30.107: Verifying bundle payload...
>
> FATAL ERROR: dir_scan: failed to change permissions for directory out/, because Not a directory
> Failed to extract bundle: Failed to run unsquashfs: Child process exited with code 1

This will be the case when 'out' is an existing file since `g_file_test("out/")` will not return `TRUE` for a file. Fix this by simply stripping a trailing slash.

 src/main: extract: strip trailing output bundle slash

While at it, also fix the `resolve_path()` documentation (which caught my eye while thinking about whether to handle the slash there or in main).

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -769,20 +769,28 @@ static gboolean extract_start(int argc, char **argv)
 		return TRUE;
 	}
 
-	g_debug("input bundle: %s", argv[2]);
-	g_debug("output dir: %s", argv[3]);
+	g_autofree gchar* inbundle = g_strdup(argv[2]);
+	g_autofree gchar* outdir = g_strdup(argv[3]);
+
+	/* strip trailing slash for later path existence check */
+	if (g_str_has_suffix(outdir, "/")) {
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -769,20 +769,28 @@ static gboolean extract_start(int argc, char **argv)
 		return TRUE;
 	}
 
-	g_debug("input bundle: %s", argv[2]);
-	g_debug("output dir: %s", argv[3]);
+	gchar* inbundle = argv[2];
+	gchar* outdir = argv[3];
```
jluebbe: These should be `const` or a copy if we need to modify them.

jluebbe (changes requested):

jluebbe (approved):

PR #1877, opened 2026-02-13
.github/workflows: use separate workflow for building docs

ejoerns:
This avoids full builds for docs-only changes.


jluebbe (approved):

ejoerns:
Maybe, but also since I changed something outside `docs/`.

PR #1884, opened 2026-03-03
data/rauc.service.in: configure Restart and RestartSec

ejoerns:
If the RAUC systemd service ran and then crashed for some reason, it should be restarted.

While this might not strictly be necessary if D-Bus activation is used, it is useful if not.
With the added polling support, having a running RAUC service also becomes more relevant (even though it is still stateless across restarts).

Wait at least 5 seconds to not restart too often.


ejoerns:
Yes, thought about this, too, but wasn't sure. I had 30 seconds in another version but maybe something like "5 minutes" is still sufficient for actual devices in the field.

jluebbe (approved):

PR #1894, opened 2026-03-18
Create the `hashref` type

jweyer8:
### What do you use the feature for?
Cloning the active (booted) slot to the inactive slot in an A/B layout without bundling a full image. A bundle with type=hashref contains only the expected SHA-256 of the active slot's content. This is useful for use cases where the desired inactive-slot content is already present on the running device.

### How does RAUC benefit from the feature?
It enables a new class of minimal bundles that express a content dependency rather than carrying payload. Instead of transferring a potentially large image over the network or storing it in the bundle, RAUC can clone an existing slot device directly. This reduces bundle size to nearly zero for the affected image and removes the network/storage cost of re-delivering content that is already on the device. 

### How did you verify the feature works?
Unit tests were added to manifest.c update_handler.c. The changes have also
 been tested on our hardware


jluebbe:
Did you consider using [adaptive updates](https://rauc.readthedocs.io/en/latest/advanced.html#adaptive-updates) (see also the [related blog post](https://pengutronix.de/en/blog/2022-10-12-rauc-adaptive-updates.html))?

It might give you most of the benefit you're looking for with this change. If adaptive updates doesn't fit your case, I'd be very interested to know why. :)

jluebbe:
To ensure that we really get the expected contents, we should probably hash data during copying and abort if that detects any changes. Do you see any problems with that?

jweyer8:
@jluebbe We did look into adaptive updates. The main reason we decided against that approach is that we want to keep the RAUC bundle as small as possible.

Our use case is a bit different:

Our hardware is deployed in air-gapped environments. To perform an update, we provide the customer with a RAUC bundle, which they copy onto a USB drive and insert into the hardware. The hardware then serves the RAUC bundle to all connected devices.

The issue with adaptive updates is that the bundle still has to contain the entire rootfs, which significantly increases its size. We'd really like to avoid large bundles because distributing them to customers can be challenging. Also, the vast majority of our updates only contain firmware or application changes on the appfs slot, so including the rootfs and other unchanged slots is wasteful in terms of both bundle size and update time.

Beyond our potentially unique situation, I would think that reducing bundle size would be beneficial to all customers if only to save server space. I also don't really see any downsides to this approach, but I could certainly be missing something. Obviously, this approach would only work for customers with readonly rootfs, but I would guess that this isn't too uncommon.

jweyer8:
> To ensure that we really get the expected contents, we should probably hash data during copying and abort if that detects any changes. Do you see any problems with that?

Off the top of my head, I don't see any issues here, probably a good move. I will try to find some time to look into this further.

PR #1752, opened 2025-07-04
eMMC Update Handler Refactoring

ejoerns:
Next to fixing a minor bug and improving manual testing of eMMC boot partition handling, this removes some unnecessary code duplication and prepares for a broader use of the eMMC update handling.

PR #1901, opened 2026-03-27
fix: 'assertion 'slot->name' failed' error for bootloader #1893

### LINKED ISSUE #1893: eMMC bootloader hook does not work with RAUC 1.15
**Describe the bug**
A device of type `boot-emmc` can't have a hook script with RAUC 1.15

**Background information**
RAUC 1.15 with Yocto, using meta-rauc layer.

**To Reproduce**
the RAUC system.conf should have this:
```
[slot.bootloader.0]
device=/dev/mmcblk0
type=boot-emmc
```

The yocto build has a hook defined:
```
inherit bundle

[...]

RAUC_SLOT_bootloader = "u-boot"
RAUC_SLOT_bootloader[type] = "boot"
RAUC_SLOT_bootloader[file] = "u-boot-sunxi-with-spl.bin"
RAUC_SLOT_bootloader[hooks] = "post-install"
```

**Expected behavior**
Post-install hook is called after installing uboot.  This works fine with RAUC 1.14

**Logs**
The following will be printed by RAUC:
```
(rauc:873): rauc-CRITICAL **: 17:31:29.154: run_slot_hook_extra_env: assertion 'slot->name' failed
```

**Additional context**
see also: https://github.com/rauc/rauc/discussions/1891

flobz:
Hello,
Just a fix for 'slot->name' failed' error for bootloader.

Fixes: #1893

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -2369,6 +2369,10 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot, 
 	 * partition as device.
 	 */
 	part_slot = g_new0(RaucSlot, 1);
+	part_slot->name = dest_slot->name;
+	part_slot->sclass = dest_slot->sclass;
+	part_slot->type = g_strdup(dest_slot->type);
+	part_slot->state = dest_slot->state;
 	part_slot->device = g_strdup_printf(
 			"%sboot%d",
 			realdev,
--- a/test/test_write_slot.py
+++ b/test/test_write_slot.py
@@ -1,5 +1,6 @@
 import os
 from subprocess import check_call
+from textwrap import dedent
 
 from helper import run
 from conftest import needs_emmc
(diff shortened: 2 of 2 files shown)


ejoerns:
The current handling with the temporary slot has never felt perfect, but this case shows that it really lacks completeness.
The solution you propose is a valid option, though I would then prefer defining and using a slot duplication method explicitly (and replacing only the device afterwards).

However, I guess [my original version](https://github.com/rauc/rauc/compare/980d062142c4ca8100ffef1da1b76f2e3741c3a9..1245bf56d21b48b8ab654a860fe78ef56c287ce4) of the [rework](https://github.com/rauc/rauc/pull/1752), that I was [convinced to drop](https://github.com/rauc/rauc/pull/1752#discussion_r2210596282) in the final phase of refactoring would have already solved the issue, too, since it uses the original slot with all fields set.

jluebbe (approved):

jluebbe:
I force-pushed to add a commit message

PR #1902, opened 2026-03-29
fix a few sign conversion warnings

jluebbe:
These are reported with `-Wsign-conversion`.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -550,7 +550,7 @@ GBytes *cms_sign(GBytes *content, gboolean detached, const gchar *certfile, cons
 	g_autoptr(R_X509_STACK_POP) intercerts = NULL;
 	g_autoptr(CMS_ContentInfo) cms = NULL;
 	GBytes *res = NULL;
-	int flags = CMS_BINARY | CMS_NOSMIMECAP;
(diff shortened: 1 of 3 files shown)


pH5 (approved):

src/signature.c
```diff
@@ -895,8 +895,10 @@ gchar* sigdata_to_string(GBytes *sig, GError **error)
 	return ret;
 }
 
-static void bio_print_recipient(BIO *text, guint id, gchar* algorithm, ASN1_OCTET_STRING *keyid, X509_NAME *issuer, ASN1_INTEGER *sno)
+static void bio_print_recipient(BIO *text, int id, gchar* algorithm, ASN1_OCTET_STRING *keyid, X509_NAME *issuer, ASN1_INTEGER *sno)
```
ejoerns, line 898: I would have argued that calling it with `guint` would be more consistent (especially since its used in only a single loop), but I see the potential reasoning for using this solution is that `sk_CMS_RecipientInfo_value()` takes an int as argument.

However, it seems a bit unrelated to the other two flag changes in this commit and might deserve a separate one.

ejoerns (approved):
Rest looks fine to me, too.

PR #1917, opened 2026-04-15
Revise "Data Storage and Migration" and add Examples for Migration via Hooks

ejoerns:
The flow and structure of the "Data Storage and Migration" chapter was quite inconsistent.
Cleaned this up before adding the relevant new bits: Examples for hook-based rootfs migration and migration of redundant data partitions using hooks and the new `empty-fs` image type.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -939,8 +939,12 @@ In these cases, a separate data partition is useful.
 Using Data Partitions
 ~~~~~~~~~~~~~~~~~~~~~
 
-If the update concept uses full redundant root file systems,
-there are also good reasons for using a redundant data storage, too.
+A separate data partition provides a dedicated, independent storage location
+for your persistent data outside the rootfs image.
+
+While a single data partition is the simplest setup, using redundant data
+partitions, one per rootfs slot, can be preferable when data migration and
+fallback behaviour are a concern.
 Read below about the possible impact on data migration.
 
 To let your system access the separate storage location, it has to be mounted
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -918,11 +918,7 @@ When performing the update, you have to ensure that the files you need to
 preserve are copied to the target slot after having written
 the system data to it.
 
-RAUC provides support for executing *hooks* from different slot installation
-stages.
-For migrating data from your old rootfs to your updated rootfs,
-simply specify a slot post-install hook.
-Read the :ref:`Hooks <sec-hooks>` chapter on how to create one.
+See :ref:`sec-data-migration-rootfs` for information migrating data in the rootfs.
 
 However, there are several reasons why you may not want to or cannot store
 your data inside the root file system:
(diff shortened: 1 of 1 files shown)


docs/advanced.rst
```diff
@@ -890,44 +890,58 @@ If needed, this could be implemented in a web application or using a reflink-cap
...
 
-RAUC provides support for executing *hooks* from different slot installation
-stages.
-For migrating data from your old rootfs to your updated rootfs,
-simply specify a slot post-install hook.
-Read the :ref:`Hooks <sec-hooks>` chapter on how to create one.
-
-Using Data Partitions
-~~~~~~~~~~~~~~~~~~~~~
+See :ref:`sec-data-migration-rootfs` for information on migration rootfs data.
```
jluebbe, line 921:
```suggestion
See :ref:`sec-data-migration-rootfs` for information on migrating data in the rootfs.
```

docs/advanced.rst
```diff
@@ -978,34 +1015,118 @@ are listed below:
...
+.. code-block:: cfg
 
-  #!/bin/bash
+   [image.data]
+   filename=rootfs.img
+   hooks=post-install
 
-  ROOTFS_DEV=<determine rootfs by using proc cmdline or mount>
-  TEST_DEV=<obtain parent rootfs device for currently processed device (%k)>
+.. note:: This requires a writeable rootfs slot (type).
```
jluebbe, line 1039:
```suggestion
.. note:: This requires a rootfs slot type which can be mounted read-write (e.g. ext4).
```

docs/advanced.rst
```diff
@@ -978,34 +1015,118 @@ are listed below:
...
 
-  if [[ $ROOTFS_DEV == $TEST_DEV ]]; then
-  	echo 1
-  else
-  	echo 0
-  fi
+The hook will run after the new rootfs image has been written to the target
+slot, with the target slot still mounted.
```
jluebbe, line 1042:
```suggestion
slot, with the target slot mounted.
```

When using images for ext4 slots, it will be mounted after writing the image, so "still" is not really correct.

docs/advanced.rst
```diff
@@ -978,34 +1015,118 @@ are listed below:
...
+                   echo "Preserving /etc/passwd and /etc/shadow"
+                   cp -a /etc/passwd "$RAUC_SLOT_MOUNT_POINT/etc/"
+                   cp -a /etc/shadow "$RAUC_SLOT_MOUNT_POINT/etc/"
+                   ;;
+           *)
+                   exit 1
+                   ;;
+   esac
+
+   exit 0
```
jluebbe: Add a note that this is focused on the basic functionality and integration with RAUC, and that error handling was intentionally left out.

docs/advanced.rst
```diff
@@ -1040,6 +1040,65 @@ A hook script could look as follows:
 
    exit 0
 
+Migration for Redundant Data Partitions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Migrating data between redundant data partitions is quite specific.
```
jluebbe, line 1075: Specific to what? Perhaps drop this sentence?

docs/advanced.rst
```diff
@@ -978,34 +1015,118 @@ are listed below:
...
+
+   [slot.data.1]
+   type=ext4
+   ...
+
+The manifest should then specify ``type=empty-fs``,
+which will make RAUC format the target slot depending on its slot type
+(only ext4 is supported for now).
+For the actual data copying or migration, a ``post-install`` hook must be used,
+which will be called with the target slot already mounted at ``$RAUC_SLOT_MOUNT_POINT``:
```
jluebbe, line 1095:
```suggestion
which will be called with the target slot mounted at ``$RAUC_SLOT_MOUNT_POINT``:
```

docs/advanced.rst
```diff
@@ -978,34 +1015,118 @@ are listed below:
...
+
```
jluebbe, line 1096: Perhaps add: "Conceptually, RAUC behaves just as if the bundle contains an empty tar archive for this slot."

docs/advanced.rst
```diff
@@ -978,34 +1015,118 @@ are listed below:
...
+                   ;;
+   esac
+
+   exit 0
+
+.. note::
+
+   This example assumes the currently active data partition is mounted at ``/data``.
+   The ``--delete`` flag ensures the target slot is an exact mirror of the source,
+   removing any stale files left from a previous installation.
```
jluebbe, line 1128: Also add a note about error handling.

docs/advanced.rst
```diff
@@ -978,34 +1015,118 @@ are listed below:
...
+
+An example hook could look as follows:
+
+.. code-block:: sh
+
+   #!/bin/sh
+
+   case "$1" in
+           slot-post-install)
+                   # only data should be handled
```
jluebbe, line 1111:
```suggestion
                   # only the data slots should be handled
```

docs/advanced.rst
```diff
@@ -957,6 +957,35 @@ You can do this by using:
 It depends on the amount and type of data you want to handle which option you
 should choose.
 
+Managing a ``/dev/data`` Symbolic Link
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For redundant data partitions the active rootfs slot has to mount the correct
+data partition dynamically.
+For example with ubifs, a udev rule set can be used for this::
```
jluebbe, line 965: Shouldn't we recommend the symlinks RAUC creates in `/run/rauc` for that? Can be a separate PR though.

jluebbe (changes requested):

docs/advanced.rst
ejoerns: I switched to make it less look like a comprehensive example instead with adding ellipsis and noting 'simplified shell script code'. This also aligns better with existing examples.

docs/advanced.rst
ejoerns: Without this, also the following sentence wouldn't make sense. Thus I tried to be more specific about what's specific here instead.

docs/advanced.rst
ejoerns: See above.

docs/advanced.rst
ejoerns: I've added a commit addressing this.

docs/advanced.rst
```diff
@@ -978,34 +1031,118 @@ are listed below:
...
+A simplified hook shell script code could look as follows:
+
+.. code-block:: sh
+
+   case "$1" in
+
+           [...]
+
+           slot-post-install)
+                   # only the data slots should be handled
```
jluebbe:
```suggestion
                   # only the rootfs slots should be handled
```

jluebbe (changes requested):

jluebbe (approved):

PR #1920, opened 2026-04-19
Add optional password prompting for Basic Auth and mask credentials in log output

obbardc:
This PR improves handling of HTTP Basic Authentication for remote bundle access to avoid leaking credentials in logs and also provides a secure way for users to supply passwords interactively with `rauc status` and `rauc info` instead of embedding them in the URL in cmdline arguments which may be logged and thus leak.

Draft as I haven't fully checked all code paths yet. Please let me know if you don't think this is useful and I will close it.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -499,7 +499,8 @@ Authentication
 ~~~~~~~~~~~~~~
 
 To use Basic Authentication, you can add the username and password to the bundle
-URL (``rauc install https//user:password@example.com/update.raucb``).
+URL (``rauc install https://user:password@example.com/update.raucb``).
+The password is masked as ``******`` in log output to avoid leaking credentials.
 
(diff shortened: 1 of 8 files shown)


jluebbe:
I find hiding the password in logs useful and we should do that. `r_censor_url()` and the related changes should be a separate PR and we can merge that first.

We shouldn't modify URLs by hand though. I'd prefer updating the glib dependency to 2.66 (see https://github.com/rauc/rauc/commit/fa224b83a48c09ff46eb4ab6bbec6d11b5e232d4) and rely on the [GURI functions](https://docs.gtk.org/glib/struct.Uri.html#parsing-uris). The HTTP User can be supplied in the URL, so we don't need a new CLI option for that. Also if GURI can distinguish between missing and empty passwords (`https://foo@example.com/` vs `https://foo:@example.com/`) we could also automatically as for the password if it is missing (not empty) and avoid that CLI option as well.

PR #1922, opened 2026-04-23
src/install: make RAUC_SYSTEM_COMPATIBLE available to system handlers

ejoerns:
This commit adds the `RAUC_SYSTEM_COMPATIBLE` environment variable to system handlers.


jluebbe (changes requested):
The commit message should mention why handlers need access to this.

ejoerns:
@jluebbe I thought it might be sufficient to align this with e.g. `RAUC_SYSTEM_VARIANT` which also did not have more justification in 1b20840cee1dc904463b4e88eeccdac8c4a60659. I explicitly noted this now. Additionally, I've added some use-case motivation.

jluebbe (approved):

PR #1892, opened 2026-03-10
qemu-test: add SSH access support for the test VM

jluebbe:
Add an `ssh` option to `qemu-test` that enables SSH access into the running QEMU VM on port 2222. When used, this:

- Generates an ed25519 keypairs (host and user) on first use
- Forwards host port 2222 to guest port 22
- Configures and starts `sshd` inside the guest
- Sets up authorized_keys for root in the guest

A `qemu-ssh` helper script is provided to connect with all necessary options pre-configured.

Also add a `background` option that starts the VM as a daemon with serial output redirected to `qemu-test-log`. This also implies starting the ssh daemon. A QMP unix socket (`.qemu-test-qmp`) is always set up to allow VM control.

The `qemu-test-stop` helper script sends the QMP quit command via `socat` to gracefully shut down the VM.

Usage:
```
  ./qemu-test shell ssh   # start VM with SSH enabled
  ./qemu-test background  # start VM in background with SSH enabled
  ./qemu-ssh              # connect to the running VM
  tail -f qemu-test-log   # watch serial output (background mode)
  ./qemu-test-stop        # stop the VM (background mode)
```

PR #1924, opened 2026-05-07
document qemu-ssh and improve environment handling

jluebbe:
This addresses https://github.com/rauc/rauc/pull/1892#issuecomment-4245075960.

### DIFF
--- a/.qemu-test-sshd-config
+++ b/.qemu-test-sshd-config
@@ -3,6 +3,7 @@ HostKey /etc/ssh/ssh_host_ed25519_key
 AuthorizedKeysFile /root/.ssh/authorized_keys
 PermitRootLogin prohibit-password
 PermitUserEnvironment yes
+AcceptEnv HOST_PWD PWD TERM
 StrictModes no
 PasswordAuthentication no
 ChallengeResponseAuthentication no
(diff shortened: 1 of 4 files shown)


.qemu-test-sshd-config
```diff
@@ -3,6 +3,7 @@ HostKey /etc/ssh/ssh_host_ed25519_key
 AuthorizedKeysFile /root/.ssh/authorized_keys
 PermitRootLogin prohibit-password
 PermitUserEnvironment yes
+AcceptEnv HOST_PWD PWD TERM
```
ejoerns, line 6: `PWD` isn't really required here (and in `SendEnv`), is it?

ejoerns:
Looks fine, otherwise.

.qemu-test-sshd-config
jluebbe: It's still useful for debugging and doesn't hurt, especially on a dev setup like this.

ejoerns: Would have expected it is overridden in the shell instantly, but maybe I'm wrong.

ejoerns: Anyway, it shouldn't break anything, thus let's get this merged.

ejoerns (approved):

PR #1933, opened 2026-05-13
.github/workflows: add workflow for scorecard analysis

jluebbe:
This checks several best practices and generates a report. Running the workflow will add more details to https://securityscorecards.dev/viewer/?uri=github.com/rauc/rauc.

See https://github.com/ossf/scorecard?tab=readme-ov-file#what-is-scorecard for more details.

### DIFF
--- a/.github/workflows/scorecard-analysis.yml
+++ b/.github/workflows/scorecard-analysis.yml
@@ -1,0 +1,57 @@
+name: Scorecard analysis workflow
+on:
+  push:
+    # Only the default branch is supported.
+    branches:
+    - master
+  schedule:
+    # Weekly on Saturdays.
+    - cron:  '30 1 * * 6'
+  workflow_dispatch: {}
+
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard-analysis.yml
```diff
@@ -0,0 +1,57 @@
+name: Scorecard analysis workflow
+on:
+  push:
+    # Only the default branch is supported.
+    branches:
+    - master
+  schedule:
+    # Weekly on Saturdays.
+    - cron:  '30 1 * * 6'
+  workflow_dispatch: {}
```
pH5, line 10: What is the point of the `workflow_dispatch` trigger? It's not in the [reference example](https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml) and the [README](https://github.com/ossf/scorecard-action/blob/main/README.md) calls it experimental.

jluebbe: This can be used to manually trigger a run of this workflow. In my tests it seemed to work fine.

pH5 (approved):

PR #1940, opened 2026-05-15
.github/workflows: restrict permissions to read contents

jluebbe:
This removes unnecessary permissions from our GitHub actions workflows.

Skip the container workflow for now. It will be updated in a separate commit.


pH5 (approved):
Looks good to me.

PR #1942, opened 2026-05-15
src/utils: rename value_check_tab_whitespace and check for any whitespace

jluebbe:
Rename the helper to value_check_whitespace and reject any character matched by g_ascii_isspace (space, tab, CR, LF, VT, FF) instead of just space and tab. This makes the helper safe to reuse for fields where any embedded whitespace is unsafe (such as device paths emitted as shell environment variables or space-joined output).

It is currently only used to check the bootname and any of the previously unchecked characters would already have caused problems, so this should not cause compatibility problems.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -408,17 +408,18 @@ gchar * key_file_consume_string(
 	return result;
 }
 
-gboolean value_check_tab_whitespace(const gchar *str, GError **error)
+gboolean value_check_whitespace(const gchar *str, GError **error)
 {
 	g_return_val_if_fail(str != NULL, FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
-	if (strchr(str, '\t') || strchr(str, ' ')) {
-		g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_PARSE,
-				"The value '%s' can not contain tab or whitespace characters",
-				str
-				);
-		return FALSE;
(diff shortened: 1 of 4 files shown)


src/utils.c
```diff
@@ -408,17 +408,18 @@ gchar * key_file_consume_string(
...
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
-	if (strchr(str, '\t') || strchr(str, ' ')) {
-		g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_PARSE,
-				"The value '%s' can not contain tab or whitespace characters",
-				str
-				);
-		return FALSE;
+	for (const gchar *p = str; *p; p++) {
+		if (g_ascii_isspace((guchar) *p)) {
```
pH5: [Unnecessary](https://gitlab.gnome.org/GNOME/glib/-/blob/glib-2-64/glib/gstrfuncs.h?ref_type=heads#L81) cast.

src/utils.c
```diff
@@ -408,17 +408,18 @@ gchar * key_file_consume_string(
...
+			g_set_error(error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_PARSE,
+					"The value '%s' can not contain whitespace characters",
```
pH5: `s/can/must` while we're at it?

pH5 (changes requested):

pH5 (approved):

PR #1947, opened 2026-05-19
support overriding kernel commandline and runtime dir

jluebbe:
This makes it preparing the test environment easier and more deterministic.

### DIFF
--- a/include/context.h
+++ b/include/context.h
@@ -63,6 +63,8 @@ typedef struct {
 	gchar *machine_id;
 	/* kernel commandline (real or testing override) */
 	gchar *cmdline;
+	/* runtime direcrory (/run/rauc or testing override) */
+	gchar *runtime_directory;
 
 	gchar *system_serial;
 	gchar *system_version;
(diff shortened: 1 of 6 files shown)


include/context.h
```diff
@@ -61,6 +61,10 @@ typedef struct {
 	gchar *bootslot;
 	gchar *boot_id;
 	gchar *machine_id;
+	/* kernel commandline (real or testing override) */
+	gchar *cmdline;
```
pH5, line 65: Why store `cmdline` in the context? This could be a local variable in `r_context_configure_target()`, passed as parameter into `get_bootname()`.

Edit: Actually, it looks like this patch could be dropped altogether, cmdline is only ever used in `get_bootname()`.

pH5 (changes requested):

include/context.h
jluebbe: In upcoming changes we'll need to access the cmdline from update handlers as well. Storing it in the context gives us a single source of truth.

pH5: > In upcoming changes we'll need to access the cmdline from update handlers as well.

I'd add this to the patch description.

jluebbe: I've update the commit message.

ejoerns (approved):

pH5 (approved):

PR #1957, opened 2026-05-29
tests/doc: use spec-conform UKI profile invocation parameter

Bastian-Krause:
The "UAPI.5 Unified Kernel Images (UKI)" [1] reads:

> A profile is (optionally) selected by prefixing the EFI stub’s invocation parameters (“command line”) with `@0 `, `@1 `, `@2 `, (i.e. an `@` character, the numeric profile index, and a space character) in order to select the desired profile."

While at least the systemd-stub implementation works without a trailing space, document and test the spec-conform invocation.

Note that neither `efi-loader` nor `efi-cmdline` are currently persisted/printed in the `test/bin/efibootmgr` mock, so the change in the tests has no direct effect and is only made for the sake of correctness.

[1] https://uapi-group.org/specifications/specs/unified_kernel_image/#multi-profile-ukis

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -1321,8 +1321,8 @@ partition.
    ``efi-cmdline`` or efibootmgr's ``--unicode``) can point to a
    `UKI profile
    <https://uapi-group.org/specifications/specs/unified_kernel_image/#multi-profile-ukis>`_
-   (e.g. ``@1``) defining the corresponding ``root=`` and ``rauc.slot=`` kernel
-   command line parameters.
+   (e.g. ``@1\s`` in RAUC's system configuration) defining the corresponding
+   ``root=`` and ``rauc.slot=`` kernel command line parameters.
 
 You can inspect and verify your settings by running:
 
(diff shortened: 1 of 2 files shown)


hnez (approved):
I was a bit confused when reading the escaped space `\s`  in the `.rst` file assumed that it would be parsed by reStructuredText to get a literal space in the output, because `.rst` never quite behaves like I expect it to.

Then I saw the test and it became clear that the `\s` is parsed by the config parser instead.

Is there some other way to encode a trailing space in the config file format? Like i.e. `efi-cmdline="@1 "`? That would be a bit less confusing in my point of view (the space character may be a bit easy to miss in that case though).

Long story short. The changes in this PR are technically correct, which is the best kind of correct, so you have my approval.

Bastian-Krause:
I was unsure if pointing to https://docs.gtk.org/glib/struct.KeyFile.html would be appropriate in RAUC's docs because it feels like an implementation detail.

Here's the relevant part:
> Newline, tab, carriage return and backslash characters in value are escaped as \n, \t, \r, and \\\\, respectively. To preserve leading spaces in values, these can also be escaped as \s.

Bastian-Krause:
Testing this one more time: I don't think we need the escaping at all. I thought trailing spaces would be dropped, but that does not seem to be the case. Do you think that's better? Quoting does not work in GLib's KeyFile.

jluebbe (approved):

hnez:
> Do you think that's better?

No. That does not sound good. There is some markup language that respects trailing whitespace and in my opinion that is the craziest thing ever. Let's just keep it as is.

No more objections from my side.

ejoerns (approved):

PR #1961, opened 2026-06-08
service: reload polling config on SIGHUP

### LINKED ISSUE #1709: Add support configuration reloading for the service
With the upcoming polling support in #1672, it would be useful to allow changes to the configuration without restarting the service.

We should add a SIGHUP handler to trigger a config reload. If RAUC is currently busy, it should postpone the reload until the ongoing operation has completed. After the config reload, the slot status als needs to be reloaded.

Simply restarting the service would also work, but that would potentially interrupt an ongoing installation.

preko-p:
Fixes #1709.

## Summary

- Add service-mode `SIGHUP` handling for polling-only `system.conf` reloads.
- Reload only `[polling]` keys and reject polling enable/disable at runtime.
- Reject every non-polling config change instead of performing a general service config reload.
- Defer reload while RAUC is busy and drain it once the service is idle.
- Reset stale polling status/cache/error state while preserving pending automatic reboot and installation-running state.
- Document the SIGHUP polling reload operator contract.

## Notes

This intentionally does not call `r_context_configure()` on the reload path and does not replace the live config object wholesale.

Rejected reloads update the shared `Installer.LastError` property in addition to the service log because `SIGHUP` has no direct reply channel. Accepted reloads clear that property. Please confirm whether this D-Bus visibility is preferred over log-only behavior.

The non-polling rejection is fail-closed and may reject comment-only edits or reordering outside `[polling]`; this is documented so operators do not confuse it with a general config reload.

## Verification

- `git diff --check`
- `PYTHONPYCACHEPREFIX=/tmp/rauc-pycache python3 -m py_compile test/test_polling.py`

I could not run the full RAUC local suite in my environment because `meson`, `ninja`, `pytest`, `pkg-config`, `uncrustify`, `dasbus`, and required C headers were unavailable. The PR should be gated on upstream CI / a properly provisioned local RAUC environment for the C tests, polling pytest suite, service tests, docs build, and formatting.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -637,6 +637,42 @@ As long as any of the file listed in ``inhibit-files`` exist, no polling or
 installation is started.
 Note that a running poll or installation is not aborted.
 
+.. _sec-polling-runtime-reload:
+
+Runtime Reload with SIGHUP
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When RAUC runs in service mode and polling was enabled when the service
+started, sending ``SIGHUP`` to the service requests a polling-only reload of
+``system.conf``.
+This can be used to update polling policy without restarting the service.
+It is only active for polling-enabled service mode; it does not provide a
+general configuration reload and does not enable polling on a service that
+started without it.
+
+The reloadable ``[polling]`` keys are ``url``, ``interval-sec``,
+``max-interval-sec``, ``inhibit-files``, ``candidate-criteria``,
+``install-criteria``, ``reboot-criteria`` and ``reboot-cmd``.
+Adding or removing the ``[polling]`` section, or adding or removing the
+required ``polling.url``, is rejected.
+Any change outside ``[polling]`` is also rejected fail-closed; this may include
+comment-only edits or section/key reordering outside ``[polling]``.
+
+If RAUC is installing when ``SIGHUP`` is received, the reload is deferred until
+the service is idle.
+An accepted reload clears cached polling bundle/status/error/backoff state and
+schedules a new initial poll with the updated interval settings.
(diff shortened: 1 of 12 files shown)


jluebbe:
Thanks for this PR! What does your test setup look like beyond the CI? Which cases have to tried on real hardware?

Also, did you use any AI tools to assist with the development of this PR? If so, which one(s)?

preko-p:
Thanks for asking. Beyond CI, I did not run this on real hardware yet. My local validation was limited because this runner did not have the full RAUC build/test stack available. The local checks I ran before opening the PR were:

- `git diff --check`
- Python syntax checks for the changed polling pytest file

I could not run the full RAUC C tests, polling pytest suite, service tests, docs build, or uncrustify locally because the required tooling/dependencies were unavailable in that environment, so I would treat CI and/or a properly provisioned RAUC dev setup as the real gate here.

The cases I intended the added tests to cover are:

- accepting a polling-only SIGHUP reload
- rejecting non-polling config changes fail-closed
- rejecting polling enable/disable at runtime
- deferring reload while an install is busy
- draining the deferred reload after both polling-triggered and D-Bus-triggered installs
- preserving a pending reboot after an accepted reload

No real hardware cases were tried yet. The hardware-sensitive cases I would want checked before relying on this operationally are at least: SIGHUP during an active install on the target, an accepted polling-only config change followed by the next poll, and a rejected non-polling config change leaving the old polling behavior active.

AI assistance disclosure: yes, I used Codex CLI to help implement the patch and Claude Opus 4.8 to review it. I then reviewed the result and kept the PR disclosure conservative about what was and was not tested locally.

jluebbe:
Before we can review this, it needs to be tested in a real environment and the changes should be split up into meaningful commits (if possible).

PR #1969, opened 2026-06-17
src/update_handler: avoid redundant flash_erase call

jluebbe:
The `--erase-all` option for `flashcp` does the same as the 'flash_erase <dev> 0 0' we currently run do before each nor_write_slot() call. Previously, we erased all blocks covered by the image twice, as `flash_cp` already erases the blocks it writes to by default.

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -1185,9 +1185,10 @@ static gboolean nor_write_slot(const gchar *image, const gchar *device, GError *
 {
 	GError *ierror = NULL;
 	gboolean res = FALSE;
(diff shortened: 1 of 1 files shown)


ejoerns:
That `--erase-all` eliminates the need to call `flash_format_slot()` is quite clear.
To understand why we erased twice, it might be worth shortly noting that 'flashcp' erases at least the blocks covered by the file to write.

Nit: In the commit message, just drop the "do" before "before".

Otherwise: Looks fine to me.

jluebbe:
I've updated the commit message.

ejoerns (approved):

PR #1972, opened 2026-06-30
improve RaucInstallArgs->cleanup callback

jluebbe:
This should have not effect on behavior and make it clearer how the cleanup callback is called.


jluebbe:
@ejoerns mentioned that the cleanup function is also called from main. I need to check this again.

jluebbe:
@ejoerns I've changed the commit message to clarify how the callback is called and why I change the signature.

pH5 (approved):
The first patch has typos in the commit message: _etiher_ -> _either_, _It's_ -> _Its_. Contents are good.

PR #1977, opened 2026-07-07
data/de.pengutronix.rauc.conf: Allow only root to talk to rauc over DBus

lnussel:
Random unprivileged user accounts shouldn't be allowed to trigger rauc. So limit access to the dbus service to `root` by default. Adding those rules may cause regressions in deployments that rely on rauc being wide open though.

### DIFF
--- a/data/de.pengutronix.rauc.conf
+++ b/data/de.pengutronix.rauc.conf
@@ -2,14 +2,37 @@
  "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
  "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
 <busconfig>
-  <!-- This config allows anyone to control rauc -->
+  <!-- This config allows only root to control rauc -->
   <!-- It is usually installed to /usr/share/dbus-1/system.d -->
 
   <policy context="default">
-    <allow send_destination="de.pengutronix.rauc"/>
+    <deny send_destination="de.pengutronix.rauc"/>
+    <allow receive_sender="de.pengutronix.rauc"/>
+
+    <!-- allow introspection -->
+    <allow send_destination="de.pengutronix.rauc"
+	   send_interface="org.freedesktop.DBus.Introspectable"/>
+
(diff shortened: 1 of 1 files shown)


lnussel:
unrelated test failure due to openssl on trixie?

jluebbe:
All update bundles are signed, so only authentic updates can be installed. Having a non-root UI application or network daemon trigger the update is a common use-case. With this change, you'd need full root privileges just to trigger an update, which is something I'd recommend against.

Having an example dbus config for the case where only a specific user or group can trigger an installation would be useful for more restrictive setups, though.

lnussel:
The pr is both limiting the access to root by default and providing an example :)

The current config unfortunately does not easily allow to limit access by means of drop-in. The default context allows anyone to send. Since dbus-daemon seems to rely on fs ordering when reading configs, we can't be sure an override would work (https://dbus.freedesktop.org/doc/dbus-daemon.1.html). AFAICT the default has to be restrictive so vendors can add snippets with user policy to open access selectively again.

Anyway, the rauc dbus interface is not a simple install() resp update() method that does whatever is needed as configured by root. Instead one passes a URL that is then fetched by the rauc service (as root?). So this can be used to attack networks from within by unprivileged, sandboxed accounts that may not even have any network access but need to talk to dbus.
Also there does not seem to be a requirement to go strictly forward (and there's a parameter to disable min version check) so triggering downgrades to potentially vulnerable previous versions seems possible.

jluebbe:
Changing the default would likely break existing deployments in unexpectedly.

If you need to have a different default in your use-case, you'd simply deploy that custom configuration in your image. Approach is the same for D-Bus policy files, RAUC's system.conf (e.g. for the minimum version checks) or any other system configuration affecting RAUC. I see restricting D-Bus messages to RAUC similar to configuring a network firewall: evaluating the requirements and implementing the restrictions is an essential part of building an embedded system.

Regarding the network update: The network access is done by and unpriviledged helper process and the payload is authenticated via dm-verity.

I'd be happy to take your restrictive config as an additional example. In the medium term, having support for PolicyKit in RAUC would be nice.

PR #1986, opened 2026-07-14
src/signature: validate signing time attribute type before use

zayeem06:
When use-bundle-signing-time is enabled, cms_verify_bytes reads the pkcs9 signingTime attribute out of the CMS before the signature is verified and dereferences so->value.utctime after only a NULL check. Because the attribute value is an ASN.1 ANY, a bundle whose signingTime is encoded as a non-time tag such as a BOOLEAN leaves that union member holding an int instead of a pointer, so the NULL check passes and ASN1_TIME_to_tm then dereferences a bogus pointer and crashes during verification. This checks the tag is UTCTime or GeneralizedTime before touching the value, and reads it through so->value.asn1_string.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -1484,8 +1484,22 @@ gboolean cms_verify_bytes(GBytes *content, GBytes *sig, X509_STORE *store, CMS_C
 			goto out;
 		}
 
+		/* The signingTime attribute value is an ASN.1 ANY that is decoded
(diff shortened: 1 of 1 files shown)


jluebbe (approved):

jluebbe:
@zayeem06 I've added the PR description to the commit.

ejoerns (approved):

PR #1988, opened 2026-07-14
src/signature: various cleanups to the signing time code

jluebbe:


jluebbe:
This was inspired by #1986.

ejoerns (approved):

PR #1989, opened 2026-07-15
src/signature: reject embedded NUL in signer CN check

zayeem06:
cms_check_signer_cns compares the signer certificate CN against the allowed-signer-cns list with g_strv_contains, which treats the ASN.1 string as a NUL-terminated C string. A signer whose certificate chains to the keyring but is not in the list can pass the check with a CN like "Test Org Release-1\0-attacker", because the comparison stops at the embedded NUL and matches the allowed prefix. Read the ASN1 length and skip any CN containing an embedded NUL before comparing.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -1124,8 +1124,22 @@ static gboolean cms_check_signer_cns(CMS_ContentInfo *cms, GError **error)
 			continue;
 
 		const X509_NAME_ENTRY *cn = X509_NAME_get_entry(current_signer, index);
+		const ASN1_STRING *cn_asn1 = X509_NAME_ENTRY_get_data(cn);
+		const unsigned char* cn_value = ASN1_STRING_get0_data(cn_asn1);
+		int cn_length = ASN1_STRING_length(cn_asn1);
+		// the ASN.1 string is compared as a NUL-terminated C string below, so a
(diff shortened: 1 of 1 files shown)


jluebbe:
Do you see any valid reason to have NUL bytes in the CN? If not, I'd prefer to just reject any such certificates.

zayeem06:
No, there's no legitimate reason for a NUL byte in the CN. Changed it to reject the certificate outright when the CN contains an embedded NUL, rather than skipping the entry and moving on.

jluebbe:
Thanks, this looks good now. Could you squash the commits and add a commit description?

jluebbe:
As I understand it, the trusted CA iself would have to sign different certificates with the same CN prefix up to the first NUL byte to make this in issue in practice. Do you agree?

zayeem06:
Squashed into one commit and added a description.

Yes, that matches how I see it. It only bites when the keyring trusts a CA that issues certs to more than one signer and the allowed-signer-cns list is what restricts which of those a given device accepts. To exploit it the CA has to actually issue a cert whose CN, up to the first NUL, equals an allowlisted entry while the full CN differs. If the CA rejects embedded NULs this never happens, but relying on the CA to do that is exactly the assumption this guards against, so I think it's worth rejecting them here regardless.

jluebbe (approved):

PR #1991, opened 2026-07-15
docs: reference: Small fixes

pslldq:
Small fixes regarding the documentation for the `type` key in the configuration files when used with artifact repositories.

Given that the `type` key in bundle manifest files for images installed into artifact repositories is probably also something to have in the future, I've worded it as "not yet supported".

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -1096,6 +1096,10 @@ The following fields are supported for image sections:
   Detailed information about the supported types is described in the
   :ref:`Supported Image Types <sec-ref-supported-image-types>` section.
 
+  .. note::
+    This key is not yet supported for artifacts and will cause an error if
+    used.
+
 ``sha256`` (generated)
   sha256 of image file.
   RAUC determines this value automatically when creating
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -800,7 +800,6 @@ The name must be different from any slot class names.
 
 ``type=<type>`` (required)
   The type of this repository.
-  Currently supported values are ``files`` and ``trees``.
 
   ``files``
     each artifact is a single file


docs/reference.rst
```diff
@@ -1096,6 +1095,10 @@ The following fields are supported for image sections:
   Detailed information about the supported types is described in the
   :ref:`Supported Image Types <sec-ref-supported-image-types>` section.
 
+  .. note::
+    This key is not yet supported for images installed into artifact repositories
```
```suggestion
    This key is not yet supported for artifacts.
```

This would be the wording we use in other places in the documentation, too.

ejoerns:
I noted that `rauc info` currently prints the 'type' for artifacts, e.g. as `tar (detected)`.
Since the purpose of 'type' is undefined for artifacts, we should maybe also avoid displaying it for now?

jluebbe:
> I noted that `rauc info` currently prints the 'type' for artifacts, e.g. as `tar (detected)`. Since the purpose of 'type' is undefined for artifacts, we should maybe also avoid displaying it for now?

I'd leave it as-is instead of reverting that again after solving type handling for artifacts.

ejoerns:
> > I noted that `rauc info` currently prints the 'type' for artifacts, e.g. as `tar (detected)`. Since the purpose of 'type' is undefined for artifacts, we should maybe also avoid displaying it for now?
> 
> I'd leave it as-is instead of reverting that again after solving type handling for artifacts.

No reverting, I just thought of adding an `if` for now:

``` diff
index dc6ca0d8..8f97123a 100644
--- a/src/main.c
+++ b/src/main.c
@@ -1114,7 +1114,8 @@ static gchar *info_formatter_readable(RaucManifest *manifest)
                } else {
                        g_string_append_printf(text, "    (no image file)\n");
                }
-               g_string_append_printf(text, "    Type:      %s%s\n", img->type, img->type_from_fileext ? " (detected)" : "");
+               if (!img->artifact)
+                       g_string_append_printf(text, "    Type:      %s%s\n", img->type, img->type_from_fileext ? " (detected)" : "");
                if (img->filename) {
                        g_autofree gchar* formatted_size = g_format_size_full(img->checksum.size, G_FORMAT_SIZE_LONG_FORMAT);
                        g_string_append_printf(text, "    Checksum:  %s\n", img->checksum.digest);
```

If `type` support for artifacts is addressed and implemented soon, this might not be necessary, though.

ejoerns:
Note: I'll address this myself. No need to fix it here.

jluebbe (approved):

ejoerns (approved):

jluebbe:
I've applied the suggested change myself.

PR #1992, opened 2026-07-16
src/manifest: reject '/' in image filenames from bundled manifests

zayeem06:
check_manifest_bundled backs check_manifest_internal and check_manifest_external, but unlike check_manifest_input it never rejects an image filename containing '/'. Those filenames become path components: install.c uses image->filename verbatim when it is absolute and otherwise joins it onto the bundle directory, and the casync conversion path joins it onto the extracted content dir, so a bundle-supplied manifest with `filename=/etc/shadow` or `filename=../../../../tmp/x` resolves outside the bundle. This applies the same rejection check_manifest_input already does, to the image filename and to the converted output names that install.c builds paths from in the same way. Worth checking that last part: 'rauc bundle' derives the converted names from the image filename, so they should never legitimately contain a slash, and no bundle rauc can create is affected since check_manifest_input already forbids it.

### DIFF
--- a/src/manifest.c
+++ b/src/manifest.c
@@ -629,6 +629,11 @@ static gboolean check_manifest_bundled(const RaucManifest *mf, GError **error)
 		if (!image->filename)
 			continue;
 
+		if (strchr(image->filename, '/')) {
+			g_set_error(error, R_MANIFEST_ERROR, R_MANIFEST_CHECK_ERROR, "Image filename %s must not contain '/'", image->filename);
+			return FALSE;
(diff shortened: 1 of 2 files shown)


ejoerns (approved):
As noted, this will not apply to a bundle created by RAUC. Intentionally creating an artificial broken bundle might be up to the signer's choice/responsibility.
On the other hand, it's not wrong as a sanity check, thus fine for me.

jluebbe (approved):

PR #1994, opened 2026-07-17
src/artifacts_composefs: validate object names read from the image

zayeem06:
composefs_collect_objects takes the object names out of the image.cfs in the bundle, and r_composefs_artifact_install then uses them as path components: the dirname goes to g_mkdir_with_parents below the repo object store, and copy_file joins the same name onto both the bundle and the repo store. libcomposefs sets the payload verbatim from the trusted.overlay.redirect xattr and only strips a leading '/', and g_build_filename does not normalize '..', so an image with a redirect like ../../../../etc/cron.d/x makes rauc create directories and copy files outside the artifact repository; r_composefs_artifact_repo_prune reads the same names. This validates the name against the layout mkcomposefs --digest-store produces and r_composefs_artifact_repo_prepare already scans for, so no bundle rauc can create is affected, and that last part is the bit worth double checking. The load failure now reports an error rather than looking like an image with no objects, since collecting the objects has to propagate one anyway.

### DIFF
--- a/src/artifacts_composefs.c
+++ b/src/artifacts_composefs.c
@@ -1,6 +1,7 @@
 #include <asm-generic/errno-base.h>
 #include <errno.h>
 #include <stdint.h>
+#include <string.h>
 #include <unistd.h>
 #include <fcntl.h>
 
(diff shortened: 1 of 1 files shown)


jluebbe (approved):

jluebbe:
@zayeem06 Can you share the approach you're using to to find these issues?

zayeem06:
This issue was identified using BugQore, our proprietary Multi-Agent Remediation Engine, which performs repository-scale semantic program analysis, contextual dependency modeling, execution-path reasoning, and multi-stage validation to identify and remediate high-confidence security findings.

If your organization is exploring advanced AI-driven software assurance capabilities, we'd be pleased to discuss how BugQore can be integrated into your engineering and security workflows.

PR #1995, opened 2026-07-17
add per-slot runtime-state storage in /run/rauc

jluebbe:
Add a small subsystem that lets handlers persist per-slot key/value pairs in `/run/rauc/runtime-state` using the same `GKeyFile` approach as the existing status file. The state is wiped on reboot (tmpfs) but survives a 'rauc' process restart within the same boot, which is what the upcoming boot-mtd-fallback recovery logic needs.

Public API (`include/runtime_state.h`):
```
  gboolean r_runtime_state_init(const gchar *path, GError **error);
  void     r_runtime_state_slot_set(RaucSlot *, const gchar *key, const gchar *value);
  gboolean r_runtime_state_slot_get(RaucSlot *, const gchar *key, gchar **value);
  gboolean r_runtime_state_commit(GError **error);
  void     r_runtime_state_cleanup(void);
```
Groups follow the existing `slot.<slot.name>` convention from the `system.conf` sections. A corrupt file is renamed to `<path>.broken` on init, matching the status-file recovery pattern.

I've used Claude to generate this code based on my specs. I've reviewed, tested and slightly reworked the result.

Assisted-by: Claude:claude-opus-4.7

### DIFF
--- a/include/runtime_state.h
+++ b/include/runtime_state.h
@@ -1,0 +1,65 @@
+#pragma once
+
+#include <glib.h>
+
+#include "slot.h"
+
+/**
+ * Initialize the runtime-state subsystem.
+ *
+ * Allocates an internal GKeyFile and remembers the path. If @path already
+ * exists it is loaded; missing files yield an empty GKeyFile and success. A
+ * corrupt file is renamed to <path>.broken and a fresh empty GKeyFile is used.
+ *
+ * @param path Path to the runtime-state file
+ * @param error Return location for a GError, or NULL
+ *
+ * @return TRUE on success (including the "file does not exist" case),
+ *         FALSE on a non-recoverable error
+ */
+gboolean r_runtime_state_init(const gchar *path, GError **error)
+G_GNUC_WARN_UNUSED_RESULT;
+
+/**
+ * Set a per-slot key/value pair.
+ *
+ * If @value is NULL, the key is removed from the slot's group.
+ * Changes are held in memory until r_runtime_state_commit() is called.
+ *
+ * @param slot Slot whose name selects the group
+ * @param key Key name within the slot group
+ * @param value Value, or NULL to remove the key
+ */
+void r_runtime_state_slot_set(RaucSlot *slot, const gchar *key, const gchar *value);
+
+/**
+ * Get a per-slot key value.
+ *
+ * @param slot Slot whose name selects the group
+ * @param key Key name within the slot group
+ * @param value Out-parameter, newly-allocated value on TRUE, NULL on FALSE
+ *
+ * @return TRUE if the key exists, FALSE otherwise (also FALSE if init
+ *         has not been called)
+ */
+gboolean r_runtime_state_slot_get(RaucSlot *slot, const gchar *key, gchar **value)
+G_GNUC_WARN_UNUSED_RESULT;
+
+/**
+ * Persist the current runtime-state.
+ *
+ * Creates the parent directory if needed.
+ *
+ * @param error Return location for a GError, or NULL
+ *
+ * @return TRUE on success, FALSE on error
+ */
+gboolean r_runtime_state_commit(GError **error)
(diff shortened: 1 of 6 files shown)


src/runtime_state.c
```diff
@@ -0,0 +1,134 @@
...
+ * r_context() would recurse) and after configure (when r_context_conf()
+ * would re-set pending=TRUE and risk a spurious re-configure). */
+static GMutex runtime_state_lock;
+static GKeyFile *runtime_state_keyfile = NULL;
+static gchar *runtime_state_path = NULL;
+
+static gchar *make_group_name(RaucSlot *slot)
+{
+	return g_strdup_printf(RAUC_SLOT_PREFIX ".%s", slot->name);
+}
```
pH5, line 18: Why is this a helper function? Alternatively, why doesn't `save_slot_status_globally()` use a helper function for this?

src/runtime_state.c
```diff
@@ -0,0 +1,134 @@
...
+#include "runtime_state.h"
+#include "status_file.h"
+
+/* Module-private state. Kept here rather than on the RaucContext struct
+ * because the API must remain usable both during configure_target (when
+ * r_context() would recurse) and after configure (when r_context_conf()
+ * would re-set pending=TRUE and risk a spurious re-configure). */
+static GMutex runtime_state_lock;
+static GKeyFile *runtime_state_keyfile = NULL;
+static gchar *runtime_state_path = NULL;
```
pH5, line 13: Should these be bundled into a struct?

src/runtime_state.c
```diff
@@ -0,0 +1,134 @@
...
+}
+
+gboolean r_runtime_state_init(const gchar *path, GError **error)
+{
+	GError *ierror = NULL;
+
+	g_return_val_if_fail(path, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	G_GNUC_UNUSED g_autoptr(GMutexLocker) locker = g_mutex_locker_new(&runtime_state_lock);
```
pH5, line 27: I'm surprised the compiler warns about this even though `GMutexLocker` is `void`. I suppose these can be changed to `G_MUTEX_AUTO_LOCK` at some point in the future.

src/runtime_state.c
```diff
@@ -0,0 +1,134 @@
...
+	g_autoptr(GError) ierror = NULL;
+
+	g_return_val_if_fail(slot, FALSE);
+	g_return_val_if_fail(key, FALSE);
+	g_return_val_if_fail(value && *value == NULL, FALSE);
+
+	G_GNUC_UNUSED g_autoptr(GMutexLocker) locker = g_mutex_locker_new(&runtime_state_lock);
+
+	if (!runtime_state_keyfile)
+		return FALSE;
```
pH5, line 83: Why is this not `g_return_val_if_fail`?

include/runtime_state.h
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Get a per-slot key value.
+ *
+ * @param slot Slot whose name selects the group
+ * @param key Key name within the slot group
+ * @param value Out-parameter, newly-allocated value on TRUE, NULL on FALSE
+ *
+ * @return TRUE if the key exists, FALSE otherwise (also FALSE if init
+ *         has not been called)
```
pH5, line 43: Why does it seem to be expected use to call `_get` without `_init`?

src/runtime_state.c
```diff
@@ -0,0 +1,134 @@
+#include <errno.h>
+#include <glib/gstdio.h>
+
+#include "runtime_state.h"
+#include "status_file.h"
+
+/* Module-private state. Kept here rather than on the RaucContext struct
```
pH5, line 7: Is this 'module-private' as in `static`? This sounds a bit confusing,since clearly the held state is global, it's just only accessible via the `r_runtime_state_` API.

PR #1998, opened 2026-07-18
src/mbr: fix out-of-bounds stack write in validate_region

zayeem06:
get_number_of_sectors() handed the BLKGETSIZE ioctl a pointer to a 4-byte guint, but that ioctl is declared _IO(0x12,96) and the kernel handler stores an unsigned long through the pointer, so on LP64 validate_region() takes an 8-byte write into a 4-byte stack slot every time a boot partition switch runs on an MBR slot. The returned value is also counted in 512-byte sectors while sector_size here holds the logical block size from BLKSSZGET, so multiplying the two overstates the device by a factor of 8 on 4Kn devices and a region reaching past the end of the disk gets through the size guard. Using the existing get_device_size() helper removes both, since it reads the size in bytes via BLKGETSIZE64 into a guint64; the thing worth checking is that comparing start + size against bytes rather than sectors still reads correctly to you, and the boot_switch tests exercise this path.

### DIFF
--- a/src/mbr.c
+++ b/src/mbr.c
@@ -44,22 +44,6 @@ struct mbr {
 G_STATIC_ASSERT(sizeof(struct mbr) == 512);
 #pragma pack(pop)
 
-static gboolean get_number_of_sectors(gint fd, guint *sectors,
-		GError **error)
-{
-	g_return_val_if_fail(sectors, FALSE);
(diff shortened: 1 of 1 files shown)


jluebbe:
Force-pushed to line-wrap the commit message.

jluebbe (approved):

PR #1999, opened 2026-07-19
src/update_utils: stop copying once the announced image size is reached

zayeem06:
**Unbounded copy in r_copy_stream_with_progress**

The `size` argument was only feeding the progress percentage, so the copy ran to EOF of the input even though every caller had already checked the destination against that same `image->checksum.size`, which means an image file larger than the size recorded in the manifest is written past the region that was checked (past the inactive half for the boot-switch handlers, past `size-limit` for `img_to_raw`) and `copy_raw_image()` only notices once the data has reached the device. Bounding it in the shared helper covers all four callers; no bundle created by RAUC can hit this, so treat it as a sanity check rather than something valid bundles run into.

### DIFF
--- a/src/update_utils.c
+++ b/src/update_utils.c
@@ -112,6 +112,15 @@ gboolean r_copy_stream_with_progress(GInputStream *in_stream, GOutputStream *out
 			g_propagate_error(error, ierror);
 			return FALSE;
 		}
+
+		/* Callers check the destination against the announced size, so
+		 * never write more than that, even if the input has more data. */
+		if (sum_size + in_size > size) {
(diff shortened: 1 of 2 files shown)


ejoerns (approved):
Just added the file scope prefix to the commit title.

Fix looks good. Checking the size indirectly got lost when moving from `g_output_stream_splice()` to the custom copying for introducing progress reporting in 7e633ec01b49540b248c147febe328a70ea1e7b1.

ejoerns:
This ran into [ruff 0.16](https://github.com/astral-sh/ruff/releases/tag/0.16.0), released a few hours ago.
I'll have a look if that's fixable with manageable effort or requires custom config.

PR #2000, opened 2026-07-20
bootloaders/efi: propagate inner error from efi_bootorder_set()

94xhn:
While reading through the EFI bootchooser backend I noticed `efi_modify_persistent_bootorder()` calls `efi_bootorder_set(order, NULL)` and just throws away the inner error. Right after that it does `g_propagate_prefixed_error(error, ierror, "Modifying bootorder failed: ")`, which unconditionally dereferences `ierror` - but `ierror` is always NULL here since nothing ever gets a chance to set it. GLib's GError functions assert that the source error is non-NULL, so this doesn't turn into a nice error message, it aborts the process.

This path is reached from `r_mark_good()`/`r_mark_bad()`/`r_mark_active()` on every boot when `bootloader=efi`, so any time the underlying `efibootmgr --bootorder ...` call fails for a real-world reason (NVRAM full, firmware refusing the write, insufficient permissions, etc.) the rauc mark-good service or CLI crashes instead of just reporting the failure.

Fix is a one-liner: pass `&ierror` instead of `NULL` so the inner error actually gets captured before being propagated.

I reproduced it locally by building with meson/ninja and making the `efibootmgr` test mock fail the `--bootorder` call - before the fix `bootchooser-test` aborts with `GLib-FATAL-CRITICAL: g_propagate_error: assertion 'src != NULL' failed`, after the fix it just returns FALSE with a proper error message. Added a regression test (`/bootchooser/efi-bootorder-failure`) covering this along with a small env-var-gated failure trigger in the `efibootmgr` mock script, both included in this PR.

### DIFF
--- a/src/bootloaders/efi.c
+++ b/src/bootloaders/efi.c
@@ -484,7 +484,7 @@ static gboolean efi_modify_persistent_bootorder(RaucSlot *slot, gboolean prepend
 
 	g_autofree gchar *order = g_strjoinv(",", (gchar**)bootorder->pdata);
 
-	if (!efi_bootorder_set(order, NULL)) {
+	if (!efi_bootorder_set(order, &ierror)) {
 		g_propagate_prefixed_error(error, ierror, "Modifying bootorder failed: ");
 		return FALSE;
 	}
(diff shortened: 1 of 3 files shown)


ejoerns (approved):
@94xhn Thank you for finding and fixing this and congratulations to opening the 2000th PR here ;)

94xhn:
Thank you for the review! Nice coincidence that this was PR #2000.

jluebbe (approved):

PR #2001, opened 2026-07-20
src/manifest: validate image digest format in bundled manifests

zayeem06:
check_manifest_bundled() requires an image to carry a sha256 digest but never checks what that string looks like, and the digest is used as a path component in two places: r_artifact_repo_insert() builds artifact->path with g_strdup_printf("%s/.artifact-%s-%s", repo->path, name, digest), and r_slot_get_checksum_data_directory() and r_slot_move_checksum_data_directory() build "hash-<digest>" and join it onto slot->data_directory with g_build_filename(), neither of which normalizes '..'. A manifest with sha256=../../../../tmp/x therefore makes rauc mkdir, extract, rename and rm_tree outside the artifact repository and outside the slot data directory; the digest also reaches the slot status file through install.c and comes back as the rename source on the next install. Rejecting a digest that is not the plain hex string of the expected length for its checksum type covers both sinks in one place, and compute_checksum() only ever emits that, so bundles created by rauc are unaffected.

Worth noting for review that this needs a manifest a signer produced by hand rather than through 'rauc bundle', so like the filename check it is a sanity check on a field that is used to build paths. The existing tests used a '0815' placeholder digest, so those fixtures are updated to a real one.


ejoerns (approved):

ejoerns:
@zayeem06 could you add your Signed-off-by so we can merge this?

zayeem06:
Added the sign-off, should be good to merge now.

PR #2004, opened 2026-07-22
src/bundle: shell-quote paths in casync_make_arch

zayeem06:
casync_make_arch joins its argument vector into an `sh -c` command line, so an image filename like `rootfs.tar;touch /tmp/pwned;x` in the manifest of the bundle passed to `rauc convert --to=casync` runs arbitrary commands, fixed by shell-quoting the interpolated paths and extra args.

### DIFF
--- a/src/bundle.c
+++ b/src/bundle.c
@@ -241,22 +241,24 @@ static gboolean casync_make_arch(const gchar *idxpath, const gchar *contentpath,
 		goto out;
 	}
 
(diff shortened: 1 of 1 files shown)


ejoerns (approved):

ejoerns:
I just noted that the entire code path for archive conversion is not covered by tests, yet.

Created #2014 to address this.

PR #2010, opened 2026-07-28
src/main: always create symlinks for slot independent artifacts

pslldq:
Create the symlinks for slot independent artifacts even when RAUC cannot determine the currently booted slot (e.g. when booting via NFS).

As a side effect the artifacts directory is also always created and isn't missing when no booted slot could be determined.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2415,9 +2415,6 @@ static void create_run_links(void)
 		}
 	}
 
-	if (!booted_slot)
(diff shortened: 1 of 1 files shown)


jluebbe (approved):

jluebbe:
Rebased to fix the ruff style check.

jluebbe (approved):

PR #2019, opened 2026-07-31
src/polling: fix build if streaming is disabled

### LINKED ISSUE #2018: Build failure when streaming is disabled
**Describe the bug**
Link stage fails if streaming is disabled.

**Background information**
master @ 98adedba6e3ddb47155e332b1042b68496679f5c

Introduced in 7f5a37b83965a496133639bc8b2f8c15589410d1

**To Reproduce**
```
meson setup -Dgpt=enabled -Djson=enabled -Dnetwork=false -Dservice=true -Dstreaming=false --prefix=/usr build && ninja -C build
...
/usr/bin/ld: librauc.a.p/src_polling.c.o: in function `polling_fetch':
..build/../src/polling.c:113:(.text.polling_fetch+0x123): undefined reference to `r_nbd_error_quark'
/usr/bin/ld: ..build/../src/polling.c:118:(.text.polling_fetch+0x183): undefined reference to `r_nbd_error_quark'
```

**Expected behavior**
Build succeeds

**Logs**

**Additional context**

MatthiasWauer:
In case steaming is disabled the NBD errors are not defined,  leading to a linker failure

```c
// ndb.h
#define R_NBD_ERROR r_nbd_error_quark()
//ndb.c
r_nbd_error_quark(void) {...}
```

Fix it by including the checks for NDB errors only when streaming is enabled.

Fixes: #2018

### DIFF
--- a/src/polling.c
+++ b/src/polling.c
@@ -110,6 +110,7 @@ static gboolean polling_fetch(RPollingContext *polling_context, GError **error)
 
 	g_autoptr(RaucBundle) bundle = NULL;
 	if (!check_bundle(r_context()->config->polling_url, &bundle, CHECK_BUNDLE_DEFAULT, &access_args, &ierror)) {
+#if ENABLE_STREAMING
 		if (g_error_matches(ierror, R_NBD_ERROR, R_NBD_ERROR_NO_CONTENT)) {
 			g_clear_error(&ierror);
 			g_message("polling: no bundle available");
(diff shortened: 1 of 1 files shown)


ejoerns:
I'm not sure if we should compile `polling.c` at all if streaming is disabled, since polling doesn't work without streaming.

However, if we disable the module, this would require some changes to `src/service.c` which calls `r_polling_setup()`, etc.

@jluebbe Do you have a preferred solution?

jluebbe:
Polling without streaming is not useful. We should be able to have stubs in `include/polling.h`.

Because `src/config.h` accesses `r_polling_supported_install_criteria` (and others), parts of `parse_polling_section()` need to be made compile-time conditional, too.

We should also add a CI build without network or streaming to CI, but that should be a different PR.

PR #2020, opened 2026-07-31
src/update_handler: emmc: allow bootloader updates, if user partition was active

### LINKED ISSUE #2017: Allow `type=boot-emmc` bootloader updates while EMMC user parition is in use.
**Describe the bug**
There was a change in behaviour of rauc with commit 7a463272c97e1c18b2e14387daf5adb053c67325 (in 1.13), that rejects bootloader updates with `type=boot-emmc`, when the active bootloader is on the user partition.

**Background information**
rauc 1.15.2

**To Reproduce**
Install an update with `type=boot-emmc` bootloader while the running system has the EMMC user partition active.

**Expected behavior**
Rauc allows/performs the update, because there is no risk that the slot update does something wrong if the user partition is enabled. I do not see a reason why rauc should reject it.

**Logs**
```
 43% Copying image to bootloader.0
 70% Copying image to bootloader.0 failed.
rauc-event-Message: 09:02:49.076: Installation 393da191 failed: Installation error: Failed updating slot bootloader.0: Active eMMC partition is UDA when boot0/boot1 was expected.
```

**Additional context**

I relied on the behaviour mostly for initial installation, where the bootloader is from on the user partition.

In addition, field updates with devices that initially never supported bootloader updates and use the user partition becomes more complicated. Any intermediate update must now flash the bootloader to a EMMC boot partition and switch to it with external tooling (e.g. mmc), instead of simply adding the slot information.

MatthiasWauer:
Commit 7a463272c (src/emmc: error-out for invalid bootpart in r_emmc_read_bootpart()) changed the behaviour to disallow bootloader updates in case the user partition was currently active.

The change restores the behaviour to allow that and log a warning only. Also match the implementation detail, that the boot partition 1 is used in that case.

Fixes: #2017

### DIFF
--- a/src/update_handler.c
+++ b/src/update_handler.c
@@ -2348,11 +2348,22 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot, 
 	/* read active boot partition from ext_csd */
 	res = r_emmc_read_bootpart(realdev, &part_active, &ierror);
 	if (!res) {
-		g_propagate_error(error, ierror);
-		return FALSE;
-	}
-
-	if (part_active == -1) {
+		/* An active user partition (UDA) is not a valid boot partition,
+		 * but was historically supported. Preserve that behaviour by
+		 * only warning and continuing instead of aborting. Reserved or
+		 * otherwise invalid values are still rejected below. */
+		if (g_error_matches(ierror, R_EMMC_ERROR, R_EMMC_ERROR_BOOTPART_UDA)) {
+			g_warning("%s Check your setup! Continuing using boot1 for the update.", ierror->message);
+			g_clear_error(&ierror);
+			/* For simplicity: Consider boot0 to be active, so that the
+			 * (inactive) boot1 partition is used as update target, matching
+			 * the behaviour prior to the UDA sanity check being added. */
+			part_active = 0;
+		} else {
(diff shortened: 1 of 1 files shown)


src/update_handler.c
```diff
@@ -2348,11 +2348,22 @@ static gboolean img_to_boot_emmc_handler(RaucImage *image, RaucSlot *dest_slot,
...
-	}
-
-	if (part_active == -1) {
+		/* An active user partition (UDA) is not a valid boot partition,
+		 * but was historically supported. Preserve that behaviour by
+		 * only warning and continuing instead of aborting. Reserved or
+		 * otherwise invalid values are still rejected below. */
+		if (g_error_matches(ierror, R_EMMC_ERROR, R_EMMC_ERROR_BOOTPART_UDA)) {
+			g_clear_error(&ierror);
+			g_warning("Active eMMC partition is the user partition (UDA), which is not a valid boot partition. Continuing using boot1 for the update.");
```
ejoerns: I'd use the error message we already have and just add the information about how we handle it:

```suggestion
			g_warning("%s Check your setup! Continuing using boot1 for the update.", ierror->message); 
			g_clear_error(&ierror);			
```

MatthiasWauer: Applied suggestion.

PR #2023, opened 2026-08-17
src/manifest: reject '..' in hook filename from bundled manifests

zayeem06:
The image filename in a bundled manifest is already rejected when it contains '/', but the hook filename isn't, even though install.c joins it onto the bundle content dir with g_build_filename() and runs the result directly as the hook executable. A manifest with filename=../../../../some/script under [hooks] therefore runs a program from outside the content dir as root, which is code execution rather than the file placement the image filename check prevents. This adds the same '/' rejection for hook_name in check_manifest_bundled and check_manifest_input so it lands the same way the image filename check does, with a regression test that exercises it through check_manifest_internal.


jluebbe:
This would be too strict, as `some-dir/hook.sh` is a valid hook script name that could be in use somewhere.

Instead, we should reject any path that contains `..` as a component.

zayeem06:
Good point, a sub-directory like some-dir/hook.sh should stay valid. Reworked it to reject only a '..' component in the hook filename instead of any '/', in both check paths. Pushed, and updated the test accordingly.

jluebbe (approved):

